PR #1, opened 2018-05-14
Update pipeline for new repo

jmcatamney:
Now that gpupgrade no longer needs to pull in the GPDB source, we can revert a
lot of the changes made to the original multi-cluster testing pipeline pulled
from https://github.com/greenplum-db/gp-performance-testing/tree/dev/pg_upgrade.

Specifically, this commit updates the pipeline file to do the following:
- Update task fields for the AWS -> GCP migration
- Remove the compile job
- Pull down a 6.X binary to replace the GPDB 6.X source
- Pull out the inline Bash setup code into a setup_clusters.sh script

The pipeline still doesn't run any tests once the clusters are set up; the
intention is for this to be a platform for manual testing, and automated tests
will be added to the pipeline later.

Authored-by: Jamie McAtamney <jmcatamney@pivotal.io>


doty-pivotal:
Looks ok.  I think that I don't fully understand the intended use.  With the make targets, would we have 1 pipeline for each pair of GPDB versions.  If so, ok, but I think we may change that workflow in the future as we iterated on our pipeline.

jmcatamney:
@doty-pivotal The multiple make targets are a carry-over from the old version, which had them as a carry-over from Jacob's scripts.  I left them in for now since it required less modification and it's an easy way to get multiple side-by-side gpupgrade pipelines for different stories, and I do intend for those to go away at the point when we add 4->5, 4->6, and 5->6 jobs to the same pipeline.

Eulerizeit:
I don't care if gp_upgrade is in a separate repo, however, could we have a
conversation on the mailing list about how we are going to deal with
compatibility  and versioning that addresses the issues we are currently
having with single repos that work with multiple db versions and how we EOL
various db versions?

To be clear, I don't think the pain point is the repo rather how are we
going to bundle and ship.

Rob

PR #5, opened 2018-06-04
Some more Makefile improvements

jchampio:
Get rid of most of the copy-paste for build targets, and add correct deps for the installation recipes.

Ideas for future work:
1. Continue reducing duplication, this time for the install step.
2. Decide why we're using a branch name for the Mac/Linux builds (in fact, why do those targets exist in the first place?).
3. Move some configure-type information (where do we install? what hosts are we talking to? which platform are we building?) out of the Makefile into an actual configure script.
4. Don't use `$GOBIN` as a build staging area; that's in my `PATH` and it shouldn't be touched unless I explicitly install. 😄

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -65,10 +65,18 @@ protobuf :
 		mockgen -source idl/cli_to_hub.pb.go  > mock_idl/cli_to_hub_mock.pb.go
 		mockgen -source idl/hub_to_agent.pb.go  > mock_idl/hub_to_agent_mock.pb.go
 
-build :
-		go build $(GOFLAGS) -o $(BIN_DIR)/$(AGENT) -ldflags $(UPGRADE_VERSION_STR) $(AGENT_PACKAGE)
-		go build $(GOFLAGS) -o $(BIN_DIR)/$(CLI) -ldflags $(UPGRADE_VERSION_STR) $(CLI_PACKAGE)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -65,22 +65,26 @@ protobuf :
...
+build_mac: OS := MAC
+build_linux build_mac: build
+
+agent-package: EXE_NAME := $(AGENT)
+cli-package: EXE_NAME := $(CLI)
+hub-package: EXE_NAME := $(HUB)
+
+$(PACKAGES): %-package:
+	$(PREFIX) go build $(GOFLAGS) -o $(BIN_DIR)/$(EXE_NAME)$(POSTFIX) -ldflags $(UPGRADE_VERSION_STR) github.com/greenplum-db/gpupgrade/$*
+
```
nadeemg: Is `$(PACKAGES)` a list  consisting of `agent-package, cli-package, hub-package`? If yes, should the `%-package` be just a `%`? Also, what does the `$*` at the end of the line do?

jchampio: > Is $(PACKAGES) a list consisting of agent-package, cli-package, hub-package?

Yes.
> If yes, should the %-package be just a %? Also, what does the $* at the end of the line do?

This construction is called a [static pattern rule](https://www.gnu.org/software/make/manual/html_node/Static-Usage.html). The purpose is to match the prefix part (`agent`, `hub`, or `cli`) and then substitute that later for the `$*` special variable.

PR #6, opened 2018-06-06
Some more small fixes

jchampio:
1. If the `RootCmd` of `gpupgrade_hub_main` errors out before we initialize `gplog`, we can't actually log an error or we'll crash. Fix this by doing a basic initialization if there is no logger set up.

2. Correct the error message if `StatusUpgrade` fails. We were getting errors that basically said `Unable to connect to hub: hub handed us an error over the connection we just claimed we couldn't make`.

### DIFF
--- a/hub/gpupgrade_hub_main.go
+++ b/hub/gpupgrade_hub_main.go
@@ -62,6 +62,14 @@ func main() {
 	RootCmd.PersistentFlags().StringVar(&logdir, "log-directory", "", "gpupgrade_hub log directory")
 
 	if err := RootCmd.Execute(); err != nil {
+		if gplog.GetLogger() == nil {
+			// In case we didn't get through RootCmd.Execute(), set up logging
+			// here. Otherwise we crash.
+			// XXX it'd be really nice to have a "ReinitializeLogging" building
+			// block somewhere.
+			gplog.InitializeLogging("gpupgrade_hub", "")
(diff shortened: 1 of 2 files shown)


hub/gpupgrade_hub_main.go
```diff
@@ -62,6 +62,14 @@ func main() {
 	RootCmd.PersistentFlags().StringVar(&logdir, "log-directory", "", "gpupgrade_hub log directory")
 
 	if err := RootCmd.Execute(); err != nil {
+		if gplog.GetLogger() == nil {
+			// In case we didn't get through RootCmd.Execute(), set up logging
+			// here. Otherwise we crash.
+			// XXX it'd be really nice to have a "ReinitializeLogging" building
```
slari-pivotal: Minor comment. Can we use different characters. e.g. TODO etc. rather than XXX.  Thanks.

jchampio: Do we have a set of debt markers already in use? The three I'm most used to are
- `TODO`: when debt has a clear path forward that we're in control of. Not necessarily negative in meaning.
- `XXX`: when something smells wrong, but it's more nebulous and may not have a clear downside yet. Tends to be more negative.
- `FIXME` when there's a clear downside/bug/misbehavior that we're deciding not to fix now. Pretty much always negative.

PR #7, opened 2018-06-06
Setup config before hub start

nadeemg:
Before this commit the hub started up without a config and waited for a user to run `check config`. So the hub could fail in various places if a `check config` had not yet been run. And these failures were often silent because they were expected in the normal workflow and that masks unexpected failures.

This commit adds a step which writes a minimal config file before the hub is started, which simplifies config handling for the rest of the hub code.

Since the existence of the config file no longer signals completion of `check config`, we've made the status reporting for the `check config` step consistent with others. 

Lastly, the CLI was giving the user a usage message even in cases of real errors. This fix may need to be applied elsewhere in the CLI.

### DIFF
--- a/hub/services/hub_check_config.go
+++ b/hub/services/hub_check_config.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/greenplum-db/gpupgrade/db"
 	"github.com/greenplum-db/gpupgrade/hub/configutils"
+	"github.com/greenplum-db/gpupgrade/hub/upgradestatus"
 	pb "github.com/greenplum-db/gpupgrade/idl"
 	"github.com/greenplum-db/gpupgrade/utils"
 
@@ -48,10 +49,23 @@ func (h *Hub) CheckConfig(ctx context.Context,
 	in *pb.CheckConfigRequest) (*pb.CheckConfigReply, error) {
 	gplog.Info("starting CheckConfig()")
 
+	c := upgradestatus.NewChecklistManager(h.conf.StateDir)
+	checkConfigStep := "check-config"
+
+	err := c.ResetStateDir(checkConfigStep)
+	if err != nil {
+		gplog.Error("error from ResetStateDir " + err.Error())
+	}
+	err = c.MarkInProgress(checkConfigStep)
+	if err != nil {
+		gplog.Error("error from MarkInProgress " + err.Error())
+	}
+
 	dbConnector := db.NewDBConn("localhost", int(in.DbPort), "template1")
 	defer dbConnector.Close()
-	err := dbConnector.Connect(1)
+	err = dbConnector.Connect(1)
 	if err != nil {
+		c.MarkFailed(checkConfigStep)
 		gplog.Error(err.Error())
 		return &pb.CheckConfigReply{}, utils.DatabaseConnectionError{Parent: err}
 	}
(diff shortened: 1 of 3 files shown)


cli/commanders/preparer.go
```diff
@@ -103,3 +108,33 @@ func HowManyHubsRunning() (int, error) {
 	// only needed if the command errors, but somehow put a parsable & negative value on stdout
 	return -1, err
 }
+
+func DoInit(stateDir string, oldBinDir string) error {
+	err := os.Mkdir(stateDir, 0700)
+	if os.IsExist(err) {
+		return fmt.Errorf("gpupgrade state dir already exists (%s). Did you already run gpupgrade prepare init?", stateDir)
```
jmcatamney: Could we rephrase the error message to `gpupgrade state dir %s already exists` so it flows a bit better?

hub/services/hub_check_config.go
```diff
@@ -48,32 +49,42 @@ func (h *Hub) CheckConfig(ctx context.Context,
 	in *pb.CheckConfigRequest) (*pb.CheckConfigReply, error) {
 	gplog.Info("starting CheckConfig()")
 
+	c := upgradestatus.NewChecklistManager(h.conf.StateDir)
+	checkConfigStep := "check-config"
+
+	err := c.ResetStateDir(checkConfigStep)
+	if err != nil {
+		gplog.Error("error from ResetStateDir " + err.Error())
+	}
```
slari-pivotal: If we cannot clear the directory, then that means that older state files might be there. In that case, subsequent status checks may be misleading.

Does it make sense to return the error to the user here and not proceed further?

integrations/integrations_suite_test.go
```diff
@@ -66,10 +69,21 @@ var _ = BeforeSuite(func() {
 })
 
 var _ = BeforeEach(func() {
+	testWorkspaceDir, err := ioutil.TempDir("", "")
+	Expect(err).ToNot(HaveOccurred())
+	testStateDir = filepath.Join(testWorkspaceDir, ".gpupgrade")
```
slari-pivotal: Shouldn't we call `utils.GetStateDir()` instead of `.gpupgrade`?

hub/services/hub_check_config.go
jchampio: Yes, I think it does. We don't currently do that in any of the other places we call `ResetStateDir()`, so I think this pattern needs to be revisited -- maybe we need another "step state" that means "internal error"? I'll add a chore.

integrations/integrations_suite_test.go
nadeemg: In this case, we're making a `.gpupgrade` dir under a temp dir, for the test to use to avoid polluting `~/.gpupgrade`

jchampio: In this case, an ephemeral state directory is being constructed for the integration tests. (We're doing this so that the integration tests don't stomp on an actual gpupgrade installation.)

`utils.GetStateDir()` will give us the "official" state directory, which won't actually point to the directory we want until we set `GPUPGRADE_HOME` down below.

jchampio:
Merged into master with 39d2348.

cli/commanders/preparer.go
```diff
@@ -103,3 +108,33 @@ func HowManyHubsRunning() (int, error) {
...
+
+	segConfig := make(configutils.SegmentConfiguration, 0)
+
+	configJSON := &configutils.ClusterConfig{
+		SegConfig: segConfig,
+		BinDir:    oldBinDir,
+	}
+
+	err = services.SaveQueryResultToJSON(configJSON, configFileHandle)
+	if err != nil {
```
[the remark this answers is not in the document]
nadeemg: good catch

jchampio:
I accidentally force-pushed away the commit history for this branch (I've never actually done that before; maybe I triggered a bug in GitHub?). Not a big deal for the master branch, which has a fine state, but if you want to see the original commit history, see [this branch](https://github.com/jchampio/gpupgrade/commits/dev/init-config-once).

PR #8, opened 2018-06-07
Refactor GetStatus patterns; add services.Step

jchampio:
There are two conceptual changes here:

1. Don't claim to return an error from the `GetStatus` helpers, if we're just going to return `nil` all the time and ignore that error anyway in most/all of the callers. This isn't a long-term change -- we *do* want to handle errors here -- but we don't really have a good concept yet of how these errors should be correctly bubbled up. Since this "fake error" makes the next refactoring step messy, get rid of it for now.

2. Refactor away the duplication from the `Hub.StatusUpgrade()` function, by introducing a `Step` struct. This struct encapsulates a logical step in the status check -- its name, gRPC code, and implementation -- and provides a nice interface that makes it easy to loop through and check each status in order.

### DIFF
--- a/agent/services/conversion_status.go
+++ b/agent/services/conversion_status.go
@@ -5,7 +5,6 @@ import (
 
 	"errors"
 	"fmt"
-	"github.com/greenplum-db/gp-common-go-libs/gplog"
 	"github.com/greenplum-db/gpupgrade/hub/upgradestatus"
 	pb "github.com/greenplum-db/gpupgrade/idl"
 	"path/filepath"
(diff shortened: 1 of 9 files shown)


jchampio:
@jmcatamney FYI

jchampio:
@khuddlefish I certainly don't mind more eyes! But I don't want to spam you and @chrishajas if it's detrimental to your workflow. What was the intended use of the management utilities group?

khuddlefish:
I didn't actually know it existed until you used it to tag us a few days ago :P

jchampio:
@khuddlefish Fair enough. 😄 I will try to find whoever owns the group membership list and we can try to get it changed if you'd like!

jchampio:
Crash fixes are coming in the follow-up refactor, which needs this to continue forward, so I've merged.

PR #9, opened 2018-06-08
Refactor ClusterPair

jmcatamney:
This commit refactors ClusterPair to be a pair of cluster.Cluster objects
instead of holding all of the segment information itself, to allow us to
leverage the functionality in the cluster package.
    
This has several consequences:
- Segment configuration information is now retrieved once and stored in memory,
  rather than being read from the config file every time.
- The configutils package has been removed and the Reader functionality has
  been added to ClusterPair.
- Quite a few testing-related structs are no longer necessary and have also been
  removed.
- Several tests have been modified to use the cluster's Executor for shelling
  out to Bash instead of CommandExecer, due to its simpler and more flexible
  interface (and CommandExecer will be removed entirely in a later commit).
- Helper functions for creating test Clusters and test ClusterPairs have been
  added to testutils to simplify test setup.
    
This commit breaks several tests and leaves them marked as Pending, because the
refactor uncovered flakiness and invalid behavior that would require too
extensive a refactor to fix, and this commit is already fairly large.  These
tests will be addressed in later refactor commits.


hub/services/cluster_pair.go
```diff
@@ -0,0 +1,191 @@
...
+	clusterConfig := &ClusterConfig{BinDir: binDir}
+	for _, contentID := range c.ContentIDs {
+		segConfigs = append(segConfigs, c.Segments[contentID])
+	}
+	clusterConfig.SegConfigs = segConfigs
+	contents, err := json.Marshal(clusterConfig)
+	if err != nil {
+		return err
+	}
+	err = utils.System.WriteFile(configFilePath, contents, 0644)
```
doty-pivotal: So we are going to pick up the "atomic" write work that @kyeap-pivotal and @jchampio when we rebase that branch...

hub/services/cluster_pair.go
```diff
@@ -0,0 +1,191 @@
+package services
```
nadeemg: I'm having trouble thinking of ClusterPair as a service. What was your thinking for moving this file into the services directory?

hub/services/hub_check_config.go
```diff
@@ -84,52 +52,15 @@ func (h *Hub) CheckConfig(ctx context.Context,
...
-		return errors.New(errMsg)
+		errMsg := fmt.Sprintf("Unable to get segment configuration for old cluster: %s", err.Error())
+		return clusterPair, errors.New(errMsg)
 	}
+	clusterPair.OldCluster = cluster.NewCluster(segConfigs)
+	clusterPair.OldBinDir = oldBinDir
 
-	return nil
+	err = clusterPair.WriteOldConfig(stateDir)
+	return clusterPair, err
```
nadeemg: Do we need to return clusterPair here?

hub/services/cluster_pair.go
jmcatamney: There were two reasons:

1) `ClusterPair` was previously `cluster.Pair`, meaning we'd have a `cluster` package in gpupgrade and be using the `cluster` package from gp-common-go-libs, which would cause import conflicts and could possibly cause some confusion.  Moving it out of the `cluster` package reduces ambiguity, and there wasn't really an obvious name to use instead.

2) After this refactor, there are only three directories in hub: `services`, `cluster_ssher` (which is going away soon in a follow-up refactor, now that we have Cluster's remote execution methods), and `upgradestatus` (which contains some `services`-like code like `shutdown_clusters.go` which could be folded into the files in `services`, and some miscellaneous code like the ChecklistManager).  At this stage, there's not much of a point to separating things out since 90+% of the code is in `services`, so we figured we'd move ClusterPair out and then after the other refactors we could remove the `services` directory.

hub/services/cluster_pair.go
jmcatamney: Correct.

kyeap-pivotal: Yes we will pick up that change once the `re-init-hub-cluster-pair` branch get in.

hub/services/hub_check_config.go
jmcatamney: Oh wait--if you mean that the clusterPair should be updated already because we passed in a pointer, yes, you're probably right, we structured it that way before making it a pointer receiver.  We'll make sure it updates without the return, and if so we'll change that.

jmcatamney:
@jchampio Regarding your second point, we did add a `ClusterConfig` struct with a cluster.Cluster and a BinDir string to enable JSON marshaling/unmarshaling, so that capability is there.  We didn't see any other points in this refactor where using a ClusterConfig instead of a ClusterPair would help (since it would add an extra layer between ClusterPair and the Cluster or BinDir), but it might prove useful in a future refactor.

PR #10, opened 2018-06-14
Refactor tests to use MockChecklistManager

slari-pivotal:
Previously, our tests were checking the status of individual steps by
calling out to status upgrade in the CLI, which required a lot of
mocking and made some tests flaky.  This commit refactors the tests to
use a MockChecklistManager, which simplifies the testing and reduces the
test time.

Additionally, the hub was refactored to take a ChecklistWriter
separately from the one in its RemoteExecutor, and several functions
were moved from the ClusterPair object to the Hub, in preparation for a
later refactor of ClusterSsher.

Finally, checks were added to the status upgrade step to not attempt to
gather status information for steps that could not have been run yet if
check config or prepare start-agents were not yet run, and the tests
were modified accordingly.

Co-authored-by: Jim Doty <jdoty@pivotal.io>
Co-authored-by: Shoaib Lari <slari@pivotal.io>


Makefile
```diff
@@ -20,7 +20,7 @@ MAC_PREFIX := env GOOS=darwin GOARCH=amd64
 LINUX_POSTFIX := .linux.$(BRANCH)
 MAC_POSTFIX := .darwin.$(BRANCH)
 
-GOFLAGS :=
+GOFLAGS := -gcflags="all=-N -l"
```
jchampio: Was this an intentional part of this commit? I don't see any mention in the commit message.

hub/services/hub_status_upgrade.go
```diff
@@ -68,21 +68,33 @@ func shutdownStatus(s Step, h *Hub) *pb.UpgradeStepStatus {
 
 // pgUpgradeStatus checks the pg_upgrade progress files for its status.
 func pgUpgradeStatus(s Step, h *Hub) *pb.UpgradeStepStatus {
+	status := &pb.UpgradeStepStatus{
+		Step: s.StepCode,
+	}
+	// We don't need to check the pg_upgrade status if there's no configuration yet
+	checkConfigStep := Step{"check-config", pb.UpgradeSteps_CHECK_CONFIG, stateCheckStatus}
+	if checkConfigStep.getStatus(checkConfigStep, h).Status != pb.StepStatus_COMPLETE {
```
jchampio: What's driving this change? Is the check for pg_upgrade status expensive, or misleading if check-config hasn't completed?

hub/services/hub_status_upgrade.go
```diff
@@ -68,21 +68,33 @@ func shutdownStatus(s Step, h *Hub) *pb.UpgradeStepStatus {
 
 // pgUpgradeStatus checks the pg_upgrade progress files for its status.
 func pgUpgradeStatus(s Step, h *Hub) *pb.UpgradeStepStatus {
+	status := &pb.UpgradeStepStatus{
+		Step: s.StepCode,
+	}
+	// We don't need to check the pg_upgrade status if there's no configuration yet
+	checkConfigStep := Step{"check-config", pb.UpgradeSteps_CHECK_CONFIG, stateCheckStatus}
```
jchampio: Step names probably need to use the new `upgradestatus.NAME` constants.

hub/services/hub_prepare_shutdown_clusters.go
```diff
@@ -1,35 +1,79 @@
...
+}
+
+func (h *Hub) ShutdownClusters() {
+	step := upgradestatus.SHUTDOWN_CLUSTERS
+
+	h.checklistWriter.ResetStateDir(step)
+	h.checklistWriter.MarkInProgress(step)
+
+	var errShutdownOld error
+	errShutdownOld = StopCluster(h.clusterPair.OldCluster, h.clusterPair.OldBinDir)
```
jchampio: Style suggestion: `newerr` and `olderr` are just fine for names; we already know they relate to "shutdown" from the context, and there's no need to spend our precious columns on them.

hub/services/hub_prepare_shutdown_clusters.go
```diff
@@ -1,35 +1,79 @@
...
+	}
 
-	 * Use go routine along with using files as a way to keep track of gpstop state
-	 */
+	masterDataDir := c.GetDirForContent(-1)
+	gpstopShellArgs := fmt.Sprintf("source %s/../greenplum_path.sh; %s/gpstop -a -d %s",
+		binDir, binDir, masterDataDir)
 
-	// XXX: May be tell user to run status, or if that seems stuck, check gpAdminLogs/gpupgrade_hub*.log
+	gplog.Info("gpstop args: %+v", gpstopShellArgs)
```
jchampio: This feels like something that belongs in `ExecuteLocalCommand`.

hub/services/hub_prepare_shutdown_clusters.go
```diff
@@ -1,35 +1,79 @@
...
+func StopCluster(c *cluster.Cluster, binDir string) error {
+	if !IsPostmasterRunning(c) {
+		return nil
+	}
 
-	 * Use go routine along with using files as a way to keep track of gpstop state
-	 */
+	masterDataDir := c.GetDirForContent(-1)
+	gpstopShellArgs := fmt.Sprintf("source %s/../greenplum_path.sh; %s/gpstop -a -d %s",
+		binDir, binDir, masterDataDir)
```
jchampio: Should be able to remove the duplication with
```
fmt.Sprintf("source %s/../greenplum_path.sh; %[1]s/gpstop -a -d %s", binDir, masterDataDir)
```

hub/services/hub_prepare_shutdown_clusters_test.go
```diff
@@ -38,28 +36,64 @@ var _ = Describe("PrepareShutdownClusters", func() {
...
+	It("isPostmasterRunning() succeeds", func() {
+		testExecutor := &testhelper.TestExecutor{}
 
-		_, err := hub.PrepareShutdownClusters(nil, &pb.PrepareShutdownClustersRequest{})
-		Expect(err).To(BeNil())
+		cluster := testutils.CreateSampleCluster(-1, 25437, "hostone",
+			"/master/datadir")
+		cluster.Executor = testExecutor
+		postmasterRunning := services.IsPostmasterRunning(cluster)
+		Expect(testExecutor.LocalCommands[0]).To(ContainSubstring("pgrep"))
```
jchampio: I think this is an example of testing something other than what we want to be testing. Does running `pgrep` guarantee that we're actually working?

hub/services/hub_prepare_shutdown_clusters_test.go
```diff
@@ -38,28 +36,64 @@ var _ = Describe("PrepareShutdownClusters", func() {
...
+
+		cluster := testutils.CreateSampleCluster(-1, 25437, "hostone",
+			"/master/datadir")
+		cluster.Executor = testExecutor
+
+		err := services.StopCluster(cluster, "/fake/bindir")
+
+		Expect(testExecutor.NumExecutions).To(Equal(2))
+		Expect(testExecutor.LocalCommands[0]).To(ContainSubstring("pgrep"))
+		Expect(testExecutor.LocalCommands[1]).To(ContainSubstring("gpstop"))
```
jchampio: Ditto. We're testing implementation details here, but not actually verifying functionality.

hub/services/hub_status_upgrade.go
```diff
@@ -68,21 +68,33 @@ func shutdownStatus(s Step, h *Hub) *pb.UpgradeStepStatus {
...
-	// certainly not be ignoring an error here.
-	conversionStatus, _ := h.StatusConversion(nil, &pb.StatusConversionRequest{})
 	status := &pb.UpgradeStepStatus{
 		Step: s.StepCode,
 	}
-
-	statuses := strings.Join(conversionStatus.GetConversionStatuses(), " ")
+	// We can't check the status of agent processes if the agents haven't been started yet
+	startAgentsStep := Step{"start-agents", pb.UpgradeSteps_PREPARE_START_AGENTS, stateCheckStatus}
+	if startAgentsStep.getStatus(startAgentsStep, h).Status != pb.StepStatus_COMPLETE {
```
jchampio: Same question as above: what's driving this change?

Makefile
slari-pivotal: Yes. We added these flags so that we can use Delve.

hub/services/hub_status_upgrade_test.go
```diff
@@ -177,132 +179,145 @@ var _ = Describe("status upgrade", func() {
...
-			outChan <- []byte("123")
+			})
 
-			utils.System.IsNotExist = func(error) bool {
-				return false
-			}
-			utils.System.FilePathGlob = func(string) ([]string, error) {
-				return []string{"somefile.inprogress"}, nil
-			}
+			It("reports that master upgrade is running when pg_upgrade/*.inprogress files exists", func() {
```
jchampio: This test name seems to be incorrect now -- `.inprogress` vs. `in.progress`?

Also, why did we switch from `.inprogress`?

hub/services/hub.go
```diff
@@ -56,14 +58,15 @@ type HubConfig struct {
 	LogDir         string
 }
 
-func NewHub(pair *ClusterPair, grpcDialer dialer, execer helpers.CommandExecer, conf *HubConfig, executor RemoteExecutor) *Hub {
+func NewHub(pair *ClusterPair, grpcDialer dialer, execer helpers.CommandExecer, conf *HubConfig, executor RemoteExecutor, checklistWriter cluster_ssher.ChecklistWriter) *Hub {
```
jchampio: Future work: one of the keys of maintainable dependency injection is that we refactor once the number of dependencies gets to be out of control, and I think we're starting to cross that line. If we haven't already. :smile:

kyeap-pivotal: The next refactor chore to remove cluster ssher is one of the stories that will help with the number of injections into the hub. Once cluster ssher is removed we should be able to remove the RemoteExecutor, and the commandExecer very soon after once all the remaining commands use the execers on the ClusterPair.

hub/services/hub_check_seginstall_test.go
```diff
@@ -40,7 +40,7 @@ var _ = Describe("PrepareSeginstall", func() {
 			Err: errChan,
 			Out: outChan,
 		})
-		hub = services.NewHub(testutils.CreateSampleClusterPair(), grpc.DialContext, commandExecer.Exec, conf, stubRemoteExecutor)
+		hub = services.NewHub(testutils.CreateSampleClusterPair(), grpc.DialContext, commandExecer.Exec, conf, stubRemoteExecutor, nil)
```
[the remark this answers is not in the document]
kyeap-pivotal: It seems like the concern is more the practice of passing in nil interfaces in general? Would refactoring the NewHub method to not accept nil parameters be a better practice?

hub/services/hub_prepare_shutdown_clusters_test.go
kyeap-pivotal: IsPostmasterRunning is doing a pgrep on the postmaster.pid file, and the test is confirming that the pgrep actually ran. Its possible that the postmaster PID exists but the postmaster isn't functional. Is that the case you're asking about? Should we instead exec a query? Or is there a better way of ensuring that the postmaster is running?

jchampio: > the test is confirming that the pgrep actually ran.

Well, sort of. It confirms that we ran a command containing the string `"pgrep"`, which is subtly different and IMHO not nearly as useful: If we pass bad options to `pgrep` and it never runs, this test will continue to pass. If we choose a method other than `pgrep` to check that the postmaster is up, someone will have to rewrite this test.

I think if we want to test to make sure that our check-postmaster step is working, we need to actually test that it checks for the postmaster. We can certainly push the system-interfacing building block (that is, the thing that actually queries the processes running on the system) down as far as possible, away from this code, but eventually we need to test that it interfaces with the system directly. And once we have that test in place, mocked tests like this will lose their value.

hub/services/hub_check_seginstall_test.go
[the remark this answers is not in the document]
jchampio: nil interfaces are a different concern, I think. The issue I have here is that we've constructed a half-valid object: some methods called on it will panic. We "know" which ones to avoid -- today -- but some poor soul tomorrow will run into the end of the illusion. If we want to allow nil parameters to the constructor, other code needs to handle the possibility that some of the struct members are nil.

One of the nicest patterns I've carried from C++ is the idea that an "object" should never exist in an invalid state, simply because it inserts a ticking time bomb into your code. It's easier to guarantee that in a language with exceptions, perhaps, but we might give it a try here.

hub/services/hub_prepare_shutdown_clusters.go
```diff
@@ -1,35 +1,79 @@
...
 	gplog.Info("starting PrepareShutdownClusters()")
 
-	// will be initialized for future uses also? We think so -- it should
-	oldPostmasterRunning, newPostmasterRunning := h.clusterPair.EitherPostmasterRunning()
-	if oldPostmasterRunning || newPostmasterRunning {
-		pathToGpstopStateDir := path.Join(h.conf.StateDir, "gpstop")
-		go h.clusterPair.StopEverything(pathToGpstopStateDir, oldPostmasterRunning, newPostmasterRunning)
-	} else {
-		gplog.Info("PrepareShutdownClusters: neither postmaster was running, nothing to do")
+	go h.ShutdownClusters()
```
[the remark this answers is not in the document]
slari-pivotal: Right now, the only thing ending goroutines is the process ending, and it's difficult to ensure they end in our current model.  A follow-up discussion would be good.

hub/services/hub_status_upgrade.go
slari-pivotal: We'll fix this.

hub/services/hub_status_upgrade.go
```diff
@@ -68,21 +68,33 @@ func shutdownStatus(s Step, h *Hub) *pb.UpgradeStepStatus {
...
+	}
 	state := upgradestatus.NewPGUpgradeStatusChecker(s.StatePath(h), h.clusterPair.OldCluster.GetDirForContent(-1), h.commandExecer)
 	return state.GetStatus()
 }
 
 // conversionStatus queries all segments for their upgrade status and
 // consolidates them into a single pass/fail.
 func conversionStatus(s Step, h *Hub) *pb.UpgradeStepStatus {
-	// FIXME: this status query involves RPC communication; we should almost
-	// certainly not be ignoring an error here.
```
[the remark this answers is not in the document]
slari-pivotal: We meant to remove the comment and check the error, but did not.  We'll add a check for that.

hub/services/hub_status_upgrade.go
slari-pivotal: There's no need to make the RPC call if the agents have not been started yet, and if we try it kills the hub, so we return PENDING in that case.

hub/services/hub_status_upgrade.go
slari-pivotal: We need the master data directory to check the pg_upgrade status, which we don't have if check-config has not completed.

hub/services/hub_prepare_shutdown_clusters_test.go
slari-pivotal: There are several places that we're testing a shelled-out command by just checking that the command was called at all.  We should probably make a follow-up chore to handle those better.

hub/services/hub_status_upgrade_test.go
slari-pivotal: It was just a typo; the actual value of the string was not being used.  We'll change it back to `.inprogress`.

hub/services/hub_prepare_shutdown_clusters.go
slari-pivotal: Since `ExecuteLocalCommand` just executes a string, it doesn't know what binary is being executed.  If it picked the first token in this case, for example, it would think it was executing `source`.

Adding a `gplog.Debug` to `ExecuteLocalCommand` to log the command would be a good idea.

hub/services/hub_prepare_shutdown_clusters.go
slari-pivotal: We'll change these.

hub/services/hub_prepare_shutdown_clusters.go
slari-pivotal: We'll fix this.

hub/services/hub_prepare_shutdown_clusters.go
```diff
@@ -52,8 +52,7 @@ func StopCluster(c *cluster.Cluster, binDir string) error {
 	}
 
 	masterDataDir := c.GetDirForContent(-1)
-	gpstopShellArgs := fmt.Sprintf("source %s/../greenplum_path.sh; %s/gpstop -a -d %s",
-		binDir, binDir, masterDataDir)
+	gpstopShellArgs := fmt.Sprintf("source %[1]s/../greenplum_path.sh; %[1]s/gpstop -a -d %[2]s", binDir, masterDataDir)
```
jchampio: For future reference, the index always starts from one and moves forward for all bare `%s` specifiers. The only time you have to specify an index using `%[i]s` is when you want to deviate from that rule.

hub/services/hub_status_upgrade.go
jchampio: Is this one coming in a future commit?

hub/services/hub_status_upgrade.go
jchampio: Can we add a bug to Tracker? "Hub crashes on unexpected RPC call" is not a good failure mode :smile: Once that's fixed, we might be able to revisit this code -- the fewer required interdependencies, the easier it is to reason about what's going on.

hub/services/hub_status_conversion_test.go
```diff
@@ -42,7 +42,8 @@ var _ = Describe("hub", func() {
 		}
 		stubRemoteExecutor = testutils.NewStubRemoteExecutor()
 
-		hub = services.NewHub(clusterPair, grpc.DialContext, nil, conf, stubRemoteExecutor)
+		hub = services.NewHub(clusterPair, grpc.DialContext, nil, conf,
+			stubRemoteExecutor, nil)
```
jchampio: Ditto.

hub/services/hub_prepare_start_agents_test.go
```diff
@@ -39,7 +39,8 @@ var _ = Describe("PrepareStartAgents", func() {
 			Out: outChan,
 		})
 		clusterPair := testutils.CreateSampleClusterPair()
-		hub = services.NewHub(clusterPair, grpc.DialContext, commandExecer.Exec, conf, stubRemoteExecutor)
+		hub = services.NewHub(clusterPair, grpc.DialContext, commandExecer.Exec,
+			conf, stubRemoteExecutor, nil)
```
jchampio: This `nil` is still hanging around.

hub/services/hub_prepare_shutdown_clusters.go
jmcatamney: We put in the explicit `[2]` mostly for clarity, since the automatic handling of bare specifiers isn't immediately obvious if the reader didn't know about that.

hub/services/hub_prepare_start_agents_test.go
jmcatamney: Whoops!  Fixed this and the `status_conversion_test` one.

hub/services/hub_status_upgrade.go
jmcatamney: Just missed it last time.  It's fixed now.

hub/services/hub_status_upgrade.go
jmcatamney: It's less "unexpected RPC call" and more "we need a good way to identify when an RPC call fails because the agents are down/broken."  But yes, a bug or chore would make sense.  I've put a chore in the icebox to be prioritized.

jmcatamney:
@jchampio Addressed your follow-up comments.  How's everything look now?

jchampio:
@jmcatamney Looks good, thanks! I will follow up some time this week to replace the other `Step` names.

PR #11, opened 2018-06-15
Remove dead code

doty-pivotal:


integrations/prepare_start_hub_test.go
```diff
@@ -39,6 +39,7 @@ var _ = Describe("Start Hub", func() {
 	It("does not return an error if a non-gpupgrade_hub process with gpupgrade_hub in the name is running", func() {
 		path := filepath.Join(testWorkspaceDir, "gpupgrade_hub_test_log")
 		f, err := os.Create(path)
+		defer os.Remove(path)
```
jchampio: If the test workspace isn't being removed, something is wrong. The point of that directory was so that we didn't have to worry about cleanup in the tests. Where are you seeing this stick around?

jchampio: Whoops. `:=` shadowing is going to be my downfall, I can feel it...

Thanks for catching that!

PR #13, opened 2018-06-19
Refactor validate-start-cluster

jmcatamney:
This series of commits refactors validate-start-cluster to source greenplum_path.sh when running gpstart instead of using the current PYTHONPATH, as PYTHONPATH will differ based on whether the old or new cluster's PYTHONPATH is currently exported.

The main commits for this refactor are the "Remove flags" and "Refactor" commits; the "Modify tests" commit fixes the formerly-broken validate-start-cluster test so that this refactor could be tested, and the "Change directory names" commit is minor test cleanup added on since I was touching the bin directories in the "Remove flags" commit.

### DIFF
--- a/hub/services/hub_upgrade_validate_start_cluster_test.go
+++ b/hub/services/hub_upgrade_validate_start_cluster_test.go
@@ -69,8 +69,8 @@ var _ = Describe("upgrade validate start cluster", func() {
 
 		Eventually(func() bool { return cm.IsComplete("validate-start-cluster") }).Should(BeTrue())
 		Expect(testExecutor.NumExecutions).To(Equal(1))
-		Expect(testExecutor.LocalCommands[0]).To(ContainSubstring("PYTHONPATH="))
-		-Expect(testExecutor.LocalCommands[0]).To(ContainSubstring("&& bin/gpstart -a -d data"))
+		Expect(testExecutor.LocalCommands[0]).To(ContainSubstring("source /new/bindir/../greenplum_path.sh"))
+		Expect(testExecutor.LocalCommands[0]).To(ContainSubstring("/new/bindir/gpstart -a -d /new/datadir"))
 	})
 
 	It("sets status to FAILED when the validate start cluster request returns an error", func() {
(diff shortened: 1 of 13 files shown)


integrations/upgrade_convert_primaries_test.go
```diff
@@ -37,11 +37,6 @@ var _ = Describe("upgrade convert primaries", func() {
 
 	BeforeEach(func() {
 		var err error
-		oldBinDir, err = ioutil.TempDir("", "")
-		Expect(err).ToNot(HaveOccurred())
-		newBinDir, err = ioutil.TempDir("", "")
-		Expect(err).ToNot(HaveOccurred())
```
jchampio: Do the corresponding vars above need to be removed?

hub/services/hub_upgrade_validate_start_cluster_test.go
```diff
@@ -64,49 +62,16 @@ var _ = Describe("upgrade validate start cluster", func() {
 	})
 
 	//This test is flaking... It will be rewritten once the checklist manager refactor comes in.
```
jchampio: Looks like this can go away.

integrations/upgrade_convert_primaries_test.go
doty-pivotal: Looks like [updates status to FAILED if it fails to run](https://github.com/greenplum-db/gpupgrade/blob/6fddcabcba71ebeee545096e94a241eaa7028c9d/integrations/upgrade_convert_primaries_test.go#L138-L143): 

```
...
	It("updates status to FAILED if it fails to run", func() {
...
		upgradeConvertPrimaries := runCommand(
			"upgrade",
			"convert-primaries",
			"--old-bindir", oldBinDir,
			"--new-bindir", newBinDir,
		)
...
```
Is still using the vars.

However, they could be moved to be `:=`'ed in this one test if that would read better.

jchampio: Oh, good catch. So... it looks like nothing is setting them? Are we just passing empty strings as the arguments?

jmcatamney: Removed them.

hub/services/hub_upgrade_validate_start_cluster_test.go
jmcatamney: Removed it.

PR #16, opened 2018-06-20
Implement --daemonize for Agent and Hub

jchampio:
This patchset implements a `fork()`-less fork of a process. The system works as follows:
- The parent process is called with the `--daemonize` option, which signals it to fork a child.
- The child process is started with the `--daemon` option, which signals it to close its stdout/err streams before performing its main server loop.
- The parent, via `utils.Daemonize()`, echoes all stdout/err from the child, and waits for the streams to be disconnected before exiting and allowing the child to continue independently.
- If the child would rather exit with an error, it writes bytes to the stderr stream, and the parent's `Daemonize()` will wait on the child to obtain the exit code. (This won't wait forever; there's a timeout to prevent a stray write to stderr from hanging the call.) 

`--daemonize` has been implemented for both the hub and the agent.

The final two patches in the set are fixes to agent bugs encountered during this work.

### DIFF
--- a/agent/gpupgrade_agent_main.go
+++ b/agent/gpupgrade_agent_main.go
@@ -1,10 +1,13 @@
 package main
 
 import (
+	"fmt"
 	"os"
 	"os/exec"
 	"syscall"
 	"time"
+
+	"golang.org/x/crypto/ssh/terminal"
 
 	"github.com/greenplum-db/gp-common-go-libs/gplog"
 	"github.com/greenplum-db/gpupgrade/agent/services"
@@ -30,6 +33,11 @@ func main() {
 		Long:  `Start the Command Listener (blocks)`,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			gplog.InitializeLogging("gpupgrade_agent", logdir)
+
+			if daemon && terminal.IsTerminal(int(os.Stdout.Fd())) {
+				// Shouldn't be calling this from the command line.
+				return fmt.Errorf("--daemon is an internal option (did you mean --daemonize?)")
+			}
 
 			// TODO: this is all copy-pasted from the hub code. Consolidate!
 			if daemonize {
--- a/agent/gpupgrade_agent_main.go
+++ b/agent/gpupgrade_agent_main.go
@@ -3,6 +3,8 @@ package main
 import (
 	"os"
 	"os/exec"
+	"syscall"
+	"time"
 
 	"github.com/greenplum-db/gp-common-go-libs/gplog"
 	"github.com/greenplum-db/gpupgrade/agent/services"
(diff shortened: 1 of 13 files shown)


agent/gpupgrade_agent_main.go
```diff
@@ -46,8 +91,18 @@ func main() {
 
 	RootCmd.Flags().StringVar(&logdir, "log-directory", "", "command_listener log directory")
 	RootCmd.Flags().StringVar(&statedir, "state-directory", utils.GetStateDir(), "Agent state directory")
+	RootCmd.PersistentFlags().BoolVar(&daemonize, "daemonize", false, "start hub in the background")
+	RootCmd.PersistentFlags().BoolVar(&daemon, "daemon", false, "disconnect standard streams (internal option; use --daemonize instead)")
```
jmcatamney: `--daemonize` and `--daemon` are pretty easy to mix up, particularly if autocompletion is involved, and I don't know if just telling users a certain flag shouldn't be set manually is good enough to prevent them from doing so, accidentally or no.

Cobra has an option to make a subcommand non-user-visible by setting its Hidden field to `true`.  Maybe this should be set up as a hidden `gpupgrade_hub daemon` subcommand instead, so users don't know it's there and accidentally running `gpupgrade_hub --daemon` instead of `gpupgrade_hub --daemonize` won't do anything?

agent/gpupgrade_agent_main.go
```diff
@@ -46,8 +91,18 @@ func main() {
...
 
 	if err := RootCmd.Execute(); err != nil {
+		if gplog.GetLogger() == nil {
+			// In case we didn't get through RootCmd.Execute(), set up logging
+			// here. Otherwise we crash.
+			// XXX it'd be really nice to have a "ReinitializeLogging" building
```
jmcatamney: For reference, `gplog.SetLogger(nil)` will clear the existing logger and allow `InitializeLogging` to set a new one; that should probably be put in the comments somewhere.  Adding a `ReinitializeLogging` wrapper might be useful, though.

integrations/hub_test.go
```diff
@@ -0,0 +1,76 @@
...
+
+	Expect(err).ToNot(HaveOccurred())
+	session.Wait()
+
+	Expect(checkPortIsAvailable(port)).To(BeTrue())
+}
+
+var _ = Describe("gpupgrade_hub", func() {
+
+	// XXX We should be testing the locally built artifacts, and killing only
```
jmcatamney: Would it make sense to have the daemon return its PID to the parent process and display or log it somewhere, so that tests can kill individual hubs and users can identify daemons if they want to e.g. script around the upgrade process?

agent/gpupgrade_agent_main.go
skahler-pivotal: I agree this is makes me a bit uncomfortable having two very similar flags and one that is the partial name of the other. When we have an internal option is a way to make sure it is only executed internally?

jchampio: I'm not too wedded to `--daemon` as the magic switch. I was hoping that there was a way to set an _option_ to "hidden", but I didn't see it at a glance. I'd much rather hide it and not document it to the end user -- but to that point...

> I don't know if just telling users a certain flag shouldn't be set manually is good enough to prevent them from doing so

I guess I wasn't as concerned about this because I didn't feel that users would ever be running `gpupgrade_hub` or `gpupgrade_agent` themselves; I saw it as a support/development option. `prepare start-hub` and `prepare start-agent` are going to be our documented springboards, right?

> When we have an internal option is a way to make sure it is only executed internally?

There are a couple tricks we could do, yeah. Either a magic environment variable, or doing what some other command line utilities do on *nix and checking to see if we're being called from a TTY, and if so bailing out with an error.

integrations/hub_test.go
jchampio: If we want to do that I think it makes more sense to throw our effort behind a PID file lookup; that way you didn't have to be the one to start the daemon to know how to find it.

jchampio:
Latest commit puts us one step better; here's the help output now:
```
gpupgrade_hub --help
Start the gpupgrade_hub (blocks)

Usage:
  gpupgrade_hub [flags]

Flags:
      --daemonize              start hub in the background
  -h, --help                   help for gpupgrade_hub
      --log-directory string   gpupgrade_hub log directory
```
@skahler-pivotal The `--daemon` option can still be passed by accident, though. Would you like that to be fixed for acceptance of this story?

@jmcatamney FYI I'm dragging my feet on implementing a separate subcommand, because I'm hoping to keep the daemonization logic contained in a single Cobra building block for later ease of refactoring. (My desired end goal is to have the hub and agent inherit from a `Command` subclass that implements `--daemonize` entirely for the caller.) It's entirely possible I am overanalyzing. :smile:

jchampio:
You know what, I've just gone ahead and fixed it. 😄 You can no longer use `--daemon` from a terminal:
```
$ gpupgrade_hub --daemon
Error: --daemon is an internal option (did you mean --daemonize?)
Usage:
  gpupgrade_hub [flags]
  <snip>
```
Does this address the remaining concerns?

jmcatamney:
Yep, I think that should cover it.  I don't know if autocomplete would complete internal options; if it does, we might want to swap the flags so the first one to be completed is the one we want them to use, but if it doesn't then it's fine as-is.

agent/gpupgrade_agent_main.go
```diff
@@ -90,9 +90,10 @@ func main() {
 	}
 
 	RootCmd.Flags().StringVar(&logdir, "log-directory", "", "command_listener log directory")
-	RootCmd.Flags().StringVar(&statedir, "state-directory", utils.GetStateDir(), "Agent state directory")
```
jchampio: Whoops, lost this by accident. Will fix before pushing.

PR #18, opened 2018-06-25
Cluster_ssher refactor

kyeap-pivotal:
- Move Seginstall and prepare start agents methods off the cluster ssher
- Refactor to use the cluster execer on cluster_pair
- Update their respective unit and integration tests
- Remove cluster_ssher and update NewHub signature


hub/services/hub_check_seginstall_test.go
```diff
@@ -1,53 +1,53 @@
...
 
-	Describe("CheckSeginstall", func() {
-		It("returns a gRPC reply object, if the software verification gets underway asynch", func() {
-			_, err := hub.CheckSeginstall(nil, &pb.CheckSeginstallRequest{})
-			Expect(err).ToNot(HaveOccurred())
-			Eventually(stubRemoteExecutor.VerifySoftwareHosts).Should(Receive(Equal([]string{"hostone"})))
-		})
+	It("shells out to cluster and verifies gpupgrade_agent is installed on master and hosts", func() {
+		cp.OldCluster = testutils.CreateMultinodeSampleCluster()
+		testExecutorOld := &testhelper.TestExecutor{}
```
jmcatamney: Could we call this `testExecutor` like in the other tests?  If I see a `testExecutorOld` I would expect to also see a `testExecutorNew` used for something else, and most of our commands use either the old or new cluster but not both.

hub/services/hub_prepare_start_agents_test.go
```diff
@@ -1,56 +1,53 @@
...
+		testExecutorOld.ClusterOutput = &cluster.RemoteOutput{0, 0, nil, nil, nil, nil}
+		cp.OldCluster.Executor = testExecutorOld
 
+		services.StartAgents(cp, cm)
+
+		Expect(testExecutorOld.NumExecutions).To(Equal(1))
+
+		startAgentsCmd := fmt.Sprintf("%s/bin/gpupgrade_agent --daemonize", os.Getenv("GPHOME"))
+		expectedCmdMap := make(map[int]string, 0)
+		expectedCmdMap[-1] = "bash -c " + startAgentsCmd
```
jmcatamney: See previous comment on validating commands.

kyeap-pivotal:
Updated based on Jamie's comments

hub/services/hub_check_seginstall.go
```diff
@@ -2,17 +2,71 @@ package services
...
+		return
+
+	fail:
+		err = h.checklistWriter.MarkFailed(upgradestatus.SEGINSTALL)
+		if err != nil {
+			gplog.Error(err.Error())
+		}
+	}()
+
+	// go h.remoteExecutor.VerifySoftware(h.clusterPair.GetHostnames())
```
jchampio: This can go away.

hub/services/hub_check_seginstall.go
```diff
@@ -14,59 +17,50 @@ import (
...
+func VerifyAgentsInstalled(cp *ClusterPair, cw cluster_ssher.ChecklistWriter) {
+	err := cw.ResetStateDir(upgradestatus.SEGINSTALL)
+	if err != nil {
+		gplog.Error(err.Error())
 		return
+	}
 
-	fail:
-		err = h.checklistWriter.MarkFailed(upgradestatus.SEGINSTALL)
+	err = cw.MarkInProgress(upgradestatus.SEGINSTALL)
```
jchampio: Seems like the reset/in-progress marks can happen before we spin up a goroutine. That way there are no races -- when we return from `CheckSeginstall()`, we're guaranteed to be in progress for whomever is watching.

hub/services/hub_prepare_start_agents.go
```diff
@@ -11,7 +18,47 @@ import (
...
 }
+
+func StartAgents(cp *ClusterPair, cw cluster_ssher.ChecklistWriter) {
+	err := cw.ResetStateDir(upgradestatus.START_AGENTS)
+	if err != nil {
+		gplog.Error(err.Error())
+		return
+	}
+
+	err = cw.MarkInProgress(upgradestatus.START_AGENTS)
```
jchampio: Ditto.

Gopkg.lock
```diff
@@ -248,6 +254,6 @@
 [solve-meta]
   analyzer-name = "dep"
   analyzer-version = 1
-  inputs-digest = "d4d488dd285feaaee3212fdfee085bf0d580a8022f6de9d6feec1771259d97a4"
+  inputs-digest = "07ef036d5e006d6872fe589ad973df85a76ab9ea9226603df65c2d426aa78848"
```
jchampio: Sorry about that -- I forgot to check in the lockfile with my `x/crypto` addition. It's fixed on master now and should go away with a rebase.

hub/services/hub_check_seginstall.go
```diff
@@ -12,7 +17,49 @@ import (
...
+	if err != nil {
+		gplog.Error(err.Error())
+		return
+	}
+
+	err = cw.MarkInProgress(upgradestatus.SEGINSTALL)
+	if err != nil {
+		gplog.Error(err.Error())
+		return
+	}
```
jchampio: Seems like we can move the reset/in-progress markings outside the goroutine. That way there are no races -- when we return from `CheckSeginstall()`, all watchers see the step as in-progress.

hub/services/hub_prepare_start_agents.go
jchampio: Seems like GitHub keeps reordering my review comments... This was referencing the comment below, with the goroutine and the races.

PR #19, opened 2018-06-27
Refactor hub unit tests

kyeap-pivotal:
Hub unit tests are slow because they are waiting for the grpc dialer to timeout (3 seconds each test).

Changes:
 - Mock grpcDialer for faster hub unit tests
 - Remove redundant tests
 - Move method EnsureConnsAreReady off the hub
 - Refactor EnsureConnsAreReady

I am unsure about the change in the second commit that removes the retries in EnsureConnsAreReady. I would not be opposed to reverting it. Reverting would add 1.5 seconds back into the hub unit tests.

hub unit tests run in ~130ms down from 16 seconds

### DIFF
--- a/hub/services/hub_test.go
+++ b/hub/services/hub_test.go
@@ -1,9 +1,11 @@
 package services_test
 
 import (
-	"github.com/greenplum-db/gp-common-go-libs/cluster"
+	"errors"
+
 	"github.com/greenplum-db/gpupgrade/hub/services"
 	"github.com/greenplum-db/gpupgrade/testutils"
+	"golang.org/x/net/context"
 
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/connectivity"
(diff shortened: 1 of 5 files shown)


hub/services/hub_test.go
```diff
@@ -53,8 +55,7 @@ var _ = Describe("Hub", func() {
 		Eventually(func() connectivity.State { return conns[0].Conn.GetState() }).Should(Equal(connectivity.Shutdown))
 	})
 
-	It("retrieves the agent connections from the config file reader", func() {
-		clusterPair.OldCluster.Segments[1] = cluster.SegConfig{Hostname: "localhost"}
+	It("retrieves the agent connections from the cluster file reader", func() {
```
jmcatamney: `cluster file reader` doesn't make much sense here.  Something like `retrieves the agent connections for the hosts in the cluster` would be better.

jmcatamney:
Regarding the `EnsureConnsAreReady` time, as far as I know "3 retries at 0.5 seconds apiece" was an arbitrary choice.  It should be fine to remove that for now and put back some retry logic if we start seeing connection issues in our end-to-end tests.

kyeap-pivotal:
Changed the test name based on Jamie's comment.

Before: Hub Services Suite - 58/59 specs... SUCCESS! 16.654310432s PASS

After:    Hub Services Suite - 55/56 specs... SUCCESS! 112.431613ms   PASS

All of the slow downs were from tests either waiting for grpc timeouts or retry timeout loops. In the method `h.AgentConns()` and its helper function `EnsureConnsAreReady()`

PR #20, opened 2018-06-29
Fix status reporting for running pg_upgrade on master and primaries

nadeemg:
The code for reporting status on `convert-primaries` was named badly and was using constructs defined for `convert-master`.This PR renames some things and reports status correctly. 

While here also fix a test and move ClusterPair out to the utils directory.


idl/cli_to_hub.pb.go
```diff
@@ -1261,97 +1245,97 @@ func (m *PrepareInitClusterReply) XXX_DiscardUnknown() {
...
-type UpgradeConvertMasterRequest struct {
-	OldBinDir            string   `protobuf:"bytes,1,opt,name=OldBinDir" json:"OldBinDir,omitempty"`
-	OldDataDir           string   `protobuf:"bytes,2,opt,name=OldDataDir" json:"OldDataDir,omitempty"`
-	NewBinDir            string   `protobuf:"bytes,3,opt,name=NewBinDir" json:"NewBinDir,omitempty"`
-	NewDataDir           string   `protobuf:"bytes,4,opt,name=NewDataDir" json:"NewDataDir,omitempty"`
+type UpgradeConvertSegmentRequest struct {
+	OldBinDir            string   `protobuf:"bytes,1,opt,name=OldBinDir,proto3" json:"OldBinDir,omitempty"`
+	OldDataDir           string   `protobuf:"bytes,2,opt,name=OldDataDir,proto3" json:"OldDataDir,omitempty"`
+	NewBinDir            string   `protobuf:"bytes,3,opt,name=NewBinDir,proto3" json:"NewBinDir,omitempty"`
+	NewDataDir           string   `protobuf:"bytes,4,opt,name=NewDataDir,proto3" json:"NewDataDir,omitempty"`
```
jchampio: Is there anything we should be watching out for in the autogenerated pieces? This `proto3` change, for instance?

nadeemg: Good catch. Seems due to this (commit)[https://github.com/golang/protobuf/commit/05f48f4eaf0e05663b562bab533cdd472238ce29]. Since we've been on `proto3` all along, doesn't seem like a big deal

nadeemg:
> I feel like I am missing something here (the why behind the rename of master->segment). The CLI command convert-master calls the subConvertSegment, which calls the commanders.NewUpgrader(client).ConvertSegment(...), which calls the u.client.UpgradeConvertSegment(...) which reaches out to the hub func (h *Hub) UpgradeConvertSegment(...) which calls the convertMaster(...) function which only converts the master. All the primaries are converted by the convertPrimaries code path, which reaches out to the agents. Is this refactor preparation for consolidating those?

The rename is just meant to address status reporting. `status upgrade` was using `upgrader.ConvertMaster` to report status of  `UpgradeSteps_MASTERUPGRADE` and `UpgradeSteps_CONVERSION`.

But you've caught where a global search-and-replace went too far. The rename in `cli_to_hub.proto` is unintentional, and I'll undo that.

hub/upgradestatus/pg_upgrade_status_checker.go
```diff
@@ -34,42 +44,40 @@ func NewPGUpgradeStatusChecker(pgUpgradePath, oldDataDir string, execer helpers.
 	- pg_upgrade will not fail without error before writing an inprogress file
 	- when a new pg_upgrade is started it deletes all *.done and *.inprogress files
 */
-func (c *ConvertMaster) GetStatus() *pb.UpgradeStepStatus {
-	var masterUpgradeStatus *pb.UpgradeStepStatus
+func (c *ConvertSegment) GetStatus() *pb.UpgradeStepStatus {
+	//make default reply object
+	masterUpgradeStatus := &pb.UpgradeStepStatus{
```
kyeap-pivotal: Doesn't make sense anymore to call this masterUpgradeStatus anymore.

hub/upgradestatus/pg_upgrade_status_checker.go
```diff
@@ -34,42 +44,40 @@ func NewPGUpgradeStatusChecker(pgUpgradePath, oldDataDir string, execer helpers.
...
+		Status: pb.StepStatus_FAILED,
+	}
+	//update Step in reply object to match the conversion
+	if c.segType == MASTER {
+		masterUpgradeStatus.Step = pb.UpgradeSteps_MASTERUPGRADE
+	} else if c.segType == PRIMARY {
+		masterUpgradeStatus.Step = pb.UpgradeSteps_CONVERSION
```
kyeap-pivotal: So... It looks like we have an extra step in cli_to_hub.proto. The `CONVERSION` step should not exist at all and should be the updating the `CONVERT_PRIMARIES` step.

hub/upgradestatus/pg_upgrade_status_checker.go
nadeemg: Agreed!

PR #22, opened 2018-07-03
Introduce centralized state management with Checklist

jchampio:
Step state handling for `status upgrade` -- both reading and writing -- are now encapsulated in the top-level `Checklist`. Here are the key things that gives us:

- Since steps are now associated with their ProtoBuf code in the `Checklist`, we don't need to pass the codes around when fetching status. This lets us remove plenty of boilerplate code, as well as a couple of helper structs that didn't really serve any purpose except to remember which step we were on.
- State reading and writing are pulled into separate interfaces that encapsulate each operation. The API that writes state to the filesystem no longer requires callers to know the name or any other details of the step they're writing to; they just write to a step. Same for readers (see for instance the simplification of `StatusUpgrade`'s implementation).
- Centralization of state mocking during tests has flushed out coverage gaps and other deficiencies.

There's still plenty of work to do, but this should be a good step forward so we see what we like about the new interface and what we don't.

### DIFF
--- a/hub/services/hub_prepare_init_cluster.go
+++ b/hub/services/hub_prepare_init_cluster.go
@@ -37,7 +37,7 @@ func SaveTargetClusterConfig(clusterPair *utils.ClusterPair, dbConnector *dbconn
 func (h *Hub) PrepareInitCluster(ctx context.Context, in *pb.PrepareInitClusterRequest) (*pb.PrepareInitClusterReply, error) {
 	gplog.Info("starting PrepareInitCluster()")
 
-	step := h.checklist.StepWriter(upgradestatus.INIT_CLUSTER)
+	step := h.checklist.GetStepWriter(upgradestatus.INIT_CLUSTER)
 	step.MarkInProgress()
 
 	dbConnector := db.NewDBConn("localhost", int(in.DbPort), "template1")
--- a/hub/services/hub_prepare_init_cluster.go
+++ b/hub/services/hub_prepare_init_cluster.go
@@ -37,14 +37,15 @@ func SaveTargetClusterConfig(clusterPair *utils.ClusterPair, dbConnector *dbconn
 func (h *Hub) PrepareInitCluster(ctx context.Context, in *pb.PrepareInitClusterRequest) (*pb.PrepareInitClusterReply, error) {
 	gplog.Info("starting PrepareInitCluster()")
 
-	h.checklistWriter.MarkInProgress(upgradestatus.INIT_CLUSTER)
+	step := h.checklistWriter.StepWriter(upgradestatus.INIT_CLUSTER)
+	step.MarkInProgress()
 
 	dbConnector := db.NewDBConn("localhost", int(in.DbPort), "template1")
 	defer dbConnector.Close()
 	err := dbConnector.Connect(1)
 	if err != nil {
 		gplog.Error(err.Error())
-		h.checklistWriter.MarkFailed(upgradestatus.INIT_CLUSTER)
+		step.MarkFailed()
 		return &pb.PrepareInitClusterReply{}, utils.DatabaseConnectionError{Parent: err}
 	}
 	dbConnector.Version.Initialize(dbConnector)
@@ -52,10 +53,10 @@ func (h *Hub) PrepareInitCluster(ctx context.Context, in *pb.PrepareInitClusterR
 	err = SaveTargetClusterConfig(h.clusterPair, dbConnector, h.conf.StateDir, in.NewBinDir)
 	if err != nil {
 		gplog.Error(err.Error())
-		h.checklistWriter.MarkFailed(upgradestatus.INIT_CLUSTER)
+		step.MarkFailed()
 		return &pb.PrepareInitClusterReply{}, err
 	}
 
-	h.checklistWriter.MarkComplete(upgradestatus.INIT_CLUSTER)
+	step.MarkComplete()
 	return &pb.PrepareInitClusterReply{}, nil
 }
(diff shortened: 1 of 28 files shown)


testutils/mock_checklist_manager.go
```diff
@@ -56,11 +83,14 @@ func (cm *MockChecklistManager) IsPending(step string) bool {
 }
 
 // Check that the step was running and is now complete
+// XXX Why must we check that inprogress was set? Is that part of the actual
```
jmcatamney: Yes, we want to make sure all of the steps mark things in progress once started so that anyone running `gpupgrade status upgrade` sees a `RUNNING` status entry; forgetting to mark that would leave it in `PENDING` and make the user think it's not working.

jchampio: Gotcha. Maybe a `WasRunning` might help with intent? I'll meditate on it some more...

hub/services/hub_prepare_init_cluster.go
```diff
@@ -37,25 +37,26 @@ func SaveTargetClusterConfig(clusterPair *utils.ClusterPair, dbConnector *dbconn
 func (h *Hub) PrepareInitCluster(ctx context.Context, in *pb.PrepareInitClusterRequest) (*pb.PrepareInitClusterReply, error) {
 	gplog.Info("starting PrepareInitCluster()")
 
-	h.checklistWriter.MarkInProgress(upgradestatus.INIT_CLUSTER)
+	step := h.checklistWriter.StepWriter(upgradestatus.INIT_CLUSTER)
+	step.MarkInProgress()
```
nadeemg: do we need to `step.ResetStateDir` before this?

hub/upgradestatus/checklist_manager.go
```diff
@@ -23,67 +22,84 @@ const (
...
 		pathToStateDir: stateDirPath,
-		inProgress:     "in.progress",
-		failed:         "failed",
-		completed:      "completed",
 	}
 }
 
-func (c *ChecklistManager) MarkFailed(step string) error {
-	err := utils.System.Remove(filepath.Join(c.pathToStateDir, step, c.inProgress))
+func (c *ChecklistManager) StepWriter(step string) StateWriter {
```
nadeemg: Is this func name confusing? Since there is a struct below, also called StepWriter, I'd be tempted to rename this.

hub/upgradestatus/checklist_manager.go
```diff
@@ -23,67 +22,84 @@ const (
 	RECONFIGURE_PORTS      = "reconfigure-ports"
 )
 
+const (
```
nadeemg: If this block was exported, we could remove some hard-coded strings from `checklist_manager_test`

testutils/mock_checklist_manager.go
jmcatamney: If you mean calling a `WasRunning` helper function in `IsComplete`/`IsFailed` instead of directly referencing `mapInProgress`, sure, I don't have strong opinions on the implementations.

hub/upgradestatus/checklist_manager.go
jchampio: My name-fu was broken that day. I'm not a fan of the name `StepWriter` either, so if you have any suggestions I'm all ears.

hub/upgradestatus/checklist_manager.go
jchampio: The constants are exported, so that's already possible. That's coming in a follow-up commit @kyeap-pivotal made.

jmcatamney: Agreed.  Making these all-caps exported variables to match the step names above would make sense.

hub/upgradestatus/checklist_manager.go
nadeemg: My name-fu is still MIA, but `NewStateWriter`?

hub/upgradestatus/checklist_manager.go
jchampio: Is it okay to deal with the progress-file-naming in a different chore? We have different steps looking for different names: `in.progress`, `.inprogress`, etc., and I'm not sure how far down the rabbit hole I want to go with this one.

jmcatamney: Since this PR is blocking another story and fixing this could take a while, I'm fine with addressing this later.  I'll make a chore for it.

testutils/mock_checklist_manager.go
jchampio: I mean that `WasRunning` would be a separate method on the mock; some tests care that the steps were called in order, and some tests don't.

jmcatamney: Works for me.

hub/upgradestatus/checklist_manager.go
```diff
@@ -35,13 +36,38 @@ type StateWriter interface {
...
+
 type ChecklistManager struct {
 	pathToStateDir string
+	codemap        map[string]pb.UpgradeSteps // maps step name to step code
 }
 
 func NewChecklistManager(stateDirPath string) *ChecklistManager {
 	return &ChecklistManager{
 		pathToStateDir: stateDirPath,
+		codemap: map[string]pb.UpgradeSteps{
```
nadeemg: Another thing we could look at later, maybe this map literal could be an exported constant?

jchampio: `codemap` is remove in a later commit; it's unnecessary now that we've got `LoadSteps`.

testutils/mock_checklist_manager.go
```diff
@@ -56,11 +83,14 @@ func (cm *MockChecklistManager) IsPending(step string) bool {
...
+// "API" guarantee?
```
nadeemg: 👍

hub/upgradestatus/checklist_manager.go
jchampio: `NewXxx` seems to be idiomatic for constructors, and I don't think we want to guarantee that a new object is created here. As much as I hate getter prefixes, maybe `GetStepReader` and `GetStepWriter` is the way to go?

jchampio:
Rerolled on top of master, and renamed the `Checklist` methods per feedback.

PR #23, opened 2018-07-03
Move daemonization code into its own package

jchampio:
`MakeDaemonizable` is the key function here. Here's how you use it (from the new godoc):

```
var shouldDaemonize bool
cmd := &cobra.Command{
		err := listenOnServerPort()
		if err != nil {
			// Cobra will write an error message to stderr for us.
			return err
		}

		if shouldDaemonize {
			fmt.Println("daemonized successfully")
			daemon.Daemonize() // disconnect standard streams
		}

		return serverMainLoop()
	},
}

daemon.MakeDaemonizable(cmd, &shouldDaemonize) // add --daemonize

err := cmd.Execute()
if err != nil {
	if err == daemon.SuccessfullyDaemonized {
		// not actually an "error", just exit cleanly
	} else {
		// handle actual error
		...
	}
}
```


utils/daemon/daemon.go
```diff
@@ -0,0 +1,412 @@
...
+	cmd := &cobra.Command{
+		RunE: func(cmd *cobra.Command, args []string) error {
+			err := listenOnServerPort()
+			if err != nil {
+				// Cobra will write an error message to stderr for us.
+				return err
+			}
+
+			if shouldDaemonize {
+				fmt.Println("daemonized successfully")
```
jmcatamney: It's a bit of a nitpick, but the print statement should probably be moved down to the `if err == daemon.SuccessfullyDaemonized` block, since it hasn't succeeded yet.

jchampio: That code only runs in the child, and from the child's point of view, that *is* success. The parent just doesn't know it yet. But that is an important distinction, yeah -- when you're writing your code, you need to make sure that there's no way for the two processes to disagree.

jchampio: I'm staring at what I just wrote, and what the code says, and... I don't think it makes sense to anyone but me. Let me improve the message being printed :smile:

jchampio:
@jmcatamney How's that?

PR #25, opened 2018-07-06
Unify integration test initialization

jmcatamney:
Previously, every integration test file performed its own setup and teardown in BeforeEach and AfterEach, most of which was just copy-pasted between test files, so any refactor touching the hub or agent would generally involve making the same change in lots of files.

This series of commits moves all of the duplicated logic to the BeforeEach and AfterEach in integration_suite_test.go, so that the initialization happens only once and refactors only need to make a given change once.

There are no observed test conflicts due to this change, and no test slowdowns due to performing unnecessary initialization for some tests.

### DIFF
--- a/integrations/integrations_suite_test.go
+++ b/integrations/integrations_suite_test.go
@@ -11,11 +11,17 @@ import (
 	"testing"
 	"time"
 
+	"github.com/greenplum-db/gp-common-go-libs/cluster"
 	"github.com/greenplum-db/gp-common-go-libs/testhelper"
+	agentServices "github.com/greenplum-db/gpupgrade/agent/services"
+	"github.com/greenplum-db/gpupgrade/hub/services"
+	"github.com/greenplum-db/gpupgrade/testutils"
+	"github.com/greenplum-db/gpupgrade/utils"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	"github.com/onsi/gomega/gbytes"
 	. "github.com/onsi/gomega/gexec"
+	"google.golang.org/grpc"
 )
 
 func TestCommands(t *testing.T) {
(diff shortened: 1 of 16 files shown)


integrations/integrations_suite_test.go
```diff
@@ -118,10 +121,19 @@ var _ = BeforeEach(func() {
...
 	hub = services.NewHub(cp, grpc.DialContext, conf, cm)
 	go hub.Start()
+
+	agentConfig := agentServices.AgentConfig{
+		Port:     hubToAgentPort,
+		StateDir: testStateDir,
+	}
+	agentExecutor = &testhelper.TestExecutor{}
+	agent = agentServices.NewAgentServer(agentExecutor, agentConfig)
+	go agent.Start()
```
jchampio: I'm kind of -0.5 on this change; it won't hurt properly written tests and implementations, sure, but it might mask cases where the code being tested accidentally relies on the agent in situations where it shouldn't. We're seeing that already with some magic envvars, and I'm hoping to move in the opposite direction. Curious to see what other people think.

nadeemg: @jchampio would it be better if we separated tests that relied on just a hub from tests that relied on hub and agent?

jchampio: Depends on what you mean by "separate" -- I think separating them explicitly with additional `BeforeEach` logic, as we do currently, is good enough for this case.

kyeap-pivotal: I don't mind keeping the agent declaration and the configuration of agent in here. I would say move the one line `go agent.Start()` back into separate `BeforeEach`s. It's only being used in 3 files now, and should eventually be removed as `share_oid`, `convert_primaries`, and `conversionStatus` become more unit testable.

jchampio: @kyeap-pivotal I like that; +1.

integrations/integrations_suite_test.go
```diff
@@ -76,13 +94,50 @@ var _ = BeforeEach(func() {
...
+	Expect(err).ToNot(HaveOccurred())
+
+	conf := &services.HubConfig{
+		CliToHubPort:   cliToHubPort,
+		HubToAgentPort: hubToAgentPort,
+		StateDir:       testStateDir,
+	}
+
+	cm = testutils.NewMockChecklistManager()
+	cp = testutils.InitClusterPairFromDB()
```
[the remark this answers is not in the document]
jmcatamney: None of the tests actually care whether they init from a real database or just use a sample cluster constructed from `testutils.CreateMultinodeSampleCluster`.  I can change it to use that, if you'd like.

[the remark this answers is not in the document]
jchampio: Oh, sweet. In that case, yeah, let's break the dependency there.

integrations/integrations_suite_test.go
jmcatamney: I don't mind pulling the agent logic out and sticking it in a `InitAndStartAgent` function or the like, so individual tests can choose whether to use an agent or not but there's no agent logic duplication.  How's that sound?

jchampio: Seems reasonable to me (though `go agent.Start()` might already be enough of an "interface" in this case).

jmcatamney: Didn't refresh the page and see Kevin's comment.  Moving the `agent.Start` call works too.

jmcatamney:
@jchampio @kyeap-pivotal Suggested changes are pushed, take a look.

PR #26, opened 2018-07-06
Export filename constants (and use them)

jchampio:
I pulled the constants into a subpackage purely for the readability of the code; `file.InProgress` felt better to me than `upgradestatus.InProgressFile` or whatever. Maybe we can rename `upgradestatus` to something more happy in the future, and lose the subpackage.


jmcatamney:
I'd be in favor of changing `upgradestatus` to just `status`; `status.InProgressFile` isn't too bad, I think.

PR #28, opened 2018-07-06
Run `dep ensure`, if necessary, during build

jchampio:
Track the last time we ran `dep ensure` with a hidden marker file. If the Gopkg files have changed since then, there's a good chance we need to grab new dependencies.

This isn't perfect -- you can still fool the system by touching files and modifying timestamps -- but the cost of failure here is low.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 vendor
 .idea
+.Gopkg.updated
--- a/Makefile
+++ b/Makefile
@@ -80,7 +80,7 @@ agent-package: EXE_NAME := $(AGENT)
 cli-package: EXE_NAME := $(CLI)
 hub-package: EXE_NAME := $(HUB)
 
-$(PACKAGES): %-package:
+$(PACKAGES): %-package: .Gopkg.updated
 	$(PREFIX) go build $(GOFLAGS) -o $(BIN_DIR)/$(EXE_NAME)$(POSTFIX) -ldflags $(UPGRADE_VERSION_STR) github.com/greenplum-db/gpupgrade/$*
 
 install_agent: agent-package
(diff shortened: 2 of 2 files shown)


jchampio:
@doty-pivotal Is that behavior guaranteed? If files are modified but not added/removed, I thought `mtime` remained constant. But yeah, using `vendor` as the target would be ideal, if it works.

jchampio:
https://github.com/golang/dep/issues/121 is somewhat illuminating:

> without a verification mechanism, the only safe choice dep can make is to drop and recreate vendor in its entirety on every operation. This is wasteful and slow; any operation that avoids unnecessary disk writes to determine if vendor needs updating will be faster, and would be a huge boon.

So we could rely on it *now*, but I'm not sure we can continue to do so. Thoughts?

doty-pivotal:
Some digging, 

```
pushd /tmp
  mkdir vendor
  mtime=`stat -f %m /tmp/vendor`

  sleep 2
  touch vendor/fileA
  _new_mtime=$(stat -f %m /tmp/vendor)
  if [[ ${mtime} < ${_new_mtime} ]];then
    echo "add file updated mtime"
    mtime=`stat -f %m /tmp/vendor`
  else
    echo "add file did not update mtime"
  fi

  sleep 2
  echo "Mod File" > vendor/fileA 
  _new_mtime=$(stat -f %m /tmp/vendor)
  if [[ ${mtime} < ${_new_mtime} ]];then
    echo "modified file updated mtime"
    mtime=`stat -f %m /tmp/vendor`
  else
    echo "modified file did not update mtime"
  fi

  sleep 2
  rm vendor/fileA
  _new_mtime=$(stat -f %m /tmp/vendor)
  if [[ ${mtime} < ${_new_mtime} ]];then
    echo "rm updated mtime"
    mtime=`stat -f %m /tmp/vendor`
  else
    echo "rm did not update mtime"
  fi

  # Clean up
  rmdir vendor
  rm -rf /tmp/vendor
popd
```

yields: 

```
/tmp /tmp
add file updated mtime
modified file did not update mtime
rm updated mtime
/tmp
```

PR #31, opened 2018-07-11
Create a new cluster as part of prepare init-cluster

jmcatamney:
Previously, the prepare init-cluster step assumed that a new cluster was already created and running.  This commit generates a gpinitsystem config file from the old cluster and runs gpinitsystem to create a new cluster.


hub/services/hub_status_upgrade.go
```diff
@@ -34,4 +33,4 @@ func GetPrepareNewClusterConfigStatus(statedir string) pb.StepStatus {
 	}
 
 	return pb.StepStatus_COMPLETE
-}
+}*/
```
slari-pivotal: This function is commented out.  Shouldn't we delete it?

jmcatamney: Yep, just missed that.  We'll remove it.

hub/services/hub_prepare_init_cluster.go
```diff
@@ -35,28 +40,216 @@ func SaveTargetClusterConfig(clusterPair *utils.ClusterPair, dbConnector *dbconn
...
+}
 
-	dbConnector := db.NewDBConn("localhost", int(in.DbPort), "template1")
+func (h *Hub) InitCluster(dbConnector *dbconn.DBConn) {
 	defer dbConnector.Close()
-	err := dbConnector.Connect(1)
+	step := h.checklist.GetStepWriter(upgradestatus.INIT_CLUSTER)
+	err := step.ResetStateDir()
 	if err != nil {
 		gplog.Error(err.Error())
```
doty-pivotal: Should there be a `step.MarkFailed()`here?

hub/services/hub_prepare_init_cluster.go
```diff
@@ -35,28 +40,216 @@ func SaveTargetClusterConfig(clusterPair *utils.ClusterPair, dbConnector *dbconn
...
+		return
+	}
+
+	err = step.MarkInProgress()
+	if err != nil {
+		gplog.Error(err.Error())
```
doty-pivotal: Same question - Although here if we can't `MarkInProgress`, it seems likely that we wouldn't be able to `MarkFailed`

hub/services/hub_prepare_init_cluster.go
```diff
@@ -35,28 +40,216 @@ func SaveTargetClusterConfig(clusterPair *utils.ClusterPair, dbConnector *dbconn
...
+	_, err = utils.System.Stat(newMasterDataDir)
+	if os.IsNotExist(err) {
+		err = os.Mkdir(newMasterDataDir, 0755)
+		if err != nil {
+			gplog.Error("Could not create new master data directory: %v", err)
+			step.MarkFailed()
+			return
+		}
+	} else if err != nil {
+		gplog.Error("Error statting new master data directory %s: %s",
```
doty-pivotal: I am curious what folks tend to do here stat<i>t</i>{ed,ing} or  stat<i>'</i>ed,'ing?

hub/services/hub_prepare_init_cluster.go
```diff
@@ -35,28 +40,216 @@ func SaveTargetClusterConfig(clusterPair *utils.ClusterPair, dbConnector *dbconn
...
+		gplog.Error("Could not create segment data directories : %v", err)
+		step.MarkFailed()
+		return
+	}
+
+	// gpinitsystem the new cluster
+	cmdStr := fmt.Sprintf("gpinitsystem -a -c %s -h %s", gpinitsystemFilepath, hostnameFilepath)
+	output, err := h.clusterPair.OldCluster.Executor.ExecuteLocalCommand(cmdStr)
+	if err != nil {
+		// gpinitsystem has a return code of 1 for warnings, so we can ignore that return code
```
doty-pivotal: Would it be ok to log the fact that there was a warning?

doty-pivotal:
New question - Should we inspect and log `err`'s from the `step.Mark*()` functions?

utils/cluster_pair.go
```diff
@@ -138,3 +138,33 @@ func (cp *ClusterPair) GetHostnames() []string {
 	}
 	return hostnames
 }
+
+// FIXME:
```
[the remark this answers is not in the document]
jmcatamney: Maybe?  They only work with the old cluster because they're only used for creating the new cluster; I don't know if a new cluster version of them would be used anywhere.

hub/services/hub_prepare_init_cluster.go
jmcatamney: That was our thought for this and ResetStateDir; whatever error would prevent these from working would probably prevent MarkFailed, but we can attempt it for ResetStateDir.

hub/services/hub_prepare_init_cluster.go
jmcatamney: Couldn't hurt.

hub/services/hub_prepare_init_cluster.go
jmcatamney: I prefer two `t`s, but have no strong opinions as to what we use in messages.

doty-pivotal: Not super authoritative but https://en.wiktionary.org/wiki/stat#English seems to agree.  Grammar is surprisingly difficult to Google.

jmcatamney:
@doty-pivotal Probably not as part of this story, since it would involve touching all of the other commands as well, but a chore for that would be good.

agent/services/create_data_dirs.go
```diff
@@ -14,9 +15,19 @@ func (s *AgentServer) CreateSegmentDataDirectories(ctx context.Context, in *pb.C
 
 	datadirs := in.Datadirs
 	for _, segDataDir := range datadirs {
-		err := os.Mkdir(segDataDir, 0755)
-		if err != nil {
+		_, err := utils.System.Stat(segDataDir)
+		if os.IsNotExist(err) {
+			err = os.Mkdir(segDataDir, 0755)
+			gplog.Info("Successfully created directory %s", segDataDir)
```
nadeemg: Can we move this line after the err check?

kyeap-pivotal:
For such a large story, it feels odd that there is only one integration test. It looks like there could be unit tests for `CreateSegmentDataDirectories`, `getMasterSegPrefix`, `GetOldSegmentDataDirs`, the building of `gpinitsystemConfig` and eventually also the population of `portbase` . Are we considering the logic of these functions too simple to yield useful unit tests? 

nit: We also have two functions/methods names`CreateSegmentDataDirectories`. One helper function on the hub and one as a method on the agents.

hub/services/hub_prepare_init_cluster.go
```diff
@@ -60,3 +169,52 @@ func (h *Hub) PrepareInitCluster(ctx context.Context, in *pb.PrepareInitClusterR
 	step.MarkComplete()
 	return &pb.PrepareInitClusterReply{}, nil
 }
+
+func getMasterSegPrefix(datadir string) (string, error) {
```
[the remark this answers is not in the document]
jchampio: Yeah, we did.

hub/services/hub_prepare_init_cluster.go
jchampio: I think many projects sidestep this question :smile:
https://www.google.com/search?q=failed+to+stat
https://www.google.com/search?q=could+not+stat

jmcatamney:
@kyeap-pivotal I kind of assumed more unit tests were added in the period before I hopped back on the story, and didn't look too closely to see what we had there since we were mostly focused on getting the integration test working.  I definitely agree that more unit tests would be good, and we can work on adding those today.

hub/services/hub_prepare_init_cluster.go
```diff
@@ -35,28 +40,219 @@ func SaveTargetClusterConfig(clusterPair *utils.ClusterPair, dbConnector *dbconn
...
+
+	wg.Wait()
+	close(errChan)
+
+	for err := range errChan {
+		if err != nil {
+			return errors.Errorf("Error creating segment data directories: %s",
+				err.Error())
+		}
+	}
```
jchampio: Thought for later -- we should be able to multiplex errors somehow. Right now we will lose any segment errors past the first one, when we really want to bubble them all up to the user. (This is a problem with Greenplum too.)

Looks like other people have a similar need; Hashicorp has a `multierror` package, for instance.

hub/services/hub_prepare_init_cluster.go
```diff
@@ -50,162 +51,195 @@ func (h *Hub) PrepareInitCluster(ctx context.Context, in *pb.PrepareInitClusterR
...
+	agentConns := []*Connection{}
+	gpinitsystemFilepath := filepath.Join(h.conf.StateDir, "gpinitsystem_config")
+	hostnameFilepath := filepath.Join(h.conf.StateDir, "hostfile")
+
+	/*
+	 * To avoid boilerplate error handling after each function call, we create
+	 * helper functions for each logical unit, pass in all arguments by reference,
+	 * return an error, and rely on left-associative OR operators to exit the flow
+	 * at the correct point.
+	 */
```
jchampio: How about returning an error directly from `InitCluster` instead? Then we can move the "error wrapping" code into the `go` call in `PrepareInitCluster, which becomes something like
```go
step := h.checklist.GetStepWriter(upgradestatus.INIT_CLUSTER)
step.ResetStateDir()
step.MarkInProgress() // eliding error handling for brevity

go func() {
    err := h.InitCluster(dbConnector)
    if err != nil {
        gplog.Error(err)
        step.MarkFailed()
    } else {
        step.MarkComplete()
    }
}()
```
and we don't have to pass pointers to everything, or chain a dozen calls together.

utils/cluster_pair.go
```diff
@@ -138,3 +138,33 @@ func (cp *ClusterPair) GetHostnames() []string {
...
+// The functions below are specific to the old cluster. They can
+// should be parameterized to work for both old and new clusters.
```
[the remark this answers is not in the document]
jmcatamney: Unfortunately you can't add a method to `cluster` outside of the cluster package, so setting it up that way wouldn't work.  We could do something like `PortBase(clusters.OldCluster)`, though.

hub/services/hub_prepare_init_cluster.go
```diff
@@ -35,28 +40,219 @@ func SaveTargetClusterConfig(clusterPair *utils.ClusterPair, dbConnector *dbconn
 }
 
 func (h *Hub) PrepareInitCluster(ctx context.Context, in *pb.PrepareInitClusterRequest) (*pb.PrepareInitClusterReply, error) {
-	gplog.Info("starting PrepareInitCluster()")
+	gplog.Info("Running PrepareInitCluster()")
+	dbConnector := db.NewDBConn("localhost", int(h.clusterPair.OldCluster.GetPortForContent(-1)),
+		"template1")
 
-	step := h.checklist.GetStepWriter(upgradestatus.INIT_CLUSTER)
-	step.MarkInProgress()
```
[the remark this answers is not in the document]
jmcatamney: We moved that into `InitCluster` so we could explicitly test that, but I'm fine with making it a precondition instead, if no one else minds.

hub/services/hub_prepare_init_cluster.go
```diff
@@ -35,28 +40,219 @@ func SaveTargetClusterConfig(clusterPair *utils.ClusterPair, dbConnector *dbconn
...
+		return
+	}
+
+	// gpinitsystem the new cluster
+	cmdStr := fmt.Sprintf("gpinitsystem -a -c %s -h %s", gpinitsystemFilepath, hostnameFilepath)
+	output, err := h.clusterPair.OldCluster.Executor.ExecuteLocalCommand(cmdStr)
+	if err != nil {
+		// gpinitsystem has a return code of 1 for warnings, so we can ignore that return code
+		// Only a return code of 2 indicates an actual failure
+		if err.Error() == "exit status 2" {
```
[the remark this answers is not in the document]
jmcatamney: 1) The executor doesn't give an error code because we haven't had a need to check that separately from the presence or absence of an error before now.  And we still wouldn't care if it weren't for ~those meddling kids~ gpinitsystem warning all the time, so I don't know if it's worth modifying `ExecuteLocalCommand` to give the return code when 99% of the time we'd be underscore-ing it out.

2) Hmm.  It looks like gpinitsystem catches some, but not all, termination methods with a `[FATAL]:-Received INT or TERM signal Script Exiting!` log message and a return code of 2; we thought it caught all of them.

The best approach here might be checking for `[ERROR]` or `[FATAL]` in the output for now, and if we do decide to change `ExecuteLocalCommand` we can go back and change that.  Thoughts?

hub/services/hub_prepare_init_cluster.go
```diff
@@ -35,28 +40,219 @@ func SaveTargetClusterConfig(clusterPair *utils.ClusterPair, dbConnector *dbconn
...
+	segPrefix, err := getMasterSegPrefix(oldMasterDataDir)
+	if err != nil {
+		gplog.Error("Could not get master segment prefix: %s", err)
+		step.MarkFailed()
+		return
+	}
+
+	gplog.Info("Data Dir: %s", oldMasterDataDir)
+	gplog.Info("segPrefix: %v", segPrefix)
+	gpinitsystemConfig = append(gpinitsystemConfig, "SEG_PREFIX="+segPrefix)
```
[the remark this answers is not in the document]
jmcatamney: Since these are just temporary directories that will be renamed later, we weren't worrying about preserving the old names.  But after talking with Scott, we're going to switch it to `PRIMARY_ARRAY` to simplify the later rename step.

hub/services/hub_prepare_init_cluster.go
```diff
@@ -35,28 +40,219 @@ func SaveTargetClusterConfig(clusterPair *utils.ClusterPair, dbConnector *dbconn
...
+	gpinitsystemConfig = append(gpinitsystemConfig, masterDataDirDeclare)
+
+	//set master port
+	// FIXME: Arbitrary assumption.	 Do something smarter later
+	masterPort := h.clusterPair.GetOldMasterPort() + 1
+	gpinitsystemConfig = append(gpinitsystemConfig, fmt.Sprintf("MASTER_PORT=%d", masterPort))
+
+	gpinitsystemConfig = append(gpinitsystemConfig, "TRUSTED_SHELL=ssh")
+	gpinitsystemConfig = append(gpinitsystemConfig, "CHECK_POINT_SEGMENTS=8")
+	gpinitsystemConfig = append(gpinitsystemConfig, "ENCODING=UNICODE")
```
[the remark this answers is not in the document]
jmcatamney: Whoops, those were just supposed to be hardcoded for the first pass.  We'll fix that.

hub/services/hub_prepare_init_cluster.go
jmcatamney: We do error multiplexing in [ExecuteClusterCommand](https://github.com/greenplum-db/gp-common-go-libs/blob/master/cluster/cluster.go) and the associated wrapper functions, so we could implement a pattern something like that here.

hub/services/hub_prepare_init_cluster.go
jmcatamney: We set things up this ways an experiment to avoid the 
```
err := Foo()
if err != nil {
    Boilerplate()
    return err
}
err := Bar()
if err != nil {
    Boilerplate()
    return err
}
err := Baz()
if err != nil {
    Boilerplate()
    return err
}
...
```
pattern; moving the `MarkFailed`/`MarkComplete` to `PrepareInitCluster` fixes part of the boilerplate, but not the error check after every function.  If the team doesn't like the pattern, we can revert it.

hub/services/hub_prepare_init_cluster.go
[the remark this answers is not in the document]
jchampio: Can we improve gpinitsystem so it's easier to script?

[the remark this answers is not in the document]
jmcatamney: I'm sure that's in the backlog somewhere.  Really, just removing some of the superfluous warnings would help.

agent/services/create_data_dirs.go
```diff
@@ -0,0 +1,34 @@
...
+			if err != nil {
+				gplog.Error("Error creating new segment data directory %s: %s", segDataDir, err)
+				return &pb.CreateSegmentDataDirReply{}, err
+			}
+		} else if err != nil {
+			gplog.Error("Error statting new segment data directory %s: %s", segDataDir, err)
+			return &pb.CreateSegmentDataDirReply{}, err
+		} else {
+			gplog.Info("Directory %s already exists", segDataDir)
```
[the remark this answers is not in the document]
jchampio: Oh, I see. Yeah, probably a good idea to clarify which directory is being created here.

hub/services/hub_prepare_init_cluster.go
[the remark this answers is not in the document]
jchampio: Yeah...

Checking for error messages in the stdout seems like we're going even further in the direction of "brittle". If we change to explicitly checking for error code 1, and treat all other error codes as a hard error, I'd be okay with that for now.

hub/services/hub_prepare_init_cluster.go
jchampio: IMO the error-check boilerplate in Go isn't too bad -- annoying, yes, and hopefully someone will improve on it in a future version of the language, but it's readable. This does not read as well for me, personally, and I'm worried that the pointer-passing is going to hide relationships between the functions and make it easy to add accidental side effects.

slari-pivotal:
@jchampio @kyeap-pivotal @doty-pivotal @nadeemg We believe we have addressed all of the PR comments, and have pushed several WIP commits.  Take a look and see if we missed anything, and if not we will squash the WIP commits with better messages.

hub/services/hub_prepare_init_cluster.go
```diff
@@ -45,41 +45,62 @@ func (h *Hub) PrepareInitCluster(ctx context.Context, in *pb.PrepareInitClusterR
 	dbConnector := db.NewDBConn("localhost", int(h.clusterPair.OldCluster.GetPortForContent(-1)),
 		"template1")
```
jchampio: I think the `Reset/MarkInProgress` steps are missing here.

agent/services/create_data_dirs.go
jchampio: This still needs to be moved.

hub/services/hub_prepare_init_cluster.go
```diff
@@ -45,41 +45,62 @@ func (h *Hub) PrepareInitCluster(ctx context.Context, in *pb.PrepareInitClusterR
...
 
-	go h.InitCluster(dbConnector)
+	go func() {
+		step := h.checklist.GetStepWriter(upgradestatus.INIT_CLUSTER)
+		err := h.InitCluster(dbConnector)
+		if err != nil {
+			step.MarkFailed()
```
jchampio: Don't we need to log the error here too?

hub/services/hub_prepare_init_cluster.go
```diff
@@ -115,6 +125,27 @@ func initializeState(step upgradestatus.StateWriter) error {
...
+	}{}
+	encodingStruct := struct {
+		Encoding string `db:"server_encoding"`
+	}{}
+	err := dbConnector.Get(&checkpointStruct, "show checkpoint_segments")
+	if err != nil {
+		return gpinitsystemConfig, wrapAndLogError("Could not retrieve checkpoint segments", err)
+	}
+	checkpointSegments := fmt.Sprintf("CHECK_POINT_SEGMENTS=%s", checkpointStruct.Segments)
+	err = dbConnector.Get(&encodingStruct, "show server_encoding")
```
jchampio: I think you should be able to query for all settings at once -- and save yourself a query, a struct, and struct tags in the process:
```go
settings := struct {
    Segments, Encoding string
}
query := `SELECT current_setting('checkpoint_segments') as segments, current_setting('server_encoding') as encoding`
err := dbConnector.Get(&settings, query)
```

hub/services/hub_prepare_init_cluster.go
```diff
@@ -89,9 +99,9 @@ func (h *Hub) InitCluster(dbConnector *dbconn.DBConn) error {
 	if err != nil {
 		return err
 	}
-	err = h.GetAndSaveTargetClusterConfig(dbConnector)
+	err = SaveTargetClusterConfig(h.clusterPair, dbConnector, h.conf.StateDir, h.clusterPair.NewBinDir)
 	if err != nil {
-		return err
+		return wrapAndLogError("Could not save new cluster configuration", err)
```
jchampio: Is it time for us to switch to `errors.Wrap()`?

It seems like we should push the "overall error" logging all the way up to the top level, and log at the point-of-call only when we have debugging info and other items that require the full context.

hub/services/hub_prepare_init_cluster.go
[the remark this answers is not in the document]
jchampio: It sounds like there's an "or" here -- can we do both?

hub/services/hub_prepare_init_cluster.go
```diff
@@ -35,28 +40,250 @@ func SaveTargetClusterConfig(clusterPair *utils.ClusterPair, dbConnector *dbconn
...
+	return nil
+}
+
+func (h *Hub) RunInitsystemForNewCluster(gpinitsystemFilepath string) error {
+	// gpinitsystem the new cluster
+	cmdStr := fmt.Sprintf("gpinitsystem -a -I %s", gpinitsystemFilepath)
+	output, err := h.clusterPair.OldCluster.Executor.ExecuteLocalCommand(cmdStr)
+	if err != nil {
+		// gpinitsystem has a return code of 1 for warnings, so we can ignore that return code
+		if err.Error() != "exit status 1" {
```
jchampio: This check is still inverted.

PR #34, opened 2018-07-17
Add end-to-end tests based on BATS

jchampio:
https://github.com/bats-core/bats-core

For your consideration. Three commits:

1. Moves the missing-options tests completely out of Go and into BATS, and adds these new end-to-end tests to `make check`.
2. Reimplements the `start-hub` tests, which didn't actually mock anything or rely on Go in really any way, as BATS tests. You can see the difference in implementation side-by-side.
3. Implements new functionality via TDD and the new suite, as a showcase.


doty-pivotal:
Looks good.  After playing around with it, I was bit every time by not having run `make install` since these tests are using the binaries in the `$PATH`.

Here are the pro's and con's that I see:

Pro's:
+ If we are exercising the CLI, `bash` provides the least overhead.
+ Tests that manipulate the env (sys admin type stuff) are easy to write in bash
+ Writing the `bash` tests will highlight some of the UI rough edges

Con's:
+ There is no other GPDB team using BATS, introduces +1 testing framework into our org's mix.
+ `bash` is great until you cross a complexity threshold, then the language becomes "fun".
+ Between test hygiene is important, since a bash script can have side-effects in shell running the test. I acknowledge that this is not limited to BATs, but when writing bash, its so "easy" to export X...

I reached out to @tanzeeb, and his feedback on BATS was positive.  His two big suggestions were that we should be mindful of pollution between tests runs (his team's use case in that past make docker-izing that a good idea, our use case may make that difficult), and being cognizant of the `bash` version matters when running BATS.

jchampio:
@doty-pivotal Nice summary. Let me respond to some of the points; they're important:

>  I was bit every time by not having run make install since these tests are using the binaries in the $PATH.

Yes. I view this as a must-fix: you should not have to install anything to run `make check`, as a rule. We have a chore to fix that, and this problem needs to be fixed at the same time.

> There is no other GPDB team using BATS, introduces +1 testing framework into our org's mix.

Yep. Luckily, the core testing "language" here is Makefile-style Bash, which we all need to know how to write anyway -- but having the additional framework is an added cost, no matter how you slice it.

> bash is great until you cross a complexity threshold, then the language becomes "fun".

Agreed. I see this as related to your "highlight some of the UI rough edges" statement, though -- if the only way to end-to-end test a feature is to write some sort of Bash insanity, maybe something is wrong with the feature.

> Between test hygiene is important, since a bash script can have side-effects in shell running the test. I acknowledge that this is not limited to BATs, but when writing bash, its so "easy" to export X...

BATS actually does this right -- each test executes in a separate process:
```
$ cat ./test/test.bats
#! /usr/bin/env bats

@test "one" {
    echo "# $TEST_VAR" 1>&3
    export TEST_VAR="1"
}

@test "two" {
    echo "# $TEST_VAR" 1>&3
    export TEST_VAR="2"
}

@test "three" {
    echo "# $TEST_VAR" 1>&3
    export TEST_VAR="3"
}

$ TEST_VAR=0 ./test/test.bats -t
1..3
# 0
ok 1 one
# 0
ok 2 two
# 0
ok 3 three
```
So while we still need to worry about test pollution, we shouldn't have to worry any more than we already do with other test frameworks.

> being cognizant of the bash version matters when running BATS.

Ooh, nice catch. I hadn't considered that.

doty-pivotal:
> if the only way to end-to-end test a feature is to write some sort of Bash insanity, maybe something is wrong with the feature.

+10 to this

> which we all need to know how to write anyway

I agree.

doty-pivotal:
> we shouldn't have to worry any more than we already do with other test frameworks.

Thanks for the example, that made that issue clear.

doty-pivotal:
I really like the idea of BATS, especially if we use it in the correct context, which this experiment is.

jchampio:
The first pieces of this experiment have been merged into master. Thanks for all the feedback everyone!

PR #35, opened 2018-07-18
Make StatusConversion more unit testable

kyeap-pivotal:
Tried to break status conversion into a more unit testable piece. 
Made StatusConversion more unit testable. (it a looks a little more complicated though?)
I managed to break the dependency of `agent.Start()` for the two `StatusConversion` tests

Not sure if this should be merged, but would love to get your thoughts.

### DIFF
--- a/hub/services/hub.go
+++ b/hub/services/hub.go
@@ -45,6 +45,7 @@ type Hub struct {
 
 type Connection struct {
 	Conn          *grpc.ClientConn
+	AgentClient   pb.AgentClient
 	Hostname      string
 	CancelContext func()
 }
@@ -136,6 +137,7 @@ func (h *Hub) AgentConns() ([]*Connection, error) {
 		}
 		h.agentConns = append(h.agentConns, &Connection{
 			Conn:          conn,
+			AgentClient:   pb.NewAgentClient(conn),
 			Hostname:      host,
 			CancelContext: cancelFunc,
 		})
--- a/hub/services/hub_status_conversion_test.go
+++ b/hub/services/hub_status_conversion_test.go
@@ -3,7 +3,9 @@ package services_test
 import (
 	"errors"
 
+	"github.com/golang/mock/gomock"
 	"github.com/greenplum-db/gp-common-go-libs/cluster"
+	mockpb "github.com/greenplum-db/gpupgrade/mock_idl"
 	"github.com/greenplum-db/gpupgrade/testutils"
 
 	"google.golang.org/grpc"
@@ -19,10 +21,14 @@ import (
 
 var _ = Describe("hub", func() {
 	var (
-		hub         *services.Hub
-		agentA      *testutils.MockAgentServer
-		clusterPair *utils.ClusterPair
-		cm          *testutils.MockChecklistManager
+		hub               *services.Hub
+		agentA            *testutils.MockAgentServer
+		clusterPair       *utils.ClusterPair
+		cm                *testutils.MockChecklistManager
+		ctrl              *gomock.Controller
+		mockAgent1        *mockpb.MockAgentClient
+		agentConnections  []*services.Connection
+		hostToSegmentsMap map[string][]cluster.SegConfig
 	)
 
 	BeforeEach(func() {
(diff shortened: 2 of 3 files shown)


hub/services/hub.go
```diff
@@ -113,6 +114,10 @@ func (h *Hub) Stop() {
 	}
 }
 
+// func (h *Hub) SetAgentConns(agentConnections []*Connection) {
```
jmcatamney: This commented-out section can be removed.

hub/services/hub_status_conversion_test.go
```diff
@@ -43,47 +49,77 @@ var _ = Describe("hub", func() {
...
+			)
+
+			_, err := services.GetConversionStatusFromPrimaries(agentConnections, hostToSegmentsMap)
+			Expect(err).To(HaveOccurred())
+		})
 	})
 
 	Describe("PrimaryConversionStatus", func() {
 		It("returns FAILED if any agents report failure", func() {
+			// statusMessages := []string{"FAILED", "RUNNING"}
```
jmcatamney: This test should either be fixed or `XIt`-ed instead of commented out, so we don't get a false green.

hub/services/hub_status_conversion_test.go
```diff
@@ -19,10 +21,14 @@ import (
...
-		cm          *testutils.MockChecklistManager
+		hub               *services.Hub
+		agentA            *testutils.MockAgentServer
+		clusterPair       *utils.ClusterPair
+		cm                *testutils.MockChecklistManager
+		ctrl              *gomock.Controller
+		mockAgent1        *mockpb.MockAgentClient
+		agentConnections  []*services.Connection
+		hostToSegmentsMap map[string][]cluster.SegConfig
 	)
```
nadeemg: Are you seeing the real gain here that we can move away from using the `testutils.MockAgentServer` and instead start using `mockpb.MockAgentClient`? I think we've talked about that as a team before and landed here, so seems like a good step.

hub/services/hub_status_conversion_test.go
```diff
@@ -43,47 +49,77 @@ var _ = Describe("hub", func() {
...
+			// )
+
+			// hub.SetAgentConns(agentConnections)
+
+			// status := services.PrimaryConversionStatus(hub)
+			// Expect(status).To(Equal(pb.StepStatus_FAILED))
+
 			agentA.StatusConversionResponse = &pb.CheckConversionStatusReply{
 				Statuses: []string{"FAILED", "RUNNING"},
 			}
```
nadeemg: Does this need to continue using the `testutils.MockAgentServer`?

hub/services/hub.go
kyeap-pivotal: This was part of my attempt to inject fake AgentClients. This is definitely not going in.

hub/services/hub_status_conversion_test.go
kyeap-pivotal: The current test being run is unmodified. This commented out section is what was written in an attempt to replace the test that requires an agent to be spun up. I discovered that even after injecting in fake AgentClients I had no way to get past the `connectivity.Ready` checks. So the test would panic.

hub/services/hub_status_conversion_test.go
kyeap-pivotal: Yes this file still needs `testutils.MockAgentServer` because when testing `PrimaryConversionStatus`, `h.AgentConns` will be called. `h.AgentConns` requires a live agent to be running for tests.

hub/services/hub_status_conversion_test.go
kyeap-pivotal: There are several positive gains. We won't have to rely on and maintain a mock agent server for the sole purpose of testing. It also means that with no agents running in the background all the time, there will be no chance of them affecting tests unintentionally. It will also force the code that is currently using `testutils.MockAgentServer` to be broken down in such a way in which testing will be safer with less dependencies, easier to unit test, and faster to test.

kyeap-pivotal:
Sorry I didn't make this clear earlier. I will be removing all commented code if we are comfortable with merging this.

kyeap-pivotal:
Removed the commented code. and made StatusConversion unit testing explicit when checking incoming arguments.

jchampio:
Small [regression](https://prod.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade/jobs/unit-tests/builds/8) after this was merged.

PR #36, opened 2018-07-18
pipeline: deploy to prod by default

jchampio:
This just sets the default `FLY_TARGET`; you can still override it to whatever you want. This pipeline is now deployed [here](https://prod.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade).


jchampio:
Switched to `gpdb-prod`.

PR #38, opened 2018-07-18
Create end-to-end test suite in Ginkgo

jmcatamney:
This commit pulls the insufficient arguments tests and start-hub tests out into
a separate end-to-end suite that builds the binaries and calls them, so that all
of the tests that build gpupgrade and call the binary are in one place.

Functional changes from the original tests are as follows:
- Build the binaries before each test run to ensure up-to-date binaries.
- Use DescribeTable to simplify highly-redundant flag tests.
- Add a _suite_test.go file for handling file and process cleanup.

---

This PR is an alternative take on #34 using pure Ginkgo, to help the team determine which approach they'd like to go with.


jchampio:
Nice, this is a great comparison! Thanks for taking the time to give this feature parity with the other PR.

Based on this, I'm definitely preferring Bash over Go for this sort of testing, mostly because of the implementation difference between happy-path (and simple sad-path) tests. For example,
```sh
gpupgrade prepare start-hub
ps -ef | grep -Gq "[g]pupgrade_hub --daemon$"
```
versus
```go
_, err := gpupgrade("prepare", "start-hub")
Expect(err).ToNot(HaveOccurred())
running := verifyProcessIsRunning("gpupgrade_hub --daemon")
Expect(running).To(BeTrue())
```
Go gives us those powerful assertions, but it's exactly that power that we want to avoid in the end-to-end tests, so I anticipate that a huge portion of the test code is going to be `Expect(err).ToNot(HaveOccurred())` and related. I think that will affect how happy we are creating these tests.

nadeemg:
I think this turned out much cleaner that I imagined it would. I'm on the fence now. Can we talk about this as a group before making a decision? Maybe if we have time in retro tomorrow?

jmcatamney:
We're closing this PR in favor of the BATS tests, but keeping the branch around for future reference.

PR #39, opened 2018-07-20
Fix race conditions between hub start and stop functions.

slari-pivotal:
The gpupgrade pipeline https://prod.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade is failing on the CI, but not locally. The reason for this is because of a race condition between the calls to hub.Stop() and hub.Start() in the test. The reason is that hub.Stop() also closes agent connections. However, in the test, hub.Start() is called from a go routine which can sometimes start after the call to hub.Stop() has happened. Hence, the failure in hub.Stop() in the test.

We have fixed this by using a channel/flag to avoid the race condition.

Additional tests are also added.

### DIFF
--- a/hub/gpupgrade_hub_main.go
+++ b/hub/gpupgrade_hub_main.go
@@ -85,7 +85,10 @@ func main() {
 				hub.MakeDaemon()
 			}
 
-			hub.Start()
+			err := hub.Start()
+			if err != nil {
+				return err
+			}
 
 			hub.Stop()
 
--- a/hub/services/hub.go
+++ b/hub/services/hub.go
@@ -26,6 +26,9 @@ import (
 
 var DialTimeout = 3 * time.Second
 
+// Returned from Hub.Start() if Hub.Stop() has already been called.
+var HubStopped = errors.New("hub is stopped")
+
 type Dialer func(ctx context.Context, target string, opts ...grpc.DialOption) (*grpc.ClientConn, error)
 
 type Hub struct {
@@ -36,9 +39,18 @@ type Hub struct {
 	grpcDialer  Dialer
 	checklist   upgradestatus.Checklist
 
-	mu      sync.Mutex
-	server  *grpc.Server
-	lis     net.Listener
+	mu     sync.Mutex
+	server *grpc.Server
+	lis    net.Listener
+
+	// This is used both as a channel to communicate from Start() to
+	// Stop() to indicate to Stop() that it can finally terminate
+	// and also as a flag to communicate from Stop() to Start() that
+	// Stop() had already beed called, so no need to do anything further
+	// in Start().
+	// Note that when used as a flag, nil value means that Stop() has
+	// been called.
+
 	stopped chan struct{}
 	daemon  bool
 }
@@ -75,18 +87,19 @@ func (h *Hub) MakeDaemon() {
 	h.daemon = true
 }
 
-func (h *Hub) Start() {
+func (h *Hub) Start() error {
 	lis, err := net.Listen("tcp", ":"+strconv.Itoa(h.conf.CliToHubPort))
 	if err != nil {
-		gplog.Fatal(err, "failed to listen")
+		return errors.Wrap(err, "failed to listen")
 	}
 
 	server := grpc.NewServer()
+
 	h.mu.Lock()
 	if h.stopped == nil {
 		// Stop() has already been called; return without serving.
 		h.mu.Unlock()
-		return
+		return HubStopped
 	}
 	h.server = server
 	h.lis = lis
@@ -102,10 +115,13 @@ func (h *Hub) Start() {
 
 	err = server.Serve(lis)
 	if err != nil {
-		gplog.Fatal(err, "failed to serve", err)
+		err = errors.Wrap(err, "failed to serve")
 	}
 
+	// inform Stop() that is it is OK to stop now
 	h.stopped <- struct{}{}
+
+	return err
 }
 
 func (h *Hub) Stop() {
@@ -116,7 +132,7 @@ func (h *Hub) Stop() {
 
 	if h.server != nil {
 		h.server.Stop()
-		<-h.stopped
+		<-h.stopped // block until it is OK to stop
 	}
 
 	// Mark this server stopped so that a concurrent Start() doesn't try to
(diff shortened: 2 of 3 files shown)


hub/services/hub_test.go
```diff
@@ -34,9 +38,66 @@ var _ = Describe("Hub", func() {
...
+			err = hub.Start()
+		}()
+		//Using Eventually ensures the test will not stall forever if this test fails.
+		Eventually(func() error { return err }).Should(HaveOccurred())
+		Expect(err.Error()).To(ContainSubstring("failed to listen"))
+	})
+
+	// This is inherently testing a race. It will give false successes instead
+	// of false failures, so DO NOT ignore transient failures in this test!
+	It("will return from Start() if Stop is called concurrently", func() {
```
nadeemg: This test is trying to hit the block on lines 98-102 in hub.go? Is there anything we could do to make this more deterministic? If not, I'd vote for deleting it.

hub/services/hub.go
```diff
@@ -26,6 +26,9 @@ import (
 
 var DialTimeout = 3 * time.Second
 
+// Returned from Hub.Start() if Hub.Stop() has already been called.
+var HubStopped = errors.New("hub is stopped")
+
```
nadeemg: Is this global var only being used on line 101 below?

hub/services/hub.go
```diff
@@ -145,7 +161,9 @@ func (h *Hub) AgentConns() ([]*Connection, error) {
 	hostnames := h.clusterPair.GetHostnames()
 	for _, host := range hostnames {
 		ctx, cancelFunc := context.WithTimeout(context.Background(), DialTimeout)
-		conn, err := h.grpcDialer(ctx, host+":"+strconv.Itoa(h.conf.HubToAgentPort), grpc.WithInsecure(), grpc.WithBlock())
+		conn, err := h.grpcDialer(ctx,
+			host+":"+strconv.Itoa(h.conf.HubToAgentPort), grpc.WithInsecure(),
```
jchampio: If we're breaking the line anyway, can we group the configuration?
```go
conn, err := h.grpcDialer(ctx,
    host + ":" + strconv.Itoa(h.conf.HubToAgentPort),
    grpc.WithInsecure(), grpc.WithBlock())
```

hub/services/hub_test.go
```diff
@@ -34,9 +38,50 @@ var _ = Describe("Hub", func() {
...
+		Eventually(func() error { return err }).Should(HaveOccurred())
+		Expect(err.Error()).To(ContainSubstring("failed to listen"))
+	})
+
+	// This is inherently testing a race. It will give false successes instead
+	// of false failures, so DO NOT ignore transient failures in this test!
 	It("will return from Start() if Stop is called concurrently", func() {
 		hubConfig := &services.HubConfig{
-			HubToAgentPort: port,
+			CliToHubPort: cliToHubPort,
```
jchampio: Why the difference here?

hub/gpupgrade_hub_main.go
```diff
@@ -85,7 +85,10 @@ func main() {
 				hub.MakeDaemon()
 			}
 
-			hub.Start()
+			err := hub.Start()
+			if err != nil {
+				return err
+			}
 
 			hub.Stop()
```
jchampio: Next step (not part of this PR) is to just make 
```go
return hub.Start()
```
safe. If `Start()` returns an error, then either it failed to start entirely (so we shouldn't need to call `Stop()`) or someone else already called `Stop()` (ditto).

hub/services/hub_test.go
```diff
@@ -34,9 +38,66 @@ var _ = Describe("Hub", func() {
...
+		}
+		hub := services.NewHub(clusterPair, grpc.DialContext, hubConfig, nil)
+
+		hub.Stop()
+		go func() {
+			err = hub.Start()
+		}()
+		//Using Eventually ensures the test will not stall forever if this test fails.
+		Eventually(func() error { return err }).Should(HaveOccurred())
```
jchampio: This should be using `services.HubStopped` for comparison, to make sure we're using the right code path.

hub/services/hub_test.go
```diff
@@ -17,13 +19,15 @@ import (
 
 var _ = Describe("Hub", func() {
 	var (
-		agentA      *testutils.MockAgentServer
-		port        int
-		clusterPair *utils.ClusterPair
+		agentA         *testutils.MockAgentServer
+		cliToHubPort   int
+		hubToAgentPort int
```
jchampio: If we're going with prefixes, can we go with something like `hubPort` and `agentPort` instead? Save some precious columns :smile:

hub/services/hub.go
jchampio: (see [my comment below](https://github.com/greenplum-db/gpupgrade/pull/39#discussion_r204079991))

hub/services/hub_test.go
jchampio: This is a tripwire -- it helped us prove we'd fixed the problem, and if we regress it should occasionally fail to show us something is in need of investigation. Is there a specific reason you want to get rid of it?

nadeemg: Transient failures is too close to a flakey test. What is stopping is us from making this deterministic?

jchampio: > What is stopping is us from making this deterministic?
Because it's not testing a deterministic code path; the test is "make sure we don't hang when two paths are called concurrently". Adding determinism -- either by serializing the calls or forcing additional test-only serialization between the two -- will not test what was failing.

hub/services/hub_test.go
kyeap-pivotal: It doesn't make any difference to the logic path, it could even be `&services.HubConfig{}`. Either way the port we pass into `CliToHubPort` is 0 (which gets us an open port), but since this is testing general race conditions pertaining to `hub.Start()` and `hub.Stop()`, setting the port that the hub listens on rather than setting the port agents listen on is much more relevant.

jchampio: Oh, cool. If nothing uses that port, I think passing in a zero-valued `HubConfig` is probably best for now.

hub/services/hub_test.go
nadeemg: I've seen latches used to test scenarios like this. Have we already considered that kind of a test?

slari-pivotal: As we read it, currently there should not be any race between `Start()` and `Stop()` because if `Stop()` acquires the `h.mu.lock` before `Start()`, then `Stop()` will run to completion and `Start()` will exit out when it gets the lock.  Similarly, if `Start()` gets the lock before `Stop()`, then `Start()` would run to completion and `Stop()` would follow after that.

In case, a code path is added to `Stop()` that never releases the mutex, then a hang in `Start()` is possible, and the `Eventually(done).Should(Receive())` should catch that.

Similarly, if call to`hub.Start()` hangs for some other reason then the `Eventually()` should catch that.

nadeemg: Talked this over with @jchampio and he's going to think a little more about what we could do here. We are agreed that its not nice to have a test fail and leave it up to the dev to figure out which of the n previous green commits might have introduced a bug. We're also agreed that we dont want to jump through hoops just to get a deterministic test if that's as hard to do as it seems.

I'm fine with this going in as is.

PR #45, opened 2018-07-26
Unify unit test initialization

jmcatamney:
Previously, every unit test file performed its own setup and teardown in
BeforeEach and AfterEach, most of which was just copy-pasted between test files,
so any refactor touching the hub or agent would generally involve making the
same change in lots of files.

In a similar manner to a previous commit[1], this one moves all the duplicated
logic to the BeforeEach and AfterEach in services_suite_test.go, so that the
initialization happens only once and refactors only need to make a given change
once.

[1] 01fba51a72708669f5b2afd8e275b933714edeef

This PR relies on code changes made in #44, and will be merged only after that PR is merged.

### DIFF
--- a/hub/services/hub_upgrade_convert_primaries_test.go
+++ b/hub/services/hub_upgrade_convert_primaries_test.go
@@ -2,63 +2,27 @@ package services_test
 
 import (
 	"errors"
-	"io/ioutil"
+	"path/filepath"
 
-	"github.com/greenplum-db/gpupgrade/hub/services"
 	pb "github.com/greenplum-db/gpupgrade/idl"
-	"github.com/greenplum-db/gpupgrade/testutils"
 
 	"github.com/greenplum-db/gp-common-go-libs/cluster"
-	"github.com/greenplum-db/gp-common-go-libs/testhelper"
-	"google.golang.org/grpc"
 
-	"github.com/greenplum-db/gpupgrade/utils"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 )
 
 var _ = Describe("hub.UpgradeConvertPrimaries()", func() {
-	var (
-		dir       string
-		hub       *services.Hub
-		mockAgent *testutils.MockAgentServer
-		port      int
-		source    *utils.Cluster
-		target    *utils.Cluster
-		cm        *testutils.MockChecklistManager
-	)
-
 	BeforeEach(func() {
-		testhelper.SetupTestLogger()
-
-		mockAgent, port = testutils.NewMockAgentServer()
-
-		var err error
-		dir, err = ioutil.TempDir("", "")
-		Expect(err).ToNot(HaveOccurred())
-		conf := &services.HubConfig{
-			StateDir:       dir,
-			HubToAgentPort: port,
-		}
-
-		source, target = testutils.CreateMultinodeSampleClusterPair("/tmp")
 		seg1 := target.Segments[0]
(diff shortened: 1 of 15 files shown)


hub/services/hub_upgrade_convert_primaries_test.go
```diff
@@ -2,72 +2,36 @@ package services_test
...
-		Expect(err).ToNot(HaveOccurred())
-		conf := &services.HubConfig{
-			StateDir:       dir,
-			HubToAgentPort: port,
-		}
-
-		source, target = testutils.CreateMultinodeSampleClusterPair("/tmp")
 		seg1 := target.Segments[0]
-		seg1.DataDir = "/tmp/seg1_upgrade"
+		seg1.DataDir = fmt.Sprintf("%s/seg1_upgrade", dir)
```
jchampio: `filepath.Join()`, maybe? There are a handful of new `fmt` calls like this one.

hub/services/hub_upgrade_reconfigure_ports_test.go
```diff
@@ -43,35 +31,27 @@ var _ = Describe("UpgradeReconfigurePorts", func() {
 			}
 		}
 
-		source, target := testutils.CreateSampleClusterPair()
 		testExecutor = &testhelper.TestExecutor{}
-		source.Segments[1] = cluster.SegConfig{Hostname: "hosttwo"}
+		//source.Segments[1] = cluster.SegConfig{Hostname: "hosttwo"}
```
jchampio: This should be deleted.

hub/services/hub_upgrade_convert_primaries_test.go
jmcatamney: Will do.

hub/services/hub_upgrade_reconfigure_ports_test.go
jmcatamney: Good catch, thanks.

jmcatamney:
@jchampio 
> The distance between declaration and use for our magic globals got much longer. We're now up to twelve globals that people need to keep track of, and they're in a completely separate file

> Why do so many unit tests require a fully constructed Hub? Will this problem go away once we remove Hub as the focal point for every helper function? Shouldn't we be testing "the Hub" as a logical unit, which would vastly reduce the number of files that need to spin one up?

The first issue is somewhat mitigated by the second.  All those globals don't really need to be kept track of, exactly, because they're not directly used for much and are only needed because pretty much everything requires the hub at the moment.  As functions are moved off of the hub, most of those should indeed be able to be moved out of the suite file and into hub-specific tests.

> I'm not -1 on this, because I think this is an improvement over what we had, but it's a step towards a local maximum and away from a global one.

It's not an ideal situation right now, but I do think that by consolidating everything and lowering the friction for refactors it's a step toward a global maximum as well.

kyeap-pivotal:
@jchampio 
> Why do so many unit tests require a fully constructed Hub? 

We didn't sufficiently separate out business logic from the grpc infrastructure early on. Its getting better quickly which is why about half of the unit tests don't require the hub anymore as mentioned by @nadeemg 

> Will this problem go away once we remove Hub as the focal point for every helper function? 

 That seems to be how the code is evolving.

I agree that this will help speed up refactors. And its ok to merge this in as long as we think its an improvement (and it seems that we do), but not ideal for unit testing. I like the idea of re-evaluating each step and slowly separating everything out into isolated unit testing environments after merging this in.

PR #46, opened 2018-07-26
Fix code coverage script

jmcatamney:
The previous run_coverage.sh script was broken and didn't do what we wanted
anyway, so this commit copies show_coverage.sh from gpbackup and modifies it to
work with gpupgrade's directories.

Authored-by: Jamie McAtamney <jmcatamney@pivotal.io>

### DIFF
--- a/scripts/show_coverage.sh
+++ b/scripts/show_coverage.sh
@@ -1,0 +1,40 @@
+#!/bin/bash
+
+DIR="github.com/greenplum-db/gpupgrade"
+
+# Generate code coverage statistics for one package, write the coverage
+# statistics to a file, and print the coverage percentage to the shell
+go test -coverpkg $DIR/utils $DIR/utils \
+        -coverprofile=/tmp/coverage.out 2> /dev/null \
+        | awk '{print $2 " unit test " $4 "\t\t" $5}' \
+        | awk -F"/" '{print $4}'
+
+PACKAGES=("agent/services"
+"cli/commanders"
+"hub/services"
+"hub/upgradestatus"
+)
+
+for PACKAGE in "${PACKAGES[@]}"; do
+  TESTFILE=unit_$(dirname "$PACKAGE")_$(basename "$PACKAGE").out
+
+  # Generate code coverage statistics for the rest of the packages in the same
+  # way as above, accounting for the multi-part package specification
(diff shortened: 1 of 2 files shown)


scripts/show_coverage.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+DIR="github.com/greenplum-db/gpupgrade"
+RESULTS="/tmp/results.out"
```
jchampio: We write stdout to the `RESULTS` file, then `cat` it, then remove it. I think the result would be the same if `RESULTS` was stripped out completely.

scripts/show_coverage.sh
```diff
@@ -0,0 +1,20 @@
...
+# Generate code coverage statistics for one package, write the coverage statistics to a file, and print the coverage percentage to the shell
+go test -coverpkg $DIR/utils $DIR/utils -coverprofile=/tmp/coverage.out 2> /dev/null | awk '{print $2 " unit test " $4 "\t\t" $5}' | awk -F"/" '{print $4}' > $RESULTS
```
jchampio: Let's not discard `stderr` here; it's not interfering with the pipeline or the `stdout` capture.

scripts/show_coverage.sh
```diff
@@ -0,0 +1,20 @@
...
+for PACKAGE in "agent/services" "cli/commanders" "hub/services" "hub/upgradestatus"; do
```
jchampio: Why do we ignore some packages?

scripts/show_coverage.sh
```diff
@@ -0,0 +1,20 @@
...
+  # Filter out the first "mode: set" line from each coverage file and concatenate them all
+  cat "/tmp/$TESTFILE" | awk '{if($1!="mode:") {print $1 " " $2 " " $3}}' >> /tmp/coverage.out
+  rm "/tmp/$TESTFILE"
+done
+
+# Print the total coverage percentage and generate a coverage HTML page
+go tool cover -func=/tmp/coverage.out | awk '{if($1=="total:") {print $1 "\t\t\t\t\t" $3}}' >> $RESULTS
+cat $RESULTS
+rm $RESULTS
+exit 0
```
jchampio: `exit 0` happens by default.

scripts/show_coverage.sh
nadeemg: agreed

PR #48, opened 2018-07-27
ci: cleanup, use resource mapping, and use built in go env

doty-pivotal:
Looking, in particular, for feedback on the last commit.


Makefile
```diff
@@ -21,20 +21,22 @@ MAC_POSTFIX := .darwin.$(BRANCH)
...
 		go get github.com/alecthomas/gometalinter
 		gometalinter --install
-		go get github.com/golang/dep/cmd/dep
-		dep ensure
 # Counterfeiter is not a proper dependency of the app. It is only used occasionally to generate a test class that
 # is then checked in.  At the time of that generation, it can be added back to run the dependency list, temporarily.
 #		go get github.com/maxbrunsfeld/counterfeiter
 
-depend : dependencies
+depend : workstation-dependencies ci-dependencies
```
jchampio: Can we have `make depend` do what `ci-dependencies` does, and have a `dev-depend` or something that installs the generators?

doty-pivotal: Yes.  For the initial pass I was attempting to not change the `make depend` behavior.  I have no strong opinions on the actual names.  I will see if others chime in, otherwise I will update to your suggested names.

PR #49, opened 2018-07-27
fixup: fix up linter warnings

doty-pivotal:
At the start of this there were many linting warnings/errors.  This set of commits address all but a few.

Up for discussion in particular are the won't fixes: 

-----

```
hub/upgradestatus/checklist_manager.go:53:2:warning: don't use underscores in Go names; struct field Name_ should be Name (golint)
hub/upgradestatus/checklist_manager.go:54:2:warning: don't use underscores in Go names; struct field Code_ should be Code (golint)
hub/upgradestatus/checklist_manager.go:55:2:warning: don't use underscores in Go names; struct field Status_ should be Status (golint)
```
These are named this way to avoid type `Step` having both field and method named `Name`.

-----
```
utils/daemon/daemon_test.go:19:1:warning: receiver name should not be an underscore (golint)
utils/daemon/daemon_test.go:37:1:warning: receiver name should not be an underscore (golint)
```
~~The test just needs the helper function doesn't need the receiver, so we don't give it a name~~ figured this one out.

-----
```
utils/daemon/daemon.go:213:75:warning: parameter timeout always receives 2000000000 (unparam)
```
the linter does not seem to recognize that we are using the timeout in the tests.

### DIFF
--- a/hub/gpupgrade_hub_main.go
+++ b/hub/gpupgrade_hub_main.go
@@ -82,6 +82,7 @@ func main() {
 				return services.PrimaryConversionStatus(hub)
 			}
 
+			// {Name_: *, Code_: *, Status_:, *}
 			cm.LoadSteps([]upgradestatus.Step{
 				{upgradestatus.CONFIG, pb.UpgradeSteps_CONFIG, stateCheck},
 				{upgradestatus.SEGINSTALL, pb.UpgradeSteps_SEGINSTALL, stateCheck},
--- a/hub/gpupgrade_hub_main.go
+++ b/hub/gpupgrade_hub_main.go
@@ -116,7 +116,7 @@ func main() {
 	daemon.MakeDaemonizable(RootCmd, &shouldDaemonize)
 
 	err := RootCmd.Execute()
-	if err != nil && err != daemon.SuccessfullyDaemonized {
+	if err != nil && err != daemon.ErrSuccessfullyDaemonized {
 		if gplog.GetLogger() == nil {
 			// In case we didn't get through RootCmd.Execute(), set up logging
 			// here. Otherwise we crash.
--- a/Makefile
+++ b/Makefile
@@ -34,7 +34,7 @@ format:
 		goimports -w .
 
 lint:
-		! gofmt -l agent/ cli/ db/ helpers/ hub/ install/ integrations/ shellparsers/ testutils/ utils/ | read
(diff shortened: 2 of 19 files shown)


jchampio:
I haven't reviewed everything yet, but I just found out about [receiver elision](https://github.com/golang/lint/issues/310#issuecomment-317071854), so it looks like
> receiver name should not be an underscore

should probably be fixed.

hub/gpupgrade_hub_main.go
```diff
@@ -83,16 +83,16 @@ func main() {
...
-				{upgradestatus.SEGINSTALL, pb.UpgradeSteps_SEGINSTALL, stateCheck},
-				{upgradestatus.INIT_CLUSTER, pb.UpgradeSteps_INIT_CLUSTER, stateCheck},
-				{upgradestatus.SHUTDOWN_CLUSTERS, pb.UpgradeSteps_SHUTDOWN_CLUSTERS, shutDownStatus},
-				{upgradestatus.CONVERT_MASTER, pb.UpgradeSteps_CONVERT_MASTER, convertMasterStatus},
-				{upgradestatus.START_AGENTS, pb.UpgradeSteps_START_AGENTS, stateCheck},
-				{upgradestatus.SHARE_OIDS, pb.UpgradeSteps_SHARE_OIDS, stateCheck},
-				{upgradestatus.CONVERT_PRIMARIES, pb.UpgradeSteps_CONVERT_PRIMARIES, convertPrimariesStatus},
-				{upgradestatus.VALIDATE_START_CLUSTER, pb.UpgradeSteps_VALIDATE_START_CLUSTER, stateCheck},
-				{upgradestatus.RECONFIGURE_PORTS, pb.UpgradeSteps_RECONFIGURE_PORTS, stateCheck},
+				{Name_: upgradestatus.CONFIG, Code_: pb.UpgradeSteps_CONFIG, Status_: stateCheck},
```
jchampio: I am -1ish on this (the "unkeyed fields") commit. I agree with the linter's goal, but I think that following its advice in this case has made the code much less readable. Let's instead work on improving the `LoadSteps` API so that you don't have to pass in magic structs.

jchampio: (after discussion -- we'll remove it from the main wall-o-text here, but keep the other changes in that commit, which help readability instead of hindering.)

PR #51, opened 2018-07-30
run integration tests in ci

kyeap-pivotal:
Currently the binaries being consumed are from the release candidates buckets. Eventually we would most likely want to use the most up to date binaries from the latest green commit.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ GOFLAGS := -gcflags="all=-N -l"
 depend:
 		go get github.com/onsi/ginkgo/ginkgo
 		go get github.com/golang/dep/cmd/dep
-		dep ensure
+		dep ensure -v
 
 depend-dev: depend
 		go get -u github.com/golang/protobuf/protoc-gen-go
@@ -117,7 +117,7 @@ clean:
 # dependency on this marker to run a `dep ensure` (if necessary) before your
 # recipe is run.
 .Gopkg.updated: Gopkg.lock Gopkg.toml
-	dep ensure
+	dep ensure -v
 	touch $@
 
 # You can override these from the command line.
@@ -142,6 +142,8 @@ set-pipeline:
 	fly -t $(FLY_TARGET) set-pipeline -p $(PIPELINE_NAME) \
 		-c ci/pipeline.yml \
 		-l ~/workspace/continuous-integration/secrets/$(SECRETS_FILE) \
+		-l ~/workspace/continuous-integration/secrets/gpdb_common-ci-secrets.yml \
+		-l ~/workspace/continuous-integration/secrets/gpdb_master-ci-secrets.yml
 		-v gpupgrade-git-remote=$(GIT_URI) \
 		-v gpupgrade-git-branch=$(BRANCH)
 
--- a/ci/pipeline.yml
+++ b/ci/pipeline.yml
@@ -13,19 +13,49 @@ resources:
     uri: {{gpupgrade-git-remote}}
     branch: {{gpupgrade-git-branch}}
 
+- name: gpdb_src
+  type: git
+  source:
+    uri: https://github.com/greenplum-db/gpdb
+    branch: master
+
 - name: slack-alert
   type: slack-notification
   source:
     url: {{cm_webhook_url}}
 
+- name: bin_gpdb_centos6
+  type: s3
+  source:
+    access_key_id: {{bucket-access-key-id}}
+    bucket: {{bucket-name}}
+    region_name: {{aws-region}}
+    secret_access_key: {{bucket-secret-access-key}}
+    versioned_file: {{bin_gpdb_centos_versioned_file}}
+
+
 jobs:
 - name: unit-tests
   plan:
-  - get: gpupgrade
-    resource: gpupgrade_src
+  - get: gpupgrade_src
     trigger: true
   - task: unit-tests
-    file: gpupgrade/ci/tasks/unit-tests.yml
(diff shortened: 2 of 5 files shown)


Makefile
```diff
@@ -142,6 +142,8 @@ set-pipeline:
 	fly -t $(FLY_TARGET) set-pipeline -p $(PIPELINE_NAME) \
 		-c ci/pipeline.yml \
 		-l ~/workspace/continuous-integration/secrets/$(SECRETS_FILE) \
+		-l ~/workspace/continuous-integration/secrets/gpdb_common-ci-secrets.yml \
+		-l ~/workspace/continuous-integration/secrets/gpdb_master-ci-secrets.yml
```
doty-pivotal: After this commit goes in, can you please update the [README.md](https://github.com/greenplum-db/continuous-integration#gpupgrade)?

ci/pipeline.yml
```diff
@@ -30,6 +46,21 @@ jobs:
       do:
       - *slack_alert
 
+- name: integration-tests
+  plan:
+  - aggregate:
+    - get: gpupgrade
```
doty-pivotal: does this need to be:
```
  - get: gpupgrade
    resource: gpupgrade_src
    trigger: true
```
?

ci/scripts/integration-tests.bash
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash -l
```
doty-pivotal: I don't understand why we want to run this script with the `--login` option.

ci/scripts/integration-tests.bash
```diff
@@ -0,0 +1,21 @@
...
+# make depend fails if run as gpadmin with a dep ensure git-remote-https signal 11 failure
+GOPATH="$PWD/go" PATH="$PWD/go/bin:$PATH" make -C go/src/github.com/greenplum-db/gpupgrade depend
+
+# build and setup demo cluster
+source gpdb_src/concourse/scripts/common.bash
+time install_gpdb
+time ./gpdb_src/concourse/scripts/setup_gpadmin_user.bash "centos"
+time make_cluster
+
+chown -R gpadmin:gpadmin go
```
doty-pivotal: Not saying no, but this can be as surprisingly expensive operation inside a container running on an OverlayFS (as we are in Concourse).  Can we be more targeted?

ci/scripts/integration-tests.bash
```diff
@@ -0,0 +1,21 @@
...
+
+su gpadmin -c 'export GOPATH=$PWD/go && \
```
doty-pivotal: Again, not a no, but I would like to see if we like any other pattern on this one.  Do we like using a `HEREDOC` to pass the commands to `su`? 

Another possibility is to drop a file like:

```
cat << EOF > /path/to/run_integration.sh
....
EOF

chmod +x /path/to/run_integration.sh
su gpadmin -c /path/to/run_integration.sh
```

I think the first is a bit cleaner.

ci/tasks/unit-tests.yml
```diff
@@ -3,8 +3,8 @@ PLATFORM: linux
 image_resource:
   type: docker-image
   source:
-    repository: golang
-    tag: '1.10.2'
+    repository: pivotaldata/centos-gpdb-dev
```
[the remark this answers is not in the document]
nadeemg: Good catch. We missed it because we were just thinking about the integration tests. 

So it would be better to pull two containers, run unit tests on the `golang` container and the integration tests on this one?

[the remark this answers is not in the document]
jchampio: Yes, I think so.

ci/scripts/integration-tests.bash
kyeap-pivotal: permissions to create folders `pkg` and `bin` are needed in the `$GOPATH` folder in order to build gpupgrade.

ci/scripts/integration-tests.bash
kyeap-pivotal: done.

Makefile
kyeap-pivotal: added as a task to the tracker.

ci/scripts/integration-tests.bash
doty-pivotal: I wish we could leverage run as the `gpadmin`[1] like:

```
- task: integration-tests
    file: gpupgrade_src/ci/tasks/integration-tests.yml
    user: gpadmin
```
but as far as I can tell this doesn't actually get around the `chown` issue, because Concourse won't mount the volumes with permissions other than root[2].

[1] https://concourse-ci.org/tasks.html#task-run
[2] https://github.com/concourse/concourse/issues/403

ci/pipeline.yml
```diff
@@ -13,19 +13,49 @@ resources:
...
     url: {{cm_webhook_url}}
 
+- name: bin_gpdb_centos6
+  type: s3
+  source:
+    access_key_id: {{bucket-access-key-id}}
+    bucket: {{gpdb-stable-builds-bucket-name}}
+    region_name: {{aws-region}}
+    secret_access_key: {{bucket-secret-access-key}}
+    versioned_file: release_candidates/bin_gpdb_centos6/gpdb6/debug/bin_gpdb.tar.gz
```
doty-pivotal: Do we need to worry about the `.../debug/...` in the path?

kyeap-pivotal: IIRC I asked toolsmiths a similar question and it seems that It just means there would be debug capabilities. This needs to eventually change to trigger on every gpdb commit.

PR #53, opened 2018-07-31
Remove upgradestatus.Step

jchampio:
This popped back onto my plate since [the linter started pointing out how ugly it was](https://github.com/greenplum-db/gpupgrade/pull/49). :smile: This patchset gets rid of `Step` as an exported struct and adds smart helper functions to register step intent as well as implementation (i.e. "is this step writable and backed on the filesystem, or is it read-only with a custom status function?")

With this change, `shutdown-clusters` now blows up as expected, since it's not supposed to be writing to a read-only step. There's some additional small cleanup of code as well.

### DIFF
--- a/hub/gpupgrade_hub_main.go
+++ b/hub/gpupgrade_hub_main.go
@@ -55,46 +55,39 @@ func main() {
 
 			hub := services.NewHub(source, target, grpc.DialContext, conf, cm)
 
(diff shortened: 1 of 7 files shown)


jchampio:
Merged in 06089d3c735d0b160358e53f51614e9798785584.

PR #61, opened 2018-08-06
Fix and refactor init-cluster

jmcatamney:
Prepare init-cluster had a bug where it failed to connect to the target cluster after `init`ing it, so the target segment configuration was incorrect.  In the process of fixing the bug, we realized/decided that it was hard to test init-cluster in its current form, and so we did two things to address this:
- We refactored the configuration retrieval logic and addressed some TODOs; more detail can be found in the individual commit messages.
- We added an end-to-end `install-check` test that we could run (independently of the other BATS tests) to step through the gpupgrade process up to the latest working step as a sanity check that all of the steps were working together.

Feedback on both of these things would be appreciated, particularly regarding whether the team feels that `install-check` would be useful in its current form.


Makefile
```diff
@@ -103,6 +103,13 @@ install: cli-package hub-package install_agent
 		cp -p $(CLI) $(GPHOME)/bin/$(CLI)
 		cp -p $(HUB) $(GPHOME)/bin/$(HUB)
 
+install-check: install
+		@echo "---------------------------------------"
+		@echo "# FIXME: Running cmd directly from make hangs after test completes."
```
jchampio: See note below.

install-check.bats
```diff
@@ -0,0 +1,72 @@
...
+}
+
+# XXX: For now, leave system alone the end and push all cleanup into the setup().
+
+@test "init-cluster can successfully create the target cluster and retrieve its configuration" {
+    gpupgrade prepare init \
+              --new-bindir "$GPHOME"/bin \
+              --old-bindir "$GPHOME"/bin
+
+    gpupgrade prepare start-hub
```
jchampio: You'll need to close out FD 3 here, because we're forking a new child that will inherit it, and BATS will hang waiting for that child to exit at the end of the test. Here's an example from `start-hub.bats`:
```sh
    # The '3>&-' below is there because we must close fd 3 before forking new
    # processes in a BATS suite. For a full explanation, see
    #    https://github.com/bats-core/bats-core#file-descriptor-3-read-this-if-bats-hangs
    gpupgrade prepare start-hub 3>&-
```
You could also kill off the hub at the end of the test, I think.

doty-pivotal: Looks like we can't run make in parallel with the script in the current state...

hub/services/check_object_count.go
```diff
@@ -16,9 +15,7 @@ func (h *Hub) CheckObjectCount(ctx context.Context,
 
 	gplog.Info("starting CheckObjectCount")
 
-	masterPort := h.source.MasterPort()
-
-	dbConnector := db.NewDBConn("localhost", masterPort, "template1")
+	dbConnector := h.source.NewDBConn()
 	defer dbConnector.Close()
 	err := dbConnector.Connect(1)
```
jchampio: If the pattern is always "make a new connection, connect it", then maybe `Cluster.Connect()` would be a better API?

hub/services/prepare_init_cluster.go
```diff
@@ -9,116 +9,100 @@ import (
...
 )
 
-// TODO: consolidate with RetrieveAndSaveOldConfig(); it's basically the same
-// code
-func SaveTargetClusterConfig(target *utils.Cluster, dbConnector *dbconn.DBConn, stateDir string) error {
-	err := os.MkdirAll(stateDir, 0700)
+func (h *Hub) PrepareInitCluster(ctx context.Context, in *pb.PrepareInitClusterRequest) (*pb.PrepareInitClusterReply, error) {
+	gplog.Info("Running PrepareInitCluster()")
+
+	go h.AsyncInit()
```
jchampio: This is a step backwards. We want to separate step status handling from the business logic so eventually we can do something like
```go
hub.ExecuteAsyncStep(upgradestatus.INIT_CLUSTER, func() error {
    return h.InitCluster(dbConnector)
})
```
so that the Hub (or something it controls) can
- take care of step initialization
- set up panic logging on the goroutine
- mark the step complete or failed based on the returned error

and then we can still write unit tests around the implementation.

utils/cluster.go
```diff
@@ -121,3 +141,47 @@ func (c *Cluster) ExecuteOnAllHosts(desc string, cmd func(contentID int) string)
...
+		User:     username,
+		DBName:   "postgres",
+		Host:     c.MasterHost(),
+		Port:     c.MasterPort(),
+		Tx:       nil,
+		Version:  dbconn.GPDBVersion{},
+	}
+}
+
+func (c *Cluster) RefreshConfig(dbConnector *dbconn.DBConn) error {
```
jchampio: This feels like something that should be a constructor; i.e. you shouldn't have to create an unusable cluster and fill it in two separate steps.
```go
c := cluster.FromDB("localhost", 5432)
```
or
```go
dbConn := db.Connect("localhost", 5432)
c := cluster.FromDB(dbConn)
```
or something.

utils/cluster.go
```diff
@@ -19,6 +22,19 @@ type Cluster struct {
 	ConfigPath string
 }
 
+func NewMasterOnlyCluster(port int, host string, binDir string, configPath string) *Cluster {
+	master := cluster.SegConfig{
+		DbID:      1,
```
jchampio: We can't assume the DBID of a master is 1. AFAIK the only correct way to figure that out is to connect and retrieve it. But it seems like this master-only cluster thing might be able to disappear entirely if we remove the chicken-and-egg problem.

hub/services/check_config.go
```diff
@@ -40,24 +45,14 @@ func (h *Hub) CheckConfig(ctx context.Context, _ *pb.CheckConfigRequest) (*pb.Ch
 	return successReply, nil
 }
 
-// RetrieveAndSaveSourceConfig() fills in the rest of the clusterPair.OldCluster by
-// querying the database located at its host and port. The results will
-// additionally be written to disk.
-func RetrieveAndSaveSourceConfig(source *utils.Cluster) error {
-	dbConnector := db.NewDBConn("localhost", 0, "template1")
-	err := dbConnector.Connect(1)
+func initializeState(step upgradestatus.StateWriter) {
```
jchampio: Time to combine `ResetStateDir` and `MarkInProgress`, perhaps?

PR #62, opened 2018-08-06
Add shutdown-clusters to install-check

doty-pivotal:
Shutdown Clusters is currently working, so we can add to the `install-check`


install-check.bats
```diff
@@ -32,15 +32,26 @@ setup() {
...
-    echo "# Waiting for init to complete" 1>&3
+
+    EventuallyStepCompletes "Initialize upgrade target cluster"
+
+    gpupgrade prepare shutdown-clusters
+    EventuallyStepCompletes "Shutdown clusters"
+
+}
+
+EventuallyStepCompletes() {
```
jchampio: This, and the `sleep 1` above, makes me think we need a `--wait` option for steps.

doty-pivotal: @skahler-pivotal - would it be ok to re-evaluate if `start-agents` and `init-cluster` should be run asynchronously?

skahler-pivotal: What I had pictured was that by default they would run async.

A need to see things go to completion may force the need to return a job/action id and then be able to poll the status of that action.

doty-pivotal:
While re-roll'ing #61 I want these changes in from the beginning, so I will be cherry-picking the changes over, and closing this PR.

PR #63, opened 2018-08-09
Use correct username for share-oids

kyeap-pivotal:
Use correct username for share-oids

Share-oids was using gpadmin to copy share-oid files.
This caused an error because gpadmin user was not guaranteed
to exist. Changed to use the user used to be the environment
variable PGUSER.

Small refactor to the logic that determines current user while we were in there.

### DIFF
--- a/utils/cluster.go
+++ b/utils/cluster.go
@@ -6,8 +6,6 @@ import (
 
 	"github.com/greenplum-db/gp-common-go-libs/cluster"
 	"github.com/greenplum-db/gp-common-go-libs/dbconn"
-	"github.com/greenplum-db/gp-common-go-libs/gplog"
-	"github.com/greenplum-db/gp-common-go-libs/operating"
 	"github.com/pkg/errors"
 )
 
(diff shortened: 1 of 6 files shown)


utils/cluster.go
```diff
@@ -114,11 +112,11 @@ func (c *Cluster) PrimaryHostnames() []string {
 }
 
 // SegmentsOn returns the configurations of segments that are running on a given
-// host. An error will be returned for unknown hostnames.
+// host excluding the master. An error will be returned for unknown hostnames.
 func (c Cluster) SegmentsOn(hostname string) ([]cluster.SegConfig, error) {
```
jchampio: The `SegmentsOn()` changes seem to be a non sequitur? What do they have to do with the username?

doty-pivotal:
Getting rid of `gpadmin` as a hardcoded user +100.  

Here is what I am wondering.  The `PGUSER` variable tends to be used by tools to determine what **database user** the tool should connect to.  For example, we all know that `psql` assumes that the database user and the OS user are the same string, unless you indicate otherwise, through `PGPORT` or by passing a `-U` option upon invocation.

This `share-oids` step, as I understand it, is concerned with the **os user**, and not the database user.  Therefore, I am wondering if overloading the `PGPORT` is a bad idea.  Would it not be acceptable to, as a first pass, assume that the user was the same as of the process running these commands?  Or at least the same as the user that the Hub/Agent was started as?  It seems like that might match the OS convention that gives ownership of files by default to the user that created them.

utils/cluster.go
kyeap-pivotal: We squashed a commit into the wrong branch.  They're not related, we'll pull them out.

jmcatamney:
Because we're not going to use the database user, and the NewDBConn move to Cluster was reverted by #65, this PR can be closed; we'll fix the issue another way.

PR #65, opened 2018-08-10
Fix and refactor init-cluster, take 2

jchampio:
This is a partial re-roll of #61 , intended to unblock those who are waiting for `prepare init-cluster` to work. It does not include many of the API refactorings that Jim has been working on; they are left as an exercise for him 😄 

Note that `install-check` has been renamed `installcheck` to match the typical Makefile convention, and it does not install for you (since the point of `installcheck` is to [test what is installed](https://www.gnu.org/software/make/manual/html_node/Standard-Targets.html), not what is currently in your source tree).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -103,6 +103,13 @@ install: cli-package hub-package install_agent
 		cp -p $(CLI) $(GPHOME)/bin/$(CLI)
 		cp -p $(HUB) $(GPHOME)/bin/$(HUB)
 
+# We intentionally do not depend on install here -- the point of installcheck is
+# to test whatever has already been installed.
+installcheck:
+		@echo "--------------------------------------------------------------"
+		@echo "# FIXME: Make, if run in parallel, hangs after test completes."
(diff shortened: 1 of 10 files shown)


Makefile
```diff
@@ -103,6 +103,13 @@ install: cli-package hub-package install_agent
 		cp -p $(CLI) $(GPHOME)/bin/$(CLI)
 		cp -p $(HUB) $(GPHOME)/bin/$(HUB)
 
+# We intentionally do not depend on install here -- the point of installcheck is
+# to test whatever has already been installed.
+installcheck:
```
kyeap-pivotal: Since this has been changed to `installcheck`, the mention of `install-check` in the commit message should probably be changed too.

kyeap-pivotal: Also, we should rename `install-check.bats` to `installcheck.bats`.

jchampio: Good catch; I fixed it in one place but missed others. Re-rolled.

PR #66, opened 2018-08-13
Move closer to do-all: get as far as convert master

doty-pivotal:
Comments in commits tell the details of the story.  

The biggest change in design is that when the hub serviced the ConvertMaster request, it would attempt to `nohup` the `pg_upgrade` process.  The way this was implemented meant that we had difficulty knowing when the process failed.  We changed the pattern and "merely" execute the `ConvertMaster()` function from a goroutine, and have that function return an error.  This makes the `CONVERT_MASTER` step writeable.

### DIFF
--- a/hub/services/upgrade_convert_master.go
+++ b/hub/services/upgrade_convert_master.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"path/filepath"
 
+	"github.com/greenplum-db/gpupgrade/hub/upgradestatus"
 	pb "github.com/greenplum-db/gpupgrade/idl"
 	"github.com/greenplum-db/gpupgrade/utils"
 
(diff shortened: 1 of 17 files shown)


hub/services/upgrade_convert_master.go
```diff
@@ -7,40 +7,53 @@ import (
...
 func (h *Hub) ConvertMaster() error {
+	gplog.Info("Starting master upgrade")
+
 	pathToUpgradeWD := filepath.Join(h.conf.StateDir, upgradestatus.CONVERT_MASTER)
 	err := utils.System.MkdirAll(pathToUpgradeWD, 0700)
 	if err != nil {
-		errMsg := fmt.Sprintf("mkdir %s failed: %v. Is there an pg_upgrade in progress?", pathToUpgradeWD, err)
-		gplog.Error(errMsg)
-		return errors.New(errMsg)
+		return errors.Wrapf(err, "mkdir %s failed. Is there an pg_upgrade in progress?", pathToUpgradeWD)
```
jchampio: Not sure the "Is there an pg_upgrade in progress?" prompt makes sense here anymore -- `os.MkdirAll` will succeed if the directory already exists.

hub/services/upgrade_convert_master.go
```diff
@@ -49,14 +62,13 @@ func (h *Hub) ConvertMaster() error {
...
-	gplog.Info("Convert Master upgrade command: %#v", upgradeCmd)
+	gplog.Info("Convert Master upgrade command: %#v", pgUpgradeCmd)
 
-	//export ENV VARS instead of passing on cmd line?
-	err = utils.System.RunCommandAsync(upgradeCmd, pgUpgradeLog)
+	output, err := h.source.Executor.ExecuteLocalCommand(pgUpgradeCmd)
 	if err != nil {
-		gplog.Error("Error when starting the upgrade: %s", err)
-		return err
+		gplog.Error(output)
```
jchampio: Might be good to have some context ("pg_upgrade failed" or similar) printed in the log on the same line, in case `Info` isn't enabled.

PR #67, opened 2018-08-13
Fix share oids

doty-pivotal:
Previously we hardcoded the username to `gpadmin`, but now we are going to remove that, and rely on the fact that rsync will assume that the username corresponds to the user that owns the process, which in this case is the username used to spawn the hub.

This is a re-roll of #63, that was much smaller due to several previous refactors that have gone into master since it was opened the first time.

### DIFF
--- a/hub/services/upgrade_share_oids.go
+++ b/hub/services/upgrade_share_oids.go
@@ -36,13 +36,12 @@ func (h *Hub) shareOidFiles() {
 
 	hostnames := h.source.PrimaryHostnames()
 
-	user := "gpadmin"
 	rsyncFlags := "-rzpogt"
 	sourceDir := filepath.Join(h.conf.StateDir, upgradestatus.CONVERT_MASTER)
 
 	anyFailed := false
 	for _, host := range hostnames {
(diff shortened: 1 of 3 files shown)


hub/services/upgrade_convert_master.go
```diff
@@ -49,14 +62,13 @@ func (h *Hub) ConvertMaster() error {
...
-	gplog.Info("Convert Master upgrade command: %#v", upgradeCmd)
+	gplog.Info("Convert Master upgrade command: %#v", pgUpgradeCmd)
 
-	//export ENV VARS instead of passing on cmd line?
-	err = utils.System.RunCommandAsync(upgradeCmd, pgUpgradeLog)
+	output, err := h.source.Executor.ExecuteLocalCommand(pgUpgradeCmd)
 	if err != nil {
-		gplog.Error("Error when starting the upgrade: %s", err)
-		return err
+		gplog.Error(output)
```
slari-pivotal: Shouldnot we be loggoing the err rather than the output?

jmcatamney: In this case, `output` has the actual error message (as `ExecuteLocalCommand` calls `CombinedOutput`, which folds stderr into stdout), while err just has `exit status 1` or the like.

PR #68, opened 2018-08-14
run gpinitsystem using target not source binaries

doty-pivotal:
gpinitsystem has to be run against the target cluster's bin dir, or else
you get a properly configured cluster of the wrong GPDB version. This
commit makes that a reality.

Co-authored-by: Jim Doty <jdoty@pivotal.io>
Co-authored-by: Shoaib Lari <slari@pivotal.io>

### DIFF
--- a/hub/services/prepare_init_cluster_test.go
+++ b/hub/services/prepare_init_cluster_test.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"path/filepath"
 
 	"github.com/pkg/errors"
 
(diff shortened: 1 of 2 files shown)


hub/services/prepare_init_cluster_test.go
```diff
@@ -147,30 +152,42 @@ var _ = Describe("Hub prepare init-cluster", func() {
...
+
 		It("successfully runs gpinitsystem", func() {
 			testExecutor.LocalError = errors.New("exit status 1")
 			err := hub.RunInitsystemForNewCluster("filepath")
+
 			Expect(err).To(BeNil())
 			testhelper.ExpectRegexp(stdout, "[WARNING]:-gpinitsystem completed with warnings")
 		})
+
+		It("should set use executables is the source's bindir", func() {
```
jchampio: s/is/in/

PR #69, opened 2018-08-16
Demo fixes, prettified

jchampio:
Here is the list of fixes we made to get the demo working. As of the last commit, `make installcheck` will run through all steps for 5x, but not yet for 6x (since share-oids hasn't been fixed yet).

See commit messages for details. The really big changes are
- `utils.Cluster` has a new `Version` field
- `utils.ClusterFromDB()` is a new helper for the construction of a `utils.Cluster` using a `DBConn`.


doty-pivotal:
I took a first pass, nothing jumped out.  If this hangs around I will take a closer look.

hub/services/check_config.go
```diff
@@ -40,24 +41,19 @@ func (h *Hub) CheckConfig(ctx context.Context, _ *pb.CheckConfigRequest) (*pb.Ch
...
-// RetrieveAndSaveSourceConfig() fills in the rest of the clusterPair.OldCluster by
-// querying the database located at its host and port. The results will
-// additionally be written to disk.
-func RetrieveAndSaveSourceConfig(source *utils.Cluster) error {
-	dbConnector := db.NewDBConn("localhost", 0, "template1")
-	err := dbConnector.Connect(1)
+// ReloadAndCommitCluster() will fill in a utils.Cluster using a database
+// connection and additionally write the results to disk.
+func ReloadAndCommitCluster(cluster *utils.Cluster, conn *dbconn.DBConn) error {
+	newCluster, err := utils.ClusterFromDB(conn, cluster.BinDir, cluster.ConfigPath)
```
doty-pivotal: This function assumes that the cluster that we pass in has a `BinDir` and a `ConfigPath` set.  For now we do, but neither here, nor in `ClusterFromDB` do we check that the values are not zero valued,  Thoughts on adding that to one place or the other?

agent/services/convert_primary.go
```diff
@@ -2,57 +2,84 @@ package services
...
+	err := s.UpgradeSegments(in)
+
+	if err != nil {
+		return &pb.UpgradeConvertPrimarySegmentsReply{}, err
+	}
+
+	return &pb.UpgradeConvertPrimarySegmentsReply{}, nil
+}
+
+func (s *AgentServer) UpgradeSegments(in *pb.UpgradeConvertPrimarySegmentsRequest) error {
```
doty-pivotal: In other places we unpack the request and pass in the data to the helper function, as opposed to passing in the `pb.*Request` object.  Any thought on passing in the `in.Version` and `in.DataPairs`?

hub/services/check_config.go
jmcatamney: Those are filled in and saved all the way back in `prepare init` and that's unlikely to change, so it should be safe to assume that those are both set, but checking that they haven't been zeroed in the meantime might be a good idea.

agent/services/convert_primary.go
jmcatamney: No reason not to.  We can change that.

hub/services/check_config.go
doty-pivotal: I was thinking the same.

PR #70, opened 2018-08-16
Generate bash completion

jchampio:
It's easy! To get the goodies,

    source ./cli/bash/bash-completion.sh

Since the root `cobra.Command` is now needed from two executables, we've moved it into the `cli/commands` package.


chrishajas:
Does this also autocomplete flags?

jchampio:
@chrishajas Yep!

```sh
$ gpupgrade prepare init <tab><tab><tab>
--new-bindir=  --old-bindir=

nadeemg:
Can we add a line to the README also?

jchampio:
@nadeemg Done.

PR #71, opened 2018-08-20
Update prepare init-cluster message

kalensk:
Co-authored-by: Kalen Krempely <kkrempely@pivotal.io>

### DIFF
--- a/cli/commanders/preparer.go
+++ b/cli/commanders/preparer.go
@@ -71,7 +71,7 @@ func (p Preparer) InitCluster() error {
 		return err
 	}
 
-	gplog.Info("Gleaning the new cluster config")
+	gplog.Info("Starting new cluster initialization")
 	return nil
 }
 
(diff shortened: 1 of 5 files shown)


cli/commanders/preparer.go
```diff
@@ -71,7 +71,7 @@ func (p Preparer) InitCluster() error {
 		return err
 	}
 
-	gplog.Info("Gleaning the new cluster config")
+	gplog.Info("Attempting to initialize upgrade target cluster")
```
slari-pivotal: How about something like:

"Starting target cluster initialization for upgrade"

jchampio:
I thought we were keeping the old/new nomenclature for now?

jmcatamney:
@slari-pivotal @jchampio `Attempting to initialize upgrade target cluster` is the message requested in the story, so we went with that.  We don't have strong feelings on the verbiage, and would be fine with changing it to `Starting new cluster initialization` or similar.

@skahler-pivotal Would you prefer we keep the message as specified or change it?

skahler-pivotal:
Let's change it to fit what we had discussed and I think 

`Starting new cluster initialization`

is the better fit. We should also change the `status upgrade` message to match

`Initialize upgrade target cluster` -> `Initialize new cluster`

PR #73, opened 2018-08-24
Modify share-oids behavior based on GPDB version

jmcatamney:
The behavior of pg_upgrade with respect to OID files has changed in master, so
this commit modifies share-oids to either copy the OID files from the master to
the segments (for 4.3 and 5 clusters) or copy the entire directory over (for 6
clusters), preserving any necessary configuration files in the latter case.


agent/services/share_oids.go
```diff
@@ -0,0 +1,66 @@
...
+
+	"github.com/greenplum-db/gp-common-go-libs/gplog"
+	"path/filepath"
+)
+
+func (s *AgentServer) CopyMasterDirectoryToSegmentDirectories(ctx context.Context, in *pb.CopyMasterDirRequest) (*pb.CopyMasterDirReply, error) {
+	gplog.Info("got a request to copy the master data directory to the segment hosts from the hub")
+
+	// TODO: See if there additional files to preserve.
+	filesToPreserve := []string{"postgresql.conf", "pg_hba.conf"}
```
jchampio: `test_gpdb.sh` shows `postmaster.opts` and `gp_replication.conf` as additional preservation candidates.

agent/services/share_oids.go
```diff
@@ -0,0 +1,66 @@
...
+			err = errors.Wrapf(err, "Could not back up segment data directory")
+			break
+		}
+		// Overwrite the segment directory with the rsync'd master directory
+		output, err = s.executor.ExecuteLocalCommand(fmt.Sprintf("cp -a %s %s", masterDir, segDataDir))
+		if err != nil {
+			err = errors.Wrapf(err, "Could not copy master data directory to segment data directory: %s", output)
+			break
+		}
+		// Put segment config files back
```
jchampio: Daniel's test script also removed the following files:
- `gp_dbid`
- `gpssh.conf`
- `gpperfmon`

Do we still need that step, or have I missed it somewhere?

hub/services/upgrade_share_oids.go
```diff
@@ -66,3 +93,67 @@ func (h *Hub) shareOidFiles() {
...
+ * Generate a list of content IDs such that running ExecuteClusterCommand
+ * against them will execute once per host.
+ *
+ * TODO: Fix this logic once the cluster per-host execution logic is refactored.
+ */
+func contentsByHost(c *utils.Cluster) []int {
+	hostSegMap := make(map[string]int, 0)
+	for content, seg := range c.Segments {
+		if content == -1 {
+			continue
```
jchampio: nit: The exclusion of the master host isn't really clear from the documentation/name of the function.

hub/services/upgrade_share_oids.go
```diff
@@ -35,24 +37,49 @@ func (h *Hub) shareOidFiles() {
...
+		sourceDir := h.target.MasterDataDir()
+		commandMap := make(map[int][]string, len(h.target.ContentIDs)-1)
 
-		rsyncCommand := strings.Join([]string{"rsync", rsyncFlags, filepath.Join(sourceDir, "pg_upgrade_dump_*_oids.sql"), destinationDirectory}, " ")
-		gplog.Info("share oids command: %+v", rsyncCommand)
+		for _, content := range h.target.ContentIDs {
+			if content == -1 {
+				continue
+			}
+			destinationDirectory := h.target.GetHostForContent(content) + ":/tmp/masterDirCopy"
```
jchampio: Would the agent state directory be a better temp location? We know it exists already.

hub/services/upgrade_share_oids.go
```diff
@@ -66,3 +93,67 @@ func (h *Hub) shareOidFiles() {
...
+	segmentDataDirMap := map[string][]string{}
+	for _, content := range target.ContentIDs {
+		if content != -1 {
+			segment := target.Segments[content]
+			segmentDataDirMap[segment.Hostname] = append(segmentDataDirMap[segment.Hostname], target.GetDirForContent(content))
+		}
+	}
+
+	errMsg := "Error copying master data directory to segment data directories"
+	masterDir := target.MasterDataDir()
```
jchampio: Seems like this should be the temporary holding directory instead?

integrations/upgrade_share_oids_test.go
```diff
@@ -13,26 +18,50 @@ import (
...
 	BeforeEach(func() {
 		go agent.Start()
+		utils.System.Stat = func(name string) (os.FileInfo, error) {
+			return nil, nil
+		}
 	})
-	It("updates status PENDING to RUNNING then to COMPLETE if successful", func() {
 
+	It("updates status PENDING to RUNNING then to COMPLETE if successful for 5.X and earlier", func() {
+		source.Version = dbconn.GPDBVersion{
```
jchampio: nit: We can use `.NewVersion()` now. (here and elsewhere)

agent/services/share_oids.go
jmcatamney: Nope, you didn't miss it.  We'll remove those too.

hub/services/upgrade_share_oids.go
jmcatamney: Added an explanatory comment.

hub/services/upgrade_share_oids.go
jmcatamney: Whether we use `/tmp/masterDirCopy` or `~/.gpupgrade/masterDirCopy`, rsync is going to have to create `masterDirCopy`, and we thought it best not to put non-state-related things in the state directory.

hub/services/upgrade_share_oids.go
jmcatamney: Fixed.

integrations/upgrade_share_oids_test.go
jmcatamney: Good catch, this was branched before the versioning commit was merged.

hub/services/upgrade_share_oids.go
jchampio: Long-term, I think it would be wise for us to not rely on `/tmp`'s existence, availability, or security properties. It's not a gating concern for this particular PR, though.

jmcatamney:
@jchampio Just pushed a commit addressing your comments, and that also did some minor refactoring.  Take another look and let me know what you think.

hub/services/upgrade_share_oids_test.go
```diff
@@ -13,38 +15,46 @@ import (
...
 	BeforeEach(func() {
 		testExecutor = &testhelper.TestExecutor{}
+		mockOutput = &cluster.RemoteOutput{}
+		testExecutor.ClusterOutput = mockOutput
 		source.Executor = testExecutor
 	})
 
-	It("copies files to each primary host", func() {
+	It("copies the master data directory to each primary host in 6.0 or later", func() {
+		source.Version = dbconn.GPDBVersion{
```
jchampio: `.NewVersion()`

hub/services/upgrade_share_oids_test.go
```diff
@@ -13,38 +15,46 @@ import (
...
+		resultMap := testExecutor.ClusterCommands[0]
+		Expect(resultMap[0]).To(Equal([]string{"rsync", "-rzpogt", masterDataDir, "host1:/tmp/masterDirCopy"}))
+		Expect(resultMap[1]).To(Equal([]string{"rsync", "-rzpogt", masterDataDir, "host2:/tmp/masterDirCopy"}))
 	})
 
-	It("copies all files even if rsync fails for a host", func() {
-		testExecutor.LocalError = errors.New("failure")
-
+	It("copies files to each primary host in 5.X or earlier", func() {
+		source.Version = dbconn.GPDBVersion{
```
jchampio: `.NewVersion()`

hub/services/upgrade_share_oids.go
jchampio: I think this may have gotten dropped?

hub/services/upgrade_share_oids_test.go
jmcatamney: Hmm, looks like I forgot to `git add` these two and the function comment.  They're in now.

PR #74, opened 2018-08-27
Fixes for primary upgrade

jchampio:
Follow-ups to #73. See commit messages.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -55,6 +55,29 @@
   version = "v1.0"
 
 [[projects]]
+  name = "github.com/jackc/pgx"
+  packages = [
+    ".",
+    "chunkreader",
+    "internal/sanitize",
+    "pgio",
+    "pgproto3",
+    "pgtype",
+    "stdlib"
+  ]
+  revision = "89f1e6ac7276b61d885db5e5aed6fcbedd1c7e31"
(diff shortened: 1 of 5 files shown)


Gopkg.lock
```diff
@@ -54,6 +54,29 @@
   revision = "76626ae9c91c4f2a10f34cad8ce83ea42c93bb75"
   version = "v1.0"
 
+[[projects]]
```
jmcatamney: gpupgrade doesn't use `jackc/pgx` or `jmoiron/sqlx` directly anymore and they're not in the `Gopkg.toml` file, so should they be in the gpupgrade lock file at all or only in the gp-common-go-libs lock file?

jchampio: I think lock files are supposed to be transitively complete -- all the dependencies of the dependencies of the dependencies come in.

jchampio:
I added one more commit, which reenables `--progress` mode for pg_upgrade. (You'll need to update to a recent GPDB master tip.)

With that, `make installcheck` passes with GPDB 6 on my machine.

Gopkg.lock
jmcatamney: In that case, looks good to me.

jchampio:
Thanks for the review!

PR #75, opened 2018-08-28
end-to-end: fix race condition in start-hub tests

jchampio:
The end-to-end development pipeline was failing on the test

    start-hub does not return an error if an unrelated process has gpupgrade_hub in its name

well over 80% of the time. I've never seen it on anything except the CI,
and I couldn't ever reproduce when hijacked into Concourse, which was
maddening.

Turns out it's most likely a good old-fashioned race condition. Just
because the background process has forked, and we have a PID for it,
doesn't mean that its process name has been set yet. Unfortunately, that
means my current best effort solution is a poll on `ps -ef`.

There is some additional good news: as part of the headdesk ritual, I
found that job control (e.g. the thing that enables `kill %1`) isn't
really guaranteed to be enabled unless we're in an interactive terminal.
That seemed like a bad thing to rely on, so I removed the jobspecs,
along with all the tempfile complications, and replaced it with a single
`exec -a`. That reduces the moving parts in this test by quite a bit.

### DIFF
--- a/test/start-hub.bats
+++ b/test/start-hub.bats
@@ -58,19 +58,23 @@ teardown() {
 
 @test "start-hub does not return an error if an unrelated process has gpupgrade_hub in its name" {
     # Create a long-running process with gpupgrade_hub in the name.
-    tmpdir=`mktemp -d`
-    tmpfile="${tmpdir}/gpupgrade_hub_test_log"
-    touch $tmpfile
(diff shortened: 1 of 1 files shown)


jchampio:
Thanks for the review! Merged in 85bdc75290ca36ce2fc3be4caa2b38c0b5ee421d.

PR #79, opened 2018-09-04
CI: various improvements to help debug failures

jchampio:
See commit messages. These all helped when debugging the installcheck tests in the new pipeline.

### DIFF
--- a/installcheck.bats
+++ b/installcheck.bats
@@ -75,7 +75,7 @@ EventuallyStepCompletes() {
             break
         fi
 
-        sleep "$i"
+        sleep 1
     done
 
     [ "$observed_complete" != "false" ]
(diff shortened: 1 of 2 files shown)


installcheck.bats
```diff
@@ -60,7 +60,7 @@ EventuallyStepCompletes() {
     local observed_complete="false"
     for i in {1..60}; do
         run gpupgrade status upgrade
-        [ "$status" -eq 0 ]
+        [ "$status" -eq 0 ] || (echo "$output" && false)
```
nadeemg: If status is not 0, then echo output, ok. But what is the false doing?

jchampio: A non-zero exit code from any of these lines will stop the test, so `false` here is essentially "fail the test".

jchampio:
thanks for the review!

PR #80, opened 2018-09-04
Run BATS end-to-end tests and installcheck as part of the CI

jchampio:
The roadmap for these commits is as follows:
- Install BATS for the test jobs.
- Run `make check` as part of both jobs. This replaces the unit/integration distinction with noinstall/install, where we either have GPDB installed or we don't.
- Remove the custom TAP output we had going in the installcheck suite; it was interfering with the real BATS output and making things harder to debug. I've moved it to standard output instead.
- Run installcheck as part of the `install` test job.

The resulting pipeline is available [here](https://dev.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:dev:ci).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -109,7 +109,7 @@ install: cli-package hub-package install_agent
 installcheck:
 		@echo "--------------------------------------------------------------"
 		@echo "# FIXME: Make, if run in parallel, hangs after test completes."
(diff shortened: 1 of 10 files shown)


jchampio:
Thanks for the review!

PR #81, opened 2018-09-12
Fixes to status conversion and status upgrade

nadeemg:
See individual commit messages for details.


agent/services/conversion_status.go
```diff
@@ -14,21 +13,25 @@ func (s *AgentServer) CheckConversionStatus(ctx context.Context, in *pb.CheckCon
 	if len(in.GetSegments()) == 0 {
 		return nil, errors.New("no segment information was passed to the agent")
 	}
-	format := "%s - DBID %d - CONTENT ID %d - PRIMARY - %s"
+	//format := "%s - DBID %d - CONTENT ID %d - PRIMARY - %s"
```
jchampio: This can go away.

hub/services/status_conversion.go
```diff
@@ -79,15 +74,15 @@ func PrimaryConversionStatus(hub *Hub) pb.StepStatus {
...
-		return pb.StepStatus_PENDING
+	finalStatus := pb.StepStatus_PENDING
+	for _, status := range conversionStatus.GetConversionStatuses() {
+		if status.Status == pb.StepStatus_FAILED {
+			return pb.StepStatus_FAILED
+		} else if status.Status == pb.StepStatus_RUNNING {
+			finalStatus = pb.StepStatus_RUNNING
+		} else if status.Status == pb.StepStatus_COMPLETE && finalStatus != pb.StepStatus_RUNNING {
+			finalStatus = pb.StepStatus_COMPLETE
+		}
```
jchampio: With this algorithm, I think a status array of `[ PENDING, PENDING, COMPLETE ]` will return `COMPLETE`. (I guess that case needs to be added to the tests?)

hub/gpupgrade_hub_main.go
```diff
@@ -64,11 +64,11 @@ func main() {
 			// pull these into a Hub method or helper function, but currently the
 			// interfaces aren't well componentized.
 			cm.AddWritableStep(upgradestatus.CONFIG, pb.UpgradeSteps_CONFIG)
+			cm.AddWritableStep(upgradestatus.START_AGENTS, pb.UpgradeSteps_START_AGENTS)
```
jchampio: I thought we had decided to sort at the CLI level instead of the hub? That way you just have to rebuild the `gpupgrade` binary to change the UI, instead of reinstalling and restarting a hub.

idl/cli_to_hub.proto
```diff
@@ -65,11 +65,11 @@ message UpgradeStepStatus {
 enum UpgradeSteps {
     UNKNOWN_STEP = 0; // http://androiddevblog.com/protocol-buffers-pitfall-adding-enum-values/
     CONFIG = 1;
-    SEGINSTALL = 2;
-    INIT_CLUSTER = 3;
-    SHUTDOWN_CLUSTERS = 4;
-    CONVERT_MASTER = 5;
-    START_AGENTS = 6;
+    START_AGENTS = 2;
```
kyeap-pivotal: `start-agents` step happens before the `seginstall` step that checks if the agents are installed?

cli/commanders/reporter_test.go
```diff
@@ -16,6 +18,18 @@ import (
 	"github.com/onsi/gomega/gbytes"
 )
 
+var (
+	stdoutRead  *os.File
+	stdoutWrite *os.File
+	stdoutSaved *os.File
+)
+
+func getStdoutContents() string {
```
dkrieger-pvtl: I am wondering about the synchronization between the pipe writer and the pipe reader...that is, ensuring that all writes done by the Writer are read by the pipe Reader.  Although not in the official Go documentation, online reading suggests that, as an in-memory pipe, os.Pipe will "do the right thing" when the writer side Close "happens before" the reader side read.  So, especially for a test, this is enough for me.   Comments welcome.

hub/services/status_conversion.go
dkrieger-pvtl: I agree with @jchampio in this case.  Since our logic is really over the status of all agents taken as a whole, I'd prefer code that does this:
```
if (FAILED in statuses) return FAILED
else if (any RUNNING in status) return RUNNING
else if (all COMPLETE in status) return COMPLETE
else if (any PENDING in status) return COMPLETE
else panic('my world isn't sane')
```
There may be a cleaner way to do this, perhaps in a tagless switch.

agent/services/conversion_status.go
dkrieger-pvtl: comment removed as per request.

hub/gpupgrade_hub_main.go
jchampio: The change in code here is going to change the output order for `status upgrade`, whether the story covers it or not. My preference would be, if we're going to change the `status upgrade` order yet again, let's do it right this time.

jmcatamney:
@jchampio We've added CLI-side status ordering and some status-checking test cases.  Take a look at the new commits when you have a chance.

hub/services/status_conversion.go
```diff
@@ -74,15 +74,32 @@ func PrimaryConversionStatus(hub *Hub) pb.StepStatus {
...
+			hasFAILED = true
 		} else if status.Status == pb.StepStatus_RUNNING {
-			finalStatus = pb.StepStatus_RUNNING
-		} else if status.Status == pb.StepStatus_COMPLETE && finalStatus != pb.StepStatus_RUNNING {
-			finalStatus = pb.StepStatus_COMPLETE
+			hasRUNNING = true
+		} else if status.Status == pb.StepStatus_PENDING {
+			hasPENDING = true
+		} else if status.Status == pb.StepStatus_UNKNOWN_STATUS {
+			hasUKNOWN_STATUS = true
```
jchampio: s/UKNOWN/UNKNOWN/

But! You can get rid of the `hasUKNOWN_STATUS` entirely if you structure the `if` like this:
```
if status.Status == pb.StepStatus_COMPLETE {
    /* Set nothing. If no flags are set, everything completed. */
} else if status.Status == pb.StepStatus_RUNNING {
    hasRUNNING = true
} else if status.Status == pb.StepStatus_PENDING {
    hasPENDING = true
} else {
    /* Any other status means overall failure. */
    hasFAILED = true
}
```

That in turn lets you get rid of the `fallthrough` case below, which isn't really helping the logic.

hub/services/status_conversion_test.go
```diff
@@ -70,47 +71,40 @@ var _ = Describe("hub", func() {
...
+			[]*pb.PrimaryStatus{{Status: pb.StepStatus_COMPLETE}, {Status: pb.StepStatus_COMPLETE}},
+			pb.StepStatus_COMPLETE),
+		Entry("returns PENDING if no agents report any other state",
+			[]*pb.PrimaryStatus{{Status: pb.StepStatus_PENDING}, {Status: pb.StepStatus_PENDING}},
+			pb.StepStatus_PENDING),
+		Entry("returns PENDING if PENDING,PENDING,COMPLETE",
+			[]*pb.PrimaryStatus{{Status: pb.StepStatus_PENDING}, {Status: pb.StepStatus_PENDING},{Status: pb.StepStatus_COMPLETE}},
+			pb.StepStatus_PENDING),
+		Entry("returns RUNNING if PENDING,RUNNING,COMPLETE",
+			[]*pb.PrimaryStatus{{Status: pb.StepStatus_PENDING}, {Status: pb.StepStatus_RUNNING},{Status: pb.StepStatus_COMPLETE}},
```
jchampio: The code might need a `go fmt`.

hub/services/status_conversion.go
```diff
@@ -79,15 +74,32 @@ func PrimaryConversionStatus(hub *Hub) pb.StepStatus {
...
-	case strings.Contains(statuses, "COMPLETE"):
-		return pb.StepStatus_COMPLETE
-	default:
+	case hasPENDING:
 		return pb.StepStatus_PENDING
+	default:
+		return pb.StepStatus_COMPLETE
 	}
+
+	return pb.StepStatus_UNKNOWN_STATUS
```
jchampio: This code is unreachable.

hub/services/status_conversion.go
jmcatamney: Good idea, we'll do that.

hub/services/status_conversion.go
jmcatamney: It is, but Go requires a return statement at the end of a function even if you'll never get there. -_- We'll see if we can tweak the function to avoid doing that.

cli/commanders/reporter.go
```diff
@@ -42,6 +42,55 @@ func NewReporter(client pb.CliToHubClient) *Reporter {
 	}
 }
 
+type PrimaryStatuses []*pb.PrimaryStatus
```
dkrieger-pvtl: Style comment:  having the rpc code(OverallUpgradeStatus and OverallConversionStatus) in the same file as the UpgradeStepsMessage and UpgradeSteps order(both of which are presentation) seems incohesive.  I realize we don't want to make those two package-visible but having "active code" with logical ordering in the same file makes it harder for me to grok the code.

jchampio:
> do we need to support mix-and-match (cli,hub,agents) from different "versions"

Currently the decision is "no": we will keep all three executables in lockstep. This will need to change if we ever decide to support external clients, but we haven't decided to yet (and if that happens I assume we will want to consider not only protobuf versioning schemes, but gRPC endpoint versioning as well).

hub/services/status_conversion.go
jchampio: Spec says that a `switch` statement that returns from all branches counts as a [terminating statement](https://golang.org/ref/spec#Terminating_statements). Similar test code compiles just fine for me...?

jmcatamney: It was the linter that was complaining, not the compiler, but Kalen and I weren't able to repro that today so we took it out.

PR #82, opened 2018-09-13
gpinitsystem looked for greenplum_path.sh in wrong directory

dkrieger-pvtl:
trivial fix for setting the directory for the greenplum_path.sh file....

### DIFF
--- a/hub/services/prepare_init_cluster.go
+++ b/hub/services/prepare_init_cluster.go
@@ -207,7 +207,7 @@ func (h *Hub) CreateAllDataDirectories(agentConns []*Connection, segmentDataDirM
 
 func (h *Hub) RunInitsystemForNewCluster(gpinitsystemFilepath string) error {
 	// gpinitsystem the new cluster
-	gphome := filepath.Dir(h.target.BinDir)
+	gphome := filepath.Dir(path.Clean(h.target.BinDir))   //works around https://github.com/golang/go/issues/4837 in go10.4
 	cmdStr := fmt.Sprintf("source %s/greenplum_path.sh; %s/gpinitsystem -a -I %s",
 		gphome,
 		h.target.BinDir,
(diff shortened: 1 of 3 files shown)


hub/services/prepare_init_cluster.go
```diff
@@ -208,7 +208,7 @@ func (h *Hub) CreateAllDataDirectories(agentConns []*Connection, segmentDataDirM
 func (h *Hub) RunInitsystemForNewCluster(gpinitsystemFilepath string) error {
 	// gpinitsystem the new cluster
 	gphome := filepath.Dir(h.target.BinDir)
-	cmdStr := fmt.Sprintf("source %s/greenplum_path.sh; %s/gpinitsystem -a -I %s",
+	cmdStr := fmt.Sprintf("source %s/../greenplum_path.sh; %s/gpinitsystem -a -I %s",
```
jchampio: Something seems wrong here. The call to `filepath.Dir()` was intended to go "one directory up", but I think `BinDir` isn't being cleaned correctly first and so you may be running into https://github.com/golang/go/issues/4837.

The agent uses `filepath.Join(newBinDir, "..", "greenplum_path.sh")`, both to avoid this problem, and to avoid hardcoding file slash conventions.

dkrieger-pvtl: Something goofy is going on as it appears not to be the reference bug...but I'll add some print outs to see what our gpupgrade code sees....

```
○ → go version
go version go1.10.4 darwin/amd64

○ → which go
/usr/local/go/bin/go

○ → cat dirtest.go 
package main

import (
	"fmt"
	"path/filepath"
)
func main() {
	x := "mypath/dir1/dir2/"
	t := filepath.Dir(x)
	fmt.Println("HI MOM! ", x, "**", t)
}


 
○ → go run dirtest.go 
HI MOM!  mypath/dir1/dir2/ ** mypath/dir1/dir2
```

dkrieger-pvtl:
Hopefully this addresses your issues; we are working around the Go 1.10.4 filename.Dir() bug directly.

PR #83, opened 2018-09-24
Fix Linting & Lockfile

kalensk:
see individual commit messages


d:
"eye before ee except after cee" @jmcatamney @kalensk

Makefile
```diff
@@ -45,7 +45,7 @@ integration:
 
 # check runs all tests. Use -k to keep going after the first failure.
 .PHONY: check
-check:
+check: lint
```
jchampio: I'm all for regular (un-ignorable) linting as part of dev, but assuming gometalinter returns nonzero for issues it finds, if an open source user installs a version of gometalinter that is better (stricter) than the one we're testing with, they'll be unable to run the unit tests.

I've seen this backfire before... are we sure we want to adopt that sort of strategy long-term? (My proposed alternative would be to add it to our CI, just not as part of `make check`. Or at the very least make it equal to the unit tests in priority rather than a prereq.)

jchampio: > if you're running into a linter issue you don't want to fix right now you can use those other targets

`make unit integration` won't run the end-to-end, and though we can add yet another target for that I think it would be missing the point. `make check` should ideally tell you if the software works (within reason), and to do that it should run all the tests that are possible. "Your code doesn't subjectively lint and therefore I won't tell you whether it objectively works or not" isn't really an appropriate behavior for `make check`.

> linter errors started accumulating, so we feel that adding it back to check is worthwhile.

I agree that we need to add linting to the dev feedback cycle, and it should be as in-your-face as possible. But it shouldn't gate running the tests.

> What kind of backfires have you seen,

One big one is external users (who are not developers) ignoring failures in `make check` because "it's just the linter; it doesn't affect correctness". The first time we tell them to do that, the power and trust in `make check` goes right out the window. Don't make users (and newcomers to the project) determine which failures are more important than others; they're not the experts.

dkrieger-pvtl: I am not aware of any industry standard for testing target names, but gnu has this document: https://www.gnu.org/software/make/manual/html_node/Standard-Targets.html for their programs.  It has "make check" and "make installcheck", but note those explicitly don't include building the program(which I disagree with for us).   There's the usual "unit", "subsystem", "system", "integration" hierarchy from software engineering, so our "make check"(kind of unit and subsystem combined) makes sense. 

So we should align our Makefile testing targets to something standard, though I'm not sure what that would be.  We can look into that further.  For the issue at hand, I think that lint'ing is a dev activity that should NOT be part of 'make check' (unit testing which tests functionality, not conformance to the strict Golang specification).  But we should strive to be strictly conforming to (someone's reasonable) interpretation of the Golang spec in development.

dkrieger-pvtl:
https://github.com/greenplum-db/gpupgrade/pull/94 fixes the `make format` and `make lint` targets and deprecates this.  We still need to consider whether to have a test that fails when either of those targets fails, but that's a future discussion.  They are really "commit test" targets and perhaps are better suited to an integration pipeline test.

PR #87, opened 2018-10-13
add a README_cli.md to descibe the CLI.

dkrieger-pvtl:
A start at a document so that everyone is using consistent instructions for running gpupgrade.

### DIFF
--- a/README_cli.md
+++ b/README_cli.md
@@ -1,0 +1,260 @@
+# gpupgrade command-line interface(cli)
+
+See the accompanying ```README.md``` for more context on developing for gpugrade.
+This document discusses gpupgrade from the standpoint of someone using gpupgrade to
+actually upgrade their cluster.
+
+### Summary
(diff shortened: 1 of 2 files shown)


README_cli.md
```diff
@@ -0,0 +1,260 @@
...
+
+1. ```gpupgrade upgrade convert-primaries```
+
+1. ```gpupgrade upgrade validate-start-cluster```
+
+1. ```gpupgrade upgrade reconfigure-ports```
+
+###TODO
+
+1. why is ```Install binaries on segments``` always pending?
```
jchampio: The `TODO` section is likely to go stale almost immediately; might be best to push these things into GitHub issues instead.

dkrieger-pvtl: probably, but since I think I might be the only one still working on this(in my 10% time for a while), I'll just leave it there.

PR #89, opened 2018-11-06
CI: add a 6-to-6 upgrade test job

jchampio:
Perform a 6-to-6 self-upgrade, using the ICW sqldump generated by the GPDB master pipeline.

This implementation is similar to that of the pg_upgrade test job in the GPDB pipeline, but instead of manually coordinating pg_upgrade using a Bash script, it uses gpupgrade directly. The upgrade steps are taken from `installcheck.bats`.

Since there hasn't been a healthy RC build for a few months now, we relaxed the requirement on GPDB binaries, so now we just grab the latest build from the production bucket. (Note that, currently, there is no guarantee that the sqldump and GPDB binary artifact belong to the same build. There are some ideas floating around on how to fix this situation, including building our own, or coordinating versions using a new custom resource type.)

Testing against a recent GPDB build exposed a small bug that has also been fixed here: the `gp_replication.conf` file no longer exists now that we have `ALTER SYSTEM` from upstream. Since this was not backported to 5X, we can remove it from gpupgrade.


jmcatamney:
It's not letting me make an inline comment for some reason, but for this line in `cluster-upgrade.bash`:
```
sleep 1 # XXX make the above synchronous 
```
you can just use `wait_for_step "Agents Started on Cluster"` for that; it should still only sleep for 1 second, but won't need any changes once we make that step synchronous.

ci/pipeline.yml
```diff
@@ -80,8 +118,104 @@ jobs:
...
+    <<: *ccp_destroy
+
+anchors:
+- &ccp_default_params
+  action: create
+  delete_on_failure: true
+  generate_random_name: true
+  terraform_source: ccp_src/google/
+
+- &ccp_default_vars
```
doty-pivotal: Since there are only 2 defaults here and we are overriding 1 when we use the anchor, lets either not use the anchor, or pull the overrides down into this block.

kalensk:
I wonder if long term it would make sense to have a "library" of common functions since there seems to be a decent amount of duplicated work in cluster-upgrade.bash in gpupgrade and test_upgrade.bash in gpdb. Granted they are in different repos.

https://github.com/greenplum-db/gpupgrade/blob/e47fc62231c83a53dc54e4fe0f8f08084d419e98/ci/scripts/cluster-upgrade.bash

https://github.com/greenplum-db/gpdb/blob/master/concourse/scripts/test_upgrade.bash

jchampio:
> you can just use wait_for_step "Agents Started on Cluster" for that

@jmcatamney I seem to remember that we tried that for `make installcheck` but it didn't work. I can't remember what that reason was, though.

> but won't need any changes once we make that step synchronous.

We'll be removing `wait_for_step` once they're synchronous, right?

jchampio:
> I wonder if long term it would make sense to have a "library" of common functions since there seems to be a decent amount of duplicated work in cluster-upgrade.bash in gpupgrade and test_upgrade.bash in gpdb.

@kalensk In this case, we've duplicated the "inflate and load sqldump" code, the "dump all" code, and the "compare diffs " code. All are small (<10 lines) and since both have to be executed on other hosts -- so you'd need to either transfer the library over scp or include the ssh loop in the library API -- I'm not sure we want to pull the trigger on deduplication just yet. We should keep an eye on it, though.

ci/pipeline.yml
jchampio: Will do.

jchampio: I went for the former, since defaulting to a larger VM type seemed like it could be accidentally expensive...

PR #90, opened 2018-12-06
gpupgrade version now prints out the correct version

dkrieger-pvtl:
In addition, we add in a test to make sure the version is reasonable.


kalensk:
Looks good.

PR #91, opened 2018-12-06
add --version commandline flag for gpupgrade_agent and gpupgrade_hub

dkrieger-pvtl:
This allows us to test the installed version in a reasonable way.
Testing for these is also added.

The following unrelated changes are also included:
  --add additional comments for flying a pipeline for gpupgrade
  --increase the pipeline timeout for 'make installcheck' to 5 minutes


installcheck.bats
```diff
@@ -59,12 +59,12 @@ EventuallyStepCompletes() {
     cliStepMessage="$1"
     echo "# Waiting for \"$cliStepMessage\" to transition to complete"
     local observed_complete="false"
-    for i in {1..60}; do
+    for i in {1..300}; do
```
jchampio: This increases the timeout for all steps, not just the one that's currently flaky. So we can do this, but it's now critical that we get synchronous steps into the tool so that we can get rid of this ASAP.

kalensk: Probably obvious, but is there a middle ground where each step can optionally pass in its own timeout, and if not it defaults to 60?

jchampio: Right, but now we're losing arguably the only benefit to the timeout, which was to warn us if something started taking a ridiculously long time. If any of these steps are taking five minutes on an empty database, we *should* be failing. (Eventually, something else in the framework should help us figure out when steps start degrading in performance, but we don't have anything to do that yet.)

dkrieger-pvtl: Right, I get that, but I think perf testing should be handled separately.  The current test as designed must choose the "max time".  Even 60 seconds is way too long for some operations. I'd rather think of a way to bound the time and measure perf better.

So, for now, I'd prefer to push this change to get the pipeline passing functionally and handle the perf test separately.

jchampio: I think we're agreed then; that's what I meant when I said

> So we can do this, but it's now critical that we get synchronous steps into the tool so that we can get rid of this ASAP.

I just don't want to carry this timeout forever; this is how tests rot.

utils/version.go
```diff
@@ -0,0 +1,22 @@
...
+
+import "github.com/spf13/cobra"
+
+// This global var UpgradeVersion should have a value set at build time.
+// see Makefile for -ldflags "-X etc"
+var UpgradeVersion = ""
+
+type VersionCommand struct{}
+
+const defaultUpgradeVersionSuffix = " unknown version"
```
jchampio: Do we use this for anything else, or can we just replace its use down in line 15 with the literal string?

utils/version.go
```diff
@@ -0,0 +1,22 @@
+package utils
+
+import "github.com/spf13/cobra"
+
+// This global var UpgradeVersion should have a value set at build time.
+// see Makefile for -ldflags "-X etc"
+var UpgradeVersion = ""
+
+type VersionCommand struct{}
```
jchampio: What's this?

agent/gpupgrade_agent_main.go
```diff
@@ -22,15 +23,23 @@ func main() {
 	//}
 	var logdir, statedir string
 	var shouldDaemonize bool
+	var doLogVersionAndExit bool
 
 	var RootCmd = &cobra.Command{
 		Use:   "gpupgrade_agent ",
 		Short: "Start the Command Listener (blocks)",
 		Long:  `Start the Command Listener (blocks)`,
+		Args:  cobra.MaximumNArgs(0), //no positional args allowed
```
jchampio:
```go
Args: cobra.NoArgs
```
is probably the way to go.

hub/gpupgrade_hub_main.go
```diff
@@ -23,16 +24,24 @@ import (
 func main() {
 	var logdir string
 	var shouldDaemonize bool
+	var doLogVersionAndExit bool
```
jchampio: `printVersion` maybe?

utils/version.go
```diff
@@ -0,0 +1,22 @@
...
+const defaultUpgradeVersionSuffix = " unknown version"
+
+func VersionString(executableName string) string {
+	if UpgradeVersion == "" {
+		return executableName + defaultUpgradeVersionSuffix
+	}
+	return executableName + " version " + UpgradeVersion
+}
+
+func VersionAddCmdlineOption(cmd *cobra.Command, doLogVersionAndExit *bool) {
```
jchampio: The shared name prefix is begging to be pulled into a `version` package. `version.String()`, `version.AddFlag()`...

hub/gpupgrade_hub_main.go
```diff
@@ -23,16 +24,24 @@ import (
...
 		Long:  `Start the gpupgrade_hub (blocks)`,
+		Args:  cobra.MaximumNArgs(0), //no positional args allowed
 		RunE: func(cmd *cobra.Command, args []string) error {
 			gplog.InitializeLogging("gpupgrade_hub", logdir)
 			debug.SetTraceback("all")
 			defer log.WritePanics()
 
+			if doLogVersionAndExit {
+				fmt.Println(utils.VersionString("gpupgrade_hub"))
+				gplog.Info(utils.VersionString("gpupgrade_hub"))
```
jchampio: IMO this isn't log-worthy. Thoughts?

utils/version.go
kalensk: A good candidate for gp-common-go-libs perhaps?

utils/version.go
kalensk: I just tested locally, but seems like we can just initialize UpgradeVersion to the default string, which will get overwritten with the correct value. This, eliminates the need for the if statement and all that extra stuff.

`var UpgradeVersion = "unknown version"`

utils/version.go
dkrieger-pvtl: I presume you mean the VersionCommand struct...no idea.  I suppose `git blame` would show it's cruft left over from an old way of doing this.

utils/version.go
dkrieger-pvtl: True, this makes sense.

agent/gpupgrade_agent_main.go
dkrieger-pvtl: Yeah, I tried that first, but the usage is different.  cobra.NoArgs signature takes a command and the args string slice....see https://github.com/spf13/cobra#positional-and-custom-arguments.  The current usage here seemed easier to understand.

hub/gpupgrade_hub_main.go
dkrieger-pvtl: I like boolean types to begin with a "boolean-like" verb: is, do, if, etc. Maybe `doPrintVersion`?

utils/version.go
dkrieger-pvtl: Sure, this seemed a bit more complicated, but I could do this...maybe keep it local to gpupgrade since we don't have good guidance on what all clis should do?

hub/gpupgrade_hub_main.go
dkrieger-pvtl: I explicitly put it in the log...I actually meant to always log the version as a matter of course.  It's really useful for debugging to know what version a system is running.  I think it's best done as:
```
gplog.Info(versionInfo)
if (doLogVersionAndExit) {
   fmt.Println(versionInfo)
}
```

dkrieger-pvtl:
As for cobra's Command.Version, yes, that might help.  But the current by-hand usage is pretty simple as it is, so I didn't look at it before your suggestion.  I think the hardest part is that we use a git tag, and that must be done manually.  I like the idea of a version like "v1.5.8_2018_Nov_22_15_33_15_githash".  But for the issue at hand, I just wanted to put in a version change scheme.

agent/gpupgrade_agent_main.go
jchampio: You pass `NoArgs` itself, rather than calling it. I.e.
```go
&cobra.Command {
    Args: cobra.NoArgs,
}
```
`cobra.MaximumNArgs(0)` itself returns a helper function, which is called later.

hub/gpupgrade_hub_main.go
jchampio: Ah, I see. I used `should` as my "boolean prefix" for the one above it. `do` isn't a truthy prefix for me personally.

utils/version.go
jchampio: I personally prefer local first, then bubbling up once there is broad consensus, yeah.

agent/gpupgrade_agent_main.go
dkrieger-pvtl: Ah, I see....yes, your suggestion is simpler.

hub/gpupgrade_hub_main.go
jchampio: I agree with logging the version on server startup, for instance, but I don't think we should print to the log every time someone runs `--version` -- at that point it's just noise. The user was just told what the version was, after all.

hub/gpupgrade_hub_main.go
dkrieger-pvtl: how about 'shouldPrintVersion'?

jchampio: I like it.

PR #94, opened 2018-12-18
code clean up: formatting and linting

dkrieger-pvtl:
Modify the `make format` command to use `goimports` followed by `gofmt` and modify `make lint` to not do any code formatting.  Use the results of these tools to correct the code, either by adding line-level exceptions for gometalinter or modifying the code.

The intended workflow for development is to run `make format` and `make lint` and correct code accordingly before each commit.  This does not interact with testing the code.


kalensk:
Looks good. Hopefully we can add make format and make lint to [the CI job or something.](https://github.com/greenplum-db/gpupgrade/pull/83#discussion_r220304327)

dkrieger-pvtl:
> Looks good. Hopefully we can add make format and make lint to [the CI job or something.](https://github.com/greenplum-db/gpupgrade/pull/83#discussion_r220304327)

Yeah, I thought about it for this PR, but this one is really "get fmt and lint running" and correct current issues; we can decide separately if we want to add a pipeline task to fail a build if there are lint/fmt warnings.  I think we probably do, but that can be a separate PR.

PR #96, opened 2019-07-31
Fix pipeline and update code to work with latest version of go

slari-pivotal:
We have made the following changes to gupgrade.
- Use the correct credentials depending on the fly target
- Use the latest docker image for terraform resource that works
- use bin_gpdb from a bucket that is regularly built
- update gupgrade go code to be compatible with go 1.12

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -147,15 +147,18 @@ GIT_URI := $(shell git ls-remote --get-url)
 ifeq ($(GIT_URI),https://github.com/greenplum-db/gpupgrade)
 ifeq ($(BRANCH),master)
 PIPELINE_NAME := gpupgrade
-DEPLOY_TYPE := prod
-FLY_TARGET := gpdb-prod
+FLY_TARGET := prod
 endif
 endif
 
 # Concourse does not allow "/" in pipeline names
 PIPELINE_NAME ?= gpupgrade:$(shell git rev-parse --abbrev-ref HEAD | tr '/' ':')
-DEPLOY_TYPE ?= dev
-FLY_TARGET ?= gpdb-dev
+FLY_TARGET ?= dev
+ifeq ($(FLY_TARGET),prod)
+SECRETS_TYPE := prod
+else
+SECRETS_TYPE := dev
+endif
 
 .PHONY: set-pipeline expose-pipeline
 # TODO: Keep this in sync with the README at github.com/greenplum-db/continuous-integration
@@ -164,10 +167,10 @@ set-pipeline:
 	#NOTE-- such as https://github.com/greenplum-db/gpupgrade.git"
 	fly -t $(FLY_TARGET) set-pipeline -p $(PIPELINE_NAME) \
 		-c ci/pipeline.yml \
-		-l ~/workspace/gp-continuous-integration/secrets/gpupgrade.$(DEPLOY_TYPE).yml \
+		-l ~/workspace/gp-continuous-integration/secrets/gpupgrade.$(SECRETS_TYPE).yml \
 		-l ~/workspace/gp-continuous-integration/secrets/gpdb_common-ci-secrets.yml \
 		-l ~/workspace/gp-continuous-integration/secrets/gpdb_master-ci-secrets.prod.yml \
-		-l ~/workspace/gp-continuous-integration/secrets/ccp_ci_secrets_gpdb-$(DEPLOY_TYPE).yml \
+		-l ~/workspace/gp-continuous-integration/secrets/ccp_ci_secrets_$(FLY_TARGET).yml \
 		-v gpupgrade-git-remote=$(GIT_URI) \
 		-v gpupgrade-git-branch=$(BRANCH)
 
--- a/agent/services/agent_server.go
+++ b/agent/services/agent_server.go
@@ -80,7 +80,7 @@ func (a *AgentServer) Start() {
 
 	err = server.Serve(lis)
 	if err != nil {
-		gplog.Fatal(err, "failed to serve", err)
+		gplog.Fatal(err, "failed to serve: %s", err)
 	}
 
 	a.stopped <- struct{}{}
--- a/ci/pipeline.yml
+++ b/ci/pipeline.yml
@@ -1,5 +1,10 @@
 ---
 resource_types:
+- name: gcs
+  type: docker-image
+  source:
+    repository: frodenas/gcs-resource
+
 - name: slack-notification
   type: docker-image
   source:
@@ -38,13 +43,11 @@ resources:
     url: {{cm_webhook_url}}
 
 - name: bin_gpdb_centos6
-  type: s3
+  type: gcs
   source:
-    access_key_id: {{bucket-access-key-id}}
-    bucket: {{bucket-name}}
-    region_name: {{aws-region}}
-    secret_access_key: {{bucket-secret-access-key}}
-    versioned_file: {{bin_gpdb_centos_versioned_file}}
+    bucket: {{gcs-bucket}}
+    json_key: {{concourse-gcs-resources-service-account-key}}
+    regexp: server/published/gpdb6/server-rc-(.*)-rhel6_x86_64.debug.tar.gz
 
 - name: ccp_src
   type: git
--- a/ci/pipeline.yml
+++ b/ci/pipeline.yml
@@ -10,6 +10,7 @@ resource_types:
   type: docker-image
   source:
     repository: ljfranklin/terraform-resource
+    tag: 0.11.14
 
 resources:
 - name: gpupgrade_src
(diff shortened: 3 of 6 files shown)


ci/pipeline.yml
```diff
@@ -40,10 +41,10 @@ resources:
   type: s3
   source:
     access_key_id: {{bucket-access-key-id}}
-    bucket: {{bucket-name}}
+    bucket: {{gpdb-stable-builds-bucket-name}}
     region_name: {{aws-region}}
     secret_access_key: {{bucket-secret-access-key}}
-    versioned_file: {{bin_gpdb_centos_versioned_file}}
+    versioned_file: release_candidates/bin_gpdb_centos6/gpdb6/prod/bin_gpdb.tar.gz
```
kalensk: I don't know enough about the entire process, but are we sure we want to be hard coding this instead of using the variable like it was before?

ci/pipeline.yml
```diff
@@ -10,12 +10,13 @@ resource_types:
...
   source:
     repository: ljfranklin/terraform-resource
+    tag: 0.11.14
 
 resources:
 - name: gpupgrade_src
   type: git
   source:
-    uri: {{gpupgrade-git-remote}}
+    uri: https://github.com/greenplum-db/gpupgrade
```
kalensk: Just an observation, if we ever need to run a test pipeline on our own remotes, I am assuming we would want this to be a variable. That may be future proofing and not needed now.

slari-pivotal: Good point. We will fix it.

ci/pipeline.yml
marksliva: We would like someone from releng to review this section and let us know if it is correct @kmacoskey

jchampio: Does this prevent us from testing artifacts unless all tests on master are green? If so, I think we decided against that back when we initially set up the pipeline. IMO we want to run gpupgrade on the latest stable build regardless of whether it is a release candidate.

ci/pipeline.yml
jchampio: The `Makefile` already has code to handle this, FYI.

jchampio: (see the `set-pipeline` recipe)

jchampio: I see that this was already reverted -- sorry, GitHub didn't tell me fast enough that it was obsolete...

Makefile
```diff
@@ -155,7 +155,12 @@ endif
 # Concourse does not allow "/" in pipeline names
 PIPELINE_NAME ?= gpupgrade:$(shell git rev-parse --abbrev-ref HEAD | tr '/' ':')
 DEPLOY_TYPE ?= dev
```
jchampio: `DEPLOY_TYPE` is probably a misnomer now with the recent changes in GPDB. Can it be merged/replaced with `FLY_TARGET`?

jchampio: (@kalensk worked on those changes I think)

ci/pipeline.yml
```diff
@@ -10,6 +10,7 @@ resource_types:
   type: docker-image
   source:
     repository: ljfranklin/terraform-resource
+    tag: 0.11.14
```
jchampio: It'd be good to have an explanation in the commit message re: the Terraform compatibility issue.

Makefile
kalensk: Good call. It might make sense to use the same or similar terminology as gen_pipeline which currently calls it `pipeline_target`.

agent/services/agent_server.go
```diff
@@ -80,7 +80,7 @@ func (a *AgentServer) Start() {
 
 	err = server.Serve(lis)
 	if err != nil {
-		gplog.Fatal(err, "failed to serve", err)
+		gplog.Fatal(err, "failed to serve: %s", err)
```
jchampio: Is this really an introduced Go incompatibility, or just an improvement to `go vet`?

slari-pivotal: We were getting compilation errors without this change, so it is not just a `go vet` issue.

ci/pipeline.yml
marksliva: We pushed an empty squash commit with our new message

Makefile
marksliva: We removed DEPLOY_TYPE. With this change, people's local FLY_TARGETS must be kept in sync with the filenames in gp-continuous-integration which is something that people should already be doing.

Makefile
```diff
@@ -164,10 +167,10 @@ set-pipeline:
 	#NOTE-- such as https://github.com/greenplum-db/gpupgrade.git"
 	fly -t $(FLY_TARGET) set-pipeline -p $(PIPELINE_NAME) \
 		-c ci/pipeline.yml \
-		-l ~/workspace/gp-continuous-integration/secrets/gpupgrade.$(DEPLOY_TYPE).yml \
+		-l ~/workspace/gp-continuous-integration/secrets/gpupgrade.$(PROD_OR_DEV).yml \
```
kalensk: I am a bit confused, why is this not just `FLY_TARGET`? Why are there two variables `PROD_OR_DEV` as well as `FLY_TARGET` instead of just one? Sure this might mean we "force" people to use `fly -t dev` instead of `fly -t gpdb-dev`, but that seems fine to me - less typing and of course we are in gpdb...

Am I miss understanding something about all this?

agent/services/agent_server.go
jchampio: `make` seems to work fine for me with go 1.12...

Are you talking about the problems that Ginkgo claims are preventing compilation? If so, that's probably a `go vet` issue:

> As part of building a test binary, go test runs go vet on the package
and its test source files to identify significant problems. If go vet
finds any problems, go test reports those and does not run the test
binary.

Makefile
jchampio: `FLY_TARGET` can be `prod`, `dev`, `cm`, ... whereas the gpupgrade secrets file has only two variants (prod or dev), to deal with the Slack notification service.

agent/services/agent_server.go
marksliva: I think the errors happened when running `make check`.

jchampio: Right, that's ginkgo.

Makefile
marksliva: The reason for PROD_OR_DEV is because those are the only 2 types of secret files.

kalensk: (Total side comment, but to me it still seems like gen_pipeline is overloading two concepts - 1) pipeline type - prod vs. dev, and 2) target instance - cm, prod, dev, <team abbreviation>. For example, the prod or dev pipeline types can be on the cm target instance.)

ci/pipeline.yml
kmacoskey: @greenplum-db/gp-releng

bradfordboyle: I am pretty sure this resource definition is wrong. No GPDB6 artifacts are being published to S3 anymore. I am not even sure which pipeline is publishing that artifact.

Makefile
marksliva: We would never want to use prod buckets on the cm target or other non-prod instances. We will rename the variable to SECRETS_TYPE to make it more clear though.

kalensk: Of course! For reference, I was referring to the fact that we as `cm` would want to fly the full "prod" equivalent pipeline with "dev" buckets, which again makes things even more confusing. Hence, why I was trying to disentangle those concepts and terms.

marksliva:
> (Just to be explicit, we're waiting on a correct GCS bucket path before moving forward with this PR.)

Yes that is correct.

Makefile
nkalampalikis: We renamed the variable to SECRETS_TYPE.

agent/services/agent_server.go
nkalampalikis: Added the description to the commit.

ci/pipeline.yml
slari-pivotal: Commit https://github.com/greenplum-db/gpupgrade/commit/3e7fcf955e0c7f2a0fac7980f86e52725e2d5a40 makes upgrade pipeline use the latest `bin_gpdb` output from 6X_STABLE prod pipeline.

Makefile
```diff
@@ -147,15 +147,18 @@ GIT_URI := $(shell git ls-remote --get-url)
...
 
 # Concourse does not allow "/" in pipeline names
 PIPELINE_NAME ?= gpupgrade:$(shell git rev-parse --abbrev-ref HEAD | tr '/' ':')
-DEPLOY_TYPE ?= dev
-FLY_TARGET ?= gpdb-dev
+FLY_TARGET ?= dev
+ifeq ($(FLY_TARGET),prod)
+SECRETS_TYPE = prod
+else
+SECRETS_TYPE = dev
```
jchampio: You probably want `:=` assignment for `SECRETS_TYPE` here.

PR #100, opened 2019-08-20
start-agents: make command synchronous

jchampio:
First in what is hopefully a chain of many.

The hub will now wait to send its response until after the agents have been started. This means that any errors encountered during the parallel start are immediately bubbled up to the CLI. Started agents now log their port and PID to file to ease debugging, and the hub also logs the same agent responses so that the logs can be correlated across hosts.

The PrepareStartAgents "unit" test, which ensured that the CLI's standard output matched a hardcoded message, has been removed entirely.

### DIFF
--- a/hub/services/prepare_start_agents.go
+++ b/hub/services/prepare_start_agents.go
@@ -11,7 +11,6 @@ import (
 	"github.com/greenplum-db/gpupgrade/hub/upgradestatus"
 	"github.com/greenplum-db/gpupgrade/idl"
 	"github.com/greenplum-db/gpupgrade/utils"
-	"github.com/greenplum-db/gpupgrade/utils/log"
 )
 
 func (h *Hub) PrepareStartAgents(ctx context.Context, in *idl.PrepareStartAgentsRequest) (*idl.PrepareStartAgentsReply, error) {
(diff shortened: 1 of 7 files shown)


hub/services/prepare_start_agents.go
```diff
@@ -54,6 +50,11 @@ func StartAgents(source *utils.Cluster) error {
 	}
 	source.CheckClusterError(remoteOutput, errStr, errMessage, true)
 
+	// Log successful starts. Agents print their port and PID to stdout.
```
jmcatamney: If we want to only print successful starts here, we'd need to put a check in this loop (`if remoteOutput.Errors[content] == nil` or similar), since `remoteOutput.Stdouts` contains the stdout for every command whether it succeeded or not (even if everything went to stderr and stdout is just an empty string).  Or we could add a `[SUCCESS]`/`[FAILURE]` note to the start of the line to distinguish them, or something; I have no strong opinions on the exact output format.

Also, we should probably add a test for this output somewhere, either in the BATS test or as a new Ginkgo one.

jchampio: Ah. So `CheckClusterError` changes from panic-based errors to no error handling at all based on that last optional boolean. Delightful.

jchampio: > Also, we should probably add a test for this output somewhere, either in the BATS test or as a new Ginkgo one.

I think there is a shared preference building to not pin logging behavior in our tests, at least not for now.

jchampio: Fixed up in the latest commit.

PR #102, opened 2019-08-20
shutdown-clusters: make command synchronous

jchampio:
In the same line as the previous commits, bubble up any failures synchronously to the CLI during cluster shutdown. The existing minimal test for this feature has been removed, to be replaced with useful end-to-end tests.

This, along with the previous two synchronous PRs, unblocks testing of the upgrade-master step.

### DIFF
--- a/cli/commanders/preparer.go
+++ b/cli/commanders/preparer.go
@@ -34,9 +34,10 @@ func (p Preparer) ShutdownClusters() error {
 	_, err := p.client.PrepareShutdownClusters(context.Background(),
 		&idl.PrepareShutdownClustersRequest{})
 	if err != nil {
-		gplog.Error(err.Error())
+		return err
 	}
-	gplog.Info("request to shutdown clusters sent to hub")
+
+	fmt.Println("clusters shut down successfully")
 	return nil
 }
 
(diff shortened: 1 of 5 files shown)


cli/commanders/preparer.go
```diff
@@ -36,7 +36,8 @@ func (p Preparer) ShutdownClusters() error {
 	if err != nil {
 		gplog.Error(err.Error())
 	}
-	gplog.Info("request to shutdown clusters sent to hub")
+
+	fmt.Println("clusters shut down successfully")
```
jchampio: This is executed even if the above code hits an error, which is wrong.

jchampio:
Added one more fixup to fix a `make check` failure.

integrations/prepare_shutdown_clusters_test.go
```diff
@@ -50,7 +50,7 @@ var _ = Describe("prepare shutdown-clusters", func() {
 		testExecutorNew.LocalError = errors.New("stop failed")
 
 		prepareShutdownClustersSession := runCommand("prepare", "shutdown-clusters")
-		Eventually(prepareShutdownClustersSession).Should(Exit(0))
+		Eventually(prepareShutdownClustersSession).Should(Exit(1))
```
kalensk: To clarify my understanding... Now that we are returning actual errors by making it a synchronous command, this test makes "more sense" with the `Exit(1)` instead of the previous `Exit(0)`, correct?

jchampio: Correct.

PR #103, opened 2019-08-21
gpupgrade: use target.BinDir for all gpupgrade binaries

dkrieger-pvtl:
We were incorrectly using the source binary directory for gpupgrade_agent.

Since we want to use the target cluster's gpupgrade, we must use its gpupgrade.  Indeed, gpupgrade might not even be installed in source.

The README commit is unrelated but we might as well get it in, and it seemed too much overhead to PR it separately.

Co-authored-by: Mark Sliva <msliva@pivotal.io>

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,7 +200,7 @@ files in gpupgrade/ci/tasks, and some secrets in a private repository. To set a
 pipeline, run:
 
 ```
-fly -t [target-name] set-pipeline -p [pipeline-name] -c gpupgrade/ci/pipeline.yml -l path/to/secrets.yml
+make set-pipeline FLY_TARGET=<CONCOURSE_INSTANCE> GIT_URI=https://github.com/<GITHUB_USERNAME>/gpupgrade.git
 ```
 
 If you want to use the defaults and have access to the continuous-integration
--- a/hub/services/check_seginstall.go
+++ b/hub/services/check_seginstall.go
@@ -26,7 +26,7 @@ func (h *Hub) CheckSeginstall(ctx context.Context, in *idl.CheckSeginstallReques
 	go func() {
 		defer log.WritePanics()
 
-		if err := VerifyAgentsInstalled(h.source); err != nil {
+		if err := VerifyAgentsInstalled(h.source, h.target); err != nil {
 			gplog.Error(err.Error())
 			step.MarkFailed()
 		} else {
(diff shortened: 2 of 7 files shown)


hub/services/check_seginstall.go
```diff
@@ -37,9 +37,9 @@ func (h *Hub) CheckSeginstall(ctx context.Context, in *idl.CheckSeginstallReques
 	return &idl.CheckSeginstallReply{}, nil
 }
 
-func VerifyAgentsInstalled(source *utils.Cluster) error {
+func VerifyAgentsInstalled(source *utils.Cluster, target *utils.Cluster) error {
```
jmcatamney: The `source` and `target` clusters have identical hosts, so instead of passing both of them, we can just pass `target` and use that for the function calls on lines 45 and 54.

dkrieger-pvtl: yes, in both `VerifyAgentsInstalled` and `StartAgents`, we should only be using the `target` cluster.  good catch.  We updated the code to remove all references to the `source` cluster for these functions.  `gpupgrade_agent` should only be installed on the `target` and should only run there(though the `target` and `source` hosts are identical).

README.md
```diff
@@ -200,7 +200,8 @@ files in gpupgrade/ci/tasks, and some secrets in a private repository. To set a
 pipeline, run:
 
 ```
-fly -t [target-name] set-pipeline -p [pipeline-name] -c gpupgrade/ci/pipeline.yml -l path/to/secrets.yml
+make set-pipeline FLY_TARGET=<INSTANCE> GIT_URI=https://github.com/<USER>/gpupgrade.git
+(make set-pipeline FLY_TARGET=cm GIT_URI=https://github.com/marksliva/gpupgrade.git)
```
jchampio: I think we should avoid pinning personal details into a shipped repo.

dkrieger-pvtl: ok. done.

hub/services/check_seginstall.go
dkrieger-pvtl: Actually, though a good idea, it turns out that when `VerifyAgentsInstalled` is called, the `target` cluster(i.e. the json file) does NOT yet contain the hostnames of `source`.  Since this PR is ancillary to our current effort(5to6 pipeline), so we'll file an issue for this, but leave this as is for now.

PR #104, opened 2019-08-21
remove conditional code based on pre-6 target GPDB versions

dkrieger-pvtl:
We won't be supported pre-6 target clusters.

Co-authored-by Mark Sliva <msliva@pivotal.io>

### DIFF
--- a/agent/services/convert_primary.go
+++ b/agent/services/convert_primary.go
@@ -5,18 +5,15 @@ import (
 	"fmt"
 	"path/filepath"
 
-	"github.com/blang/semver"
+	"github.com/greenplum-db/gp-common-go-libs/gplog"
 	"github.com/greenplum-db/gpupgrade/idl"
 	"github.com/greenplum-db/gpupgrade/utils"
-	"github.com/pkg/errors"
-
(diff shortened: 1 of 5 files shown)


dkrieger-pvtl:
@jmcatamney we removed even more code for these; please re-review at your convenience. It turns out that we don't need to pass `newVersion` at all when we upgrade segments so we could remove that code too.

jchampio:
`services.MINIMUM_VERSION` probably needs to be updated with this change. Something has to bail out when a user inevitably points gpupgrade at a 5X target bin by accident.

dkrieger-pvtl:
> `services.MINIMUM_VERSION` probably needs to be updated with this change. Something has to bail out when a user inevitably points gpupgrade at a 5X target bin by accident.

`services.MINIMUM_VERSION` is for the `source` cluster, which we updated to `5.0.0` for now, until we decide on which actual minimum version to support.  Setting the version for the `target` cluster is beyond the scope of this PR, but we'll add an issue for it.

jchampio:
> services.MINIMUM_VERSION is for the source cluster

Ugh, you're right... +1 for filing an issue and moving on.

PR #105, opened 2019-08-24
exectest: a new package for testing os/exec.Command

jchampio:
Mocking out `exec.Command()` [is very difficult](https://npf.io/2015/06/testing-exec-command/
). Our current solution (`cluster.Executor`) hides all of the power of `exec.Command()` behind an overly simplistic synchronous interface that is no longer adequate for our use cases.

Instead of mocking the `exec.Command()` API, `exectest.NewCommand()` swaps out the executable that the resulting `*exec.Cmd` will actually run, redirecting to the test binary itself with an implementation inspired by posts from [Nate Finch](https://npf.io/2015/06/testing-exec-command/), [Joe Shaw](https://www.joeshaw.org/testing-with-os-exec-and-testmain/), and [Chris Hines](http://cs-guy.com/blog/2015/01/test-main/).

Try it! Break it! Tear it apart. I tried to use more idiomatic Go, both in the implementation and in the tests, but I'm rusty. So bring out the red pencils.

`godoc -http=localhost:6060` should also be able to generate the actual documentation. I tried to document the package fully, so suggestions for greater clarity would be great too.

### DIFF
--- a/testutils/exectest/command.go
+++ b/testutils/exectest/command.go
@@ -70,23 +70,8 @@ var runCalled = false
 //     cmd := execCommand("/bin/bash", "-c", "sleep 15")
 //     cmd.Run() // this invokes MyExecutable, not Bash
 //
-// If you'd like to explicitly test that an exec.Cmd is being created with the
-// arguments you expect, you may use an optional verifier function in your call
-// to NewCommand (technically any number may be passed, but you'll probably only
-// need one):
 //
-//     execCommand := exectest.NewCommand(MyExecutable,
-//         func(name string, arg ...string)) {
-//             if name != "/bin/bash" {
-//                 t.Errorf("didn't use Bash")
-//             }
-//         }
-//     )
-//
-//     execCommand("/bin/bash", "-c", "sleep 1") // succeeds
-//     execCommand("/bin/sh", "-c", "sleep 1")   // logs a test failure
-//
-func NewCommand(m Main, verifiers ...func(string, ...string)) Command {
+func NewCommand(m Main) Command {
 	// Sanity check. Ensure that our two boilerplate conditions have been met.
 	index := indexOf(m, mains)
 	if index < 0 {
@@ -98,11 +83,6 @@ func NewCommand(m Main, verifiers ...func(string, ...string)) Command {
 	}
 
 	return func(executable string, args ...string) *exec.Cmd {
-		// Verify arguments if requested.
-		for _, v := range verifiers {
-			v(executable, args...)
-		}
-
 		// Pass the original arguments to the process.
 		cmd := exec.Command(os.Args[0], args...)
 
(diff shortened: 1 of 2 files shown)


jchampio:
We fixed up two important bugs, and now that we're actively writing tests that make use of this package, I'm removing draft status.

kalensk:
I have been taking a look at this a bit last week and some today. I wish I had more feedback, but really nothing really significant comes to mind. It definitely takes a bit to fully comprehend and there is a lot there. (And I don't think there is anything to make it easier to fully grasp as that is the nature of it.)

Overall, awesome job that definitely allows us to more effectively unit test exec.Command! I appreciate the use of go unit testing style.

testutils/exectest/command.go
```diff
@@ -0,0 +1,173 @@
...
+
+// magicEnvVar is the environment variable that signals to Run() that it should
+// invoke a command and exit. It is expected to contain the integer index of the
+// Main function in mains that should be invoked.
+const magicEnvVar = "EXECTEST_MAIN_INDEX"
+
+// runCalled tracks whether or not Run() has been called by the test package.
+// It's a quick-and-dirty failsafe against developers forgetting to use the
+// handler (which would cause infinite recursion).
+var runCalled bool = false
```
marksliva: This line makes the linter fail when running `make lint`

testutils/exectest/command.go
```diff
@@ -0,0 +1,173 @@
...
+//         os.Exit(1)
+//     }
+//
+//     func init() {
+//         exectest.RegisterMains(
+//             MyExecutable,
+//         )
+//     }
+//
+// The second is that the test suite must be started using testexec.Run:
```
marksliva: typo: testexec should be exectest

testutils/exectest/command.go
```diff
@@ -0,0 +1,173 @@
...
+// be used exactly like you would use exec.Command():
+//
+//     execCommand := exectest.NewCommand(MyExecutable)
+//
+//     cmd := execCommand("/bin/bash", "-c", "sleep 15")
+//     cmd.Run() // this invokes MyExecutable, not Bash
+//
+// If you'd like to explicitly test that an exec.Cmd is being created with the
+// arguments you expect, you may use an optional verifier function in your call
+// to NewCommand (technically any number may be passed, but you'll probably only
```
marksliva: It appears that we can have multiple verifiers check that the same executable gets called multiple times with specific arguments in a particular order. Which could potentially be a common enough use case that we could document that behavior here or in a sample test.

jchampio: > check that the same executable gets called multiple times with specific arguments in a particular order

No, all the verifiers run at the same time on the same `Command()` invocation. Maybe it would make more sense to have two constructor functions, one that takes a verifier and one that does not?

```go
cmdf := NewCommand(Main)
// vs
cmdf := NewVerifiedCommand(Main, func(e string, args ...string) {
    ...
})
```

testutils/exectest/command.go
jchampio: Fixed, thanks!

testutils/exectest/command.go
jchampio: Fixed.

jchampio:
Rebased against latest master.

testutils/exectest/command.go
marksliva: I agree, one or no verifiers is easier to understand.

dkrieger-pvtl:
This looks good to me...it allows basic testing of a system executable entirely in Golang.

Improvements, which need not be done in this PR:
* we appear unable to easily modify the behavior of a mock executable with state from the main program.  That is, suppose we have an array [1,2,3,4] of return values, and we wish to return the _i_th return value on call _i_.  Since each run of NewCommand() gets its own address space, it appears the only way to do this currently is via command arguments.  But those are outside the scope of the NewCommand itself.  Contrast this, with say, python mock where you can specify a sequence of return values.
* CommandContext is not available, so we cannot cancel the mock function easily.

testutils/exectest/command.go
jchampio: Cool, I'll make that change.

jchampio:
> we appear unable to easily modify the behavior of a mock executable with state from the main program.

I think I'd want to see an example of a test case that needs that before we make the effort to put it into this particular API. At first glance that'd be pretty difficult to pull off.

> That is, suppose we have an array [1,2,3,4] of return values, and we wish to return the _i_th return value on call i.

One `Cmd` gets one `Run` before it's used up. If you mean that you want multiple calls to `exec.Command()` to return different implementations, I think that's a valid use case but I'd want to see an example where it's needed first.

> CommandContext is not available, so we cannot cancel the mock function easily.

Agreed. Luckily I think it'll be easy to add when we need it.

testutils/exectest/command.go
jchampio: Pushed a fix with latest. We decided on `NewCommandWithVerifier`.

jchampio:
Thanks for the reviews!

PR #111, opened 2019-08-27
match mockgen version to mock package version

kalensk:
This is a followup to commit 0296161adc to also ensure the mockgen version matches the mock package version being used. We also bumped the mock version in Gopkg.toml to fix a compile bug.


jchampio:
Thanks for the review!

PR #114, opened 2019-09-05
protobuf: use go generate instead of Make

jchampio:
The gRPC and mock code is now generated using a single

    go generate ./idl

instead of a Makefile command. As a result, the `mock_idl` package is moved to a subpackage of `idl`, and the mock files have been slightly renamed, to match the example code layout given by gomock.

So basically the net result of this is that
- instead of using `make protobuf` you'll use `go generate ./idl` directly,
- which will eventually allow us to migrate to `go generate ./...` to recreate every generated file in the package tree, and
- if you want to make changes to how the `idl` package generation is done, you do so in the file `idl/generate.go`. The file name is not important; only the presence of the `//go:generate` comments in it.


idl/generate.go
```diff
@@ -0,0 +1,8 @@
+package idl
+
+// Creates the .pb.go protobuf definitions.
+//go:generate protoc --go_out=plugins=grpc:. cli_to_hub.proto hub_to_agent.proto
+
+// Generates mocks for the above definitions.
+//go:generate mockgen -source cli_to_hub.pb.go   -destination mock_idl/mock_cli_to_hub.pb.go
+//go:generate mockgen -source hub_to_agent.pb.go -destination mock_idl/mock_hub_to_agent.pb.go
```
kalensk: I am still somewhat new to mockgen and go, and reading the [mockgen docs](https://github.com/golang/mock#running-mockgen) on reflect mode states:
```
[Relect mode] is enabled by passing two non-flag arguments: an import path, and a comma-separated list of symbols.

Example:

mockgen database/sql/driver Conn,Driver
```

However, the above generate.go file uses `-source`? Is reflect mode not being used, or how does it interact with go generate, or what might I be misunderstanding?

kalensk:
A bit more context for others from https://github.com/greenplum-db/gpupgrade/pull/115
> Due to a bug in mockgen, we're unable to use the "source mode" generation when using gRPC streams. We have to manually specify the mock interfaces we want to generate, using "reflect mode"; hence the changes to idl/generate.go.

kalensk:
Naive question perhaps, but why not have `make protobuf` run `go generate ./idl` for you? 

I guess that raises the broader question of what things belong in a Makefile and what things do not? That is, what should a Makefile do, and should it not do? etc. 
I am probably still grappling with these questions, or perhaps easily get confused by them.

idl/generate.go
jchampio: Correct, source mode is still used in this commit, since the bug introduced in #115 isn't relevant yet.

jchampio:
> Naive question perhaps, but why not have make protobuf run go generate ./idl for you?

I am fine with doing that if the team would prefer to keep the recipe. `go generate ./idl` is not much more typing (though it's less tab-completable if I'm honest).

> I guess that raises the broader question of what things belong in a Makefile and what things do not?

IMO, anything where the benefit of automation (and dependency resolution) outweighs the cost of Makefile maintenance. In this case I wasn't sure that the `make protobuf` alias met that bar, especially since it seemed like half the team forgot it existed over the break. :grin:

But it wouldn't be a problem to add back if it's disruptive to workflows.

slari-pivotal:
I think a make target gives us the possibility of including that target as a dependency for other things in the Makefile. And yes it can run `go generate ./idl`.

jchampio:
> I think a make target gives us the possibility of including that target as a dependency for other things in the Makefile.

That's true, but it can also be trivially added if/when we decide that it needs to be a dependency.

jchampio:
@dkrieger-pvtl 
> For distributing the code to customers, I presume we are planning to git commit our generated non-test files(gRPC generated files, etc)?

Yeah... as much as I hate checking in generated files, it is [the Go way](https://blog.golang.org/generate), since they make no distinction between a source repository and a source distribution:
>  Also, if the containing package is intended for import by go get, once the file is generated (and tested!) it must be checked into the source code repository to be available to clients.

jchampio:
Although I guess we don't plan to be run using `go get`, right...?

PR #115, opened 2019-09-05
convert-master: stream stdout/err from pg_upgrade

jchampio:
_This is a draft PR that depends on #114 and #105. [The last commit]() is the one you care about._

Stream stdout/err from pg_upgrade, using `exec.Command` directly so that we can get at the independent streams. This effectively makes convert-master synchronous. Client disconnections will not affect pg_upgrade execution; it will continue and all output will still be streamed to a log file.

Also reimplement the core of `ConvertMaster()` to be fully unit-testable. There are still end-to-end tests that we'd like to write for this case.

Due to a bug in mockgen, we're unable to use the "source mode" generation when using gRPC streams. We have to manually specify the mock interfaces we want to generate, using "reflect mode"; hence the changes to `idl/generate.go`.


hub/services/upgrade_convert_master.go
```diff
@@ -1,68 +1,156 @@
...
-		h.target.BinDir,
-		h.target.MasterDataDir(),
-		h.target.MasterPort())
-
-	gplog.Info("Convert Master upgrade command: %#v", pgUpgradeCmd)
-
-	output, err := h.source.Executor.ExecuteLocalCommand(pgUpgradeCmd)
+	// Create a log file to contain pg_upgrade output.
+	log, err := utils.System.OpenFile(
+		filepath.Join(wd, "pg_upgrade.log"),
```
dkrieger-pvtl: So there's one file that gets over-written every time we run? I'd think we want a timestamp on the file, to prevent this, as we want to keep the logs for each run for retrospective debugging.  A related issue is knowing which log file is the active one(if any is) if the hub goes down and comes back up again.

hub/services/upgrade_convert_master.go
```diff
@@ -1,68 +1,156 @@
...
+	}
+}
+
+type streamWriter struct {
+	*muxedStream
+	cType idl.Chunk_Type
+}
+
+func (w *streamWriter) Write(p []byte) (int, error) {
+	w.mutex.Lock()
```
dkrieger-pvtl: What is the necessity of this mutex?  That is, what are the execution contexts that might concurrently use this?  RIght now, I believe `pg_upgrade` is a single thread, so this should not need protection.  In the general case, this isn't true, so I agree we need to have a clear synchronization strategy.

If there are multiple execution contexts, this lock could be held for a "long" time...i/o is notoriously costly.  It might be premature, but you can imagine a queueing scheme where Write() pushes into a synchronized buffer quickly, and then in the background, another thread actually publishes to file and to gRPC.  Otherwise, you are forcing the actual execution context to take a potentially heavy-weight hit to do what would otherwise be a simple stdout/stderr write.

hub/services/upgrade_convert_master.go
```diff
@@ -1,68 +1,156 @@
...
+	defer w.mutex.Unlock()
+
+	n, err := w.writer.Write(p)
```
dkrieger-pvtl: This makes no separation between STDOUT/STDERR, so these outputs will get intermixed in the log file.  Perhaps we can write a Chunk here, so that we can easily playback the results for a cli that connects/reconnects.   But a file of protobufs doesn't have message boundaries, so we'd have to write something like (# of bytes in Chunk, Chunk) to file so we can replay the file.

hub/services/upgrade_convert_master.go
```diff
@@ -1,68 +1,156 @@
...
+	if err != nil {
+		return n, err
+	}
+
+	if w.stream != nil {
+		// Attempt to send the chunk to the client. Since the client may close
+		// the connection at any point, errors here are logged and otherwise
+		// ignored. After the first send error, no more attempts are made.
+		err = w.stream.Send(&idl.Chunk{
```
dkrieger-pvtl: This is potentially a "long" operation, so perhaps we don't want to hold the Mutex across this.  Do we need to synchronize this, or does a stream Send act like Posix?

idl/cli_to_hub.proto
```diff
@@ -127,7 +127,15 @@ message PrepareInitClusterRequest {}
 message PrepareInitClusterReply {}
 
 message UpgradeConvertMasterRequest {}
-message UpgradeConvertMasterReply {}
+message Chunk {
```
dkrieger-pvtl: Do we want to add meta-data here?  Otherwise, the cli just gets a "bunch of bytes".  I can see adding: sequence number, sender id and timestamp.  If we replay data, you'd want to cli to know which messages are old and which are current.

hub/services/upgrade_convert_master.go
```diff
@@ -1,68 +1,156 @@
...
 
-	return nil
+	pair := clusterPair{h.source, h.target}
+	return pair.ConvertMaster(stream, log, wd)
+}
+
+// muxedStream provides io.Writers that wrap both gRPC stream and a parallel
+// io.Writer (in case the gRPC stream closes) and safely serialize any
+// simultaneous writes.
+type muxedStream struct {
```
dkrieger-pvtl: I'm not sure about the term `muxedStream`.  `stream` is already a gRPC concept so maybe `loggedStream` is better?

dkrieger-pvtl:
I like this as a first step...we log stdout/stderr of `pg_upgrade` for UpgradeMaster to a file and push those as a gRPC stream to the CLI.  

More things to do, which can be deferred, as they are additional functionality not part of this PR:
1). handle a hub disconnecting and reconnecting(if we decide to handle this, or if we are requiring the hub to stay up)
2). handle multiple CLIs
3). handle a CLI reconnecting

hub/services/upgrade_convert_master.go
```diff
@@ -1,68 +1,156 @@
...
-		gplog.Error("pg_upgrade failed to start: %s", output)
-		return errors.Wrapf(err, "pg_upgrade on master segment failed")
+		return err
 	}
 
-	return nil
+	pair := clusterPair{h.source, h.target}
+	return pair.ConvertMaster(stream, log, wd)
+}
+
```
dkrieger-pvtl: Seems like all of the "mux" code can and should be in its own module.  It has nothing per se to do with UpgradeMaster.

hub/services/upgrade_convert_master.go
jchampio: Yes. At the moment I don't think we don't deal with multi-run correctly in any step, and I didn't really want to decide on the way to fix that in this already-large PR. We might decide to back up the logs before trying again, for instance, rather than timestamping.

hub/services/upgrade_convert_master.go
jchampio: Yes, I think that'd be a good improvement if we decide that reattach needs that level of replayability.

idl/cli_to_hub.proto
jchampio: > I can see adding: sequence number, sender id and timestamp.

I'm not convinced any of those are relevant for the current implementation:
- sequence number is defined by reception order and guaranteed stable by the transport
- sender id is fixed (there is by definition only one sender: the server)
- timestamp is determined by reception time; not even the hub has a definitive idea of when a line was written because of OS buffering effects

> If we replay data, you'd want to cli to know which messages are old and which are current.

Maybe. I think that belongs in its own feature if we decide it's worth the cost, and I don't think I'd want an implementation that attached timestamps to arbitrary chunk boundaries.

hub/services/upgrade_convert_master.go
dkrieger-pvtl: Right, we don't need to solve the reattach issue now.  But I think we do want to have a way to distinguish STDOUT vs STDERR in the log, even if it's to prepend "STDOUT:" or "STDERR:" to each log line depending on the chunk type.

jchampio: > even if it's to prepend "STDOUT:" or "STDERR:" to each log line depending on the chunk type.

That's tricker than it seems at first glance. Chunk boundaries aren't guaranteed to fall at the same place as line boundaries. I'd prefer to solve this, if we decide it needs to be solved, with a more expressive log file format.

hub/services/upgrade_convert_master.go
jchampio: Currently it's tightly coupled to the convert-master protobuf definitions. That'll be fixed when needed, but more importantly to me, code belongs closest to where it's used, and there's only one user right now. (Our Python, and the CLI "commanders" architecture, violate this rule and it's really a pain to navigate.)

As that changes, we can move the implementation (and test it separately as a unit, to cut down on testing elsewhere), but if there aren't any users outside this package then IMO it doesn't belong in a different one.

hub/services/upgrade_convert_master.go
jchampio: Will reply in a separate followup.

hub/services/upgrade_convert_master.go
jchampio: Will reply in a separate followup.

hub/services/upgrade_convert_master.go
jchampio: Will reply in a separate followup.

hub/services/upgrade_convert_master.go
dkrieger-pvtl: That depends on the underlying `pg_upgrade` writes, right?  That is, if pg_upgrade does `write(stdout,"foo")` and then `write(stdout,"bar\n"`) you'll get "foo" and "bar\n" as two chunks.  But if pg_upgrade does `write(stdout,"foobar\n")` you'll get "foobar\n" as a single chunk.  

Even if the line boundaries don't match chunk boundarys, it will allow a simple script to reconstruct the separate STDOUT/STDERR from the single file.

My point is that, even for simple debugging now, we probably want STDOUT/STDERR separated.  You could do that in your currently implementation by switching based on the chunk type and writing STDOUT/STDERR to separate files.

jchampio: > That is, if pg_upgrade does write(stdout,"foo") and then write(stdout,"bar\n") you'll get "foo" and "bar\n" as two chunks. But if pg_upgrade does write(stdout,"foobar\n") you'll get "foobar\n" as a single chunk.

No, that's not guaranteed. User- and kernel-space buffering effects make it pretty difficult to say anything more than "you'll get a chunk with some data in the correct order eventually". I have a test program as an example if you'd like to see it firsthand.

> You could do that in your currently implementation by switching based on the chunk type and writing STDOUT/STDERR to separate files.

In my experience knowing where the stdout and stderr interleave with respect to each other is more useful to debugging than knowing which output came from which stream. Don't get me wrong, I think _both_ are important, but I prefer to have the former if we're choosing one.

jchampio:
@dkrieger-pvtl Regarding thread-safety and multiplexing, I think there are some misconceptions that I'd like to clear up.

> RIght now, I believe pg_upgrade is a single thread, so [the multiplexed stream] should not need protection.

pg_upgrade's threading strategy isn't relevant; it's running in a different process. `exec.Cmd` reads from stdout and stderr simultaneously from parallel goroutines. We cannot write to gRPC concurrently [1], so we need the mutex to serialize the chunks.

> what are the execution contexts that might concurrently use this?

Maximum of two goroutines; one for each standard stream. In practice, since pg_upgrade isn't streaming data to both stdout and stderr from parallel threads, contention on this lock is going to be nearly nonexistent.

> It might be premature, but you can imagine a queueing scheme where Write() pushes into a synchronized buffer quickly, and then in the background, another thread actually publishes to file and to gRPC.

I would say this is absolutely premature 😁 We'd have to manage flow control and backpressure somehow anyway, to avoid out of memory conditions from that buffer, so I don't see a practical benefit in adding synchronization complexity.

> Otherwise, you are forcing the actual execution context to take a potentially heavy-weight hit to do what would otherwise be a simple stdout/stderr write.

"Heavyweight" needs to be measured relative to the I/O that pg_upgrade has to do to generate a single line of output. We're not solving streaming in the general case; we're solving it for the utilities we use.

> I'm not sure about the term muxedStream. stream is already a gRPC concept so maybe loggedStream is better?

I'm not crazy about the name either, so suggestions welcome. But the key feature of the implementation is the multiplexing of parallel channels into serial streams. If all we were doing was logging, we could use `io.MultiWriter` directly. That mutex is the core feature.

[1] https://godoc.org/google.golang.org/grpc#ServerStream:

    // It is safe to have a goroutine calling SendMsg and another goroutine
    // calling RecvMsg on the same stream at the same time, but it is not safe
    // to call SendMsg on the same stream in different goroutines.

hub/services/upgrade_convert_master.go
dkrieger-pvtl: Ah, I see.  You're saying that the underlying system call `write(stdout,"123456")` can get split up into two write calls as seen by the golang code, and hence generate two Chunks.  

As an aside, I thought that POSIX `write` was atomic in the sense that one thread writing `write(stdout,"foo")` and one thread writing `write(stdout,"bar")` could only generate to stdout "foobar" or "barfoo" but not an intermixing of the two separate writes.  Provided that each write() succeeded in writing 3 bytes.

kalensk: I agree that having the "timeline" of when a stderr message is received in relation to the surrounding stdout messages is way more useful than having two separate files.

dkrieger-pvtl:
> `exec.Cmd` reads from stdout and stderr simultaneously from parallel goroutines. We cannot write to gRPC concurrently [1], so we need the mutex to serialize the chunks.

Ok, that's interesting: https://golang.org/pkg/os/exec/#Cmd


> "Heavyweight" needs to be measured relative to the I/O that pg_upgrade has to do to generate a single line of output. We're not solving streaming in the general case; we're solving it for the utilities we use.

I see what you're saying...the connection between the write in the thread of pg_upgrade and the read/write to gRPC in the exec.Cmd goroutine are mediated by a golang-library implemented buffer. So pg_upgrade itself wouldn't be blocked per se.  A slow exec.Cmd stdout reader might eventually backup the pg_upgrade if the buffer filled, but not necessarilty: it's the exec.Cmd goroutines that would be blocked.

hub/services/upgrade_convert_master.go
jchampio: `multiplexedStream` seems good to me for a follow-up.

kalensk: Pushed a fixup with the rename!

PR #116, opened 2019-09-10
CI: use Go 1.12

jchampio:
noinstall-tests are failing because `exec.ExitError` does not have the `ExitCode` member defined; it was added to `os.ProcessState` in Go 1.12. The GPDB Docker images have already updated to 1.12.7, so use that version for now.


jchampio:
thanks for the quick review!

PR #120, opened 2019-09-13
CI: enable a 5-to-6 pipeline job

jchampio:
See individual commits; there are a few bug fixes and cleanups sprinkled in there.

This pipeline job does not go green, as we discussed -- the diff fails.


jchampio:
[Here's the dev pipeline.](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:dev:5to6)

ci/pipeline.yml
```diff
@@ -148,19 +161,117 @@ jobs:
...
+      config:
+        platform: linux
+        image_resource:
+          type: docker-image
+          source:
+            repository: alpine
+            tag: latest
+        inputs:
+          - name: bin_gpdb_new
+          - name: ccp_src
```
marksliva: The `ccp_src` input is not used in this task.

ci/pipeline.yml
```diff
@@ -148,19 +161,117 @@ jobs:
...
+          - name: cluster_env_files
+        run:
```
marksliva: This script seems long enough to move into a task.yml.

ci/tasks/cluster-upgrade.yml
```diff
@@ -6,14 +6,16 @@ image_resource:
     repository: pivotaldata/centos-gpdb-dev
     tag: "6-gcc6.2-llvm3.7"
 
+params:
+  GPHOME_OLD: /usr/local/greenplum-db-devel
```
marksliva: It would be better to explicitly set these params in the 6-to-6 task that calls this, and then make these 2 params default to nothing.

ci/pipeline.yml
kalensk: Good catch, seems like its not used. Removed.

ci/tasks/cluster-upgrade.yml
kalensk: I agree, being more explicit seems better. Pushed a fixup.

jchampio: But... why? If it's not a useful default, why have a default?

ci/pipeline.yml
jchampio: It was used until we found that the SSH setup script used Bash instead of `sh`.

ci/scripts/cluster-upgrade.bash
```diff
@@ -22,24 +100,25 @@ make
...
+make_trampoline_directories "${hosts[@]}"
+
 # Now do the upgrade.
-time ssh mdw bash <<"EOF"
-    set -eu -o pipefail
+time ssh mdw GPHOME_OLD="${GPHOME_OLD}" GPHOME_NEW="${GPHOME_NEW}" bash <<"EOF"
+    set -eux -o pipefail
 
-    source /usr/local/greenplum-db-devel/greenplum_path.sh
     export PGPORT=5432 # TODO remove the need for this
```
marksliva: This export is no longer needed.

kalensk: Awesome! Pushed a fixup.

ci/tasks/cluster-upgrade.yml
kalensk: Removed. It seems like concourse does not have a mechanism (that I am currently aware of) to indicate that a parameter is needed and not have a default value.

ci/pipeline.yml
```diff
@@ -148,19 +161,119 @@ jobs:
...
+              ssh -ttn mdw '
+                  set -ex
+                  source /usr/local/greenplum-db-devel/greenplum_path.sh
+                  MASTER_DATA_DIRECTORY=/data/gpdata/master/gpseg-1 gpstop -ai
+              '
+              for host in $(cat cluster_env_files/hostfile_all); do
+                  ssh -ttn centos@"$host" '
+                      set -ex
+                      gphome=/usr/local/greenplum-db-old
+                      sudo mv /usr/local/greenplum-db-devel $gphome
```
dkrieger-pvtl: it'd be great to use GPHOME_OLD variable here...passed to a yml script.

ci/pipeline.yml
```diff
@@ -148,19 +161,119 @@ jobs:
...
+                  MASTER_DATA_DIRECTORY=/data/gpdata/master/gpseg-1 gpstart -a
+              '
+
+              # Install the new binary.
+              for host in $(cat cluster_env_files/hostfile_all); do
+                  scp bin_gpdb_new/*.tar.gz "${host}:/tmp/bin_gpdb_new.tar.gz"
+
+                  ssh -ttn centos@"$host" '
+                      set -ex
+                      gphome=/usr/local/greenplum-db-new
```
dkrieger-pvtl: it'd be great to use GPHOME_NEW variable here...passed to a yml script.

ci/pipeline.yml
```diff
@@ -30,6 +30,22 @@ resources:
...
+
+- name: bin_gpdb5_centos6
+  type: s3
+  source:
+    access_key_id: {{bucket-access-key-id}}
+    bucket: {{gpdb-stable-builds-bucket-name}}
+    region_name: {{aws-region}}
+    secret_access_key: {{bucket-secret-access-key}}
+    versioned_file: release_candidates/bin_gpdb_centos6/gpdb5/bin_gpdb.tar.gz
+
```
dkrieger-pvtl: For the gpdb-5 and gpdb-6 versions here, these seem to use the latest release_candidate for each.  I'd think we'd want to fix gpdb-5 version at the min version of 5 that we've allowed, or at least the latest pivnet release of 5.  For 6, I'd think we'd want to use the latest pivnet release as well.  Otherwise, the gpdb version could shift from underneath us, and if we got a failure, it could be from gpdb and not gpupgrade.

ci/pipeline.yml
kalensk: We added params for GPHOME_OLD and GPHOME_NEW for self-documentation and to follow the conventions of the cluster-upgrade task. I think it makes things a bit more clearer.

ci/pipeline.yml
kalensk: We added params for GPHOME_OLD and GPHOME_NEW for self-documentation and to follow the conventions of the cluster-upgrade task. I think it makes things a bit more clearer.

ci/pipeline.yml
kalensk: Good point. We will need to figure out our exact testing matrix when we get 5-to-6 upgrade working without all the RPATH hacks and a working diff of the dumps to know what actual version is the minimum supported.

dkrieger-pvtl: That's probably too much for this PR.  Since the intent is to test gpupgrade itself, we can just fix 5 at `5.21.4` and 6 at `6 GA` and use that.

kalensk: We created a story to capture this need since it is out of scope for this particular PR.

PR #121, opened 2019-09-17
master upgrade: set LD_LIBRARY_PATH to work around RPATH issues

jchampio:
In the GPDB 6 official build, `plpython2.so` is not set up with a correct `RPATH`. Commit a9ecbe74 unset all environment variables when calling the master pg_upgrade, since we shouldn't need any of them... but until the `RPATH` issue is fixed, we'll need to hack around this by passing through the `LD_LIBRARY_PATH` from the target installation.

### DIFF
--- a/hub/services/upgrade_convert_master.go
+++ b/hub/services/upgrade_convert_master.go
@@ -1,6 +1,7 @@
 package services
 
 import (
+	"fmt"
 	"io"
 	"os"
 	"os/exec"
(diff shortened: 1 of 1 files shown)


jchampio:
[6-to-6 test pipeline job.](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:dev:fix-plpython/jobs/6-to-6/builds/1)

PR #122, opened 2019-09-19
step consolidation: initialize

dkrieger-pvtl:
The CLI will contain three steps once we are done:
* initialize
* execute
* finalize

This series of commits implements 'initialize' by subsuming the existing steps used to set up the cluster for an upgrade into a single call.  We defer as much work as possible to the hub, which has the state for the entire upgrade.

We squashed all the commits into a single one, since we are in heavy development on this project and it was expedient.  Follow up efforts will:
* add tests for hub/services/step_initialize.go
* fix the integrations.bats (see #123)
* rework the CUI


cli/tasks/task_description.go
```diff
@@ -0,0 +1,42 @@
+package main
```
jchampio: This file can be removed.

hub/services/hub.go
```diff
@@ -235,3 +235,8 @@ func (h *Hub) InitializeStep(step string) (upgradestatus.StateWriter, error) {
 
 	return stepWriter, nil
 }
+
+func DoInit(stateDir, sourceBinDir, targetBinDir string) error {
```
jchampio: This can be deleted.

test/initialize.bats
```diff
@@ -5,7 +5,7 @@ load helpers
 setup() {
     STATE_DIR=`mktemp -d`
     export GPUPGRADE_HOME="${STATE_DIR}/gpupgrade"
-    gpupgrade prepare init --old-bindir /dummy --new-bindir /dummy
+    gpupgrade initialize --old-bindir=/usr/local/gpdb6/bin/ --new-bindir=/usr/local/gpdb6/bin/
```
jchampio: All of these tests currently fail. Also -- we don't want to couple against installation paths in these tests.

hub/gpupgrade_hub_main.go
```diff
@@ -48,23 +46,15 @@ func main() {
...
-			errSource := source.Load()
-			errTarget := target.Load()
-			if errSource != nil && errTarget != nil {
-				errBoth := errors.Errorf("Source error: %s\nTarget error: %s", errSource.Error(), errTarget.Error())
-				return errors.Wrap(errBoth, "Unable to load source or target cluster configuration")
-			} else if errSource != nil {
-				return errors.Wrap(errSource, "Unable to load source cluster configuration")
-			} else if errTarget != nil {
-				return errors.Wrap(errTarget, "Unable to load target cluster configuration")
+			err := os.Mkdir(conf.StateDir, 0700)
```
jchampio: This change breaks the ability of the hub to come back up after a crash, caught by `./test/config.bats`. (I was the one who suggested this change; sorry, I forgot.)

When the hub comes up, if there's already state inside `GPUPGRADE_HOME`, it should load it. Which means that the `os.Mkdir/os.IsExist` pair needs to move over to the `Initialize` implementation.

cli/commanders/step_initialize_test.go
```diff
@@ -0,0 +1,145 @@
...
+	})
+
+	ginkgo.It("we cannot start hub if hub is already running", func() {
+		execCommandHubCount = exectest.NewCommand(HowManyHubsRunning_1_Main)
+		execCommandHubStart = exectest.NewCommand(GpupgradeHub_good_Main)
+		err := StartHub("/bindir")
+		Expect(err).To(HaveOccurred())
+	})
+
+	ginkgo.It("we cannot start hub because exec failed", func() {
```
jchampio: I know this test will be rewritten in vanilla Go, but in Ginkgo tests, `It()` represents the argument to `Describe()` and is expected to be the subject of the sentence. For example:
```go
Describe("StartHub", ...
    It("returns an error if the hub fails to start", ...
```
The idea is that the test hierarchy should form complete, coherent sentences.

cli/commanders/commanders_suite_test.go
```diff
@@ -1,6 +1,8 @@
 package commanders_test
 
 import (
+	"github.com/greenplum-db/gpupgrade/testutils/exectest"
```
jchampio: Your editor needs to be switched over to use `goimports`.

cli/commanders/step_initialize.go
```diff
@@ -0,0 +1,143 @@
...
+
+// FIXME: reconcile these with the ones in the `command` package
+
+// connTimeout retrieves the GPUPGRADE_CONNECTION_TIMEOUT environment variable,
+// interprets it as a (possibly fractional) number of seconds, and converts it
+// into a Duration. The default is one second if the envvar is unset or
+// unreadable.
+//
+// TODO: should we make this a global --option instead?
+func connTimeout() time.Duration {
```
kalensk: I would prefer this as an option rather than an environment variable. It seems clearer to me to be consistent with either command line options, or environment variables, or a config file rather than mixing some of each. It makes it more explicit as a command line option, which most customers probably won't even need to worry about adjusting from the default.

cli/commanders/step_initialize.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	return time.Duration(duration * float64(time.Second))
+}
+
+// connectToHub() performs a blocking connection to the hub, and returns a
+// CliToHubClient which wraps the resulting gRPC channel. Any errors result in
+// an os.Exit(1).
+func connectToHub() idl.CliToHubClient {
+	upgradePort := os.Getenv("GPUPGRADE_HUB_PORT")
```
kalensk: Similar comment from above about being explicit with command line options.

cli/commanders/step_initialize.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	hubAddr := "localhost:" + upgradePort
+
+	// Set up our timeout.
+	ctx, cancel := context.WithTimeout(context.Background(), connTimeout())
+	defer cancel()
+
+	// Attempt a connection.
+	conn, err := grpc.DialContext(ctx, hubAddr, grpc.WithInsecure(), grpc.WithBlock())
```
kalensk: Can we add a story or have a record somewhere to move towards using secure grpc?

cli/commanders/step_initialize.go
dkrieger-pvtl: We can't refactor everything at the same time, so I've placed this on a list for the next refactor.

cli/commanders/step_initialize.go
dkrieger-pvtl: We can't refactor everything at the same time, so I've placed this on a list for the next refactor.

cli/commanders/step_initialize.go
dkrieger-pvtl: done.

cli/commanders/step_initialize.go
```diff
@@ -0,0 +1,143 @@
...
+			gplog.Error(err.Error())
+		}
+		os.Exit(1)
+	}
+
+	return idl.NewCliToHubClient(conn)
+}
+
+func StartHub(oldBinDir, newBinDir string) error {
+	countHubs, err := HowManyHubsRunning()
```
kalensk: Can you add refactor `HowManyHubsRunning` to the "refactor list"? Either have it be called something like `isHubRunning` and raise and error if so, and in terms of implementation consider using PID files instead of grepp'ing ps.

dkrieger-pvtl: Yeah, that's an ugly one too...we really want a (bool, err) return from something like `isHubRunning`.  And, yes, PID files would be great too.

cli/commanders/step_initialize.go
```diff
@@ -0,0 +1,143 @@
...
+		gplog.Error("failed to determine if hub already running")
+		return err
+	}
+	if countHubs >= 1 {
+		gplog.Error("gpupgrade_hub process already running")
+		return errors.New("gpupgrade_hub process already running")
+	}
+
+	hub_path := path.Join(newBinDir, "gpupgrade_hub")
+	cmd := exec.Command(hub_path, "--daemonize", "--old-bindir", oldBinDir,
```
kalensk: nit: We now assume gpupgrade binaries are on PATH, so the command can now become `exec.Command("gpupgrade_hub", "--daemonize", ...`

cli/commanders/step_initialize.go
```diff
@@ -0,0 +1,143 @@
...
+			// Annotate with the Stderr capture, if we have it.
+			err = fmt.Errorf("%s: %s", err, exitErr.Stderr)
+		}
+		return err
+	}
+	gplog.Debug("gpupgrade_hub started successfully: %s", stdout)
+	return nil
+}
+
+func VerifyConnectivity(client idl.CliToHubClient) error {
```
kalensk: Another to add to the refactor list, if you don't mind. I don't understand why this function is needed, especially given the guarantees that gRPC gives us.

jchampio: It's removed in a later commit.

cli/commanders/step_initialize.go
```diff
@@ -0,0 +1,143 @@
...
+	client := connectToHub()
+
+	err = VerifyConnectivity(client)
+	if err != nil {
+		gplog.Error("gpupgrade is unable to connect via gRPC to the hub")
+		gplog.Error("%v", err)
+		os.Exit(1)
+	}
+
+	err = TellHubToInitializeUpgrade(client, oldBinDir, newBinDir)
```
kalensk: nit: Consider inlining this function since its so short and it would reduce having similar error handling in both places.

hub/services/step_initialize.go
```diff
@@ -0,0 +1,122 @@
...
+
+	err = h.PrepareStartAgentsStep()
+
+	return &idl.TellHubToInitializeUpgradeReply{}, err
+}
+
+func (h *Hub) CheckConfigStep() error {
+	gplog.Info("starting %s", upgradestatus.CONFIG)
+
+	step, err := h.InitializeStep(upgradestatus.CONFIG)
```
kalensk: Another to add to the refactor list...now that we have three steps initialize, execute, finalize, calling this "initializeStep" is super confusing since it is basically setting up the step writing capabilities for recording the status.

hub/services/step_initialize.go
```diff
@@ -0,0 +1,161 @@
...
+	err = StartAgents(h.source, h.target)
+	if err != nil {
+		gplog.Error(err.Error())
+		step.MarkFailed()
+		return err
+	}
+
+	step.MarkComplete()
+	return nil
+}
```
kalensk: I am sure this will fall out later, but I can't help from commenting that this above function is quite common. Hopefully, in the future it makes sense to pull this into a common `ApplySubStep` type of function where you pass it the subStep you want it to execute and it handles all the logging, error handling, and marking the status.

cli/commanders/step_initialize.go
dkrieger-pvtl: Have we decided how we're packaging `gpupgrade` yet?  I chose `newBinDir` to be safe, but if we package gpupgrade separately, we'll need some other way to specify it.  `PATH` is a good as any other directory, just wondering...

hub/services/step_initialize.go
dkrieger-pvtl: added this.  As an aside, I want to refactor all the "step tracking" to a separate module so that we'd just call a function steps.executeStep(myStep, upgradestatus.STEP_NAME) that would handle all the step accounting and the allowed state transitions between steps/substeps.

hub/services/step_initialize.go
dkrieger-pvtl: Done.  As an aside, a common technique in small embedded systems is to have a global state value(integer) that automatically transitions in the code(via a switch statement that examines global variables and the current state and chooses the next state).  We can use a similar idea, but with the transitions done via user commands.

kalensk:
Overall great job! This looks much cleaner and setting the stage for future refactors and clean code. Thank you :) This was a lot of work.

cli/commands/commands.go
```diff
@@ -19,7 +19,6 @@ package commands
  * 			config       gather cluster configuration
```
marksliva: This can be removed.

cli/tasks/task_description.go
dkrieger-pvtl: done...thanks for the catch...I kept _noticing_ to do this ;)...

hub/services/hub.go
dkrieger-pvtl: good catch.

hub/gpupgrade_hub_main.go
dkrieger-pvtl: This is a much bigger issue as the three-step "Initialize", "Execute" and "Finalize" plan doesn't provide a way to currently re-start the hub independently of a `gpupgrade initialize`.  For now, we disabled the one test that requires that and use `GPHOME` and `PGPORT` to all `gupgrade initialize` to run.  But the bats tests are end-to-end I thought, so we can require a running database.

cli/commanders/step_initialize.go
dkrieger-pvtl: Ok...as an aside, we will have to make sure the version-check in gpupgrade/gpupgrade_agent/gppgrade_hub works well as otherwise the user could end up running the wrong version.

cli/commanders/step_initialize.go
dkrieger-pvtl: We would prefer to keep RPC-specific code out of the Cobra commands code.

hub/gpupgrade_hub_main.go
jchampio: > This is a much bigger issue as the three-step "Initialize", "Execute" and "Finalize" plan doesn't provide a way to currently re-start the hub independently of a `gpupgrade initialize`.

Hub crashes are outside the explicit design, but it worries me that we broke a working use case. The test should be able to `gpupgrade_hub --daemonize` instead of running `gpupgrade prepare start-hub`.

jchampio: > we need to consider if we plan to handle it or make the hub more robust.

That was something the previous code did, yes. It was accidentally broken during this work. Unless you're saying we need to decide again whether we actually want that behavior?

jchampio: > But we still have the problem that we have no way per se to have the hub.Initialize() distinguish between a full gpupgrade initialize and a restart of just the hub process.

The client-led Initialize step happens once. Hub startup could happen multiple times. If you only initialize the internal state during the initialize step, then the hub can never go down without breaking the system.

dkrieger-pvtl: For a straight refactor, we would preserve all of the existing behavior as evidenced by the tests.  But that constraint will constrain how we can improve gpupgrade.  I am asserting that being about to restart the hub without restarting the entire gpupgrade, while perhaps desired in the final product, is not currently necessary to run upgrade.  We need to revisit the entire robustness of the gpupgrade.  Removing this behavior now will not constrain us.

hub/services/step_initialize.go
```diff
@@ -0,0 +1,206 @@
...
+
+	"github.com/greenplum-db/gpupgrade/hub/upgradestatus"
+	"github.com/greenplum-db/gpupgrade/idl"
+	"github.com/greenplum-db/gpupgrade/utils"
+	"github.com/pkg/errors"
+	"golang.org/x/net/context"
+)
+
+func (h *Hub) Initialize(ctx context.Context, in *idl.InitializeRequest) (*idl.InitializeReply, error) {
+	err := h.InitializeSteps()
```
marksliva: We can rename this to 'InitializeSubSteps'.

hub/services/step_initialize.go
```diff
@@ -0,0 +1,206 @@
...
+
+	// Set up the checklist steps in order.
+	//
+	// TODO: make sure the implementations here, and the Checklist below, are
+	// fully exercised in end-to-end tests. It feels like we should be able to
+	// pull these into a Hub method or helper function, but currently the
+	// interfaces aren't well componentized.
+	//
+	cm := upgradestatus.NewChecklistManager(startDir)
+	cm.AddWritableStep(upgradestatus.CONFIG, idl.UpgradeSteps_CONFIG)
```
marksliva: These are also SubSteps.

hub/gpupgrade_hub_main.go
dkrieger-pvtl: The state directory is created in the cli now; we error out if it exists already.  This enforces that we can only run gpupgrade once.  The actual steps are still created in the hub initialize routine, but the hub main routine makes sure the state directory exists on the machine.

It trivially reuses the existing directory due to the implementation of the steps...only the actual steps write the actual files.  So it doesn't matter if it's the first call or the 10th to start the hub.

cli/commands/commands.go
```diff
@@ -503,3 +400,49 @@ var version = &cobra.Command{
...
+		Short: "prepare the system for upgrade",
+		Long:  `prepare the system for upgrade`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			// If we got here, the args are okay and the user doesn't need a usage
+			// dump on failure.
+			cmd.SilenceUsage = true
+
+			err := commanders.CreateStateDir()
+			if err != nil {
+				return errors.Wrap(err, "tried to create state directory")
```
marksliva: Why aren't we logging these errors using gplog.Error and then calling `os.Exit(1)`?

jchampio: The gpupgrade CLI's main function already handles errors at the top level, so it doesn't make much sense IMO to `os.Exit(1)` (which throws away `defer` handers, etc.) in the middle of the stack.

marksliva: That makes sense. Although we should still `gplog.Error` the error before returning it.

jchampio: We could `gplog` at the top level instead, and get rid of a bunch of duplication.

marksliva: 👍

test/initialize.bats
dkrieger-pvtl: these tests were updated to reflect the new initialize step that was added.

hub/gpupgrade_hub_main.go
dkrieger-pvtl: We now create the state directory and a dummy cluster in the cli, so that the hub can start and be restarted.  Only at hub.Initialize() time do we fill in the actual cluster.

hub/services/step_initialize.go
dkrieger-pvtl: Done.

hub/services/step_initialize.go
dkrieger-pvtl: That adds in more turmoil right now to an already big PR; let's defer.  I'll add an issue to track this.

hub/services/step_initialize.go
```diff
@@ -0,0 +1,206 @@
+package services
```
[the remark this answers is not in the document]
dkrieger-pvtl: This will be added in a follow-up PR.

cli/commands/commands.go
dkrieger-pvtl: this will be deferred to a follow up PR

jchampio:
I'll commit the make check fix in my PR; no need to put it into this one.

cli/commanders/step_initialize_test.go
dkrieger-pvtl: thanks; these tests have been rewritten go-style

dkrieger-pvtl:
> Looks good. This is one of several PRs as we continue to refactor/rewrite. Ensure the tests pass before merging.

They do, except 5to6 still fails due to dump(expected) and the integrations waits #123.

PR #123, opened 2019-09-19
integrations: minimize and migrate to BATS

jchampio:
Three commits:
- minimize the `integrations` tests
- fix `make check --keep-going`
- work around #127

This should get [the pipelines green again](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:dev:integrations).


test/hub.bats
```diff
@@ -0,0 +1,57 @@
...
+    regex='pid ([[:digit:]]+)'
+    [[ $output =~ $regex ]] || fail "actual output: $output"
+
+    pid="${BASH_REMATCH[1]}"
+    procname=$(ps -o ucomm= $pid)
+    [ $procname = gpupgrade_hub ] || fail "actual process name: $procname"
+}
+
+@test "hub saves cluster configs to disk when initialized" {
+    # XXX how useful is a test for this behavior?
```
jmcatamney: This behavior was originally being tested because `prepare init` was only creating dummy files for `check config` to fill later.  Now that `initialize` is both creating the files and actually writing configuration information to them, I'd say we should either expand the test to check the file contents or remove the test.

test/hub.bats
```diff
@@ -0,0 +1,57 @@
...
+
+    gpupgrade initialize \
+        --old-bindir=$PWD \
+        --new-bindir=$PWD \
+        --old-port=$PGPORT 3>&-
+
+    run gpupgrade status upgrade
+    [ "$status" -eq 0 ] || fail "$output"
+
+    # XXX is this a useful test? Seems like it's pinning the wrong behavior.
```
jmcatamney: We eventually plan to separate out status checking into its own separate thing and we can get rid of this then, but for now having at least one test to check for status output is probably valuable.

Makefile
```diff
@@ -53,11 +53,12 @@ unit:
...
 		ginkgo -r -keepGoing -randomizeAllSpecs integrations
 
-# check runs all tests. Use -k to keep going after the first failure.
+# check runs all tests against the locally built gpupgrade binaries.
 .PHONY: check
+check: export PATH := $(CURDIR):$(PATH)
 check:
 		ginkgo -r -keepGoing -randomizeSuites -randomizeAllSpecs
-		PATH=.:$$PATH bats -r ./test
+		bats -r ./test
```
kalensk: Just curious why PATH got moved up from being local to just the bats. Is PATH now needed by the ginkgo tests since we require gpupgrade binaries to be in PATH?

jchampio: Yes.

jchampio: To expand on that, the integration tests used to build their own versions of the binaries (one of the reasons the integration tests just seem to hang there doing nothing for a while); now that they don't; it's even more important to get the `PATH` set correctly.

idl/cli_to_hub.proto
```diff
@@ -3,18 +3,15 @@ syntax = "proto3";
...
 
 service CliToHub {
-    rpc Ping(PingRequest) returns (PingReply) {}
     rpc StatusUpgrade(StatusUpgradeRequest) returns (StatusUpgradeReply) {}
     rpc StatusConversion(StatusConversionRequest) returns (StatusConversionReply) {}
-    rpc CheckConfig(CheckConfigRequest) returns (CheckConfigReply) {}
-    rpc CheckSeginstall(CheckSeginstallRequest) returns (CheckSeginstallReply) {}
     rpc CheckObjectCount(CheckObjectCountRequest) returns (CheckObjectCountReply) {}
     rpc CheckVersion (CheckVersionRequest) returns (CheckVersionReply) {}
     rpc CheckDiskSpace (CheckDiskSpaceRequest) returns (CheckDiskSpaceReply) {}
```
kalensk: Should CheckDiskSpace, CheckVersion, and CheckObjectCount be deleted in favor of being called from inside the Initialize gRPC call? I might be missing the bigger architecture picture though.

jchampio: Maybe eventually, but there are enough questions about whether we need checks to run independently of hub initialization that we punted.

jchampio:
We added a temporary (hopefully) workaround to #127 as the final commit in this chain.

jchampio:
Pushed a commit to fix the `install-tests` pipeline failure; see the commit for details. Our `PATH` problems are now making correct testing very tricky.

test/config.bats
```diff
@@ -10,9 +10,14 @@ setup() {
...
+
+    # XXX We use $PWD here instead of a real binary directory because
+    # `make check` is expected to test the locally built binaries, not the
+    # installation. This causes problems for tests that need to call GPDB
+    # executables...
     gpupgrade initialize \
-        --old-bindir "${GPHOME}/bin" \
-        --new-bindir "${GPHOME}/bin" \
+        --old-bindir "$PWD" \
+        --new-bindir "$PWD" \
```
kalensk: I am assuming we "don't" want `make install` a dependency of running these tests to "fix" this issue?... Thoughts?

test/init-cluster.bats
```diff
@@ -0,0 +1,118 @@
...
+    GPSTOP="$GPHOME"/bin/gpstop
+
+    # Store the data directories for each source segment by port.
+    run $PSQL -AtF$'\t' -p $PGPORT postgres -c "select port, datadir from gp_segment_configuration where role = 'p'"
+    [ "$status" -eq 0 ] || fail "$output"
+
+    declare -a olddirs
+    while read -r port dir; do
+        olddirs[$port]="$dir"
+    done <<< "$output"
```
kalensk: Can you explain this a bit for me. I believe `$output` is a global bats variable that gets set after running the above PSQL command. But how does the `read -r port dir` construct get extracted from `$output`. Is that bash magic? :p

test/init-cluster.bats
```diff
@@ -0,0 +1,118 @@
...
+    for port in "${!olddirs[@]}"; do
+        local olddir="${olddirs[$port]}"
+        local newdir
+
+        # Master is special -- the new master is only incremented by one.
+        # Primary ports are incremented by 4000.
+        if [ $port -eq $PGPORT ]; then
+            (( newport = $port + 1 ))
+        else
+            (( newport = $port + 4000 ))
```
kalensk: For now hardcoding to `4000` against our internal implementation is "ok", but hopefully as we improve the implementation we can find a way to test that is less tightly coupled to the implementation. Just an observation.

integrations/integrations_suite_test.go
```diff
@@ -1,183 +1,117 @@
...
-	// put the gpupgrade_hub on the path don't need to rename the cli nor put
-	// it on the path: integration tests should use RunCommand() below
-	userPreviousPathVariable = os.Getenv("PATH")
-	os.Setenv("PATH", cliDirectoryPath+":"+hubDirectoryPath+":"+userPreviousPathVariable)
-
-	testout, testerr, testlog = testhelper.SetupTestLogger()
+	// All gpupgrade binaries are expected to be on the path for integration
+	// tests. Be nice to developers and check up front; warn if the binaries
+	// being tested don't look like the ones that are built locally.
+	_, testPath, _, ok := runtime.Caller(0)
```
kalensk: I know this returns the filename as described in the [godocs](https://golang.org/pkg/runtime/#Caller), but what _exactly_ is that in this context? Can you explain what is going on here a bit more. Thanks!

test/initialize.bats
```diff
@@ -4,30 +4,75 @@ load helpers
 
 setup() {
     require_gpdb
```
marksliva: Why does `require_gpdb` skip the test instead of failing? Are the same tests run in multiple environments in ci?

test/config.bats
jchampio: Correct. The reason we run `make check` before `make install` in the CI `install-tests` is to catch these sorts of issues explicitly.

test/init-cluster.bats
jchampio: >I believe $output is a global bats variable that gets set after running the above PSQL command.

Correct.

>But how does the read -r port dir construct get extracted from $output. Is that bash magic?

Yeah, the magic is in [`read`](https://www.gnu.org/software/bash/manual/html_node/Bash-Builtins.html#index-read) and the [`<<<` herestring](https://www.gnu.org/savannah-checkouts/gnu/bash/manual/bash.html#Here-Strings).

So `$output` is supplied to the `while` loop's stdin, which then parses line by line using `read -r`, and then we put that into the `olddirs` ~map~ associative array.

test/init-cluster.bats
jchampio: Personally I'd either like the user to have control over that, or have the implementation intelligently pick ports that aren't used (or both).

integrations/integrations_suite_test.go
jchampio: The absolute path to the containing file (`integrations/integrations_suite_test.go`), in this case.

test/initialize.bats
jchampio: > Are the same tests run in multiple environments in ci?

Yes. I was just telling @dkrieger-pvtl about how helpful that is -- in this case, it lets us catch problems where we accidentally test installed binaries during `make check`.

jchampio: (It also means that you can run `make check` without having GPDB installed on your system.)

test/init-cluster.bats
kalensk: Sweet, Thanks!

test/initialize.bats
kalensk: Would it make things more clear to rename this to perhaps something like `ensure_gpdb` since require has connotations of "failing" if not set.

jchampio: Hm, maybe. I don't have that particular connotation personally. And `ensure` implies to me that GPDB will be present after the function returns.

integrations/integrations_suite_test.go
kalensk: For some reason my brain is having a bit of trouble "_quickly_" seeing what is going on in these code section. I am not exactly sure what would help. Could the above comment be a bit more expansive, such as something like:

Ensure the gpupgrade binaries being tested are in the same base directory as this test file.

test/initialize.bats
kalensk: Hmmmm, maybe `is_gpdb_present`...?

jchampio: `skip_if_no_gpdb` is pretty clear, I think.

integrations/integrations_suite_test.go
jchampio: Sure, we'll add a clarification.

jchampio:
> Im glad we removed the end-to-end "integration" tests in favor of actual end-to-end tests! :)

@kalensk Me too 😁

test/initialize.bats
jchampio: Switched to that in the latest commit.

integrations/integrations_suite_test.go
jchampio: Updated the comment; does it help?

kalensk: Thanks!

PR #125, opened 2019-09-24
Introduce execute step

marksliva:
Our design plan is for the CLI to contain three user-facing steps:

initialize (#122)
execute (this PR)
finalize 

We implement 'execute' as a single CLI command by subsuming the existing steps used to `execute` an upgrade after `initialize` is run.

The new `execute` step runs the following sub-steps:
    - InitCluster
    - ShutdownClusters
    - UpgradeMaster
    - CopyMaster
    - UpgradePrimaries
    - StartTargetCluster

Notes:
* ExecuteUpgradeMasterSubStep continues to stream stdout and stderr from
    the hub to the client, however the other steps still need this
    implemented

* Ginkgo tests will be converted over to go-style in a follow up PR.

* ExecuteCopyMasterSubStep and StartTargetClusterSubStep were converted to
    be synchronous

* Unnecessary grpc layers from the sub-steps were removed.


hub/services/execute_init_target_cluster_sub_step.go
```diff
@@ -19,38 +19,38 @@ import (
...
-	return &idl.PrepareInitClusterReply{}, err
+	return err
 }
 
-func (h *Hub) CreateTargetCluster() error {
+func (h *Hub) CreateTargetCluster(stream idl.CliToHub_ExecuteServer) error {
 	sourceDBConn := db.NewDBConn("localhost", int(h.source.MasterPort()), "template1")
 
-	targetDBConn, err := h.InitCluster(sourceDBConn)
+	targetDBConn, err := h.InitCluster(stream, sourceDBConn)
```
dkrieger-pvtl: nit: it would be slightly better style to have the db connection first in these to make the source vs target clearer:
`h.InitCluster(sourceDBConn, stream)` and
`ReloadAndCommitCluster(targetDBConn, h.target)`

hub/services/execute_init_target_cluster_sub_step.go
```diff
@@ -19,38 +19,38 @@ import (
...
 	}
 
 	return ReloadAndCommitCluster(h.target, targetDBConn)
 }
 
-func (h *Hub) InitCluster(sourceDBConn *dbconn.DBConn) (*dbconn.DBConn, error) {
+func (h *Hub) InitCluster(stream idl.CliToHub_ExecuteServer, sourceDBConn *dbconn.DBConn) (*dbconn.DBConn, error) {
 	err := sourceDBConn.Connect(1)
 	if err != nil {
 		return nil, errors.Wrap(err, "could not connect to database")
```
dkrieger-pvtl: nit: change string to `could not connect to source database`

hub/services/execute_init_target_cluster_sub_step.go
```diff
@@ -19,38 +19,38 @@ import (
 	"golang.org/x/net/context"
 )
```
dkrieger-pvtl: We only upgrade from 5.0.0 or later, so you can delete L127-132:
```
	if h.source.Version.Before("5.0.0") {
		// FIXME: we need to decide how to deal with HEAP_CHECKSUM. At the
		// moment, we assume that 4.x has checksums disabled, and 5.x and later
		// have checksums enabled.
		gpinitsystemConfig = append(gpinitsystemConfig, "HEAP_CHECKSUM=off")
	}
```


(sorry, couldn't figure out how to comment on an arbitrary line in the file)

dkrieger-pvtl: The AgentConns() concept seems broken...it appears we assume either there are no connections from hub to agents or all of them are good.  We need to think through how to handle these connections.  We could re-create them on each command(as we do for cli to hub) and have a scheme for dealing with re-started agents.  This is for a follow-up PR.

hub/services/execute_init_target_cluster_sub_step.go
```diff
@@ -19,38 +19,38 @@ import (
...
+	targetDBConn, err := h.InitCluster(stream, sourceDBConn)
 	if err != nil {
 		return errors.Wrap(err, "could not initialize the new cluster")
 	}
 
 	return ReloadAndCommitCluster(h.target, targetDBConn)
 }
 
-func (h *Hub) InitCluster(sourceDBConn *dbconn.DBConn) (*dbconn.DBConn, error) {
+func (h *Hub) InitCluster(stream idl.CliToHub_ExecuteServer, sourceDBConn *dbconn.DBConn) (*dbconn.DBConn, error) {
```
dkrieger-pvtl: This function is write for a rewrite/refactor.  Maybe for a follow up PR.  Some ideas, pending how they actually work out since it's hard for me to statically read and determine all dependencies:
* have a single function that returns all of `gpinitsystemConfig` that we need
* barring that, have each such call return a new array that gets appended to the first one.  The pattern `arr = addtoArr(arr)` is pretty awkward, and I prefer `newarr = newArrItems(); arr.append(newarr)`.
* seems like the call the `h.AgentConns()` can be moved down(or up) to just before it's actually used.
* perhaps split up this function into three: `InitCluster_GetConfigs`, `InitCluster_Create` and `InitCluster_Connect` or some such.

hub/services/execute_shutdown_clusters_sub_step.go
```diff
@@ -30,7 +27,7 @@ func (h *Hub) PrepareShutdownClusters(ctx context.Context, in *idl.PrepareShutdo
 		step.MarkComplete()
 	}
 
-	return &idl.PrepareShutdownClustersReply{}, err
+	return err
 }
 
 func (h *Hub) ShutdownClusters() error {
```
dkrieger-pvtl: In `StopCluster()`, we call `gpstop -a -d datadir` which leaves us in smart mode.  For the 6 cluster, we have our new "smart mode to user dialog" scheme if the gpstop is "too slow".  This can either be due to user connections or a slow shutdown(maybe it takes 2 minutes on a 200-node cluster?).  So do we want to `gpstop -a -d datadir -M fast`?

hub/services/execute_upgrade_master_sub_step.go
```diff
@@ -18,7 +18,7 @@ import (
 // Allow exec.Command to be mocked out by exectest.NewCommand.
 var execCommand = exec.Command
 
-func (h *Hub) UpgradeConvertMaster(in *idl.UpgradeConvertMasterRequest, stream idl.CliToHub_UpgradeConvertMasterServer) error {
+func (h *Hub) ExecuteUpgradeMasterSubStep(stream idl.CliToHub_ExecuteServer) error {
```
dkrieger-pvtl: nit.  I prefer the line-spacing here:
```
initializeStep(STEP)

err = runmystep

if err {
   fail(STEP)
} else{
   pass(STEP)
}
```

but we also sometimes eliminate the step:

```
initializeStep(STEP)

err = runmystep
if err {
   fail(STEP)
} else{
   pass(STEP)
}
```

These should be consistent for all substeps.

dkrieger-pvtl:
Perhaps for a future refactor. In `utils/upgrade.go`, we have three functions with names that make it sound to me like they actually upgrade a database(`MasterPGUpgradeDirectory`) etc.  The functions here actually return a string that is notionally a path under the state directory for tracking state.  Please rename these.  Maybe `getStateDirForMasterUpgrade`?

dkrieger-pvtl:
As is, this PR is almost a straight refactor of the existing code.  Given how much code this represents, perhaps this is fine as a first step as the testing of the new code will be an effort in and of itself.  There are, however, a large number of smaller refactors in the existing code that can be cleaned up.  Perhaps now, and perhaps in a follow up PR.

hub/services/execute_copy_master_sub_step.go
```diff
@@ -13,32 +13,28 @@ import (
```
dkrieger-pvtl: What happens if the directory we copy over (`destinationDirName := "/tmp/masterDirCopy"`) has data already in it?  We either need to be able to safely assume it is wiped or clean it up before copying over files.

kalensk:
> As is, this PR is almost a straight refactor of the existing code. Given how much code this represents, perhaps this is fine as a first step as the testing of the new code will be an effort in and of itself. There are, however, a large number of smaller refactors in the existing code that can be cleaned up. Perhaps now, and perhaps in a follow up PR.

Right our goal was to figure out small units to PR and merge, rather than doing it "all at once" or some PR that stays open "too long". There will some followup PRs, based on increased testing and some refactors, however, even some of that is "outside the scope of the story".

hub/services/execute_init_target_cluster_sub_step.go
kalensk: Why is there not an existing bug in this section of code. The line above does:
`sourceDBConn := db.NewDBConn("localhost", int(h.source.MasterPort()), "template1")`
but shouldn't that be `targetDBConn` which uses `h.target.MasterPort()` since this while method and file is trying to initialize the target (ie: new) cluster?!?

kalensk: Never mind, after taking a closer look we run gpinitsystem based on the source cluster in order to init the target cluster. (A little confusing at first.)

hub/services/execute_init_target_cluster_sub_step.go
kalensk: Renamed to `"failed to connect to old database"` since externally to customers we should use the terms old/new to be consistent, and internally in the go code we should use source/target since `new` is kind of like a keyword in creating objects that gets easily confused (hence why the code uses source and target)

hub/services/execute_init_target_cluster_sub_step.go
kalensk: Renamed the function to `InitTargetCluster` for clarity.

hub/services/execute_init_target_cluster_sub_step.go
kalensk: Removed

hub/services/execute_upgrade_master_sub_step.go
kalensk: Sure. Used the later for now for consistency.

hub/services/execute_init_target_cluster_sub_step.go
kalensk: Created a `gpupgrade Execute Refactors` story to track this.

kalensk:
> Perhaps for a future refactor. In `utils/upgrade.go`, we have three functions with names that make it sound to me like they actually upgrade a database(`MasterPGUpgradeDirectory`) etc. The functions here actually return a string that is notionally a path under the state directory for tracking state. Please rename these. Maybe `getStateDirForMasterUpgrade`?

Tracked in the "gpupgrade Execute Refactors & Rewrites" story

hub/services/execute_shutdown_clusters_sub_step.go
kalensk: Added a story to keep track of this and discuss further.

hub/services/execute_copy_master_sub_step.go
marksliva: created a story to look into this

hub/services/execute_init_target_cluster_sub_step.go
marksliva: created a story to improve this

ci/scripts/upgrade-cluster.bash
```diff
@@ -186,22 +186,7 @@ time ssh mdw GPHOME_OLD="${GPHOME_OLD}" GPHOME_NEW="${GPHOME_NEW}" bash <<"EOF"
...
-
-    gpupgrade upgrade copy-master
-    wait_for_step "Copy master data directory to segments"
-
-    gpupgrade upgrade convert-primaries
-    wait_for_step "Run pg_upgrade on primaries" 1200 # twenty minute timeout
-
-    gpupgrade upgrade validate-start-cluster
-    wait_for_step "Validate the upgraded cluster can start up"
+    gpupgrade execute
```
dkrieger-pvtl: Perhaps for another PR, but since this is the major integration test in our pipeline, perhaps we want to run the "finalize" step here aka reconfiguring the ports.

ci/scripts/upgrade-cluster.bash
```diff
@@ -186,22 +186,7 @@ time ssh mdw GPHOME_OLD="${GPHOME_OLD}" GPHOME_NEW="${GPHOME_NEW}" bash <<"EOF"
               --old-bindir ${GPHOME_OLD}/fake-bin \
               --old-port 5432
 
-    gpupgrade prepare init-cluster
```
dkrieger-pvtl: The `wait_for_step()` function above is now unused because the two calls to `gpugpgrade`(initialize,execute) are now synchronous, so you can remove it.

cli/commanders/step_execute.go
```diff
@@ -0,0 +1,38 @@
...
+	"context"
+	"io"
+	"os"
+
+	"github.com/greenplum-db/gp-common-go-libs/gplog"
+	"github.com/greenplum-db/gpupgrade/idl"
+)
+
+func Execute(client idl.CliToHubClient) error {
+	stream, err := client.Execute(context.Background(), &idl.ExecuteRequest{})
```
dkrieger-pvtl: This call uses server-side streaming.  How does the server(i.e. the hub in this case) indicate that the call is done?  In other words, I want to make sure this call is synchronous.

cli/commanders/upgrade.go
```diff
@@ -0,0 +1,28 @@
...
+
+type Upgrader struct {
+	client idl.CliToHubClient
+}
+
+func NewUpgrader(client idl.CliToHubClient) *Upgrader {
+	return &Upgrader{client: client}
+}
+
+func (u *Upgrader) ReconfigurePorts() error {
```
dkrieger-pvtl: This is part of finalize; just curious why you did this in this execute PR.  Also, consider running this in `upgrade-cluster.bash` during our pipeline test of 6-6.

cli/commanders/upgrade_test.go
```diff
@@ -0,0 +1,48 @@
...
+		})
+
+		It("returns error when ports cannot be reconfigured", func() {
+			hubClient.Err = errors.New("reconfigure ports failed")
+
+			err := upgrader.ReconfigurePorts()
+			Expect(err).To(HaveOccurred())
+		})
+	})
+})
```
dkrieger-pvtl: redo this in golang style?

hub/services/execute_upgrade_master_sub_step_test.go
```diff
@@ -0,0 +1,196 @@
...
+		scanner := bufio.NewScanner(&buf)
+		for scanner.Scan() {
+			Expect(scanner.Text()).NotTo(HavePrefix("PGPORT="),
+				"PGPORT was not stripped from the child environment")
+			Expect(scanner.Text()).NotTo(HavePrefix("PGHOST="),
+				"PGHOST was not stripped from the child environment")
+		}
+		Expect(scanner.Err()).NotTo(HaveOccurred())
+	})
+})
```
dkrieger-pvtl: redo in golang style?

hub/services/execute.go
```diff
@@ -0,0 +1,35 @@
...
+	}
+
+	err = h.ExecuteUpgradePrimariesSubStep()
+	if err != nil {
+		return err
+	}
+
+	err = h.ExecuteStartTargetClusterSubStep(stream)
+	return err
+}
```
dkrieger-pvtl: rhetorical comment that I'd love to see golang exceptions, but since the steps are sequential, I don't see a better way.  Well....
```
substeps := []struct {
                           sstep func(idl.CliToHub_ExecuteServer),
                     }{
                           h.s1,h.s2.h.s3.h.s4,h.s5,h.s5
                     }
for step := range substeps {
     err := step.sstep(stream)
     if (err != nil) {
          return nil
    }
}
```
Of course, not all the functions take a stream, but maybe they could/should.  Not sure what this really buys us but IMHO it is more readable.

hub/services/execute_init_target_cluster_sub_step.go
```diff
@@ -19,79 +20,90 @@ import (
...
 
-	targetDBConn, err := h.InitCluster(sourceDBConn)
+	targetPort, err := h.InitTargetCluster(stream, sourceDBConn)
 	if err != nil {
-		return errors.Wrap(err, "could not initialize the new cluster")
+		gplog.Error("failed to initialize new cluster due to %s", err.Error())
+		return errors.Wrap(err, "failed to initialize the new cluster")
 	}
 
+	targetDBConn := db.NewDBConn("localhost", targetPort, "template1")
```
dkrieger-pvtl: Is `localhost` a standard?  That is, must all client's linux machines contain this hostname?  I know most machines are configured this way, but not sure if they are required to.

hub/services/execute_init_target_cluster_sub_step.go
```diff
@@ -194,24 +199,21 @@ func (h *Hub) CreateAllDataDirectories(agentConns []*Connection, segmentDataDirM
...
-		// gpinitsystem has a return code of 1 for warnings, so we can ignore that return code
-		if err.Error() == "exit status 1" {
-			gplog.Warn("gpinitsystem completed with warnings")
-			return nil
-		}
-		return errors.Wrapf(err, "gpinitsystem failed: %s", output)
-	}
-	return nil
+	cmd := execCommand("bash", "-c",
+		fmt.Sprintf("source %s/greenplum_path.sh; %s/gpinitsystem -a -I %s",
```
dkrieger-pvtl: Since customers are known to modify their `greenplum_path.sh` files, I agree we should specify the absolute path of `gpinitsystem`.  That said, if they've done so in such a way that sourcing  `gphome/greenplum_path.sh` doesn't find `h.target.BinDir/gpinitsystem` first, perhaps other parts of gpupgrade would fail/work improperly.  So this is defensive but not fully protective.  

Perhaps just a comment here to this effect.

hub/services/execute_shutdown_clusters_sub_step.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+func StopCluster(stream idl.CliToHub_ExecuteServer, c *utils.Cluster) error {
+	err := IsPostmasterRunning(stream, c)
+	if err != nil {
+		return err
+	}
+
+	cmd := execCommand("bash", "-c",
+			fmt.Sprintf("source %[1]s/../greenplum_path.sh; %[1]s/gpstop -a -d %[2]s",
```
dkrieger-pvtl: See https://github.com/greenplum-db/gpupgrade/pull/125/files#r329102324.  Why specify the absolute path of gpstop if you've just sourced greenplum_path.sh?

hub/services/execute_start_target_cluster_sub_step.go
```diff
@@ -0,0 +1,45 @@
...
+	} else {
+		step.MarkComplete()
+	}
+
+	return nil
+}
+
+func (h *Hub) startNewCluster(stream idl.CliToHub_ExecuteServer) error {
+	cmd := execCommand("bash", "-c",
+		fmt.Sprintf("source %s/../greenplum_path.sh; %s/gpstart -a -d %s",
```
dkrieger-pvtl: nit: you can use the `[]` syntax of fmt.Sprintf to only specify `h.target.BinDir` once.

See https://github.com/greenplum-db/gpupgrade/pull/125/files#r329102324...why specify the path of gpstart explicitly after sourcing greenplum_path.sh?

test/execute.bats
```diff
@@ -0,0 +1,119 @@
...
+    # allow users to override data directories/ports during init-cluster?
+    delete_cluster $newport "$newmasterdir" || log "no upgraded cluster running"
+
+    gpupgrade initialize \
+        --old-bindir "$GPHOME/bin" \
+        --new-bindir "$GPHOME/bin" \
+        --old-port "$PGPORT" 3>&-
+
+    gpupgrade execute
+
```
dkrieger-pvtl: why not reconfigure ports here?

hub/services/execute_upgrade_master_sub_step_test.go
kalensk: That is for a followup PR

cli/commanders/upgrade_test.go
kalensk: That is for a followup PR

dkrieger-pvtl:
Consider running golang's code coverage tools to understand gaps in the testing.

ci/scripts/upgrade-cluster.bash
kalensk: done

cli/commanders/upgrade.go
kalensk: We didn't looks like the file was renamed from upgrader to upgrade at some point as shown on [master](https://github.com/greenplum-db/gpupgrade/blob/master/cli/commanders/upgrader.go#L83). Could be from one of the many previous PRs or fixups in between.

test/execute.bats
kalensk: added it.

hub/services/execute_init_target_cluster_sub_step.go
```diff
@@ -187,24 +199,21 @@ func (h *Hub) CreateAllDataDirectories(agentConns []*Connection, segmentDataDirM
...
```
jchampio: Needs an `&&` rather than a `;` to catch problems during source.

hub/services/execute_init_target_cluster_sub_step.go
```diff
@@ -187,24 +199,21 @@ func (h *Hub) CreateAllDataDirectories(agentConns []*Connection, segmentDataDirM
...
+			gphome,
+			h.target.BinDir,
+			gpinitsystemFilepath,
+		))
+
+	mux := newMultiplexedStream(stream, ioutil.Discard)
```
jchampio: The `ioutil.Discard` here will permanently lose any errors that happen before gpinitsystem opens up its log.

hub/services/execute_shutdown_clusters_sub_step.go
```diff
@@ -30,49 +32,50 @@ func (h *Hub) ExecuteShutdownClustersSubStep() error {
...
 	}
 
-	return nil
+	cmd := execCommand("bash", "-c",
+			fmt.Sprintf("source %[1]s/../greenplum_path.sh; %[1]s/gpstop -a -d %[2]s",
+				c.BinDir,
+				c.MasterDataDir(),
+			))
+
+	mux := newMultiplexedStream(stream, ioutil.Discard)
```
jchampio: This would also be a good one to not discard.

hub/services/execute_shutdown_clusters_sub_step.go
```diff
@@ -30,49 +32,50 @@ func (h *Hub) ExecuteShutdownClustersSubStep() error {
...
+
+	return cmd.Run()
 }
 
-func IsPostmasterRunning(c *utils.Cluster) bool {
-	masterDataDir := c.MasterDataDir()
-	checkPidCmd := fmt.Sprintf("pgrep -F %s/postmaster.pid", masterDataDir)
+func IsPostmasterRunning(stream idl.CliToHub_ExecuteServer, c *utils.Cluster) error {
+	cmd := execCommand("bash", "-c",
+		fmt.Sprintf("pgrep -F %s/postmaster.pid",
```
jchampio: I know this commit didn't introduce it, but is `pgrep -F` in any way standardized? [This man page](https://linux.die.net/man/1/pgrep) doesn't show it, for instance...

hub/services/multiplexed_stream.go
```diff
@@ -0,0 +1,64 @@
...
+
+	"github.com/greenplum-db/gp-common-go-libs/gplog"
+	"github.com/greenplum-db/gpupgrade/idl"
+)
+
+// multiplexedStream provides io.Writers that wrap both gRPC stream and a parallel
+// io.Writer (in case the gRPC stream closes) and safely serialize any
+// simultaneous writes.
+type multiplexedStream struct {
+	stream idl.CliToHub_ExecuteServer
```
jchampio: This refactor would be the perfect time to decouple `multiplexedStream` from the business logic, and define the interface that we want `stream` to implement. Looks like
```go
type Sender interface {
    func Send(*idl.Chunk) error
}
```
would be enough? That way you can also get rid of the execute-specific mocks from the tests.

hub/services/execute_init_target_cluster_sub_step_test.go
```diff
@@ -146,6 +147,10 @@ var _ = Describe("Hub prepare init-cluster", func() {
 
 		BeforeEach(func() {
 			stdout, _, _ = testhelper.SetupTestLogger()
+
+			mockStream.EXPECT().
```
jchampio: The use of a magic global here is dangerous (as I'm sure you found out with the panics). I don't think covering it up with a `BeforeEach` initialization is the right way to go; we should be explicitly controlling the interface. IMHO, that goes hand-in-hand with replacing the stream with an interface that implements `Send()`.

hub/services/execute_shutdown_clusters_sub_step_test.go
```diff
@@ -1,69 +1,108 @@
...
-
-		err := services.StopCluster(mockStream, source)
-
-		Expect(testExecutor.NumExecutions).To(Equal(1))
-		Expect(testExecutor.LocalCommands[0]).To(ContainSubstring("pgrep"))
-		Expect(err).ToNot(HaveOccurred())
+		execCommandIsPostmasterRunning = exectest.NewCommand(FailedMain)
+		var skippedStopClusterCommand = true
+		execCommandStopCluster = exectest.NewCommandWithVerifier(EmptyMain,
+			func(path string, args ...string) {
```
jchampio: Technically the verifier is only run on command creation, not command execution. It'd be completely valid for a future implementation to construct the `Cmd` and then not run it if the cluster's already down.

That said, this test works for the implementation we have, and it's cheap.

hub/services/execute_init_target_cluster_sub_step.go
kalensk: What file would you like to write to? I was hoping to create a story and address it in a followup PR.

testutils/mock_agent_server.go
```diff
@@ -91,7 +91,7 @@ func (m *MockAgentServer) PingAgents(context.Context, *idl.PingAgentsRequest) (*
 	return &idl.PingAgentsReply{}, nil
 }
 
-func (m *MockAgentServer) UpgradeConvertPrimarySegments(ctx context.Context, in *idl.UpgradeConvertPrimarySegmentsRequest) (*idl.UpgradeConvertPrimarySegmentsReply, error) {
+func (m *MockAgentServer) AgentExecuteUpgradePrimariesSubStep(ctx context.Context, in *idl.UpgradePrimariesRequest) (*idl.UpgradePrimariesReply, error) {
```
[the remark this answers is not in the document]
kalensk: For now I would like to keep it as `StepStep` if possible for a couple of reasons: 1) the execute commit follows this convention for many files (granted we can do a rename and fix it). 2) intelliJ shows an annoying squiggle line for substep thinking its misspelled.

If you feel strongly we can rename all instances of `SubStep` to `Substep` (its just "work").

agent/services/execute_upgrade_primaries_sub_step_test.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+func init() {
+	exectest.RegisterMains(
+		EmptyMain,
+		EnvironmentMain,
+	)
+}
+
+func init() {
```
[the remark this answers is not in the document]
kalensk: fixed

cli/commanders/upgrade_test.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/greenplum-db/gpupgrade/idl"
+	"github.com/greenplum-db/gpupgrade/testutils"
+	"github.com/greenplum-db/gpupgrade/utils"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("reporter", func() {
+	var (
+		ctrl   *gomock.Controller
```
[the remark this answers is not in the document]
kalensk: fixed and renamed the file back to ugprader

hub/services/execute_init_target_cluster_sub_step.go
kalensk: Updated. Even though I was trying to not change "too" much at once.

agent/services/execute_upgrade_primaries_sub_step.go
```diff
@@ -0,0 +1,96 @@
...
+func (s *AgentServer) AgentExecuteUpgradePrimariesSubStep(ctx context.Context, in *idl.UpgradePrimariesRequest) (*idl.UpgradePrimariesReply, error) {
+	gplog.Info("got a request to convert primary from the hub")
+
+	err := UpgradeSegments(in.OldBinDir, in.NewBinDir, in.DataDirPairs)
+
+	if err != nil {
+		return &idl.UpgradePrimariesReply{}, err
+	}
+
+	return &idl.UpgradePrimariesReply{}, nil
```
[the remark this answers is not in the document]
kalensk: updated

agent/services/execute_upgrade_primaries_sub_step.go
```diff
@@ -0,0 +1,96 @@
...
+			if err != nil {
+				agentErrs <- errors.Wrapf(err, "failed to upgrade primary on host %s with content %d", host, segment.Content)
+			}
+
+			wg.Done()
+		}()
+	}
+
+	wg.Wait()
+	close(agentErrs)
```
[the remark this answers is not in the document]
jchampio: I missed that you're ranging over the channel, never mind.

PR #126, opened 2019-09-25
CI: Update sql dump buckets

kalensk:
After talking with releng these are the buckets we should be consuming the sql dump artifacts from since GCS is where current artifacts are published and S3 is deprecated.

This also pushes the dump_gpdb5_simple to a bucket the prod pipeline can access.

### DIFF
--- a/ci/pipeline.yml
+++ b/ci/pipeline.yml
@@ -84,20 +84,18 @@ resources:
       bucket_path: clusters-google/
 
 - name: dump_gpdb6_icw_gporca_centos6
-  type: s3
+  type: gcs
   source:
-    access_key_id: {{bucket-access-key-id}}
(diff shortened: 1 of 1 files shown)


marksliva:
> The changes themselves appear fine, but for gpupgrade, have you flown a production pipeline and a cm pipeline to validate that both pipelines succeed with these changes? It appears the latter was having issues with access: see the 5to6 job for https://prod.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade.

OK we will fly the prod pipeline with the fix, and merge it if it works. The dev pipeline works: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:update_sql_dump_buckets

PR #128, opened 2019-09-30
Execute: stream output for single-host steps

kalensk:
Stream stdout and stderr for execute sub-steps that don't need to stream from multiple hosts.

InitTargetCluster had to be refactored to ignore warnings from gpinitsystem instead of failing, because those warnings result in an exit code of 1. This caused gpupgrade to incorrectly fail. This special logic will be removed when gpinitsystem is fixed.

### DIFF
--- a/hub/services/execute_init_target_cluster_sub_step.go
+++ b/hub/services/execute_init_target_cluster_sub_step.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"os/exec"
 	"path"
 	"path/filepath"
 	"strings"
@@ -19,7 +20,7 @@ import (
 	"golang.org/x/net/context"
 )
 
-func (h *Hub) ExecuteInitTargetClusterSubStep() error {
+func (h *Hub) ExecuteInitTargetClusterSubStep(stream idl.CliToHub_ExecuteServer) error {
 	gplog.Info("starting %s", upgradestatus.INIT_CLUSTER)
 
 	step, err := h.InitializeStep(upgradestatus.INIT_CLUSTER)
@@ -28,7 +29,7 @@ func (h *Hub) ExecuteInitTargetClusterSubStep() error {
 		return err
 	}
 
-	err = h.CreateTargetCluster()
+	err = h.CreateTargetCluster(stream)
 	if err != nil {
 		gplog.Error(err.Error())
 		step.MarkFailed()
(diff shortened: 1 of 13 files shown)


hub/services/execute_init_target_cluster_sub_step.go
```diff
@@ -39,59 +40,70 @@ func (h *Hub) ExecuteInitTargetClusterSubStep() error {
...
-		return nil, err
+	err = RunInitsystemForTargetCluster(stream, gpinitsystemFilepath, h.target.BinDir)
+
+	var gpinitsystemWarning bool
+	if exitErr, ok := err.(*exec.ExitError); ok {
+		// gpinitsystem exits with 1 on warnings and 2 on errors. Continue gpupgrade even when gpinitsystem has warnings.
+		gpinitsystemWarning = exitErr.ExitCode() == 1
+		if gpinitsystemWarning {
+			gplog.Warn("gpinitsystem had warnings and exited with status %d", exitErr.ExitCode())
+		}
```
kalensk: A story has been created to fix gpinitsystem to not exit with 1 on warnings.

hub/services/execute_init_target_cluster_sub_step_test.go
```diff
@@ -93,109 +96,100 @@ var _ = Describe("Hub prepare init-cluster", func() {
...
-			err := hub.RunInitsystemForTargetCluster("filepath")
-			Expect(err.Error()).To(Equal("gpinitsystem failed: some output: exit status 2"))
-		})
+		//It("successfully runs gpinitsystem", func() {
+		//	testExecutor.LocalError = errors.New("exit status 1")
+		//	err := hub.RunInitsystemForTargetCluster(mockStream, "filepath")
+		//
+		//	Expect(err).To(BeNil())
+		//	testhelper.ExpectRegexp(stdout, "[WARNING]:-gpinitsystem completed with warnings")
+		//})
```
kalensk: I wasn't exactly sure what to do with this test. It seems a bit redundant given the existing tests. However, there is not yet a good test that indicates that we continue on if gpinitsystem has warnings and exists with status 1. Maybe something to add.

kalensk: Is it worth adding this test if it will be removed once we address the gpinitsystem exit code story?

jchampio: Let's not. It's busy work that we plan to throw away.

marksliva: Removed this test in a fixup commit.

kalensk:
This has bit-rotted. See https://github.com/greenplum-db/gpupgrade/pull/138 for a rebased version.

PR #131, opened 2019-10-02
finalize: implement reconfigure-ports substep

jchampio:
The reconfigure-ports substep has been fully reimplemented, and the CLI command to use it is now `gpupgrade finalize`.

The core of the reconfigure-ports implementation is `ClonePortsFromCluster`, which sets up the master gp_segment_configuration on the new cluster to match the old one. Ideally it would also perform

    ALTER SYSTEM SET port

on the new master, but bugs in gpstart and gpstop currently prevent us from doing that, so for now we keep the (brittle) sed replacement strategy for postgresql.conf that was in place previously.

The new end-to-end test sets up a new "post-upgrade" cluster and runs finalize on it. The new `testutils/dump_config` utility helps with this by writing the manually-constructed cluster's JSON configuration to disk.

The test can't actually run in the CI yet due to two bugs:

- kill_agents doesn't kill agents on all hosts in the cluster, and
- we expect `gpupgrade_*` to be in `$GPHOME/bin` instead of looking for it on the `PATH`. (This is the same bug that prevents us from running the execute end-to-end test.)

`delete_cluster` has been moved to `test/helpers`, since it's used from two places now. It also performs the port lookup via PID file, instead of the environment, since we end up changing ports during execution.

The fact that we ignore error codes coming from gpinitsystem and gpdeletesystem continues to be a travesty.


jchampio:
[dev pipeline](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:dev:finalize)

kalensk:
> * kill_agents doesn't kill agents on all hosts in the cluster, and

Do we have a story to address this?

> * we expect `gpupgrade_*` to be in `$GPHOME/bin` instead of looking for it on the `PATH`. (This is the same bug that prevents us from running the execute end-to-end test.)

I know there is a PR forthcoming for this, so maybe a story is not quite needed for this.

README_cli.md
```diff
@@ -1,197 +0,0 @@
-# gpupgrade command-line interface(cli)
```
kalensk: I am so glad that README_cli.md has been removed!

cli/commands/commands.go
```diff
@@ -287,10 +287,10 @@ var subStatusUpgrade = &cobra.Command{
...
 
 //////////////////////////////////////// UPGRADE and its subcommands
-var subUpgradeReconfigurePorts = &cobra.Command{
-	Use:   "reconfigure-ports",
-	Short: "Set master port on upgraded cluster to the value from the older cluster",
-	Long:  `Set master port on upgraded cluster to the value from the older cluster`,
+var finalize = &cobra.Command{
+	Use:   "finalize",
+	Short: "finalizes the cluster after upgrade execution",
+	Long:  "finalizes the cluster after upgrade execution",
```
kalensk: Totally not for this PR, but the spacing after the colon is triggering my OCD tendencies. The commands.go file needs a quick formatting pass through in regards to using `"` instead of `, spacing, etc. Maybe I can quickly do that after this is merged.

cli/commands/commands.go
```diff
@@ -287,10 +287,10 @@ var subStatusUpgrade = &cobra.Command{
...
 	Run: func(cmd *cobra.Command, args []string) {
 		client := connectToHub()
 		err := commanders.NewUpgrader(client).ReconfigurePorts()
```
kalensk: Is there a way of updating this line of code to more closely match "finalize". That is, maybe rename the file `upgrader.go`, and possibly remove the need for yet another "upgrader" object.

jchampio: I basically want to remove all "commanders" from the code base, but I don't really want to do that in this PR unless you feel strongly about it.

cli/commands/commands.go
jchampio: Spacing's controlled by `go fmt`, so, uh, good luck!

kalensk: darn!

cli/commands/commands.go
kalensk: I feel fairly strongly, but not necessarily for this PR. Seems ok for now, as long as its on the radar.

test/helpers.bash
```diff
@@ -52,3 +52,23 @@ kill_agents() {
...
+delete_cluster() {
+    local masterdir="$1"
+
+    # Sanity check.
+    if [[ $masterdir != *_upgrade/demoDataDir* ]]; then
+        abort "cowardly refusing to delete $masterdir which does not look like an upgraded demo data directory"
+    fi
+
+    # Look up the master port (fourth line of the postmaster PID file).
+    local port=$(awk 'NR == 4 { print $0 }' < "$masterdir/postmaster.pid")
```
kalensk: nice update!

test/helpers.bash
```diff
@@ -52,3 +52,23 @@ kill_agents() {
...
+
+    local gpdeletesystem="$GPHOME"/bin/gpdeletesystem
+
+    # XXX gpdeletesystem returns 1 if there are warnings. There are always
+    # warnings. So we ignore the exit code...
+    yes | PGPORT="$port" "$gpdeletesystem" -fd "$masterdir" || true
```
kalensk: Why not inline `$gpdeletesystem`? What's the purpose of having it a variable just a line above?

test/finalize.bats
```diff
@@ -0,0 +1,97 @@
...
+
+@test "finalize modifies ports on the live target cluster" {
+    skip "this test can't work until we fix hub and agent PATH lookup"
+
+    # To avoid spinning up an entire upgrade just to test finalize, we instead
+    # create a new cluster for the test and fake the configurations to point at
+    # it.
+    #
+    # XXX we assume three primaries (demo cluster layout)
+    # XXX we hardcode ports here, so we'll fail if there are any conflicts.
```
kalensk: This is definitely ok for this PR, and a note to myself that this may potentially become an issue if the future. Just something for me to keep in mind.

test/finalize.bats
```diff
@@ -0,0 +1,97 @@
...
+
+    # XXX There are always warnings, so ignore them...
+    gpinitsystem -ac "$STATE_DIR/gpinitsystem_config" 3>&- || true
+    NEW_CLUSTER="$STATE_DIR/_upgrade/demoDataDir-1"
+
+    # Generate a new target cluster configuration that the hub can use, then
+    # restart the hub.
+    PGPORT=40000 go run ./testutils/dump_config "$GPHOME/bin" "$GPUPGRADE_HOME/target_cluster_config.json"
+    kill_hub
+    gpupgrade_hub --daemonize 3>&-
```
kalensk: Ugh, this makes me really want start/stop/restart semantics for gpupgrade* across the entire cluster. Can we add a story for that?

Also, is the reason for restarting the hub to pick up the new config? I am assuming the config is not just read when needed...should it be?

test/finalize.bats
```diff
@@ -0,0 +1,97 @@
...
+
+    gpupgrade finalize
+
+    # Check to make sure the new cluster's ports match the old one.
+    local new_ports=$(get_ports)
+    [ "$OLD_PORTS" = "$new_ports" ] || fail "actual ports: $new_ports"
+}
```
kalensk: I think this is a great check, but I am also wanting a bit more reassurance that we are actually connecting to the target cluster and ensuring it has been upgraded. (How do we know we aren't really connecting to the old cluster accidentally when querying for new_ports, etc.?) For now, perhaps adding a check on the master_data_dirs is sufficient.

hub/services/upgrade_reconfigure_ports.go
```diff
@@ -36,13 +40,82 @@ func (h *Hub) UpgradeReconfigurePorts(ctx context.Context, in *idl.UpgradeReconf
...
+		return xerrors.Errorf("%s failed to start target cluster in utility mode: %w",
+			upgradestatus.RECONFIGURE_PORTS, err)
+	}
+
+	// 3). rewrite gp_segment_configuration with the updated port number
+	connURI := fmt.Sprintf("postgresql://localhost:%d/template1?gp_session_role=utility&allow_system_table_mods=true&search_path=", h.target.MasterPort())
+	targetDB, err := sql.Open("pgx", connURI)
+	defer func() {
+		targetDB.Close() //TODO: return multierror here to capture err from Close()
+	}()
```
kalensk: Why is this wrapped in a func? Because we will want to eventually check if it errors?

hub/services/upgrade_reconfigure_ports.go
```diff
@@ -36,13 +40,82 @@ func (h *Hub) UpgradeReconfigurePorts(ctx context.Context, in *idl.UpgradeReconf
...
+	if err != nil {
+		return xerrors.Errorf("%s failed to stop cluster: %w",
+			upgradestatus.RECONFIGURE_PORTS, err)
+	}
+
+	// 2). bring up the master(fts will not "freak out", etc)
+	script := fmt.Sprintf("source %s/../greenplum_path.sh && %s/gpstart -am -d %s",
+		h.target.BinDir, h.target.BinDir, h.target.MasterDataDir())
+	cmd := exec.Command("bash", "-c", script)
+	_, err = cmd.Output()
```
kalensk: Curious why `cmd.Output()` was chosen over say `cmd.Run()`? I know that if there is an error then the caller reads stderr from the exitErr, but I believe Run also does that.

test/finalize.bats
jchampio: Agreed!

hub/services/upgrade_reconfigure_ports.go
jchampio: Yes, thanks for catching that -- the `TODO` should be fixed before we push this.

hub/services/upgrade_reconfigure_ports.go
jchampio: No, `Run()` will discard stderr unless you explicitly tell it otherwise.

jchampio: (by setting `cmd.Stderr`)

hub/services/upgrade_reconfigure_ports.go
```diff
@@ -36,13 +40,82 @@ func (h *Hub) UpgradeReconfigurePorts(ctx context.Context, in *idl.UpgradeReconf
...
+		return xerrors.Errorf("%s failed to execute sed command: %w",
+			upgradestatus.RECONFIGURE_PORTS, err)
+	}
 
-	output, err := h.source.Executor.ExecuteLocalCommand(sedCommand)
+	// 6. bring up the cluster
+	script = fmt.Sprintf("source %s/../greenplum_path.sh && %s/gpstart -a -d %s",
+		h.target.BinDir, h.target.BinDir, h.target.MasterDataDir())
+	cmd = exec.Command("bash", "-c", script)
+	_, err = cmd.Output()
```
kalensk: Similar to above in how we call `StopCluster(h.target)`, would it be useful here to call `startNewCluster()` in `execute_start_start_clsuter_sub_step.go`. But rewrite `startNewCluster` to name it `startCluster` that takes a cluster, and remove its dependency on the hub. In fact [PR 12]( https://github.com/greenplum-db/gpupgrade/pull/128/files#diff-f07c939941aba1a6916ec49b1536fb8bR34) tries to do a bit of that.

hub/services/upgrade_reconfigure_ports.go
```diff
@@ -36,13 +40,82 @@ func (h *Hub) UpgradeReconfigurePorts(ctx context.Context, in *idl.UpgradeReconf
...
-	sedCommand := fmt.Sprintf(SedAndMvString, h.target.MasterPort(), h.source.MasterPort(), h.target.MasterDataDir())
-	gplog.Debug("executing command: %+v", sedCommand) // TODO: Move this debug log into ExecuteLocalCommand()
+// reconfigurePorts executes the tricky sequence of operations required to
+// change the ports on a cluster:
+//    1). bring down the cluster
+//    2). bring up the master(fts will not "freak out", etc)
+//    3). rewrite gp_segment_configuration with the updated port number
+//    4). bring down the master
+//    5). modify the master's config file to use the new port
+//    6). bring up the cluster
```
kalensk: Total nitpick, but I think the in-line comments for these is sufficient over repeating them above the method.

hub/services/finalize.go
```diff
@@ -0,0 +1,151 @@
...
+	return nil
+}
+
+// ClonePortsFromCluster will modify the gp_segment_configuration of the passed
+// sql.DB to match the cluster port settings from the source cluster.Cluster.
+//
+// As a reminder to developers, we don't have any mirrors up at this point on
+// the target cluster. We copy only the primary information. Good thing too,
+// because cluster.Cluster doesn't give us mirror info.
+func ClonePortsFromCluster(db *sql.DB, src *cluster.Cluster) (err error) {
```
kalensk: Total nitpick, can this method be moved to the top of the file. I tend to read files from top to bottom with the "order of execution or entry point functions at the top" and helper functions below it. Just preference and not a big deal either way.

hub/services/upgrade_reconfigure_ports.go
```diff
@@ -36,13 +40,82 @@ func (h *Hub) UpgradeReconfigurePorts(ctx context.Context, in *idl.UpgradeReconf
...
+// TODO: this method needs test coverage.
+func (h *Hub) reconfigurePorts() (err error) {
```
kalensk: I am ok with this PR as long as the tests get added soon and don't "slip through the cracks". If a story needs to be made or something then that is fine as well.

test/finalize.bats
jchampio: > Ugh, this makes me really want start/stop/restart semantics for gpupgrade* across the entire cluster

Start and stop would be nice. I'm not sure if restart would be useful in production... Maybe as a debugging tool?

> Also, is the reason for restarting the hub to pick up the new config?

Yes.

> I am assuming the config is not just read when needed...should it be?

I think expecting the hub to keep up with live changes to its configuration is a recipe for disaster. And the only time we need this behavior is in testing -- configs should not be manually modified in production environments.

test/helpers.bash
jchampio: It made the line hard to read for me when it was inlined. I don't feel strongly about putting it back that way, if you prefer.

hub/services/upgrade_reconfigure_ports.go
jchampio: Yes, but I'd like to move to `cluster.Start()` and `cluster.Stop()` in the end.

hub/services/upgrade_reconfigure_ports.go
jchampio: I agree.

hub/services/upgrade_reconfigure_ports.go
jchampio: Yeah, my hope is that we can get the substep infrastructure refactored soon and then we'll have a perfect framework to handle the "serial execution testing".

jchampio: (Unit tests are not a great way to test serial execution.)

kalensk: > (Unit tests are not a great way to test serial execution.)

agreed.

hub/services/finalize.go
```diff
@@ -0,0 +1,151 @@
...
+
+	// Make sure the content IDs in gp_segment_configuration match the source
+	// cluster exactly.
+	if err := sanityCheckContentIDs(tx, src); err != nil {
+		return err
+	}
+
+	for _, content := range src.ContentIDs {
+		port := src.Segments[content].Port
+		res, err := tx.Exec("UPDATE gp_segment_configuration SET port = $1 WHERE content = $2",
```
kalensk: Note to self: I wish that dbid was unique in gp_segment_configuration such that dbid could be used and the extra checks (while good and useful) would be less needed.

test/helpers.bash
kalensk: Honestly I was more curious than anything. I might have a slight preference over inlining, but also prefer readability - probably fine as is.

hub/services/finalize_test.go
```diff
@@ -0,0 +1,283 @@
...
+			c.check(t, err)
+		})
+	}
+}
+
+// match is a helper for the errorCases table test above. You can use it as the
+// errorCase.check() callback implementation; it just uses errors.Is() to see
+// whether the actual error returned matches the expected one, and complains
+// through the testing.T if not.
+func match(expected error) func(*testing.T, error) {
```
kalensk: I might rename this method to something like `expectedError` to make things a bit more self-documenting when reading the code.

hub/services/finalize_test.go
```diff
@@ -0,0 +1,283 @@
...
+			}
+			defer finishMock(mock, t)
+
+			// prepare() sets up any mock expectations.
+			c.prepare(mock)
+
+			err = ClonePortsFromCluster(db, src)
+
+			// check() verifies that the error is the one we expect.
+			c.check(t, err)
```
kalensk: Renaming `check` to something like `verifyError` might make the code more documenting and remove the need for the comment.

hub/services/finalize_test.go
```diff
@@ -0,0 +1,283 @@
...
+			mock.ExpectBegin()
+			mock.ExpectQuery("SELECT content FROM gp_segment_configuration").
+				WillReturnRows(contents)
+			mock.ExpectRollback()
+		},
+		func(t *testing.T, err error) {
+			// XXX It'd be nice to inject our ErrSentinel into the Scan
+			// function, but there doesn't seem to be a way to do that. We
+			// instead return junk values from our query, and search for the
+			// "scan error" hardcoded string. Blech.
```
kalensk: Is the last sentence still relevant?

cli/commands/commands.go
dkrieger-pvtl: As @kalen suggests, I'd prefer if we follow the Initialize and Execute idioms by calling this:

`commanders.Finalize(client)`.

We don't need to remove all the commanders in this PR, just renaming this command.

hub/services/finalize_test.go
jchampio: I'll give `expect()` a try. No reason to put `error` in the name when the argument is itself an error.

hub/services/finalize_test.go
jchampio: `verify` sounds fine to me, but again, `error` is implied by context. And code only documents implementation; comments document intent. They complement each other.

hub/services/finalize_test.go
jchampio: Yes. Is there a reason you ask?

cli/commands/commands.go
jchampio: Will do.

hub/services/upgrade_reconfigure_ports.go
jchampio: (leaving this for a followup)

jchampio:
Updated in response to feedback. gRPC reshuffling is done in a separate commit to divide it from the new implementation.

cli/commands/commands.go
kalensk: Thanks for making this change. I find it much easier and cleaner to see all three steps together now and following a similar pattern. (Still some work to go, but it is definitely coming together.)

PR #133, opened 2019-10-02
gpupgrade initialize status output

dkrieger-pvtl:
gpupgrade initialize console output

`gpupgrade initialize` now prints to stdout a scrolling list of steps.  When a step is in a final state, it goes onto the next line of output.  Otherwise, it overwrites the current line with the updated status.

This PR code is not intended to be the final code, but a demo as to the output and to solicit feedback on the approach.

Our thinking on overall approaches is below.  We thought step 3 made the most sense.

1.  CLI parses the stateDir
  - exposes the state_checker to CLI
  - simple, out-of-band communication mechanism
  - how will a remote CLI get info? (hub or ssh)
  - how will remote cli know when there's new info(can only watch a locally mounted remote fs)?

2. CLI parses a new stream from the hub (existing stream is for logs)

- hub sends latest information(send a message to stream in addition to writing to stateDir)
- the first stream message contains the complete hub state
- how do you display status if there's no current operation in progress?
- what do you do if the hub goes down? (restart it)

3.  CLI sends new gRPC messages to Hub to obtain the information

- a new "api" to get state from hub that abstracts out stateChecker
- allows agents to also track state, if ever desired
- how do you know when there are new messages? (polling)
- if there's only one gRPC message, that CLI then parses.
- what do you if the hub goes down? (restart it)


cli/commanders/step_initialize.go
```diff
@@ -90,12 +92,48 @@ func StartHub() error {
...
+						idl.UpgradeSteps_CONFIG:       true,
+						idl.UpgradeSteps_START_AGENTS: true,
+					})
+				fmt.Fprintf(os.Stdout, "%s", curStatus)
+			}
+
+			if stopRequested {
+				return
+			}
+			time.Sleep(500 * time.Millisecond)
```
kalensk: Why is this needed? Any way to avoid having sleeps in the code?

Overall, it seems like this is an async approach to displaying status, rather than having it be serially printed which would avoid a lot of this complication.

jchampio: Yeah, I think we probably want to do this serially, by streaming status back from the hub.

cli/tui/display.go
```diff
@@ -0,0 +1,131 @@
...
+	for _, status := range statuses {
+		if included[status.GetStep()] {
+			if status.GetStatus() == idl.StepStatus_COMPLETE ||
+				status.GetStatus() == idl.StepStatus_FAILED {
+				if !displayed[status.GetStep()] {
+					fmt.Fprintf(&s, "%s\n",
+						getDisplayLine(status.GetStep(), status.GetStatus()))
+					displayed[status.GetStep()] = true
+				}
+			} else {
```
kalensk: I haven't taken a really close look at this, but the three nested `if's` is saying there is hopefully a cleaner way of approaching this.

cli/tui/display.go
```diff
@@ -0,0 +1,131 @@
...
+
+	stepString := getStepString(step)
+	statusString := getStepStatusString(status)
+
+	numBrackets := 2
+	numDots := 3
+	maxLeftEndBackoff := getMaxStepStringLength()
+	maxRightEndBackoff := getMaxStepStatusStringLength()
+
+	stepStringPad := maxLeftEndBackoff - len(stepString)
```
jchampio: This is a lot of code for padding -- do the builtin padding specifiers from `fmt.Sprintf()` not work for this use case?

utils/idl_sorting_test.go
```diff
@@ -0,0 +1,308 @@
...
+		},
+		{
+			"reverse",
+			len2PrimaryStatus_2lt1,
+			args{0, 1},
+			false,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
```
jchampio: Just as a general comment, this is a case where I think the cost of the subtest/table-driven approach isn't justified. Just run two calls to `Less` with hardcoded arguments; it's easier to read.

cli/commanders/step_initialize.go
dkrieger-pvtl: See recently added discussion in the PR Description.

marksliva: We are now sending the StepStatus through the same stream as the stdout/stderr.

cli/tui/display.go
marksliva: This logic has been simplified.

utils/idl_sorting_test.go
marksliva: We removed the subtest approach for tests that had only 1-2 examples, but kept that approach for the other tests.

cli/tui/display.go
marksliva: We have changed this to use Sprintf.

PR #134, opened 2019-10-03
execute: add status streaming and console output

jchampio:
tty


jchampio:
There's a lot of copy-paste code that we want to refactor away; we just wanted to show the general approach.

jchampio:
Pulled into https://github.com/greenplum-db/gpupgrade/pull/137.

Issue #117, opened 2019-09-11
Hub (and probably agents) relies on PATH


jchampio:
It's annoying that gpupgrade can't find its own hub (and agent?) binaries without having PATH explicitly set:
```
$ /usr/local/greenplum-db-new/bin/gpupgrade prepare start-hub
20190911:18:19:39 gpupgrade_cli:gpadmin:mdw:015332-[ERROR]:-failed to start hub (exec: "gpupgrade_hub": executable file not found in $PATH)
```

marksliva:
Fixed in PR https://github.com/greenplum-db/gpupgrade/pull/135

PR #135, opened 2019-10-08
Fix path issues

marksliva:
Allow gpupgrade to easily find the hub and agent across hosts in the cluster by requiring gpupgrade binaries to be in the same location on each host. Also require gpupgrade to be in the user's PATH, instead of being in the greenplum bin directory, since gpupgrade is independent of the server. Another change was to enable the execute.bats test since it was blocked by the PATH issues.

This fixes issue https://github.com/greenplum-db/gpupgrade/issues/117


ci/scripts/upgrade-cluster.bash
```diff
@@ -100,7 +100,8 @@ make
 # Install the artifacts onto the cluster machines.
 artifacts='gpupgrade gpupgrade_hub gpupgrade_agent'
 for host in "${hosts[@]}"; do
-    scp $artifacts "gpadmin@$host:${GPHOME_NEW}/bin/"
+    scp $artifacts "gpadmin@$host:/tmp"
+    ssh centos@$host "sudo mv /tmp/gpupgrade* /usr/local/bin"
```
jchampio: Why not just
```sh
scp $artifacts "centos@$host:/usr/local/bin"
```
?

hub/services/step_initialize.go
```diff
@@ -130,16 +131,19 @@ func StartAgents(source *utils.Cluster, target *utils.Cluster, stateDir string)
 	// started, or do we just want to stop all of them and kick back to the
 	// user?
 	logStr := "start agents on master and hosts"
+	hubPath, err := os.Executable()
+	if err != nil {
+		return errors.Errorf("failed to get the hub executable path %v", err)
+	}
+	agentPath := filepath.Join(path.Dir(hubPath), "gpupgrade_agent")
```
jchampio: Use `filepath.Dir()` rather than `path.Dir()`.

[`path` is for forward-slash-paths (e.g. URLs); `filepath` is for filesystem paths.](https://golang.org/pkg/path/) They almost always do the same thing on *nix, so it's not an emergency, but keep it in mind.

test/execute.bats
```diff
@@ -14,13 +16,17 @@ setup() {
 }
 
 teardown() {
+    skip_if_no_gpdb
```
jchampio: Wait, skipping teardown works? That's cool, and makes code like [this](https://github.com/greenplum-db/gpupgrade/blob/66527fe587c814021fa6ef62e15ac6987ed4838b/test/initialize.bats#L25) unnecessary...

test/helpers.bash
```diff
@@ -71,4 +72,9 @@ delete_cluster() {
     # XXX gpdeletesystem returns 1 if there are warnings. There are always
     # warnings. So we ignore the exit code...
     yes | PGPORT="$port" "$gpdeletesystem" -fd "$masterdir" || true
+
+    # gpdeletesystem moves the datadir's to .old rather than removing them.
```
jchampio: That... should not be the case.

ci/scripts/upgrade-cluster.bash
kalensk: One [would hope](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/kalen_path_test/jobs/6-to-6/builds/1#L5d9d3b96:21:26)...
```
+ artifacts='gpupgrade gpupgrade_hub gpupgrade_agent'
+ for host in '"${hosts[@]}"'
+ scp gpupgrade gpupgrade_hub gpupgrade_agent centos@mdw:/usr/local/bin
scp: /usr/local/bin/gpupgrade: Permission denied
scp: /usr/local/bin/gpupgrade_hub: Permission denied
scp: /usr/local/bin/gpupgrade_agent: Permission denied
```

Is there an easy and elegant way around that?

test/helpers.bash
kalensk: Running the execute test and gpdeletesystem by hand shows the following directories still present:
```
→ find . -path  "*_upgrade" -exec find {} -maxdepth 1  \;
./dbfast2_upgrade
./dbfast2_upgrade/demoDataDir1.old
./dbfast3_upgrade
./dbfast3_upgrade/demoDataDir2.old
./dbfast1_upgrade
./dbfast1_upgrade/demoDataDir0.old
./qddir_upgrade
```

Steps:
```
rm -rf ~/.gpupgrade && killall gpupgrade*

make install && gpupgrade initialize --old-bindir "$GPHOME/bin" --new-bindir "$GPHOME/bin" --old-port 15432 && ps -ef | grep gpupgrade

gpupgrade execute

export masterdir=/Users/user/workspace/gpdb6/gpAux/gpdemo/datadirs/qddir_upgrade/demoDataDir-1
port=$(awk 'NR == 4 { print $0 }' < "$masterdir/postmaster.pid")

PGPORT=$port gpdeletesystem  -fd "$masterdir"
```

brianfeaster: I knew I'd find a find within find someday!

kalensk: Comment needs to be updated since it was a misunderstanding.

jchampio: When you do, can you tag it `TODO` or `XXX`? This code can go away when we fix that bug in the agent.

hub/services/step_initialize.go
marksliva: Changed this in a fixup commit.

test/execute.bats
marksliva: Yes it works! We figured this is cleaner than refactoring and wrapping this code in a conditional.

test/helpers.bash
marksliva: Pushed a fixup that addresses this.

PR #137, opened 2019-10-09
output progress of initialize, execute and finalize

dkrieger-pvtl:
The steps of a gpupgrade now stream the status of their substeps
back to the cli, which displays them.  The currently executing
substep is shown as "IN_PROGRESS" until it completes("COMPLETE") or
fails("FAILED").

Here is the help output:

```
> gpupgrade initialize --help

Runs through pre-upgrade checks and prepares the old and new clusters for upgrade.
This step can be reverted.

Usage:
  gpupgrade initialize [flags]

Flags:
  -h, --help                help for initialize
      --new-bindir string   install directory for new gpdb version
      --old-bindir string   install directory for old gpdb version
      --old-port int        master port for old gpdb cluster

> gpupgrade execute --help

Upgrades the master and primary segments over to the new cluster.
This step can be reverted.

Usage:
  gpupgrade execute [flags]

Flags:
  -h, --help      help for execute
  -v, --verbose   print the output stream from all substeps

> gpupgrade finalize --help

Updates the port of the new cluster.
This step can not be reverted.

Usage:
  gpupgrade finalize [flags]

Flags:
  -h, --help   help for finalize
```

## Pretty GIF

output


hub/services/step_initialize.go
```diff
@@ -8,33 +8,29 @@ import (
 	"github.com/greenplum-db/gp-common-go-libs/cluster"
```
marksliva: We should name these files consistently. Either prepend the others with `step_` or remove `step_` from this file.

cli/commanders/steps.go
```diff
@@ -0,0 +1,229 @@
...
+			}
+			lastStep = x.Status.Step
+
+			fmt.Printf(FormatStatus(x.Status))
+			if verbose {
+				fmt.Println()
+			}
+
+		default:
+			panic(fmt.Sprintf("Unknown message type for Execute: %T", x))
```
marksliva: Remove `for Execute` since all 3 of the steps are calling this shared function.
Also, why don't we return an error here instead of the panic?

dkrieger-pvtl: We panic here because this would be caused by an internal, presumably unrecoverable error.  The protobuf uses `oneof` and you'd have to hand-write a protobuf buffer inside the protobuf layer to generate a different type.

cli/commanders/steps.go
```diff
@@ -0,0 +1,229 @@
...
+validation desired prior to finalizing your upgrade.
+
+WARNING: If any queries modify the database during this time, this will affect
+your revert time.
+
+If you are satisfied with the state of the cluster, run "gpupgrade finalize" on
+the command line to finish the upgrade.
+
+If you would like to return the cluster to its original state, run
+"gpupgrade revert" on the command line.`)
```
marksliva: Should we remove this part since it is not yet implemented?

jchampio: That's up to @bharath1028.

hub/services/step_initialize.go
jchampio: Done (moved to `initialize.go`).

cli/commands/commands.go
```diff
@@ -323,8 +330,19 @@ func initialize() *cobra.Command {
 			}
 
 			// TODO: how do we rollback here?
-			return commanders.NewVersionChecker(client).Execute()
+			err = commanders.NewVersionChecker(client).Execute()
```
marksliva: We can refactor NewVersionChecker to pass the client to the `Execute` function, and then we could get rid of the initializer to make this match what we're doing for commanders/steps. But since we are considering removing commanders altogether, maybe we can hold off on this for now. If we decide to refactor it, we can also refactor the other commanders that have initializers too.

jchampio: Yeah, I'd like to defer commanders refactoring to the "get rid of it" step, if that's okay.

hub/services/execute_upgrade_primaries_sub_step.go
```diff
@@ -15,9 +15,22 @@ import (
 	"github.com/greenplum-db/gpupgrade/idl"
 )
 
-func (h *Hub) ExecuteUpgradePrimariesSubStep() error {
+func (h *Hub) ExecuteUpgradePrimariesSubStep(stream idl.CliToHub_ExecuteServer) error {
```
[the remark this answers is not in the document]
jchampio: Good call; done.

PR #138, opened 2019-10-12
Execute: stream output for single-host steps

kalensk:
Stream stdout and stderr for execute sub-steps that don't need to stream from multiple hosts.

All single-host subSteps also write to execute.log. This may be useful later for gpupgrade attach.

Collapse the common parts of running a subStep into ExecuteSubStep.

multiplexedStream and its tests have been moved to execute.go since it is now used in multiple places.

[test pipeline](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/kalen_rebased-execute-stream-single-host)


hub/upgradestatus/checklist_manager.go
```diff
@@ -14,12 +14,12 @@ import (
...
 	SHUTDOWN_CLUSTERS      = "shutdown-clusters"
-	CONVERT_MASTER         = "convert-master"
+	UPGRADE_MASTER         = "upgrade-master"
 	COPY_MASTER            = "copy-master"
-	CONVERT_PRIMARIES      = "convert-primaries"
-	VALIDATE_START_CLUSTER = "validate-start-cluster"
+	UPGRADE_PRIMARIES      = "upgrade-primaries"
+	START_TARGET_CLUSTER   = "start-target-cluster"
 	RECONFIGURE_PORTS      = "reconfigure-ports"
 )
```
dkrieger-pvtl: The variable names here for the checklist(`INIT_TARGET_CLUSTER`, etc) and the step names in `enum UpgradeSteps` in `cli_to_hub.proto` should match exactly.  Otherwise, it becomes confusing as to which steps correspond.  Also, consider renaming `UPGRADE PRIMARIES to `UPGRADE_TARGET_PRIMARIES`.

hub/services/execute.go
```diff
@@ -1,35 +1,146 @@
...
-	err := h.ExecuteInitTargetClusterSubStep(stream)
+	// Create a log file to contain execute output.
+	log, err := utils.System.OpenFile(
+		filepath.Join(utils.GetStateDir(), "execute.log"),
+		os.O_WRONLY|os.O_CREATE,
+		0600,
+	)
 	if err != nil {
 		return err
 	}
```
dkrieger-pvtl: add a `defer log.Close()` here.

hub/upgradestatus/checklist_manager.go
kalensk: Just pushed a commit that addresses this.

hub/services/execute.go
```diff
@@ -1,35 +1,146 @@
...
+	h.log = log
 
-	err = h.ExecuteShutdownClustersSubStep(stream)
+	_, err = h.log.WriteString("\nExecute in progress.\n")
+	if err != nil {
+		return xerrors.Errorf("failed writing to execute log: %w", err)
+	}
+
+	h.stream = stream
```
dkrieger-pvtl: ` h.log` and `h.stream` are per-gRPC call on the hub, not part of the Hub itself.  So I'd prefer they both get passed into each `ExecuteSubStep` call, or collected into a helper struct.

dkrieger-pvtl: Is the `stream` implicitly close by the gRPC infrastructure?

kalensk: Right. I was debating between adding it to the hub or not, and was originally passing it in, but it seemed a tad weird that all the class had the same argument being passed.

I will look into changing it.

kalensk: [Correct](https://godoc.org/google.golang.org/grpc#ServerStream).

hub/services/execute.go
kalensk: fixed.

hub/services/execute.go
kalensk: Updated.

cli/commanders/reporter.go
```diff
@@ -26,12 +26,12 @@ var UpgradeStepsMessage = map[idl.UpgradeSteps]string{
 	idl.UpgradeSteps_UNKNOWN_STEP:           "- Unknown step",
 	idl.UpgradeSteps_CONFIG:                 "- Configuration Check",
 	idl.UpgradeSteps_START_AGENTS:           "- Agents Started on Cluster",
-	idl.UpgradeSteps_INIT_CLUSTER:           "- Initialize new cluster",
-	idl.UpgradeSteps_CONVERT_MASTER:         "- Run pg_upgrade on master",
+	idl.UpgradeSteps_INIT_TARGET_CLUSTER:    "- Initialize new cluster",
+	idl.UpgradeSteps_UPGRADE_MASTER:         "- Run pg_upgrade on master",
```
marksliva: Switch `UpgradeSteps_UPGRADE_MASTER` and `UpgradeSteps_SHUTDOWN_CLUSTERS` so that it matches the `UpgradeStepsOrder` steps order.

cli/commanders/steps.go
```diff
@@ -20,12 +20,12 @@ type receiver interface {
 var lines = map[idl.UpgradeSteps]string{
 	idl.UpgradeSteps_CONFIG:                 "Retrieving configs...",
 	idl.UpgradeSteps_START_AGENTS:           "Starting agents...",
-	idl.UpgradeSteps_INIT_CLUSTER:           "Creating new cluster...",
+	idl.UpgradeSteps_INIT_TARGET_CLUSTER:    "Creating new cluster...",
```
marksliva: I like these step renames. This improves the code clarity 👍

hub/services/execute.go
```diff
@@ -1,26 +1,65 @@
...
+		os.O_WRONLY|os.O_CREATE,
+		0600,
+	)
 	if err != nil {
 		return err
 	}
+	defer func() {
+		closeErr := log.Close()
+		if closeErr != nil {
+			err = multierror.Append(err,
```
marksliva: Are you sure that this error gets assigned correctly? I tried to see if it works in a basic example, however I didn't see the error from the defer func() get assigned to the `err` in the outer scope.

[test.go]
```go
package main

import (
		"fmt"
		"errors"
                "time"
)

func shouldReturnAnError() error {
		_, err := fmt.Println("HI")
		defer func () {
				err = errors.New("my-new-error")
		}()
                time.Sleep(3 * time.Second)
		return err
}

func main () {
		err := shouldReturnAnError()
		fmt.Printf("the err is %#v", err)
}
```
Result:
```
$ go run test.go
HI
the err is <nil>
```

Expected:

```
$ go run test.go
HI
the err is &errors.errorString{s:"my-new-error"}
```

cli/commanders/reporter.go
kalensk: Good catch. Fixed.

hub/services/execute.go
kalensk: I am not yet sure why your example is not working, but I believe it should be based off these [blog](https://www.joeshaw.org/dont-defer-close-on-writable-files/) [posts](https://blog.learngoprogramming.com/5-gotchas-of-defer-in-go-golang-part-iii-36a1ab3d6ef1) and [referred example](https://play.golang.org/p/PZGn5-1TXv4).

hub/services/execute.go
```diff
@@ -1,26 +1,66 @@
...
 	"github.com/greenplum-db/gpupgrade/idl"
+	"github.com/greenplum-db/gpupgrade/utils"
 )
 
+type ExecuteStream struct {
+	stream messageSender
+	log *os.File
+}
+
 func (h *Hub) Execute(request *idl.ExecuteRequest, stream idl.CliToHub_ExecuteServer) error {
```
dkrieger-pvtl: change this to have a _named_ return value, i.e. `(err error)` vs `error`, so that the defer statement can add to the return value.  See below.

hub/services/execute.go
dkrieger-pvtl: See https://blog.golang.org/defer-panic-and-recover.  `defer` can only modify a  _named_ return value:

```
package main

import (
		"fmt"
)

func nochange() (int) {
    i:=0
    defer func() {i++}()	
    return i
}

func change() (i int) {
    i=0
    defer func() {i++}()	
    return i
}

func main () {
    fmt.Printf("NOCHANGE: %d CHANGE: %d\n", nochange(),change())
}
```

outputs: `NOCHANGE: 0 CHANGE: 1`

marksliva: I see, defining (err error) at the function level is how to make it work. Like this:
```go
package main

import (
		"fmt"
		"errors"
		"time"
)

func shouldReturnAnError() (err error) {
		_, err = fmt.Println("HI")
		defer func () {
				err = errors.New("my-new-error")
		}()
		time.Sleep(3 * time.Second)
		return errors.New("my-new-error")
}

func main () {
		err := shouldReturnAnError()
		fmt.Printf("the err is %#v", err)
}
```
Now we get the expected output:
```
$ go run test.go
HI
the err is &errors.errorString{s:"my-new-error"}
```

hub/services/execute.go
```diff
@@ -1,26 +1,66 @@
...
-	err := h.ExecuteInitTargetClusterSubStep(stream)
+	// Create a log file to contain execute output.
+	log, err := utils.System.OpenFile(
+		filepath.Join(utils.GetStateDir(), "execute.log"),
```
marksliva: Why are we logging to this file? If we want to know the state of the upgrade, we can inspect the sub-step state directories. If there was other more useful information that we logged here then I might be convinced that we need it, but if not then replace `log.WriteString` with `gplog.Info`.

marksliva: Oh nevermind, I think I now see how this is set up. That log file eventually gets used in some of the substeps.

Issue #141, opened 2019-10-15
gpupgrade execute fails to initialize demo cluster


kalensk:
With the change of demo cluster ports (https://github.com/greenplum-db/gpdb/pull/8746 and https://github.com/greenplum-db/gpdb/pull/8843) gpupgrade execute fails when initializing a demo cluster. This is because we simply increment the [current master port by one](), which is now colliding with the master standby causing the following gpinitsystem errors:

```
[INFO]:-Start Function CHK_QES_FROM_INPUTFILE
[INFO]:-Checking Master host
[INFO]:-Start Function CHK_DIR
[INFO]:-End Function CHK_DIR
[INFO]:-Start Function GET_PG_PID_ACTIVE
[INFO]:-Have lock file /tmp/.s.PGSQL.7001.lock and a process running on port 7001
[INFO]:-End Function GET_PG_PID_ACTIVE
[INFO]:-Start Function ERROR_EXIT
[FATAL]:-Found indication of postmaster process on port 7001 on Master host Script Exiting
```

kalensk:
Closing as the `temp_port_range` flag helps address this. However, it would be nice to check early-on if the user specified `temp_port_range` collides with the current source cluster's ports in use.

Issue #140, opened 2019-10-15
checkpoint_segments is removed in GPDB7


kalensk:
gpupgrade execute fails when initializing a new GPDB7 cluster since it attempts to make the following call which fails: `SELECT current_setting('checkpoint_segments') AS string` from [execute_init_target_cluster_sub_step.go:GetCheckpointSegmentsAndEncoding](https://github.com/greenplum-db/gpupgrade/blob/0d0322754407158e7d43f13b1e90eb076b2e7653/hub/services/execute_init_target_cluster_sub_step.go#L80). checkpoint_segments have been removed in GPDB7 (see https://github.com/greenplum-db/gpdb/pull/8659).

I am not exactly sure what checkpoint_segments` does or if its still relevent. _If so_, it seems like it should be wrapped an an if guard. 
```
If less than GPDB7 {
    checkpointSegments, err := dbconn.SelectString(dbConnector, "SELECT current_setting('checkpoint_segments') AS string")
}
```

kalensk:
Closing as it has been fixed.

PR #144, opened 2019-10-17
Migrate execute init cluster tests from ginkgo

kalensk:
This ports the existing execute init target cluster tests to Go style from ginkgo. Some minor refactors were made to decouple methods from the Hub for testability.

***Note***: This is an initial pass of simply porting the existing tests. There may be some room for some further refactors or organization which is outside the scope of this current PR. Further updates will be made when we fix the following related issues: https://github.com/greenplum-db/gpupgrade/issues/140 https://github.com/greenplum-db/gpupgrade/issues/141

Namely, this does not bring up a real Agent Server and grpc Server which the services_suite_test did.

Gomock is used to mock the grpc client.

[test pipeline](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:port_execute_tests)


hub/services/execute_init_target_cluster_sub_step.go
```diff
@@ -119,9 +116,9 @@ func WriteInitsystemFile(gpinitsystemConfig []string, gpinitsystemFilepath strin
 	return nil
 }
 
-func (h *Hub) DeclareDataDirectories(gpinitsystemConfig []string) ([]string, map[string][]string, int) {
+func DeclareDataDirectories(sourceCluster utils.Cluster, gpinitsystemConfig []string) ([]string, map[string][]string, int) {
```
marksliva: Did we decide if we want to start with a lower-case letter for functions that are only used locally? Our repo has a mix of both styles and I'm not sure which way we prefer.

hub/services/execute_init_target_cluster_sub_step_test.go
```diff
@@ -1,165 +1,355 @@
...
-})
+func TestGetMasterSegPrefix(t *testing.T) {
+		NewGomegaWithT(t)
+		ctrl := gomock.NewController(GinkgoT())
+		defer ctrl.Finish()
+
+		testhelper.SetupTestLogger() // initialize gplog
+
+		t.Run("returns a valid seg prefix given", func(t *testing.T) {
+			errorCases := []struct {
```
marksliva: Should these be called successCases?

hub/services/execute_init_target_cluster_sub_step_test.go
```diff
@@ -1,165 +1,355 @@
...
+			return nil, errors.New("permission denied")
+		}
 
-			err := hub.CreateAllDataDirectories(fakeConns, segDataDirMap)
-			Expect(err).To(HaveOccurred())
-			Expect(err.Error()).To(ContainSubstring(createErr.Error()))
-		})
+		err := CreateAllDataDirectories([]*Connection{}, segDataDirMap, sourceMasterDataDir)
+		expectedErr := errors.New("Error statting new directory /data/qddir_upgrade: permission denied")
+		if xerrors.Is(err,expectedErr) {
```
jchampio: You mean `!xerrors.Is`, presumably?

I think this is currently not working because your two `errors.New` calls have now connection to each other. You could instead store the `permission denied` error and check it:
```go
expected := errors.New("permission denied")
utils.System.Stat = func(...) {
    return nil, expected
}
...
    ...
}
```

hub/services/execute_init_target_cluster_sub_step_test.go
```diff
@@ -1,165 +1,355 @@
...
+			t.Errorf("wanted: '%s' got: '%#v'", expectedErr, err)
```
jchampio: nit: the use of `%#v` typically sets things apart enough that you shouldn't need quoting.

hub/services/execute_init_target_cluster_sub_step_test.go
```diff
@@ -1,165 +1,355 @@
...
-)`, dir)}
-			resultConfig, resultMap, port := hub.DeclareDataDirectories([]string{})
-			Expect(resultMap).To(Equal(segDataDirMap))
-			Expect(resultConfig).To(Equal(expectedConfig))
-			Expect(port).To(Equal(15433))
-		})
+func TestCreateAllDataDirectories(t *testing.T) {
+	NewGomegaWithT(t)
+	ctrl := gomock.NewController(GinkgoT())
+	defer ctrl.Finish()
```
jchampio: Same `ctrl.Finish()`/`GinkgoT()` comment as in the other PR (here and elsewhere).

hub/services/execute_init_target_cluster_sub_step_test.go
jchampio: Just `cases` works nicely too.

hub/services/execute_init_target_cluster_sub_step_test.go
```diff
@@ -1,165 +1,355 @@
...
+			{"a path without a content identifier", "/opt/myseg"},
+			{"a path with a segment content identifier", "/opt/myseg2"},
+			{"a path that is only a content identifier", "-1"},
+			{"a path that ends in only a content identifier", "///-1"},
+		}
+
+		for _, c := range errorCases {
+			_, err := GetMasterSegPrefix(c.MasterDataDir)
+			if err == nil {
+				t.Fatalf("expected err, but got: %v", err)
```
jchampio: This will always print the same thing (`<nil>`).

hub/services/execute_init_target_cluster_sub_step_test.go
```diff
@@ -1,165 +1,355 @@
...
+			}{
+				{"an absolute path", "/data/master/gpseg-1"},
+				{"a relative path", "../master/gpseg-1"},
+				{"a implicitly relative path", "gpseg-1"},
+			}
+
+			for _, c := range errorCases {
+				segPrefix, err := GetMasterSegPrefix(c.MasterDataDir)
+				if segPrefix != "gpseg" {
+					t.Errorf("wanted: gpseg got: %s", segPrefix)
```
jchampio: Try `%q` for strings; it will delimit for you (with quotes) and show you special characters with slash-syntax like `\n`, which is helpful when trying to grok error messages from the tests. So several of the Go standard library tests do something like
```go
t.Errorf("returned %q, want %q", actual, expected)
```

hub/services/execute_init_target_cluster_sub_step_test.go
kalensk: I don't think that quite works when I tried it since CreateAllDataDirectories() wraps the error with `Error statting new directory /data/qddir_upgrade`.

```
wanted: permission denied got: Error statting new directory /data/qddir_upgrade: permission denied
```

I also tried reflect.deepEquals. So far what I came up with is gross error string compare with err.Error.

hub/services/execute_init_target_cluster_sub_step_test.go
kalensk: Fixed.

hub/services/execute_init_target_cluster_sub_step_test.go
kalensk: That's cool! I did not know that. Thanks :)

hub/services/execute_init_target_cluster_sub_step_test.go
jchampio: As long as the implementation correctly wraps the underlying error (e.g. `xerrors.Errorf("blah: %w", ...)`), `xerrors.Is` will handle it.

kalensk: Ah! Thanks. I fixed up the errors that needed it for the existing tests.

hub/services/execute_init_target_cluster_sub_step.go
```diff
@@ -130,9 +128,9 @@ func (h *Hub) DeclareDataDirectories(gpinitsystemConfig []string) ([]string, map
 	// declare segment data directories
 	segmentDataDirMap := map[string][]string{}
 	segmentDeclarations := []string{}
-	for _, content := range h.source.ContentIDs {
+	for _, content := range sourceCluster.ContentIDs {
```
dkrieger-pvtl: nit: `sourceCluster.Segments` is a `map[contentID]SegmentConfig`, so if might be clearer here to do:
```
for contentID, segment := range sourceCluster.segments
```
and modify the code accordingly.

hub/services/execute_init_target_cluster_sub_step.go
```diff
@@ -12,13 +11,15 @@ import (
 	"strings"
 	"sync"
 
+	"github.com/pkg/errors"
+	"golang.org/x/xerrors"
 	"github.com/greenplum-db/gp-common-go-libs/dbconn"
 	"github.com/greenplum-db/gp-common-go-libs/gplog"
+	"golang.org/x/net/context"
+
```
dkrieger-pvtl: These imports should be ordered as per our standard:
```
import (
	"fmt"
	"io"
	"io/ioutil"
	"os"
	"os/exec"
	"path"
	"path/filepath"
	"strings"
	"sync"

	"github.com/pkg/errors"
	"golang.org/x/net/context"
	"golang.org/x/xerrors"

	"github.com/greenplum-db/gp-common-go-libs/dbconn"
	"github.com/greenplum-db/gpupgrade/db"
)
```

hub/services/execute_init_target_cluster_sub_step.go
kalensk: I will leave that for a followup PR as this one is only porting over the existing tests and a few minor changes associated with making the go-style tests pass. I just want to be careful to not let the scope or effort blow-up for this PR as you correctly pointed out the implementation and tests need work.

hub/services/execute_init_target_cluster_sub_step.go
kalensk: Fixed with the order being: standard libs, third party (including gp-common-go-libs), gpupgrade, and dot imports.

In other PRs and commits I was incorrectly combining gp-common-go-libs with gpupgrade.

hub/services/execute_init_target_cluster_sub_step.go
```diff
@@ -119,9 +117,9 @@ func WriteInitsystemFile(gpinitsystemConfig []string, gpinitsystemFilepath strin
 	return nil
 }
 
-func (h *Hub) DeclareDataDirectories(gpinitsystemConfig []string) ([]string, map[string][]string, int) {
+func DeclareDataDirectories(sourceCluster utils.Cluster, gpinitsystemConfig []string) ([]string, map[string][]string, int) {
```
dkrieger-pvtl: nit: Looking at the usage of this function in the top-level call `InitTargetCluster`, please reorder the signature to put `gpinitsystemConfig` before `sourceCluster`.

hub/services/execute_init_target_cluster_sub_step.go
```diff
@@ -215,31 +213,31 @@ func GetMasterSegPrefix(datadir string) (string, error) {
...
+		}(conn)
 	}
 
 	wg.Wait()
 	close(errChan)
 
 	for err := range errChan {
 		if err != nil {
-			return errors.Wrap(err, "Error creating segment data directories")
+			return xerrors.Errorf("failed to create segment data directories: %w", err)
```
dkrieger-pvtl: I think we want to return a `multierror` here to capture the error on each segment host.

kalensk: If possible I would like to leave that for another PR and keep this to only porting over the existing tests and a few minor changes associated with making the go-style tests pass. I just want to be careful to not let the scope or effort blow-up for this PR as you correctly pointed out the implementation and tests need further work.

hub/services/execute_init_target_cluster_sub_step.go
kalensk: Sure thing.

hub/services/execute_init_target_cluster_sub_step_test.go
```diff
@@ -1,165 +1,411 @@
...
+		ctrl := gomock.NewController(t)
+		defer ctrl.Finish()
+
+		execCommand = exectest.NewCommandWithVerifier(gpinitsystem,
+			func(path string, args ...string) {
+				g.Expect(path).To(Equal("bash"))
+				g.Expect(args).To(Equal([]string{"-c", "source /target/greenplum_path.sh && " +
+					"/target/bin/gpinitsystem -a -I /home/gpadmin/.gpupgrade/gpinitsystem_config"}))
+			})
+
```
dkrieger-pvtl: `defer execCommand=exec.Command` here to make this test self-contained.

hub/services/execute_init_target_cluster_sub_step_test.go
```diff
@@ -1,165 +1,411 @@
...
 
-})
+	t.Run("when gpinitsystem has a warning it logs and does not return an error", func(t *testing.T) {
+		ctrl := gomock.NewController(t)
+		defer ctrl.Finish()
+
+		_, _, log := testhelper.SetupTestLogger() // initialize gplog
+
+		execCommand = exectest.NewCommand(gpinitsystem_Warnings)
+
```
dkrieger-pvtl: `defer execCommand=exec.Command` here to make this test self-contained.

hub/services/execute_init_target_cluster_sub_step_test.go
```diff
@@ -1,165 +1,411 @@
...
+			t.Errorf("want: %q got: %q", expected, actual)
+		}
+	})
+
+	t.Run("when gpinitsystem fails it returns an error", func(t *testing.T) {
+		ctrl := gomock.NewController(t)
+		defer ctrl.Finish()
+
+		execCommand = exectest.NewCommand(gpinitsystem_Errors)
+
```
dkrieger-pvtl: `defer execCommand=exec.Command` here to make this test self-contained.

hub/services/execute_init_target_cluster_sub_step.go
dkrieger-pvtl: Ok, but please add a TODO comment in the code.

hub/services/execute_init_target_cluster_sub_step_test.go
kalensk: I don't think these are needed once once #143 goes in as [Jacob pointed out](https://github.com/greenplum-db/gpupgrade/pull/145#discussion_r337181726). Let me know what you think.

> As for execCommand, #143 is going to make sure that it's always set to nil when the test executable starts using an init function.

hub/services/execute_init_target_cluster_sub_step.go
kalensk: Added a comment.

hub/services/execute_init_target_cluster_sub_step_test.go
dkrieger-pvtl: The `init()` function is only run once, at module load time.  My comment is to make the tests manifestly standalone.  If you do not reset the `execCommand` in test A, and test B uses it without setting it to something else, then test B will use test A's execCommand.  You could use an "afterEach" semantic here too, instead, but somehow `execCommand` needs to be reset to something after each test.  `nil` is useful if we are sure we do not want to use `exec.Command` in our test AND in any other test we'll be running that somehow expects it to be set to that.

jchampio: > My comment is to make the tests manifestly standalone.

That's really only important if you _want_ each test to be standalone. Go-style tests allow tighter coupling between test runs to cut down on magic boilerplate, by letting you run multiple tests against the same environment.

> If you do not reset the execCommand in test A, and test B uses it without setting it to something else, then test B will use test A's execCommand.

Yes. And if we're all aware of that, and if the tests are driven to fail first (you _are_ writing the tests to fail first, right?), then I think that's fine. In fact it makes it easier to write several tests that use the same `execCommand` value and run them serially.

Even if the consensus is to reset after every test, do _not_ reset the value to `exec.Command`. That's never an appropriate value in a unit test.

hub/services/execute_init_target_cluster_sub_step.go
```diff
@@ -119,9 +117,9 @@ func WriteInitsystemFile(gpinitsystemConfig []string, gpinitsystemFilepath strin
 	return nil
 }
 
-func (h *Hub) DeclareDataDirectories(gpinitsystemConfig []string) ([]string, map[string][]string, int) {
+func DeclareDataDirectories(gpinitsystemConfig []string, sourceCluster utils.Cluster) ([]string, map[string][]string, int) {
```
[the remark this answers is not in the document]
kalensk: Personally I think sourceCluster is more self documenting. I'll change it to source so future people don't accidentally pass in target. I suppose cluster may be redundant based on the type being a cluster.

hub/services/execute_init_target_cluster_sub_step_test.go
```diff
@@ -1,165 +1,411 @@
...
+		expectedDataDirMap := map[string][]string{
+			"host1": {"/data/dbfast1_upgrade"},
+			"host2": {"/data/dbfast2_upgrade"},
+		}
+		if !reflect.DeepEqual(expectedDataDirMap, actualSegDataDirMap) {
+			t.Errorf("wanted: %v, got: %v", expectedDataDirMap, actualSegDataDirMap)
+		}
+
+		expectedPort := 15433
+		if !reflect.DeepEqual(expectedPort, actualPort) {
```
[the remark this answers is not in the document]
kalensk: fixed

hub/services/execute_init_target_cluster_sub_step_test.go
```diff
@@ -1,165 +1,411 @@
...
+		}
+
+		if actual.ExitCode() != 2 {
+			t.Errorf("want: 2 got: %d", actual.ExitCode())
+		}
+	})
+}
+
+func TestGetMasterSegPrefix(t *testing.T) {
+		ctrl := gomock.NewController(GinkgoT())
```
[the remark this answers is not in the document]
kalensk: fixed

hub/services/execute_init_target_cluster_sub_step_test.go
```diff
@@ -1,165 +1,411 @@
...
+	defer ctrl.Finish()
+
+	mockStream := mock_idl.NewMockCliToHub_ExecuteServer(ctrl)
+	mockStream.EXPECT().
+		Send(gomock.Any()).
+		AnyTimes()
+
+	execCommand = nil
+	defer func() {
+		execCommand = exec.Command
```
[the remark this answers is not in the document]
kalensk: fixed

hub/services/execute_init_target_cluster_sub_step_test.go
```diff
@@ -1,165 +1,411 @@
...
+		gpinitsystem_Errors,
 	)
+}
+
+func TestCreateInitialInitsystemConfig(t *testing.T) {
+	testhelper.SetupTestLogger() // initialize gplog
+
+	t.Run("successfully get initial gpinitsystem config array", func(t *testing.T) {
+		ctrl := gomock.NewController(t)
+		defer ctrl.Finish()
```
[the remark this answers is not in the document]
kalensk: fixed

hub/services/execute_init_target_cluster_sub_step_test.go
```diff
@@ -1,165 +1,411 @@
...
+		err := CreateAllDataDirectories([]*Connection{}, segDataDirMap, sourceMasterDataDir)
+		if !xerrors.Is(err, expected) {
+			t.Errorf("wanted: %#v got: %#v", expected, err)
+		}
 	})
+}
+
+func TestCreateSegmentDataDirectories(t *testing.T) {
+	ctrl := gomock.NewController(GinkgoT())
+	defer ctrl.Finish()
```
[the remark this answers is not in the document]
kalensk: done

hub/services/execute_init_target_cluster_sub_step_test.go
```diff
@@ -308,16 +279,13 @@ func TestRunInitsystemForTargetCluster(t *testing.T) {
 			})
 
 		var buf bytes.Buffer
-		err := RunInitsystemForTargetCluster(mockStream, &buf, targetBin + "/", gpinitsystemConfigPath)
+		err := RunInitsystemForTargetCluster(mockStream, &buf, targetBin+"/", gpinitsystemConfigPath)
```
kalensk: from running go fmt

hub/services/execute_init_target_cluster_sub_step_test.go
```diff
@@ -356,40 +321,29 @@ func TestRunInitsystemForTargetCluster(t *testing.T) {
...
-				}
-				if err != nil {
-					t.Fatalf("couldn't get master segment prefix: %v", err)
-				}
+			if err != nil {
+				t.Fatalf("couldn't get master segment prefix: %v", err)
 			}
-		})
+		}
+	})
```
kalensk: from running go fmt and removing unneeded:
```
ctrl := gomock.NewController(GinkgoT())
defer ctrl.Finish()
```

hub/services/execute_init_target_cluster_sub_step_test.go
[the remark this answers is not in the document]
jchampio: This one is still here, though 😁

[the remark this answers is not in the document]
kalensk: Wack-a-mole. Got most of them tho.

PR #145, opened 2019-10-18
Migrate some execute tests to go style

kalensk:
This ports the following tests Go style from ginkgo:
- execute copy master
- check disk space
- check object count
- upgrade master
- execute_test

Note: This is an initial pass of simply porting the existing tests. There _may_ be some room for some further refactors or organization which is outside the scope of this current PR.

Namely, this does not bring up a real Agent Server and grpc Server which the services_suite_test did.

[test pipeline](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/kalen_port_copy_master_tests)

### DIFF
--- a/hub/services/check_disk_space_test.go
+++ b/hub/services/check_disk_space_test.go
@@ -1,67 +1,99 @@
 package services_test
 
 import (
-	"github.com/greenplum-db/gpupgrade/idl"
-
-	"github.com/greenplum-db/gpupgrade/hub/services"
+	"strings"
+	"testing"
 
 	"github.com/golang/mock/gomock"
-
-	. "github.com/onsi/ginkgo"
-	. "github.com/onsi/gomega"
+	"github.com/greenplum-db/gp-common-go-libs/testhelper"
 	"github.com/pkg/errors"
+
+	"github.com/greenplum-db/gpupgrade/hub/services"
+	"github.com/greenplum-db/gpupgrade/idl"
+	"github.com/greenplum-db/gpupgrade/idl/mock_idl"
 )
 
-var _ = Describe("object count tests", func() {
-	Describe("GetDiskUsageFromSegmentHosts", func() {
-		It("returns err msg when unable to call CheckDiskSpaceOnAgents on segment host", func() {
-			var clients []services.ClientAndHostname
+func TestCheckDiskSpace(t *testing.T) {
+	testhelper.SetupTestLogger() // initialize gplog
 
-			client.EXPECT().CheckDiskSpaceOnAgents(
-				gomock.Any(),
-				&idl.CheckDiskSpaceRequestToAgent{},
-			).Return(&idl.CheckDiskSpaceReplyFromAgent{}, errors.New("couldn't connect to hub"))
-			clients = append(clients, services.ClientAndHostname{Client: client, Hostname: "doesnotexist"})
+	t.Run("returns err msg when unable to call CheckDiskSpaceOnAgents on segment host", func(t *testing.T) {
+		ctrl := gomock.NewController(t)
+		defer ctrl.Finish()
 
-			messages := services.GetDiskSpaceFromSegmentHosts(clients)
-			Expect(len(messages)).To(Equal(1))
-			Expect(messages[0]).To(ContainSubstring("Could not get disk usage from: "))
-		})
+		var clients []services.ClientAndHostname
 
-		It("lists filesystems above usage threshold", func() {
-			var clients []services.ClientAndHostname
+		client := mock_idl.NewMockAgentClient(ctrl)
+		client.EXPECT().CheckDiskSpaceOnAgents(
+			gomock.Any(),
+			&idl.CheckDiskSpaceRequestToAgent{},
+		).Return(&idl.CheckDiskSpaceReplyFromAgent{}, errors.New("couldn't connect to hub"))
+		clients = append(clients, services.ClientAndHostname{Client: client, Hostname: "doesnotexist"})
 
-			var expectedFilesystemsUsage []*idl.FileSysUsage
-			expectedFilesystemsUsage = append(expectedFilesystemsUsage, &idl.FileSysUsage{Filesystem: "first filesystem", Usage: 90.4})
-			expectedFilesystemsUsage = append(expectedFilesystemsUsage, &idl.FileSysUsage{Filesystem: "/second/filesystem", Usage: 24.2})
+		messages := services.GetDiskSpaceFromSegmentHosts(clients)
+		if len(messages) != 1 {
+			t.Fatalf("want exactly one message; returned %#v", messages)
+		}
 
-			client.EXPECT().CheckDiskSpaceOnAgents(
-				gomock.Any(),
-				&idl.CheckDiskSpaceRequestToAgent{},
-			).Return(&idl.CheckDiskSpaceReplyFromAgent{ListOfFileSysUsage: expectedFilesystemsUsage}, nil)
-			clients = append(clients, services.ClientAndHostname{Client: client, Hostname: "doesnotexist"})
+		expected := "Could not get disk usage from: "
+		if !strings.Contains(messages[0], expected) {
+			t.Errorf("returned %#v want prefix %#v", messages[0], expected)
+		}
 
-			messages := services.GetDiskSpaceFromSegmentHosts(clients)
-			Expect(len(messages)).To(Equal(1))
-			Expect(messages[0]).To(ContainSubstring("diskspace check - doesnotexist - WARNING first filesystem 90.4 use"))
-		})
-
-		It("lists hosts for which all filesystems are below usage threshold", func() {
-			var clients []services.ClientAndHostname
-
-			var expectedFilesystemsUsage []*idl.FileSysUsage
-			expectedFilesystemsUsage = append(expectedFilesystemsUsage, &idl.FileSysUsage{Filesystem: "first filesystem", Usage: 70.4})
-			expectedFilesystemsUsage = append(expectedFilesystemsUsage, &idl.FileSysUsage{Filesystem: "/second/filesystem", Usage: 24.2})
-
-			client.EXPECT().CheckDiskSpaceOnAgents(
-				gomock.Any(),
-				&idl.CheckDiskSpaceRequestToAgent{},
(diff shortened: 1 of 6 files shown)


hub/services/execute_copy_master_sub_step_test.go
```diff
@@ -1,82 +1,126 @@
-package services_test
+package services
```
marksliva: This file should be renamed to match the file it is testing: `execute_copy_master_sub_step_service_test.go`

marksliva: Oh I see what happened. It looks like this file should replace the existing one.

hub/services/execute_test.go
```diff
@@ -122,15 +127,15 @@ var _ = Describe("ConvertMaster", func() {
...
+	teardown := func() {
 		execCommand = exec.Command
-	})
+	}
 
-	It("streams stdout and stderr to the client", func() {
-		ctrl := gomock.NewController(GinkgoT())
-		defer ctrl.Finish()
+	t.Run("streams stdout and stderr to the client", func(t *testing.T) {
+		setup()
```
jchampio: I think that the per-subtest `setup` and `teardown` are no longer particularly useful.

hub/services/execute_copy_master_sub_step_service.go
```diff
@@ -71,23 +66,23 @@ func CopyMasterDirectoryToSegmentDirectories(agentConns []*Connection, target *u
 	errMsg := "Error copying master data directory to segment data directories"
 	wg := sync.WaitGroup{}
 	errChan := make(chan error, len(agentConns))
-	for _, agentConn := range agentConns {
+
+	for _, conn := range agentConns {
 		wg.Add(1)
-		go func(c *Connection) {
+
+		go func(conn *Connection) {
```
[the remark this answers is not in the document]
jchampio: [It's one way to capture the range variable.](https://github.com/golang/go/wiki/CommonMistakes#using-goroutines-on-loop-iterator-variables) Other options include
```go
    conn := conn // capture range variable
```
or
```go
for i := range agentConns {
    conn := agentConns[i]
```

hub/services/execute_copy_master_sub_step_test.go
kalensk: > Oh I see what happened. It looks like this file should replace the existing one.

Yup! `execute_copy_master_sub_step_service_test.go` was the go-style version which got removed and renamed to just `execute_copy_master_sub_step_test.go` once all the other tests got ported over. Thus, there is just one file now of just go-tests and the name matches like the other files.

For another PR I would like to rename all the execute_* files to just their step name.

hub/services/execute_copy_master_sub_step_service.go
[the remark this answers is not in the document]
kalensk: > There is no need to pass `conn` to this function.

Right, but for this PR to avoid scoop creep I just want to port over the test and not change any functionality. If you would like me to take out the rename I can.

hub/services/check_disk_space_test.go
```diff
@@ -1,67 +1,76 @@
...
 import (
-	"github.com/greenplum-db/gpupgrade/idl"
-
-	"github.com/greenplum-db/gpupgrade/hub/services"
+	"testing"
 
 	"github.com/golang/mock/gomock"
+	"github.com/pkg/errors"
+
+	"github.com/greenplum-db/gp-common-go-libs/testhelper"
```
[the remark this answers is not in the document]
kalensk: Ah! I was grouping it with the greenplum-db packages. I can fix those.

hub/services/execute_copy_master_sub_step_service.go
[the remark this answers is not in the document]
marksliva: @jchampio good point! We can leave it as is.

hub/services/check_disk_space_test.go
```diff
@@ -1,67 +1,76 @@
-package services_test
+package services
```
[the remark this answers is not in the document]
kalensk: Don't think so. Only differences seem to be needing to qualify `services.ClientAndHostname` and `services.GetDiskSpaceFromSegmentHosts(clients)`. I can make the update

I am assuming we have a preference for `services_test` to avoid peeking into the internals unless otherwise needed (ie: testing exec.Command). And that we are ok with having a mixture of test files where some are in `services` and others in `services_test`?

hub/services/check_disk_space_test.go
```diff
@@ -1,67 +1,76 @@
...
+		client := mock_idl.NewMockAgentClient(ctrl)
+		client.EXPECT().CheckDiskSpaceOnAgents(
+			gomock.Any(),
+			&idl.CheckDiskSpaceRequestToAgent{},
+		).Return(&idl.CheckDiskSpaceReplyFromAgent{}, errors.New("couldn't connect to hub"))
+		clients = append(clients, ClientAndHostname{Client: client, Hostname: "doesnotexist"})
+
+		messages := GetDiskSpaceFromSegmentHosts(clients)
+		g.Expect(len(messages)).To(Equal(1))
+		g.Expect(messages[0]).To(ContainSubstring("Could not get disk usage from: "))
```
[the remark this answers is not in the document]
kalensk: > Totally fine to leave this as-is for this PR, if you're trying to limit the changes done at once

Yup. This PR is simply to port over the tests. However, maybe for simple changes like that I can fixup for this PR. I just want to be careful to not let the scope or effort blow-up for this PR as I am sure some of the tests or implementation need work.

hub/services/execute_copy_master_sub_step_service.go
```diff
@@ -46,12 +46,7 @@ func (h *Hub) CopyMasterDataDir(stream messageSender, log io.Writer) error {
 		}
 	}
```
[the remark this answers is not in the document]
kalensk: No idea! (Squirrels in my computer). Will fix it.

hub/services/execute_test.go
kalensk: Would you inline setup and teardown and repeat it for each test? Or keep `pair` as "globals" at the top, and simply reset `execCommand` in each test? Or ...?

jchampio: > Or keep pair as "globals" at the top, and simply reset execCommand in each test?

Yeah. `pair` should be constant, and `SetupTestLogger()` can be done once. As for `execCommand`, https://github.com/greenplum-db/gpupgrade/pull/143 is going to make sure that it's always set to `nil` when the test executable starts [using an `init` function](https://github.com/greenplum-db/gpupgrade/pull/143/files#diff-fd7d8b43d91d542ef4e4e7a4969aeeefR6).

As for `teardown`, I think we don't really want `execCommand` to be reset to the "real" `exec.Command` at any point during the tests.

kalensk:
Thanks for the comments. I believe they should all be addressed in the latest fixups.

hub/services/check_disk_space_test.go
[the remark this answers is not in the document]
jchampio: > I am assuming we have a preference for services_test to avoid peeking into the internals unless otherwise needed (ie: testing exec.Command).

Correct. And if qualifying with `services.` gets unwieldy, dot-importing `services` directly is fine IMO.

> And that we are ok with having a mixture of test files where some are in services and others in services_test?

If your test needs to test internals, you'll need `package services`. If not, prefer `services_test`. It's okay to have both as long as they're used appropriately.

hub/services/check_object_count_test.go
```diff
@@ -1,33 +1,39 @@
...
+	fakeResults = sqlmock.NewRows([]string{"count"}).
+		AddRow([]driver.Value{int32(3)}...)
+	sqlMock.ExpectQuery(".*c.relstorage NOT IN.*").
+		WillReturnRows(fakeResults)
+
+	aocount, heapcount, err := GetCountsForDb(dbConn)
+	g.Expect(err).ToNot(HaveOccurred())
+	g.Expect(aocount).To(Equal(int32(2)))
+	g.Expect(heapcount).To(Equal(int32(3)))
+}
```
jchampio: Your editor is eating end-of-file newlines.

hub/services/execute_copy_master_sub_step_test.go
```diff
@@ -1,82 +1,127 @@
...
 )
 
-var _ = Describe("ExecuteCopyMasterSubStep", func() {
-	var (
-		mockOutput   *cluster.RemoteOutput
-		testExecutor *testhelper.TestExecutor
+func TestCopyMaster(t *testing.T) {
+	g := NewGomegaWithT(t)
+	ctrl := gomock.NewController(GinkgoT())
+	defer ctrl.Finish()
```
jchampio: The `ctrl.Finish()` refactor needs to be done on all subtests throughout this PR.

hub/services/execute_upgrade_master_sub_step_test.go
```diff
@@ -77,15 +83,15 @@ var _ = Describe("ConvertMaster", func() {
 				},
 			},
 		}
-	})
+	}
 
-	AfterEach(func() {
+	teardown := func() {
```
jchampio: Same `setup`/`teardown` comment here too.

hub/services/check_disk_space_test.go
```diff
@@ -1,67 +1,76 @@
...
-			clients = append(clients, services.ClientAndHostname{Client: client, Hostname: "doesnotexist"})
-
-			messages := services.GetDiskSpaceFromSegmentHosts(clients)
-			Expect(len(messages)).To(Equal(1))
-			Expect(messages[0]).To(ContainSubstring("diskspace check - doesnotexist - OK"))
-		})
+func TestCheckDiskSpace(t *testing.T) {
+	g := NewGomegaWithT(t)
+	ctrl := gomock.NewController(GinkgoT())
+	defer ctrl.Finish()
```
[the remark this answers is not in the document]
kalensk: Should be fixed in all places in this PR.

hub/services/execute_copy_master_sub_step_test.go
kalensk: Ah. Should be done now.

hub/services/execute_upgrade_master_sub_step_test.go
kalensk: Ah. Should be done now.

PR #148, opened 2019-10-21
CI: use 6X source to test 6X binaries

jchampio:
The install tests are failing because of an incompatibility with the GPDB master branch. While we do need to fix that incompatibility, the fact that we're relying on master at all is a mistake -- we're testing using 6X binaries and should be relying on 6X source.

### DIFF
--- a/ci/pipeline.yml
+++ b/ci/pipeline.yml
@@ -24,11 +24,11 @@ resources:
     uri: {{gpupgrade-git-remote}}
     branch: {{gpupgrade-git-branch}}
 
-- name: gpdb_src
+- name: gpdb6_src
   type: git
   source:
     uri: https://github.com/greenplum-db/gpdb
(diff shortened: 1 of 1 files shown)


kalensk:
1) Do we need stories for the following issues to fix gpupgrade working with master? https://github.com/greenplum-db/gpupgrade/issues/141 https://github.com/greenplum-db/gpupgrade/issues/140

2) Do we need a story to create a testing matrix for gpupgrade? Specifically, with this PR to add back a gpupgrade with master test?

jchampio:
@dkrieger-pvtl https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:dev:fix-6-install

jchampio:
@kalensk Yes and yes, though I think the issues are already tracking them well.

PR #150, opened 2019-10-24
Update init cluster test assertions

kalensk:
Update to follow the convention "got: X want: Y"


jchampio:
You're going to hate me, but... there's no reason to have colons in the output if the tests properly use delimiting specifiers (like `%q` and the `%v` variants). Take a look at the Go standard library tests for some examples.

kalensk:
> You're going to hate me, but... there's no reason to have colons in the output if the tests properly use delimiting specifiers (like `%q` and the `%v` variants). Take a look at the Go standard library tests for some examples.

Easy fix with the power of IDE's, etc. I somehow missed that when looking at the Go standard library tests. I guess we only see what we want to see. 

On the plus side I am converging on following the convention. We are definitely making forward progress!

PR #151, opened 2019-10-29
initialize: check available disk space

jchampio:
See commit messages. The core of the implementation is the last one:
```
initialize: check available disk space

Replace the old disk space checks with disk.CheckUsage(). This
implementation is run by both the hub and the agents during initialize;
it returns a list of any filesystems that do not meet the requirements.
(For now, only data directory roots are checked; tablespaces are left
for later.)

Since we only support copy-mode for now, the default target disk
availability is 60%. This can be overridden via the --disk-free-ratio
option to initialize.

Object-count checks were unused and have been removed.
```

Current output looks like this:
```
You currently do not have enough disk space to run an upgrade.

Hostname   Filesystem  Shortfall  Available  Required
pchampion  /           176.9 GiB  287.5 GiB  464.4 GiB
```


test/checks.bats
```diff
@@ -0,0 +1,82 @@
...
+    [[ $version = *"GNU coreutils"* ]] && echo $path
+    return 0
+}
+
+@test "initialize prints disk space on failure" {
+    stat=$(gnu_stat)
+    [ -n "$stat" ] || skip "GNU stat is required for this test"
+
+    datadir=$(psql postgres -Atc "select datadir from gp_segment_configuration where role='p' and content=-1")
+    space_before=$("$stat" -f -c '%a*%S/1024' "$datadir" | bc)
```
jchampio: Test images don't have `bc`, so we'll need to find a different calculator. Maybe we should just use `bash` directly, though it'll probably be much more verbose....

cli/commands/commands.go
```diff
@@ -367,6 +338,7 @@ If you would like to return the cluster to its original state, run
 	subInit.MarkPersistentFlagRequired("new-bindir")
 	subInit.PersistentFlags().IntVar(&oldPort, "old-port", 0, "master port for old gpdb cluster")
 	subInit.MarkPersistentFlagRequired("old-port")
+	subInit.PersistentFlags().Float32Var(&diskFreeRatio, "disk-free-ratio", 0.60, "percentage of disk space that must be available (from 0.0 - 1.0)")
```
dkrieger-pvtl: In `subInit` above, check that the `diskFreeRatio` is in the range [0.0,1.0] and throw an error if not.

cli/commands/commands.go
```diff
@@ -345,9 +316,9 @@ This step can be reverted.
 			}
 
 			// TODO: how do we rollback here?
-			err = commanders.NewVersionChecker(client).Execute()
+			err = commanders.RunChecks(client, diskFreeRatio)
```
dkrieger-pvtl: We should discuss how the current sequence in this PR, `Initialize` on the hub and then `RunChecks` fits into our overall scheme.  We want to run the following checks: 

- pg_upgrade --check
- GUCs
- extensions
- disk space

The first item(pg_upgrade --check) requires a new target cluster to be initialized, as do presumably the GUCs and extensions checks.  So I was thinking of placing those within `Initialize`.  

There are many possible schemes:

1. Adding all checks to Initialize

1. splitting `Initialize` into two parts(`Initialize_1`, `checks not requiring target cluster`, `Initialize_2`, `checks requiring target cluster`.

1.

cli/commands/commands.go
jchampio: Good catch; will do.

cli/commands/commands.go
jchampio: For this particular check, if you don't have disk space, you certainly don't want to initialize a new cluster. So maybe we need two parts, but for now maybe it's enough to say "disk space has to be checked before initsystem".

jchampio:
> In a sense, having the hub/agents run a GetDiskSpace might be cleaner than a CheckDiskSpace command.

@dkrieger-pvtl We had a (brief) discussion about that when we started implementation, and decided that we wanted the hub and agents to have the final say on whether a check was passed or not, so that the CLI could remain a dumb view into the system. In this case it's just math, so it's probably not a big deal, but that's why we did it.

This architecture does have the slight advantages of inherent parallelism (though the impact of serialization would probably be miniscule) and the fact that we don't have to funnel dozens of segments' worth of data back through the hub in the success case.

Is there a clear advantage to switching? If not I'd definitely say keep it, though I'm biased. 😁

jmcatamney:
> We had a (brief) discussion about that when we started implementation, and decided that we wanted the hub and agents to have the final say on whether a check was passed or not, so that the CLI could remain a dumb view into the system

Agreed.  Keeping in mind that theoretically we'll want to support more frontends than just our CLI at some point (e.g. GPCC), the more functionality we can keep in the hub and agents, the better.

test/checks.bats
jchampio: Fixed in latest push.

cli/commands/commands.go
jchampio: Fixed in latest push.

dkrieger-pvtl:
> Is there a clear advantage to switching? If not I'd definitely say keep it as-is, though I'm biased. 😁

Right now, I don't think so.  It might be nice to show the final disk usage after `execute` or `finalize` to encourage the DBA to clean up the old cluster from the system, but we can refactor that later.

dkrieger-pvtl:
AAAAAAARGH...why is there a "Close and Comment" button right next to the "Comment" button in the GitHub UI!????!??

utils/disk/disk.go
```diff
@@ -0,0 +1,118 @@
...
+	"golang.org/x/xerrors"
+
+	sigar "github.com/cloudfoundry/gosigar"
+
+	"github.com/greenplum-db/gpupgrade/idl"
+)
+
+// SpaceFailures is a convenience alias for a map of unique filesystem
+// identifiers to disk usage descriptors. One entry is created for each
+// filesystem that doesn't have enough available disk.
```
dkrieger-pvtl: I find the first sentence cryptic.  I prefer:
```
// SpaceFailures maps a unique description of a filesystem to its disk usage.
// One entry is created for each filesystem that doesn't have enough available disk.
```

utils/disk/disk.go
```diff
@@ -0,0 +1,118 @@
...
+	failures := make(SpaceFailures)
+
+	// Find the device ID for every filesystem. We'll use these to map data
+	// directories to filesystems later.
+	fs, err := d.Filesystems()
+	if err != nil {
+		return nil, xerrors.Errorf("enumerating filesystems: %w", err)
+	}
+
+	fsByID := make(map[uint64]string)
```
dkrieger-pvtl: `fsByID` appears to exist to map the master data directory to its mount point, if there is insufficient disk space on it.  This took a bit for me to understand.  Additionally, this is only used if there is insufficient space.  Perhaps we can lazily populate it only if we have insufficient space below in this function?

utils/disk/disk.go
```diff
@@ -0,0 +1,118 @@
...
+		}
+
+		// FIXME it looks like UsePercent returns 0.0 for low usage values...
+		usedRatio := usage.UsePercent() / 100.0
+		availRatio := 1.0 - usedRatio
+
+		if availRatio < float64(required) {
+			// Exclude superuser-reserved space from our denominator.
+			usable := usage.Total - usage.Free + usage.Avail
+			required := uint64(float64(required) * float64(usable))
```
dkrieger-pvtl: using `required` both as disk space in bytes and as a ratio in the function signature is confusing.  Maybe use `usableBytes` and `requiredBytes` instead?

utils/disk/disk.go
```diff
@@ -0,0 +1,118 @@
...
+			return nil, xerrors.Errorf("getting fs usage for %s: %w", path, err)
+		}
+
+		// FIXME it looks like UsePercent returns 0.0 for low usage values...
+		usedRatio := usage.UsePercent() / 100.0
+		availRatio := 1.0 - usedRatio
+
+		if availRatio < float64(required) {
+			// Exclude superuser-reserved space from our denominator.
+			usable := usage.Total - usage.Free + usage.Avail
```
dkrieger-pvtl: This line and the one above it took me a while to figure out.  Maybe clarify:
```
// This excludes superuser-reserved space
usableBytes := usage.Used + usage.Avail
```

utils/disk/disk.go
```diff
@@ -0,0 +1,118 @@
...
+			required := uint64(float64(required) * float64(usable))
+
+			// Get the filesystem that this path belongs to.
+			stat, err := d.Stat(path)
+			if err != nil {
+				return nil, xerrors.Errorf("stat'ing %s: %w", path, err)
+			}
+
+			fs, ok := fsByID[uint64(stat.Dev)]
```
dkrieger-pvtl: This `fs` shadows the `fs` above that is equal to a filesystem list.

cli/commanders/checks.go
```diff
@@ -0,0 +1,72 @@
...
+type DiskSpaceError struct {
+	Failed disk.SpaceFailures
+}
+
+func (d DiskSpaceError) Error() string {
+	var b strings.Builder
+	b.WriteString("You currently do not have enough disk space to run an upgrade.\n\n")
+
+	b.WriteString("Expected Space Available:\n")
+	for host, disk := range d.Failed {
```
dkrieger-pvtl: `host` and `disk` are misleading as physical referents.  It should be `hostFS, diskSpace` or something similar.  Most other places in the code, we use `host` to mean an actual host or hostname.  Same with the usage just below.

hub/services/check_disk_space_test.go
```diff
@@ -1,99 +1,241 @@
...
-	"github.com/greenplum-db/gpupgrade/hub/services"
 	"github.com/greenplum-db/gpupgrade/idl"
 	"github.com/greenplum-db/gpupgrade/idl/mock_idl"
+	"github.com/greenplum-db/gpupgrade/utils"
+	"github.com/greenplum-db/gpupgrade/utils/disk"
 )
 
 func TestCheckDiskSpace(t *testing.T) {
-	testhelper.SetupTestLogger() // initialize gplog
+	var d disk50
```
dkrieger-pvtl: I thought this was a 50GB disk, not one that was 50% used.   Maybe call this `disk50Used`.

hub/services/check_disk_space_test.go
```diff
@@ -1,99 +1,241 @@
...
+	"golang.org/x/xerrors"
 
-	"github.com/greenplum-db/gpupgrade/hub/services"
 	"github.com/greenplum-db/gpupgrade/idl"
 	"github.com/greenplum-db/gpupgrade/idl/mock_idl"
+	"github.com/greenplum-db/gpupgrade/utils"
+	"github.com/greenplum-db/gpupgrade/utils/disk"
 )
 
 func TestCheckDiskSpace(t *testing.T) {
```
dkrieger-pvtl: add a simple positive case, say where the requirement is 0.25.

cli/commanders/checks.go
```diff
@@ -0,0 +1,72 @@
...
+
+type DiskSpaceError struct {
+	Failed disk.SpaceFailures
+}
+
+func (d DiskSpaceError) Error() string {
+	var b strings.Builder
+	b.WriteString("You currently do not have enough disk space to run an upgrade.\n\n")
+
+	b.WriteString("Expected Space Available:\n")
```
dkrieger-pvtl: If there are many Failed checks, there will be a long list of Expected followed by a long list of Actual.  It's probably clearer to the DBA to show the actual/expected values adjacent to each other in the list.

test/checks.bats
```diff
@@ -0,0 +1,103 @@
...
+total_disk_space() {
+    local path=$1
+
+    local bavail bsize bfree btotal
+    read -r bsize bavail bfree btotal <<< $($STAT -f -c '%S %a %f %b' "$path")
+
+    echo $(( ( $btotal - $bfree + $bavail ) * $bsize / 1024 ))
+}
+
+@test "initialize prints disk space on failure" {
```
dkrieger-pvtl: add a positive test too; that would have a disk-free-ratio of 0.0.

kalensk:
`idl/cli_to_hub.proto` still has references to `CheckObjectCount` and `CheckDiskSpace`. I believe some of these are no longer used. Can we take one more closer look and remove any unused references such as `CheckObjectCountRequest` and `CheckObjectCountReply`, etc. ?

kalensk:
**1)** nit: With the removal of the check sub command consider updating this log line to longer have a reference to it.
```
cmd/gpupgrade/main.go
33:		log.Fatal("Please specify one command of: check, config, initialize, prepare, status, upgrade, or version")

```

**2)** If so does `args.bats` need to be updated?
```
9:    [[ "$output" = *"Please specify one command of: check, config, initialize, prepare, status, upgrade, or version"* ]]
```

**3)** The comment at the top of `commands.go` references `gpupgrade check` as an example. Might be nice to update it to something else.

dkrieger-pvtl:
Overall, the design and implementation looks good.  But there are quite a few nits and implementation details that I would like to see changed.

utils/disk/disk.go
jchampio: Sounds good.

utils/disk/disk.go
jchampio: >  Something like(with some perf caching?):

The perf caching is what the current function is doing 😁 I'm not really a fan of the proposed change.

utils/disk/disk.go
jchampio: They're not bytes, but I agree it's confusing and should change.

utils/disk/disk.go
jchampio: That doesn't do the same thing.

hub/services/check_disk_space.go
```diff
@@ -1,70 +1,112 @@
...
+	"github.com/greenplum-db/gpupgrade/utils"
+	"github.com/greenplum-db/gpupgrade/utils/disk"
 )
 
-type ClientAndHostname struct {
-	Client   idl.AgentClient
-	Hostname string
-}
+func (h *Hub) CheckDiskSpace(ctx context.Context, in *idl.CheckDiskSpaceRequest) (*idl.CheckDiskSpaceReply, error) {
+	reply := new(idl.CheckDiskSpaceReply)
```
kalensk: What's the difference between `new(idl.CheckDiskSpaceReply)` and `&idl.CheckDiskSpaceReply{}`? [Effective Go](https://golang.org/doc/effective_go.html#allocation_new) seems to indicate there isn't any:

>As a limiting case, if a composite literal contains no fields at all, it creates a zero value for the type. The expressions new(File) and &File{} are equivalent.

utils/disk/disk.go
dkrieger-pvtl: My main objection to the current implementation is that it's hard for me to differentiate the disk-space logic from the fs-lookup logic.  A function that returns the `fsByID[uint64(stat.Dev)]` mapping that is called before the segment loop might be clean enough.

What's your objection to separating out the disk-space logic from the fs-name logic?

cli/commanders/checks.go
jchampio: `fs` and `space` should be good, though I'm not sure why `disk` is misleading given the context and usage.

cli/commanders/checks.go
jchampio: Yep, Jamie and I have been working on that.

test/checks.bats
jchampio: All successful initializations are positive tests.

cli/commanders/checks.go
dkrieger-pvtl: Mainly I think `host` is a bad name given its idiomatic usage in our code and generally to mean a hostname.  My objection to `disk` might be idiosyncratic, but it's not representing a disk but rather "diskInfo".

hub/services/check_disk_space.go
```diff
@@ -1,70 +1,112 @@
...
+				return
+			}
+
+			req := &idl.CheckSegmentDiskSpaceRequest{
+				Request: in,
+			}
+			for _, s := range segments {
+				req.Datadirs = append(req.Datadirs, s.DataDir)
+			}
+
```
kalensk: nitpick: I kind of wish creating req shows that Datadirs is also available to use. That is, currently one has to "know" that when you create a CheckSegmentDiskSpaceRequest{} then Datadirs is available, even though it is not obviously called out anywhere when creating the request until you look at the implementation of CheckSegmentDiskSpaceRequest.

Its more verbose, but part of me kind of likes creating my variables locally, and then building up the Request.

utils/disk/disk.go
jchampio: > it's hard for me to differentiate the disk-space logic from the fs-lookup logic.

I'd argue that this is the hard (and useful) part of the implementation; the disk space logic is just math.

> What's your objection to separating out the disk-space logic from the fs-name logic?

I want to make sure we hit the expensive statistics implementations (like filesystem discovery) exactly once per check.

hub/services/check_disk_space_test.go
jchampio: Sounds good.

utils/disk/disk.go
jchampio: Good catch.

test/checks.bats
```diff
@@ -0,0 +1,103 @@
...
+teardown() {
+    skip_if_no_gpdb
+
+    gpupgrade kill-services
+    rm -r "$STATE_DIR"
+}
+
+# require_gnu_stat tries to find a GNU stat program. If one is found, it will be
+# assigned to the STAT global variable; otherwise the current test is skipped.
+require_gnu_stat() {
```
kalensk: I like this.

hub/services/check_disk_space.go
jchampio: Correct! I like `new(Thing)` when I want a pointer to a zero-value struct, and `&Thing{Stuff: 1}` when I want to fill up the struct with values.

test/checks.bats
```diff
@@ -70,39 +104,26 @@ total_disk_space() {
...
-        min="$space_before"
-        max="$space_after"
-    else
-        min="$space_after"
-        max="$space_before"
-    fi
-    allowable_threshold=$((10 * 1024))
-    (( min -= $allowable_threshold ))
-    (( max += $allowable_threshold ))
+    $(are_equivalent_within_tolerance "total space" $required_bytes $total_space 4 1)
```
jchampio: Why the use of subshells (here and below)?

test/checks.bats
```diff
@@ -0,0 +1,131 @@
...
+    required="${BASH_REMATCH[5]}"
+    required_unit="${BASH_REMATCH[6]}"
+    required_bytes=$(convert_to_kib $required $required_unit)
+
+    total_space=$(total_disk_space "$datadir")
+    if ! are_equivalent_within_tolerance $required_bytes $total_space 0.001; then
+        fail "the required bytes ($required_bytes) are not within 0.1% of the total disk space ($total_space)"
+    fi
+
+    # Due to rounding errors throughout the chain of calculations we allow a delta of 1.
```
jchampio: Unfortunately this still isn't good enough; we need to handle the difference between e.g. `121.4 GiB` and `121.5 GiB`.

jchampio: Reverted to use the tolerance method for this comparison too.

jchampio:
> Can we take one more closer look and remove any unused references such as CheckObjectCountRequest and CheckObjectCountReply, etc. ?

Done.

> With the removal of the check sub command consider updating this log line to longer have a reference to it.

I'm purposefully ignoring that line for now; I think we want to remove it entirely and just print usage, but we should check with Bharath.

jchampio:
Rebased onto latest master. The last two `squash!` commits contain the changes since the last major round of review.

dkrieger-pvtl:
nit: speaking of the perils of dead code, the file `test/out-of-order.bats` has no running tests, but you should still add `--disk-free-ratio 0` to its use of `gpupgrade initialize` for posterity.

utils/disk/disk.go
dkrieger-pvtl: After re-reviewing the code, the combined function(both fsid to pathname lookup and diskFree logic) seems fine.

jchampio:
@dkrieger-pvtl Fixed your nit, thanks for the catch!

jchampio:
Rebased onto latest master. Our third (fourth?) attempt at validating the shortfall UI in the end-to-end tests was still flaking due to bigger rounding errors in the CI environment than we expected. So we stopped doing that test in BATS and moved it to a unit test. See the final `squash!` commit.

PR #152, opened 2019-10-31
Miscellaneous fixes

jchampio:
Prefactoring and preparation to fix the port collision problem with the latest 6X and master:
- Remove `gpupgrade status` and all associated code, since it was pinning down the substep definitions, it doesn't currently work anyway, and we want the new implementation to be completely different.
- Break the import cycle between `services` and `testutils`, to make it possible to...
- ...write a test for our ignoring of substep failures during execute.


hub/services/execute.go
```diff
@@ -65,7 +65,10 @@ func (h *Hub) Execute(request *idl.ExecuteRequest, stream idl.CliToHub_ExecuteSe
 		return err
 	}
 
-	err = h.ExecuteUpgradePrimariesSubStep(stream)
+	err = h.ExecuteSubStep(executeStream, upgradestatus.UPGRADE_PRIMARIES,
+		func(_ messageSender, _ io.Writer) error {
+			return h.ConvertPrimaries()
```
marksliva: Why is this called `ConvertPrimaries`? Could we call this `UpgradePrimaries` instead?

kalensk:
nit: `cmd/gpupgrade/main.go` references status command: `log.Fatal("Please specify one command of: check, config, initialize, prepare, status, upgrade, or version")`. I am ok with leaving it which assumes when we add it back it the new command will be named status.

hub/services/execute.go
jchampio: That's what it was called before, that's all.

PR #153, opened 2019-11-06
Gopkg.lock: update with latest dep ensure

jchampio:
This kept popping up during a PR rebase.


kalensk:
> Seems to fix whatever funkiness is going on when I checkout master, blow away the vendor directory, and run make:

Errr, I seem to have spoken too soon. That same issue appears on this branch as well. But in any case the change seems fine.

PR #154, opened 2019-11-06
execute: shut down source before initializing target

jchampio:
6X is now failing due to port collisions between the source and target clusters. We need to improve the port logic, but further discussions indicate that we're not really sure which direction we want to go. Since there's a very real possibility that users and/or developers may want to bring up the new cluster on exactly the same ports as the old one, make that possible by stopping the source before init'ing the target.

To do this, the init-target-cluster and shutdown-clusters substeps become

1. create-target-config
2. shutdown-source
3. init-target-cluster
4. shutdown-target

and the rest of this patch is the fallout from that separation, much of it having to with the tightly coupled code in the init-target-cluster implementation.

(`execute_init_target_cluster_sub_step.go` is a mess. I know it's ugly, but I really wanted to keep changes to that file to a minimum for the purposes of this fix.)


jchampio:
Note to reviewers: hiding whitespace changes in the diff really helps clarify things IMO.

PR #155, opened 2019-11-09
Add pg_upgrade --check substep to initialize step

jmcatamney:
This PR adds a substep under the `initialize` step to run `pg_upgrade --check` against the master and primaries and log the results in the state directory.
    
Because `pg_upgrade --check` requires that the target cluster be initialized for the check to work, this PR moves all cluster creation logic from `execute` to `initialize` as well.

See individual commit messages for more details.


kalensk:
If the Bats test are getting overly slow one possibility might be adding a hidden flag (?)  `--disable-checks` or `--disable-init-cluster` which would not init the target cluster. This of course would only be used for tests that do not depend on initializing the target cluster or on execute or finalize. However, this would not work if that's needed.

test/execute.bats
```diff
@@ -39,6 +39,8 @@ upgrade_datadir() {
     echo "$dir/$base"
 }
 
+# TODO: we actually need to run execute here to get this test to work
+#   what are we actually testing here?
```
jchampio: What is this comment about?

agent/services/execute_upgrade_primaries_sub_step.go
```diff
@@ -92,13 +98,53 @@ func UpgradeSegments(sourceBinDir string, targetBinDir string, segments []Segmen
...
+				defer func() {
+					if closeErr := stdout.Close(); closeErr != nil {
+						err = multierror.Append(err,
+							xerrors.Errorf("failed to close pg_upgrade_check_stdout log: %w", closeErr))
+					}
+					if closeErr := stderr.Close(); closeErr != nil {
+						err = multierror.Append(err,
+							xerrors.Errorf("failed to close pg_upgrade_check_stderr log: %w", closeErr))
+					}
+				}()
```
jchampio: What's the goal here? These changes to `err` will be lost.

agent/services/execute_upgrade_primaries_sub_step.go
```diff
@@ -92,13 +98,53 @@ func UpgradeSegments(sourceBinDir string, targetBinDir string, segments []Segmen
 			cmd.Env = append(cmd.Env, fmt.Sprintf("LD_LIBRARY_PATH=%s", path))
 		}
 
+		myContent := segment.Content
```
jchampio: nit: let's ditch filler words like `my` in variable names. If we want to capture a range variable for use in a goroutine, let's just do that with the same name we would have used anyway.

agent/services/execute_upgrade_primaries_sub_step.go
```diff
@@ -92,13 +98,53 @@ func UpgradeSegments(sourceBinDir string, targetBinDir string, segments []Segmen
...
 		go func() {
+			defer wg.Done()
+
+			if checkOnly {
+				//TODO: put this in the "official" segment dir location
+				stdout, err := utils.System.OpenFile(
+					filepath.Join(stateDir, fmt.Sprintf("pg_upgrade_check_stdout_seg_%d.log", myContent)),
+					os.O_WRONLY|os.O_CREATE,
+					0600,
+				)
```
jchampio: This seems like a lot of work to replicate what we already have working for execute in general -- stdout/err logging to file and gRPC.

dkrieger-pvtl:
> If the Bats test are getting overly slow one possibility might be adding a hidden flag (?) `--disable-checks` or `--disable-init-cluster` which would not init the target cluster. This of course would only be used for tests that do not depend on initializing the target cluster or on execute or finalize. However, this would not work if that's needed.

This PR adds a flag `only-run-pre-init` for `gpupgrade initialize` that does not create the target cluster, and most of the bats tests run with this flag.  We only leave it off for bats tests that require the target cluster to be created.  Note, with that flag set, we also run the diskCheck.

hub/services/initialize.go
```diff
@@ -16,59 +21,144 @@ import (
...
+	)
 	if err != nil {
 		return err
 	}
+	defer func() {
+		if closeErr := log.Close(); closeErr != nil {
+			err = multierror.Append(err,
+				xerrors.Errorf("failed to close initialize log: %w", closeErr))
+		}
+	}()
```
jchampio: This `err` modification will also be lost.

hub/services/initialize.go
```diff
@@ -16,59 +21,144 @@ import (
...
+	)
+	if err != nil {
+		return err
+	}
+	defer func() {
+		if closeErr := log.Close(); closeErr != nil {
+			err = multierror.Append(err,
+				xerrors.Errorf("failed to close initialize log: %w", closeErr))
+		}
+	}()
```
jchampio: Ditto.

agent/services/execute_upgrade_primaries_sub_step_test.go
```diff
@@ -82,6 +127,7 @@ var _ = Describe("UpgradeSegments", func() {
 				oldPort := fs.Int("old-port", -1, "")
 				newPort := fs.Int("new-port", -1, "")
 				mode := fs.String("mode", "", "")
+				checkOnly := fs.Bool("check", true, "")
```
jchampio: This flag defaults to true, so if it's not passed, the test won't fail.

agent/services/execute_upgrade_primaries_sub_step_test.go
```diff
@@ -93,15 +139,34 @@ var _ = Describe("UpgradeSegments", func() {
...
+		err := UpgradeSegments("/old/bin", "/new/bin", segments, "/tmp", true)
 		Expect(err).NotTo(HaveOccurred())
 	})
 
+	It("when pg_upgrade --check fails it returns an error", func() {
+		execCommand = exectest.NewCommand(FailedMain)
+
+		err := UpgradeSegments("/old/bin", "/new/bin", segments, "/tmp", true)
+		Expect(err).To(HaveOccurred())
+		Expect(err.Error()).To(ContainSubstring("failed to upgrade --check primary on host"))
```
jchampio: nit: seems like `failed to check primary` would be a fine alternative.

cli/commanders/step_initialize.go
```diff
@@ -93,3 +93,16 @@ func IsHubRunning() (bool, error) {
...
+	s := Substep("Running pre-gpinitsystem checks...")
+	defer s.Finish(&err)
+	return err
+}
+func RunPostGpInitSystemChecks() error {
+	var err error
+	s := Substep("Running post-gpinitsystem checks...")
+	defer s.Finish(&err)
+	return err
+}
```
dkrieger-pvtl: we will remove these two added functions, as they aren't called.  (I thought we had removed them before PR.)

hub/services/initialize.go
```diff
@@ -131,7 +221,10 @@ func getAgentPath() (string, error) {
 }
 
 // TODO: use the implementation in RestartAgents() for this function and combine them
-func StartAgents(source *utils.Cluster, target *utils.Cluster, stateDir string) error {
+func (h *Hub) startAgentsSubStep(stream messageSender, log io.Writer, inputs ...string) error {
```
jchampio: This is getting farther away from the `TODO`, no?

hub/services/initialize.go
```diff
@@ -16,59 +21,144 @@ import (
...
 		gplog.Error(err.Error())
 		return err
 	}
 
-	err = StartAgents(h.source, h.target, h.conf.StateDir)
+	err = subStepFunc(initializeStream.stream, initializeStream.log, args...)
 	if err != nil {
 		gplog.Error(err.Error())
 		step.MarkFailed()
-		return err
```
jchampio: Why do we no longer return an error on failure?

hub/services/initialize.go
```diff
@@ -16,59 +21,144 @@ import (
...
 	}
 
-	step.MarkComplete()
 	return nil
 }
 
-func (h *Hub) startAgentsSubStep(stream messageSender) error {
-	gplog.Info("starting %s", upgradestatus.START_AGENTS)
+// TODO: is there a way to pass in a variadic arg but still get type checking?
+func (h *Hub) InitializeSubStep(initializeStream *InitializeStream, subStep string,
```
jchampio: This implementation is identical to `ExecuteSubStep`.

dkrieger-pvtl:
We will also eliminate the varargs in the substep calls in favor of a function that contains that argument, that is:

```
	err = h.ExecuteSubStep(executeStream, upgradestatus.INIT_TARGET_CLUSTER,
		func(stream messageSender, log io.Writer) error {
			return h.CreateTargetCluster(stream, log, targetMasterPort)
		})
```

hub/services/initialize.go
```diff
@@ -16,59 +21,144 @@ import (
...
 		return err
 	}
 
-	step.MarkComplete()
 	return nil
 }
 
-func (h *Hub) startAgentsSubStep(stream messageSender) error {
-	gplog.Info("starting %s", upgradestatus.START_AGENTS)
+// TODO: is there a way to pass in a variadic arg but still get type checking?
```
jchampio: Sure, with runtime type assertions, but the existence of this comment implies that it's probably not what we want to do, right? 😁 The vast majority of the `subStepFunc` "implementations" don't use the args, so it's not the right API. Why not mirror what's been done in `Execute`?

hub/services/shutdown_clusters_sub_step.go
```diff
@@ -0,0 +1,89 @@
...
+
+const (
+	SOURCE = "SOURCE"
+	TARGET = "TARGET"
+)
+
+func (h *Hub) ShutdownClusters(stream messageSender, log io.Writer, inputs ...string) error {
+	var shutdownErr error
+
+	switch inputs[0] {
```
jchampio: Switching on a string input without a `default` case makes me nervous.

hub/services/shutdown_clusters_sub_step.go
```diff
@@ -0,0 +1,89 @@
...
+	case SOURCE:
+		err := StopCluster(stream, log, h.source)
+		if err != nil {
+			shutdownErr = multierror.Append(shutdownErr, errors.Wrap(err, "failed to stop source cluster"))
```
jchampio: Why the use of multierror?

hub/services/shutdown_clusters_sub_step.go
```diff
@@ -0,0 +1,89 @@
...
+	cmd := startStopClusterCmd("bash", "-c",
+		fmt.Sprintf("source %[1]s/../greenplum_path.sh && %[1]s/%[2]s -a -d %[3]s",
+			cluster.BinDir,
+			cmdName,
+			cluster.MasterDataDir(),
+		))
+
+	mux := newMultiplexedStream(stream, log)
+	cmd.Stdout = mux.NewStreamWriter(idl.Chunk_STDOUT)
+	cmd.Stderr = mux.NewStreamWriter(idl.Chunk_STDERR)
```
jchampio: Now is probably the time to fix this duplication.

hub/services/shutdown_clusters_sub_step_test.go
```diff
@@ -19,24 +19,24 @@ import (
...
 	exectest.RegisterMains(
+		StartClusterCmd,
 		StopClusterCmd,
 		IsPostmasterRunningCmd,
 		IsPostmasterRunningCmd_Errors,
 	)
 }
 
-func TestShutdownClusters(t *testing.T) {
+func TestShutdownStartClusters(t *testing.T) {
```
jchampio: nit: `TestShutdownStartClusters` doesn't leave me with a good idea of what's going on 😁

test/pg_upgrade.bats
```diff
@@ -0,0 +1,117 @@
...
+
+  gpupgrade initialize \
+      --old-bindir "$GPHOME/bin" \
+      --new-bindir "$GPHOME/bin" \
+      --old-port "$PGPORT" \
+      --disk-free-ratio=0 3>&-
+
+   [ $(grep -c "Clusters are compatible" "$GPUPGRADE_HOME"/initialize.log) ] || fail "error: expected file $GPUPGRADE_HOME/initialize.log to indicate clusters are compatible"
+
+   [ -e "$GPUPGRADE_HOME"/pg_upgrade_check_stdout_seg_0.log ] || fail "error: expected file $GPUPGRADE_HOME/pg_upgrade_check_stdout_seg_0.log does not exist"
```
jchampio: nit: most of these `fail` messages don't add anything to the BATS default error message. `fail` is just a helper for when you want to add more context or work around the `[[ ]]` BATS bug.

agent/services/execute_upgrade_primaries_sub_step.go
dkrieger-pvtl: This is done on the segments, and I thought we hadn't decided how to proceed with streaming and logging on them.  Hence, the local logging on each segment host.  There are issues both for the log file and streaming:
1). log file...we have each `pg_upgrade --check` running in parallel on the segment, so want to write to separate files to preserve the order of output from each host.  In principle, we could keep the output in memory and only write the results when the check is done.
2). streaming... we'd need to stream from the segment host to the master, and then to the cli process.  

We are deferring these for now.

cli/commanders/checks.go
```diff
@@ -14,11 +14,12 @@ import (
...
-	err := CheckVersion(client)
-	if err != nil {
-		return errors.Wrap(err, "checking version compatibility")
-	}
+func RunPreChecks(client idl.CliToHubClient, ratio float32) error {
+	//TODO: when do we check this?  It requires the source cluster to be up.
+	//err := CheckVersion(client)
+	//if err != nil {
+	//	return errors.Wrap(err, "checking version compatibility")
+	//}
```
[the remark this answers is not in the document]
dkrieger-pvtl: Because we need the source cluster up to run the current version check, and this PR brings the source cluster down and leaves it down.  The CheckVersion would have to be run as part of Initialize() on the Hub.  So we are skipping the version check for now, in favor of future work that would rework the checkVersion to something we actually think we should do.  For instance, we might decide to check the binaries vs the live cluster.

cli/commanders/steps.go
```diff
@@ -57,6 +60,20 @@ func Initialize(client idl.CliToHubClient, oldBinDir, newBinDir string, oldPort
 	return nil
 }
 
+func InitializeCreateCluster(client idl.CliToHubClient) (err error) {
+	stream, err := client.InitializeCreateCluster(context.Background(), &idl.InitializeCreateClusterRequest{})
+	if err != nil {
+		return errors.Wrap(err, "initializing hub2")
+	}
+
+	err = UILoop(stream, false)
```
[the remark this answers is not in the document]
dkrieger-pvtl: Because we allow a flag to `initialize` to only run the first part of initialize, that is, to not create the target cluster.  So initialize now has two parts.

cli/commands/commands.go
```diff
@@ -308,6 +318,8 @@ If you would like to return the cluster to its original state, run
 	subInit.MarkPersistentFlagRequired("new-bindir")
 	subInit.PersistentFlags().IntVar(&oldPort, "old-port", 0, "master port for old gpdb cluster")
 	subInit.MarkPersistentFlagRequired("old-port")
+	subInit.PersistentFlags().BoolVar(&onlyRunPreInit, "only-run-pre-init", false, "only run up to pre-init")
+	subInit.PersistentFlags().MarkHidden("only-run-pre-init")
```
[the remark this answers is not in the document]
dkrieger-pvtl: What's a good name?  `no-target-cluster-init`?  `no-gpinitsystem`?  We don't like this name either, but couldn't think of a good, succinct one.

hub/services/initialize.go
dkrieger-pvtl: Done.

cli/commanders/step_initialize.go
dkrieger-pvtl: Done.

hub/services/init_target_cluster_sub_step.go
```diff
@@ -61,6 +52,15 @@ func (h *Hub) writeConf(sourceDBConn *dbconn.DBConn) error {
...
 
+func (h *Hub) CreateTargetCluster(stream messageSender, log io.Writer, args ...string) error {
+	targetDBConn, err := h.InitTargetCluster(stream, log)
+	if err != nil {
+		return err
+	}
+
+	return ReloadAndCommitCluster(h.target, targetDBConn)
+}
+
```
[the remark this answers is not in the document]
dkrieger-pvtl: Can you please explain this further?  I am not sure what you mean.

cmd/gpupgrade_hub/main.go
```diff
@@ -97,6 +97,8 @@ func main() {
 			cm.AddWritableStep(upgradestatus.SHUTDOWN_SOURCE_CLUSTER, idl.UpgradeSteps_SHUTDOWN_SOURCE_CLUSTER)
 			cm.AddWritableStep(upgradestatus.INIT_TARGET_CLUSTER, idl.UpgradeSteps_INIT_TARGET_CLUSTER)
 			cm.AddWritableStep(upgradestatus.SHUTDOWN_TARGET_CLUSTER, idl.UpgradeSteps_SHUTDOWN_TARGET_CLUSTER)
+			cm.AddWritableStep(upgradestatus.CHECK_MASTER, idl.UpgradeSteps_CHECK_MASTER)
+			cm.AddWritableStep(upgradestatus.CHECK_PRIMARIES, idl.UpgradeSteps_CHECK_PRIMARIES)
```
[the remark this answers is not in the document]
dkrieger-pvtl: Because that's how it is done for the actual upgrade: `upgrader master`, `copy master`, `upgrade primaries`.  It's really a UI question, but I'd prefer to keep `check` and actual upgrade the same for now.

jchampio:
General comments from working on this branch this morning:
- The loss of `--verbose` output to debug gpinitsystem failures is pretty bad.
- `pg_upgrade --check` failures are output in triplicate (probably an uncaptured range variable).

hub/services/execute_upgrade_master_sub_step.go
```diff
@@ -2,19 +2,25 @@ package services
...
+	"github.com/greenplum-db/gpupgrade/utils"
 )
 
 // Allow exec.Command to be mocked out by exectest.NewCommand.
 var execCommand = exec.Command
 
-func (h *Hub) UpgradeMaster(stream messageSender, log io.Writer) error {
+const (
+	MASTER_CHECK   = "MASTER_CHECK"
+	MASTER_UPGRADE = "MASTER_UPGRADE"
```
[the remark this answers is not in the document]
dkrieger-pvtl: done

hub/services/execute_start_target_cluster_sub_step.go
```diff
@@ -1,21 +1,9 @@
...
-			h.target.BinDir,
-			h.target.MasterDataDir(),
-		))
-
-	mux := newMultiplexedStream(stream, log)
-	cmd.Stdout = mux.NewStreamWriter(idl.Chunk_STDOUT)
-	cmd.Stderr = mux.NewStreamWriter(idl.Chunk_STDERR)
-
-	return cmd.Run()
+func (h *Hub) StartTargetCluster(stream messageSender, log io.Writer, args ...string) error {
```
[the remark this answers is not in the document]
dkrieger-pvtl: done

test/execute.bats
dkrieger-pvtl: Unfortunately, I added that comment a month ago, and think it is moot. Removed.

test/pg_upgrade.bats
```diff
@@ -0,0 +1,117 @@
...
+    [ $(grep -c "Checking for indexes on partitioned tables                  fatal" "$GPUPGRADE_HOME"/initialize.log) ] || fail "error: expected file $GPUPGRADE_HOME/initialize.log to contain index-based failure"
+
+    # revert added index
+    gpstart -a
+    $PSQL -d postgres -p $PGPORT -c "DROP TABLE test_pg_upgrade CASCADE;"
+    gpstop -a
+
+}
+
+@test "gpupgrade initialize runs pg_upgrade --check on master and primaries" {
```
jchampio: This test doesn't clean up after itself with `NEW_CLUSTER`. Also, the use of three-space tabs is deeply distressing.

hub/services/initialize.go
dkrieger-pvtl: coding error.  Fixed.

hub/services/shutdown_clusters_sub_step.go
dkrieger-pvtl: fixed.

agent/services/execute_upgrade_primaries_sub_step_test.go
dkrieger-pvtl: We left this as is as we need to check against the runtime error in the code, which contains hostname, etc.

test/pg_upgrade.bats
```diff
@@ -0,0 +1,117 @@
...
+
+    # add in a index on a partition table, which causes pg_upgrade --check to fail
+    $PSQL -d postgres -p $PGPORT -c "CREATE TABLE test_pg_upgrade(a int) DISTRIBUTED BY (a) PARTITION BY RANGE (a)(start (1) end(4) every(1));"
+    $PSQL -d postgres -p $PGPORT -c "CREATE UNIQUE INDEX fomo ON test_pg_upgrade (a);"
+
+    gpupgrade initialize \
+        --old-bindir "$GPHOME/bin" \
+        --new-bindir "$GPHOME/bin" \
+        --old-port "$PGPORT" \
+        --disk-free-ratio=0 3>&-
```
jchampio: We ran into this one while fixing a different bug -- this call to `initialize` should fail, not succeed with a zero exit code. (See the ignored error propagation elsewhere in `InitializeSubStep()`.)

cmd/gpupgrade_hub/main.go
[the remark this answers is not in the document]
jchampio: Keep in mind that's done for a specific reason -- master upgrade must complete before starting segment upgrades because we make use of the upgraded data directory contents. Otherwise we'd run them in parallel. Is there a reason we can't run all checks at once to save on time?

test/pg_upgrade.bats
marksliva: Yes. Now that we started to return an `err` in the one spot that we had missed, it fails as expected. So we prepend it with `run` so that test execution continues, and then check for an exit status of '1' and the logs for the expected error.

hub/services/initialize.go
```diff
@@ -16,59 +20,146 @@ import (
...
+	err = h.InitializeSubStep(initializeStream, upgradestatus.CHECK_MASTER, func(stream messageSender, log io.Writer) error {
+		return h.UpgradeMaster(stream, log, true)
+	})
+	if err != nil {
+		return err
+	}
+	err = h.ConvertPrimaries(true)
+	if err != nil {
+		return err
+	}
```
jchampio: This needs to be associated with a substep.

test/pg_upgrade.bats
marksliva: fixed.

cmd/gpupgrade_hub/main.go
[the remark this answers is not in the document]
dkrieger-pvtl: I see...if we run `pg_upgrade --check` serially on the master and then segments, we notionally take twice as long.  That's a good idea.  Offhand, I cannot think of any drawbacks of this for a production cluster as the master should be on its own host.

hub/services/initialize.go
dkrieger-pvtl: Only trivially, I think.  I don't want to combine RestartAgents and this function in this PR.  Making this a receiver method here, but undoing it later, is almost a no-op.

cli/commanders/checks.go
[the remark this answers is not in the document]
jchampio: Okay. Just make sure @bharath1028 is in the loop on removed "features" so that stories can be filed and prioritized.

hub/services/init_target_cluster_sub_step.go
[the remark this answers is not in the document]
jchampio: On my first pass I misunderstood the `...args` change. I am a little confused as to why this method moved though.

hub/services/initialize.go
marksliva: Done. We also added a test for this.

cli/commanders/checks.go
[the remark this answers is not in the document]
dkrieger-pvtl: Yup, created a story to track it.

test/pg_upgrade.bats
marksliva: ok, we have removed `fail` calls. What is this BATS bug you are referring to?

hub/services/shutdown_clusters_sub_step_test.go
marksliva: fixed. If you can think of an even better name for the file, let us know.

test/pg_upgrade.bats
jchampio: When using Bash-syntax `[[ ]]` tests, for example
```bash
[[ $string = *hello* ]]
```
BATS reports any failures as belonging to the previous line, which is incredibly confusing. Adding
```bash
[[ $string = *hello* ]] | fail "'$string' did not contain 'hello'"
```
works around the issue.

hub/services/shutdown_clusters_sub_step.go
marksliva: fixed

hub/services/shutdown_clusters_sub_step.go
marksliva: Fixed. It looks like this was left over from a prior version of this function.

cmd/gpupgrade_hub/main.go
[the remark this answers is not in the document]
marksliva: Done. I refactored these into a single sub-step so that it runs both checks concurrently, and updated the step output to match our initialize designs.

agent/services/execute_upgrade_primaries_sub_step.go
```diff
@@ -46,15 +48,15 @@ func UpgradePrimaries(sourceBinDir string, targetBinDir string, dataDirPairs []*
 		segments = append(segments, Segment{DataDirPair: dataPair, UpgradeDir: dataPair.NewDataDir})
 	}
 
-	err := UpgradeSegments(sourceBinDir, targetBinDir, segments)
+	err := UpgradeSegments(sourceBinDir, targetBinDir, segments, stateDir, checkOnly)
```
jchampio: I don't want you to fix it in this PR, because Jamie and I are already working on it in ours, but there's a misunderstanding about the agent state directory here that I want to correct.

The working directory for an upgrade is supposed to be contained in `Segment.UpgradeDir`. That should point somewhere inside the `.gpupgrade` directory (e.g. `~/.gpupgrade/pg_upgrade/seg3`) and is supposed to be where all the logs get dropped. This isn't really working today; it looks like the `UpgradeDir` is incorrectly set to the new data directory, which isn't very helpful.

What's more, `pg_upgrade` gives us the `--retain` option to keep all the logs around after a successful run. Setting this is a lot easier (and gives us much more debugging power) than manually creating our own log files for now. Eventually we will _also_ want to stream back to the hub, and capture any errors that happen a level above pg_upgrade, and when that happens we will want a separate log file. But for now `--retain` should be enough.

dkrieger-pvtl:
> General comments from working on this branch this morning:
> 
> * The loss of `--verbose` output to debug gpinitsystem failures is pretty bad.

Thanks for catching this.  Fixed.

> * `pg_upgrade --check` failures are output in triplicate (probably an uncaptured range variable).

Yes, we weren't capturing the content for each segment, but using the last one.  Fixed.

hub/services/initialize.go
```diff
@@ -16,59 +21,144 @@ import (
 	"github.com/pkg/errors"
 )
 
+type InitializeStream struct {
+	stream messageSender
+	log    *os.File
+}
```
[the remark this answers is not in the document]
jchampio: Will fix in a follow-up.

agent/services/execute_upgrade_primaries_sub_step.go
jchampio: Will fix in a follow-up.

hub/services/initialize.go
jchampio: Will fix in a follow-up.

test/pg_upgrade.bats
```diff
@@ -0,0 +1,141 @@
...
+    match="Creating directories\.\.\..*COMPLETE.*
+Starting hub\.\.\..*COMPLETE.*
+Retrieving configs\.\.\..*COMPLETE.*
+Starting agents\.\.\..*COMPLETE.*
+Checking disk space\.\.\..*COMPLETE.*
+Generating new cluster configuration\.\.\..*COMPLETE.*
+Shutting down existing cluster\.\.\..*COMPLETE.*
+Creating new cluster\.\.\..*COMPLETE.*
+Shutting down new cluster\.\.\..*COMPLETE.*
+Running pg_upgrade checks\.\.\..*COMPLETE.*"
```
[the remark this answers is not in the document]
marksliva: We discovered that there was no test coverage when we refactored check_primary and check_master to a single step, so we came up with this test.
We can make a new test just for this UI output. We don't see a noticeable slowdown in `install-tests` yet, but we can refactor this test later if it becomes a problem.

test/pg_upgrade.bats
```diff
@@ -113,21 +120,23 @@ Checking primaries\.\.\..*COMPLETE.*"
...
+    [ -e "$GPUPGRADE_HOME"/pg_upgrade_check_stdout_seg_0.log ]
+    [ -e "$GPUPGRADE_HOME"/pg_upgrade_check_stdout_seg_1.log ]
+    [ -e "$GPUPGRADE_HOME"/pg_upgrade_check_stdout_seg_2.log ]
 
-    [ -e "$GPUPGRADE_HOME"/pg_upgrade_check_stdout_seg_0.log ] || fail "error: expected file $GPUPGRADE_HOME/pg_upgrade_check_stdout_seg_0.log does not exist"
-    [ -e "$GPUPGRADE_HOME"/pg_upgrade_check_stdout_seg_1.log ] || fail "error: expected file $GPUPGRADE_HOME/pg_upgrade_check_stdout_seg_1.log does not exist"
-    [ -e "$GPUPGRADE_HOME"/pg_upgrade_check_stdout_seg_2.log ] || fail "error: expected file $GPUPGRADE_HOME/pg_upgrade_check_stdout_seg_2.log does not exist"
+    [ $(grep -c "Clusters are compatible" "$GPUPGRADE_HOME"/pg_upgrade_check_stdout_seg_0.log) -eq 1 ]
+    [ $(grep -c "Clusters are compatible" "$GPUPGRADE_HOME"/pg_upgrade_check_stdout_seg_1.log) -eq 1 ]
+    [ $(grep -c "Clusters are compatible" "$GPUPGRADE_HOME"/pg_upgrade_check_stdout_seg_2.log) -eq 1 ]
```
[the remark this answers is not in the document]
marksliva: fixed.

test/pg_upgrade.bats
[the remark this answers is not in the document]
marksliva: Good point on this not really fitting into the current test. We moved it into its own test.

cli/commands/commands.go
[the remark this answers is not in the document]
marksliva: fixed.

agent/services/execute_upgrade_primaries_sub_step_test.go
marksliva: Fixed. We check the args directly instead in the false case.

agent/services/execute_upgrade_primaries_sub_step_test.go
marksliva: fixed.

hub/services/initialize.go
marksliva: fixed.

agent/services/execute_upgrade_primaries_sub_step.go
marksliva: Fixed. We added the named return parameter to the function.

hub/services/initialize.go
marksliva: fixed.

cli/commanders/steps.go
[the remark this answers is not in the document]
dkrieger-pvtl: So, the idea is to have all the logic of each step(Initialize, Execute, Finalize) inside the hub, and what we have breaks this pattern by placing the logic(before-gpinitsystem-hub, diskcheck, gpinitsystem) in the CLI.  As we discussed, we'll leave this for a future refactor.

hub/services/initialize.go
dkrieger-pvtl: The concern here was that, by making it a receiver on the Hub, that we'd be inviting use of the Hub as a "collection" of state variables.  We refactored this method into a function.

test/pg_upgrade.bats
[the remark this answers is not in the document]
dkrieger-pvtl: Ok, we've deleted this test as it is only testing in-development step ordering specifics which are likely to change.

agent/services/execute_upgrade_primaries_sub_step.go
dkrieger-pvtl: We'll leave the last item for your follow up PR and the first item for the future.

hub/services/execute.go
```diff
@@ -129,6 +107,12 @@ type multiplexedStream struct {
 	mutex  sync.Mutex
 }
 
+func attachMultiplexedStreamToCmd(cmd *exec.Cmd, stream messageSender, log io.Writer) {
+	mux := newMultiplexedStream(stream, log)
+	cmd.Stdout = mux.NewStreamWriter(idl.Chunk_STDOUT)
+	cmd.Stderr = mux.NewStreamWriter(idl.Chunk_STDERR)
+}
+
```
jchampio: Just as a general "refactoring zen" observation: by refactoring this small piece, you've hidden the bigger issue, which is that a new multiplexed stream is being created over and over again. The duplication was pointing to a real problem, but this change only makes the duplication harder to see; it doesn't remove it.

(The stream-fix chore will pull this back out so you can see what I mean; no action is needed here.)

hub/services/initialize.go
```diff
@@ -16,59 +20,140 @@ import (
...
+	err = h.InitializeSubStep(initializeStream, upgradestatus.CHECK_UPGRADE, func(stream messageSender, log io.Writer) error {
+		return h.CheckUpgrade(stream, log)
+	})
+
+	return err
 }
 
-func (h *Hub) startAgentsSubStep(stream messageSender) error {
-	gplog.Info("starting %s", upgradestatus.START_AGENTS)
+// TODO: is there a way to pass in a variadic arg but still get type checking?
```
jchampio: Comment needs to be removed.

hub/services/start_or_stop_cluster_sub_step.go
```diff
@@ -0,0 +1,74 @@
+package services
+
+import (
+	"fmt"
+	"io"
+	"os/exec"
+
+	"github.com/greenplum-db/gpupgrade/utils"
```
jchampio: Don't make a change for this PR, but add `-local github.com/greenplum-db/gpupgrade` to your IDE's `goimports` invocation to get this sorted properly.

test/pg_upgrade.bats
```diff
@@ -0,0 +1,128 @@
...
+    run $PSQL -AtF$'\t' -p $PGPORT postgres -c "SELECT port, datadir FROM gp_segment_configuration WHERE role = 'p'"
+    [ "$status" -eq 0 ] || fail "$output"
+
+    declare -a olddirs
+    while read -r port dir; do
+        olddirs[$port]="$dir"
+    done <<< "$output"
+
+    local masterdir="${olddirs[$PGPORT]}"
+    newmasterdir="$(upgrade_datadir $masterdir)"
```
jchampio: The other code does the `olddirs` manipulation for a reason. This code can just grab the master data directory:
```bash
local masterdir="$($PSQL -At -p $PGPORT postgres -c "SELECT datadir FROM gp_segment_configuration WHERE role = 'p' and content = -1")"
newmasterdir="$(upgrade_datadir $masterdir)"
```
and be done.

dkrieger-pvtl:
> Seems like `test/status.bats` snuck back in.

Grrrrr....removed now.

hub/services/initialize.go
dkrieger-pvtl: Done.

cli/commanders/steps.go
```diff
@@ -21,9 +21,10 @@ var lines = map[idl.UpgradeSteps]string{
 	idl.UpgradeSteps_CONFIG:                  "Retrieving configs...",
 	idl.UpgradeSteps_START_AGENTS:            "Starting agents...",
 	idl.UpgradeSteps_CREATE_TARGET_CONFIG:    "Generating new cluster configuration...",
-	idl.UpgradeSteps_SHUTDOWN_SOURCE_CLUSTER: "Stopping old cluster...",
+	idl.UpgradeSteps_SHUTDOWN_SOURCE_CLUSTER: "Shutting down existing cluster...",
 	idl.UpgradeSteps_INIT_TARGET_CLUSTER:     "Creating new cluster...",
-	idl.UpgradeSteps_SHUTDOWN_TARGET_CLUSTER: "Stopping new cluster...",
+	idl.UpgradeSteps_SHUTDOWN_TARGET_CLUSTER: "Shutting down new cluster...",
```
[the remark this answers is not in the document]
dkrieger-pvtl: Fixed.

hub/services/check_upgrade_sub_step.go
```diff
@@ -0,0 +1,42 @@
...
+		}
+	}()
+
+	wg.Wait()
+	close(checkErrs)
+
+	var err error
+	for checkErr := range checkErrs {
+		err = multierror.Append(err, checkErr)
+	}
```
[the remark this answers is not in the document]
dkrieger-pvtl: Fixed.

hub/services/check_upgrade_sub_step.go
```diff
@@ -0,0 +1,42 @@
+package services
+
+import (
+	"github.com/hashicorp/go-multierror"
+	"io"
+	"sync"
+)
+
+func (h *Hub) CheckUpgrade(stream messageSender, log io.Writer) error {
+	wg := sync.WaitGroup{}
```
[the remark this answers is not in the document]
dkrieger-pvtl: Fixed.

hub/services/execute.go
dkrieger-pvtl: Ok, we looked at the stream-fix chore and know what you mean.

hub/services/start_or_stop_cluster_sub_step.go
dkrieger-pvtl: ok

test/pg_upgrade.bats
dkrieger-pvtl: Done

agent/services/execute_upgrade_primaries_sub_step_test.go
jchampio: I think maybe the wrong case was fixed here?

dkrieger-pvtl: The flag defaults to `false` in pg_upgrade.  Maybe I am misunderstanding your comment.

PR #156, opened 2019-11-09
add gpupgrade initialize --ports

kalensk:
The --ports flag enables users to choose the ports for initializing the new cluster.  ~Ports are assigned on each host in order of content ID, beginning at the lowest number.~ The lowest port is used for the master, and then remaining ports are assigned on each host in order of content ID, beginning at the second-lowest number. Unneeded ports at the top of the set are ignored, but an error results if the set is too small to handle all segments on a host. If no ports are given, start at 50432 (the pg_upgrade default) and count up, as if the user specified a range of 50432-65535.

If gpinitsystem fails due to a port collision then an error is returned and the user can fix or specify a different set of ports.

Ports can be specified either individually, or as ranges, separated by commas. When specifying a range a dash is used. For example, `--ports 6000,6005-6008` uses a port set <6000,6005,6006,6007,6008> for each host.

### DIFF
--- a/cli/commands/commands.go
+++ b/cli/commands/commands.go
@@ -32,6 +32,7 @@ import (
 	"fmt"
 	"os"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/greenplum-db/gp-common-go-libs/gplog"
@@ -236,6 +237,7 @@ func initialize() *cobra.Command {
 	var diskFreeRatio float64
 	var stopBeforeClusterCreation bool
 	var verbose bool
+	var ports string
 
 	subInit := &cobra.Command{
 		Use:   "initialize",
@@ -253,6 +255,11 @@ This step can be reverted.
 				)
 			}
 
+			ports, err := parsePorts(ports)
+			if err != nil {
+				return err
+			}
+
 			// If we got here, the args are okay and the user doesn't need a usage
 			// dump on failure.
 			cmd.SilenceUsage = true
@@ -261,7 +268,7 @@ This step can be reverted.
 			fmt.Println("Initialization in progress.")
 			fmt.Println()
 
-			err := commanders.CreateStateDirAndClusterConfigs(oldBinDir, newBinDir)
+			err = commanders.CreateStateDirAndClusterConfigs(oldBinDir, newBinDir)
 			if err != nil {
 				return errors.Wrap(err, "tried to create state directory")
 			}
@@ -297,7 +304,7 @@ This step can be reverted.
 				return nil
 			}
 
-			err = commanders.InitializeCreateCluster(client, verbose)
+			err = commanders.InitializeCreateCluster(client, verbose, ports)
 			if err != nil {
 				return errors.Wrap(err, "initializing cluster")
 			}
@@ -323,6 +330,7 @@ If you would like to return the cluster to its original state, run
 	subInit.PersistentFlags().MarkHidden("stop-before-cluster-creation")
 	subInit.PersistentFlags().Float64Var(&diskFreeRatio, "disk-free-ratio", 0.60, "percentage of disk space that must be available (from 0.0 - 1.0)")
 	subInit.PersistentFlags().BoolVarP(&verbose, "verbose", "v", false, "print the output stream from all substeps")
+	subInit.Flags().StringVar(&ports, "ports", "", "set of ports to use when initializing the new cluster")
 
 	return subInit
 }
@@ -337,13 +345,11 @@ func execute() *cobra.Command {
 Upgrades the master and primary segments over to the new cluster.
 This step can be reverted.
 `,
-		Run: func(cmd *cobra.Command, args []string) {
+		RunE: func(cmd *cobra.Command, args []string) error {
+			cmd.SilenceUsage = true
+
 			client := connectToHub()
-			err := commanders.Execute(client, verbose)
-			if err != nil {
-				gplog.Error(err.Error())
-				os.Exit(1)
-			}
+			return commanders.Execute(client, verbose)
 		},
 	}
 
@@ -367,6 +373,48 @@ This step can not be reverted.
 			os.Exit(1)
 		}
 	},
+}
+
+func parsePorts(val string) ([]uint32, error) {
+	var ports []uint32
+
+	if val == "" {
+		return ports, nil
+	}
+
+	for _, p := range strings.Split(val, ",") {
+		parts := strings.Split(p, "-")
+		switch {
+		case len(parts) == 2: // this is a range
+			low, err := strconv.ParseUint(parts[0], 10, 16)
+			if err != nil {
+				return nil, xerrors.Errorf("failed to parse port range %s", p)
+			}
+
+			high, err := strconv.ParseUint(parts[1], 10, 16)
+			if err != nil {
+				return nil, xerrors.Errorf("failed to parse port range %s", p)
+			}
+
+			if low > high {
+				return nil, xerrors.Errorf("invalid port range %s", p)
+			}
+
+			for i := low; i <= high; i++ {
+				ports = append(ports, uint32(i))
+			}
+
+		default: // single port
+			port, err := strconv.ParseUint(p, 10, 16)
+			if err != nil {
+				return nil, xerrors.Errorf("failed to parse port %s", p)
+			}
+
+			ports = append(ports, uint32(port))
+		}
+	}
+
+	return ports, nil
 }
 
 var restartServices = &cobra.Command{
--- a/cli/commands/commands_test.go
+++ b/cli/commands/commands_test.go
@@ -1,0 +1,55 @@
+package commands
+
+import (
+	"reflect"
+	"testing"
+)
+
+func TestParsePorts(t *testing.T) {
+	cases := []struct {
+		input    string
+		expected []uint32
+	}{
+		{"", []uint32(nil)},
+		{"1", []uint32{1}},
+		{"1,3,5", []uint32{1, 3, 5}},
+		/* ranges */
+		{"1-5", []uint32{1, 2, 3, 4, 5}},
+		{"1-5,6-10", []uint32{1, 2, 3, 4, 5, 6, 7, 8, 9, 10}},
+		{"1-5,10,12,15-15", []uint32{1, 2, 3, 4, 5, 10, 12, 15}},
+	}
+
+	for _, c := range cases {
+		actual, err := parsePorts(c.input)
+		if err != nil {
+			t.Errorf("parsePorts(%q) returned error %#v", c.input, err)
+		}
+		if !reflect.DeepEqual(actual, c.expected) {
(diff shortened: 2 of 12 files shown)


jchampio:
Rebased onto latest master. The big change was the movement of `gpinitsystem` from the execute step to initialize.

jchampio:
[pipeline](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:port_selection)

cli/commands/commands.go
```diff
@@ -323,6 +330,7 @@ If you would like to return the cluster to its original state, run
 	subInit.PersistentFlags().MarkHidden("stop-before-cluster-creation")
 	subInit.PersistentFlags().Float32Var(&diskFreeRatio, "disk-free-ratio", 0.60, "percentage of disk space that must be available (from 0.0 - 1.0)")
 	subInit.PersistentFlags().BoolVarP(&verbose, "verbose", "v", false, "print the output stream from all substeps")
+	subInit.Flags().StringVar(&ports, "ports", "", "set of ports to use when initializing the new cluster")
```
dkrieger-pvtl: this might not be a change for this PR, but how will a user obtain documentation for the format?  That is, "12345,12350-12360", etc.  Perhaps just added `e.g. 50000,50100-50200` might be enough or perhaps nothing is required if we're adding documentation elsewhere.

cli/commands/commands_test.go
```diff
@@ -0,0 +1,55 @@
...
+		if err != nil {
+			t.Errorf("parsePorts(%q) returned error %#v", c.input, err)
+		}
+		if !reflect.DeepEqual(actual, c.expected) {
+			t.Errorf("parsePorts(%q) returned %v, want %v", c.input, actual, c.expected)
+		}
+	}
+
+	errorCases := []string{
+		"1, 3, 5",
```
dkrieger-pvtl: In other words, `1, 3, 5` fails but `1,3,5` passes?

cli/commands/commands.go
```diff
@@ -369,6 +375,48 @@ This step can not be reverted.
...
+				return nil, xerrors.Errorf("failed to parse port %s", p)
+			}
+
+			ports = append(ports, uint32(port))
+		}
+	}
+
+	return ports, nil
+}
+
```
dkrieger-pvtl: This function allows any port [0,65535], including privileged ports and and dynamic ports.  I agree that we should not make such a business decision here but allow the gpdb database initialization to fail if it doesn't like that port.  I just wanted to point this out here, and perhaps there should be a comment somewhere in the code that this is the design we are taking.

cli/commands/commands_test.go
jchampio: Correct.

hub/services/init_target_cluster_sub_step.go
```diff
@@ -131,32 +131,99 @@ func upgradeDataDir(path string) string {
...
+				maxSegs = len(segments)
+			}
 		}
+
+		for i := 0; i < maxSegs; i++ {
+			ports = append(ports, uint32(50432+i))
+		}
+	}
+
+	ports = sanitize(ports)
```
dkrieger-pvtl: I'm thinking `sanitize` should be done in the CLI as part of the input checking instead.  If a user specifies the same port multiple times, it's almost certainly an unintended user error, and we should fail fast so they can correct it.  I see no reason to permit a user to specify the same port multiple times. Presumably the database initialization code will fail if it tries to use the same port twice, and that is our real check here.

Also, `sanitize` makes sense as a concept, but all we are doing in that function here is deduplication.  Since we seem to be deferring all real checking of the ports to `gpinitsystem`, perhaps a CLI function called `checkPorts` or similar would make more sense.

cli/commands/commands.go
jchampio: > nothing is required if we're adding documentation elsewhere.

Yeah, I think we're still figuring out documentation at the high level. We'll want to do a docs pass over the current option language anyway before we release.

hub/services/init_target_cluster_sub_step.go
jchampio: > If a user specifies the same port multiple times, it's almost certainly an unintended user error, and we should fail fast so they can correct it.

`1-5,5-10` is unambiguous but still needs to be deduplicated for safety. There's not much reason to make the user jump through hoops here IMO.

> Also, sanitize makes sense as a concept, but all we are doing in that function here is deduplication.

And sorting, which is critical for the implementation.

hub/services/init_target_cluster_sub_step.go
```diff
@@ -131,32 +131,99 @@ func upgradeDataDir(path string) string {
...
+	// the source cluster. This keeps the in-memory representation of source
+	// cluster consistent with its on-disk representation.
+	copySegments := make(map[int]cluster.SegConfig)
+	for _, segments := range segmentsByHost {
+		if len(ports) < len(segments) {
+			return nil, 0, errors.New("not enough ports for each segment")
+		}
+
+		for i, segment := range segments {
+			segment.Port = int(ports[i])
```
jchampio: So unfortunately, if we assign the master's port to a segment, `gpinitsystem` fails by incorrectly stating that the `PORT_BASE` (which we did not give, on purpose) overlaps with the `MASTER_PORT`. This is the cause of the pipeline failures on the CI.

We're considering just reserving the lowest port for the master and using the next port and up for the segments, to avoid having to go through the motions of fixing this and backporting it to 6X...

jchampio:
Our latest change to this PR tweaks the port assignment slightly:
> The lowest port in the range is always assigned to the
master, and segment port assignments begin with the second-lowest port
in the range. This means that the minimum number of ports needed in the
set is two, for a system in which all segments are on different hosts.

Unfortunately, gpinitsystem won't let you have a segment with a port that is equal to the master, even if they're on separate hosts. We didn't really feel like fixing that and pushing a backport to 6X for it, when it's easy enough to work around on our side.

cli/commands/commands_test.go
```diff
@@ -0,0 +1,55 @@
...
+		input    string
+		expected []uint32
+	}{
+		{"", []uint32(nil)},
+		{"1", []uint32{1}},
+		{"1,3,5", []uint32{1, 3, 5}},
+		/* ranges */
+		{"1-5", []uint32{1, 2, 3, 4, 5}},
+		{"1-5,6-10", []uint32{1, 2, 3, 4, 5, 6, 7, 8, 9, 10}},
+		{"1-5,10,12,15-15", []uint32{1, 2, 3, 4, 5, 10, 12, 15}},
```
marksliva: We could add a case which shows that ports are not duplicated in this function, i.e.
`{"7,7", []uint32{7,7}},`

hub/services/init_target_cluster_sub_step.go
dkrieger-pvtl: `1-5,5-10` is fine, but what about `6000,6001,6001,6003`(that is, a typo)?  

It is really a UI question, and my (relatively uninformed UI opinion) is that we should force the user to give us an unambiguous list. 

This need not delay the PR, but if we cannot receive some UI input from others, we should at least add an issue to rethink this later.

jchampio: > 1-5,5-10 is fine, but what about 6000,6001,6001,6003(that is, a typo)?

My arguments are
1. There's no difference from the internal perspective.
2. Users are likely to specify the latter as `6000-6003` anyway.
3. The penalty for a typo here is low-impact. If they don't have enough ports due to the duplication, we tell them within seconds. If they do have enough ports, we're still using the range they provided; it's just slightly smaller than they thought. And this temporary range is only used during upgrade.

jchampio: (@kodonnell26 , in case David's ping didn't go through on the edit.)

dkrieger-pvtl: Though after `execute`, they can have both clusters up for a while....

jchampio: > Though after execute, they can have both clusters up for a while....

Sure. What negative consequences do you see happening in this situation?

bharath1028: Is there a reason we're specifying a port range and not just a PORT_BASE like we do in `gpinitsystem`?

dkrieger-pvtl: Ok, after discussion with Brendan from Support, repeated ports are fine as long as there are enough of them.

@bharath1028 was wondering, however, why we chose the current explicit scheme(full range for segments) instead of specifying a single `PORT_BASE` that [gpinitsystem](https://gpdb.docs.pivotal.io/6-1/utility_guide/ref/gpinitsystem.html) allows.

jchampio: @bharath1028 @dkrieger-pvtl Well, this is embarrassing. The original intent was to make sure that we didn't have to reserve a port range from `PORT_BASE`-`PORT_BASE + <number of segments>`. But it turns out that the current `PORT_BASE` scheme already does that. Apparently I have been holding on to that misconception for a long time.

The current scheme, then, only has the advantage that the temporary ports are not required to be contiguous. I still _like_ that, but I admit it's a much weaker justification.

cli/commands/commands.go
dkrieger-pvtl: As per previous discussions, this is fine...we will let the hub determine this, and the ultimate decision here is done by `gpinitsystem` in determining if the various processes can bind to their respective ports.

hub/services/init_target_cluster_sub_step.go
dkrieger-pvtl: To me, we don't need to finalize on this decision before this PR goes in...the additional flexibility of a non-contiguous port range _can_ be removed later as a product decision.  But I think we need more customer input before finalizing this.  So, provided we add an issue to revisit this decision before release, the current scheme is fine with me.

bharath1028: +1 @dkrieger-pvtl. Thanks for the heads up @jchampio!

cli/commands/commands.go
dkrieger-pvtl: As previously commented, we can just let `gpinitsystem` fail in this case.

cli/commands/commands_test.go
```diff
@@ -0,0 +1,55 @@
...
+		"5-1",
+		"1--5",
+		"1-3-5",
+		"1,,2",
+		"1,a",
+		"1-a",
+		"a-1",
+		"900000",
+		"1-900000",
+		"900000-1000000",
```
dkrieger-pvtl: nit: might be nice to add a comment to each of the lines with a port over 65535 that that's why it fails.

hub/services/init_target_cluster_sub_step.go
dkrieger-pvtl: I also understand and agree that `sanitize` should go into the hub code for security considerations.

jchampio:
Thanks for the reviews!

PR #160, opened 2019-11-16
RFC: exectest improvements

jchampio:
This is a WIP. What I'm trying to do is get rid of some of the more annoying duplications: the proliferation of empty "success" and "failure" functions, and the introduction of additional `execCommand` injection points (`startStopCmd`, `isPostmasterRunningCmd`, etc.).

### DIFF
--- a/hub/services/start_or_stop_cluster_sub_step_test.go
+++ b/hub/services/start_or_stop_cluster_sub_step_test.go
@@ -19,23 +19,6 @@ import (
 	. "github.com/onsi/gomega"
 )
 
-func StartClusterCmd()        {}
-func StopClusterCmd()         {}
-func IsPostmasterRunningCmd() {}
-func IsPostmasterRunningCmd_Errors() {
-	os.Stderr.WriteString("exit status 2")
-	os.Exit(2)
-}
-
-func init() {
-	exectest.RegisterMains(
-		StartClusterCmd,
-		StopClusterCmd,
-		IsPostmasterRunningCmd,
-		IsPostmasterRunningCmd_Errors,
-	)
-}
-
 func TestStartOrStopCluster(t *testing.T) {
 	g := NewGomegaWithT(t)
 	ctrl := gomock.NewController(GinkgoT())
@@ -67,7 +50,7 @@ func TestStartOrStopCluster(t *testing.T) {
 	}()
 
 	t.Run("isPostmasterRunning succeeds", func(t *testing.T) {
-		isPostmasterRunningCmd = exectest.NewCommandWithVerifier(IsPostmasterRunningCmd,
+		isPostmasterRunningCmd = exectest.NewCommandWithVerifier(exectest.Success,
 			func(path string, args ...string) {
 				g.Expect(path).To(Equal("bash"))
 				g.Expect(args).To(Equal([]string{"-c", "pgrep -F basedir/seg-1/postmaster.pid"}))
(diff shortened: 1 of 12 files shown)


testutils/exectest/mains_test.go
```diff
@@ -0,0 +1,41 @@
...
+func TestBuiltinMains(t *testing.T) {
+	t.Run("Success()", func(t *testing.T) {
+		success := NewCommand(Success)("/unused/path")
+		err := success.Run()
+		if err != nil {
+			t.Errorf("exited with error %v", err)
+
+			var exitErr *exec.ExitError
+			if xerrors.As(err, &exitErr) {
+				t.Logf("subprocess stderr follows:\n%s", string(exitErr.Stderr))
```
dkrieger-pvtl: The string `subprocess stderr follows` seems misleading to me...the "process" running here is `success`.  I believe you are saying the test framework is the "process" and the `Success` command running is the "subprocess".  I'd prefer this be phrased as `Success process stderr follows`.

hub/services/start_or_stop_cluster_sub_step_test.go
```diff
@@ -41,71 +43,84 @@ func TestStartOrStopCluster(t *testing.T) {
...
+		SetExecCommand(exectest.Select(func(path string, args ...string) exectest.Command {
+			if len(args) < 2 {
+				goto unexpected
+			}
+
+			switch {
+			case strings.Contains(args[1], "pgrep"):
+				return pgrep
+			case strings.Contains(args[1], "gpstop"):
+				return gpstop
```
dkrieger-pvtl: It's not clear to me that ` bash -c source %[1]s/../greenplum_path.sh && %[1]s/%[2]s -a -d %[3]s"` matches `gpstop` in this way.

dkrieger-pvtl:
Overall, I really like this PR...it cleans up a lot of our ad-hoc usage of exectest. And, given our frequent usage of external programs in gpupgrade, it is an important feature.

I am on board with the separate test package for execttest tests, with the common Success/Failure functions and the use of SetExecCommand helper in all tests.

The value of exectest.Select is not yet clear to me. I mean that literally--I am not sure what exactly it saves for us. I know there are uses of it in the code, but perhaps a before/after of the current PR but without exectest.Select would help me understand this.

kalensk:
> The value of exectest.Select is not yet clear to me.

I too am having a bit of difficultly wrapping my head around this construct...As David mentioned perhaps some examples?

dkrieger-pvtl:
> > The value of exectest.Select is not yet clear to me.
> 
> I too am having a bit of difficultly wrapping my head around this construct...As David mentioned perhaps some examples?

The code uses it i this [commit](https://github.com/greenplum-db/gpupgrade/pull/160/commits/f5fd97395204668ecd9e6fb40722ba8d22f25be2).  It seems cleaner but there's a lot of indirection.

jchampio:
@dkrieger-pvtl @kalensk 

re: `Select`, the goal of that API was to remove situations where you had to define multiple injection points in order to unit test functions that call multiple commands. `Select` allowed me to get rid of [this code](https://github.com/greenplum-db/gpupgrade/pull/160/files#diff-7a5831ae3a47811eebffef675c4e6587), for instance.

hub/services/start_or_stop_cluster_sub_step_test.go
jchampio: I'm not sure what resolution you're looking for... Are you saying that you're not sure if the gpstop command actually calls gpstop? Because that's what the test verifies.

dkrieger-pvtl: It is not clear to me that the argument lists work that way, that's all.  I think I need to run it in a debugger to see it working, as opposed to looking at it statically.

berlin-ab:
Do we want to take action on this?

jchampio:
@berlin-ab I think some pieces of this need to be done, but it's now out of date enough that it's no longer relevant. Let's close for now and discuss as tech debt.

PR #161, opened 2019-11-21
CI: filter/patch to fix the 5-to-6 pipeline diff

jchampio:
_The first patch in this set is just a prefactor to make things easier, so make sure to review per-commit._

Unfortunately, there are some diffs in the output of pg_dumpall after an upgrade from 5X to 6X, and we're probably not going to be able to change them:

- The plpgsql extension is installed by default in 6, which leads to a group of related SQL additions.
- The postgres database now has an attached comment due to a change in behavior in the 6X initdb.
- All databases will be dumped with `gp_use_legacy_hashops = 'on'` in the 6X cluster, but there has been a product decision not to dump the same setting for a 5X cluster.

To mask these changes, implement a two-step process. First use a filter executable to remove any changes that can be handled with a simple sed-like streaming approach. This still leaves diffs that are hard to handle with a stream parser, so patch out those with a record of the "acceptable diffs" before performing the final comparison.

### DIFF
--- a/ci/scripts/filter/acceptable_diff
+++ b/ci/scripts/filter/acceptable_diff
@@ -1,0 +1,16 @@
+--- old.sql	2019-11-19 16:15:52.000000000 -0700
++++ /dev/fd/63	2019-11-19 18:02:03.000000000 -0700
+@@ -131,10 +132,11 @@
+ SET xmloption = content;
+ SET client_min_messages = warning;
+ 
+-SET default_tablespace = '';
+-
+ SET default_with_oids = false;
+ 
++
++SET default_tablespace = '';
++
+ --
+ -- Name: test_table; Type: TABLE; Schema: public; Owner: gpadmin; Tablespace: 
+ --
--- a/ci/scripts/filter/filter_test.go
+++ b/ci/scripts/filter/filter_test.go
@@ -1,0 +1,117 @@
+package main_test
+
+import (
+	"bytes"
+	"testing"
+
+	main "github.com/greenplum-db/gpupgrade/ci/scripts/filter"
+)
+
+func TestFilter(t *testing.T) {
+	t.Run("it writes stdin to stdout when nothing is filtered", func(t *testing.T) {
+		var in, out bytes.Buffer
+
+		expected := "hello\n"
+		in.WriteString(expected)
+
+		main.Filter(&in, &out)
+		if out.String() != expected {
+			t.Errorf("wrote %q want %q", out.String(), expected)
+		}
+	})
+
+	t.Run("filters out legacy hashops settings", func(t *testing.T) {
+		var in, out bytes.Buffer
+
+		in.WriteString(`
+GRANT ALL ON DATABASE template1 TO gpadmin;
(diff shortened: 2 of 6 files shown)


jchampio:
[pipeline](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:dev:5-6-filter)

ci/scripts/filter/filter_test.go
```diff
@@ -0,0 +1,117 @@
...
+RESET allow_system_table_mods;
+`
+		if out.String() != expected {
+			t.Errorf("wrote %q want %q", out.String(), expected)
+			t.Logf("actual (expanded): %s", out.String())
+			t.Logf("expected (expanded): %s", expected)
+		}
+	})
+
+	t.Run("filters out empty comment lines attached to filtered SQL", func(t *testing.T) {
```
jmcatamney: Very minor nit: This would be better as "filters out empty or commented lines" or similar, as "empty comment lines" implies `--\n` to me.

ci/scripts/filter/acceptable_diff
```diff
@@ -0,0 +1,16 @@
...
+ SET xmloption = content;
+ SET client_min_messages = warning;
+ 
+-SET default_tablespace = '';
+-
+ SET default_with_oids = false;
+ 
++
++SET default_tablespace = '';
++
```
jmcatamney: How necessary is it to handle this diff this way, as opposed to just getting rid of any `SET default_tablespace = '';` lines entirely?  I would assume pg_dump would only output the line once and any other `SET default_tablespace` statements would be at a per-user level or the like, so removing it in each case would fix the diff, but it's been a while since I looked into pg_dump.

jchampio: > How necessary is it to handle this diff this way, as opposed to just getting rid of any SET default_tablespace = ''; lines entirely?

Making those sorts of calls puts us into `gpdiff` territory, where we can accidentally ignore semantic differences. For example, did the test pass because we removed `default_tablespace` in the filter, or did it pass because pg_dump stopped correctly printing those statements after an upgrade?

In cases where we've decided we just want to ignore any new additions and we're okay if those additions silently disappear, we can use the filter. For all other cases, I'd rather be explicit.

ci/scripts/filter/filter_test.go
jchampio: Yeah, we accidentally a word there.

ci/scripts/filter/acceptable_diff
jmcatamney: Sounds reasonable.  I was going to say we can rely on pg_dump to not change out from under us 'cause it's tested elsewhere, but, well....

jchampio: Keep in mind that the systems are currently tightly coupled enough that a mistake _we_ make might cause pg_dump to do something incorrectly.

ci/scripts/filter/filter_test.go
jchampio: Fixed.

PR #162, opened 2019-11-21
add new/old gp-dbid to pg_upgrade

jchampio:
This is a new required flag in the latest 6X_STABLE branch,
so we add it to all invocations of pg_upgrade to fix gpupgrade.

Co-authored-by: Mark Sliva <msliva@pivotal.io>

### DIFF
--- a/idl/hub_to_agent.proto
+++ b/idl/hub_to_agent.proto
@@ -26,6 +26,7 @@ message DataDirPair {
     int32  OldPort    = 3;
     int32  NewPort    = 4;
     int32  Content    = 5;
+    int32  DBID       = 6;
 }
 
 message UpgradePrimariesReply {}
--- a/agent/services/execute_upgrade_primaries_sub_step.go
+++ b/agent/services/execute_upgrade_primaries_sub_step.go
@@ -71,13 +71,16 @@ func UpgradeSegments(sourceBinDir string, targetBinDir string, segments []Segmen
 		wg.Add(1)
 
 		path := filepath.Join(targetBinDir, "pg_upgrade")
+		dbid := strconv.Itoa(int(segment.DBID))
 		args := []string{
 			"--old-bindir", sourceBinDir,
 			"--old-datadir", segment.OldDataDir,
 			"--old-port", strconv.Itoa(int(segment.OldPort)),
+			"--old-gp-dbid", dbid,
 			"--new-bindir", targetBinDir,
 			"--new-datadir", segment.NewDataDir,
 			"--new-port", strconv.Itoa(int(segment.NewPort)),
+			"--new-gp-dbid", dbid,
 			"--mode=segment",
 		}
 		if checkOnly {
--- a/agent/services/execute_upgrade_primaries_sub_step_test.go
+++ b/agent/services/execute_upgrade_primaries_sub_step_test.go
@@ -51,7 +51,14 @@ var _ = Describe("UpgradeSegments", func() {
 		targetBinDir = "/new/bin"
 
 		segments = []Segment{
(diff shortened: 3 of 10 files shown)


berlin-ab:
One question that came up today was "how does the end user know their gp dbids?" Is this information already known to gpugprade, or did you make assumptions here in your tests?

jchampio:
@berlin-ab 

> Is this information already known to gpugprade

Yep, we know the source cluster dbids via `gp_segment_configuration` and we have control over the target cluster via `gpinitsystem`.

> One question that came up today was "how does the end user know their gp dbids?"

Thinking out loud... in GPDB 6+, each segment already knows its `gp_dbid` (through `internal.auto.conf`). So a usability optimization for that case would be to only require the `--old-gp-dbid` if the old cluster is 5X or prior. `--new-gp-dbid` would not be required at all when upgrading to 6 or later.

berlin-ab:
> @berlin-ab
> 
> > Is this information already known to gpugprade
> 
> Yep, we know the source cluster dbids via `gp_segment_configuration` and we have control over the target cluster via `gpinitsystem`.
> 
> > One question that came up today was "how does the end user know their gp dbids?"
> 
> Thinking out loud... in GPDB 6+, each segment already knows its `gp_dbid` (through `internal.auto.conf`). So a usability optimization for that case would be to only require the `--old-gp-dbid` if the old cluster is 5X or prior. `--new-gp-dbid` would not be required at all when upgrading to 6 or later.

Thanks, I appreciate the recommendation. cc @Eulerizeit  @gaos1 

Do you know of a way that we get that information programatically from gpdb5 without starting the server?

jchampio:
@berlin-ab 
> Do you know of a way that we get that information programatically from gpdb5 without starting the server?

Not off the top of my head, though that doesn't mean there isn't one... Do 5X segments have that information stored anywhere inside them for FileRep?

jmcatamney:
> @berlin-ab
> 
> > Do you know of a way that we get that information programatically from gpdb5 without starting the server?
> 
> Not off the top of my head, though that doesn't mean there isn't one... Do 5X segments have that information stored anywhere inside them for FileRep?

5x has a `gp_dbid` file in the segment directory, though as I recall that's never actually read by the segment and it solely relies on the `--gp_dbid` flag.  One could theoretically parse that file to get the dbid, but as far as gpupgrade goes the flag is easier.

idl/hub_to_agent.proto
```diff
@@ -26,6 +26,7 @@ message DataDirPair {
     int32  OldPort    = 3;
     int32  NewPort    = 4;
     int32  Content    = 5;
+    int32  DBID       = 6;
```
dkrieger-pvtl: nit: `DBID` makes me think it is a [constant](https://golang.org/doc/effective_go.html#constants).  I'm not sure if some other mixed case looks better....`dbid` seems better to me, though.  Any capital letters there look weird to me, by `dbID` seems the best.

jchampio: Keep in mind that case is significant in Go.

Idiomatically, constants aren't all-caps in Go (though there are some instances of that in the standard library that the core team has since apologized for, heh). They use [MixedCaps](https://github.com/golang/go/wiki/CodeReviewComments#mixed-caps):
> For example an unexported constant is maxLength not MaxLength or MAX_LENGTH.

This code review snippet on [initialisms](https://github.com/golang/go/wiki/CodeReviewComments#initialisms) also applies:
> For example, "URL" should appear as "URL" or "url" (as in "urlPony", or "URLPony"), never as "Url".

dkrieger-pvtl:
> LGTM...consider renaming the `DBID` hub to agent proto field, but the current `DBID` looks just as ok as other options as far as things go.

Ok, `DBID` is fine.

PR #164, opened 2019-11-26
agent: retain internal.auto.conf during master copy

jchampio:
...so that segments' DBIDs are correctly preserved. This fixes problems with mirror attachment after upgrade.

### DIFF
--- a/test/pg_upgrade.bats
+++ b/test/pg_upgrade.bats
@@ -13,6 +13,8 @@ setup() {
     # gpdeletesystem on this cluster.
     NEW_CLUSTER=
     KEEP_STATE_DIR=1
+
+    PSQL="$GPHOME"/bin/psql
 }
 
 teardown() {
@@ -31,7 +33,6 @@ teardown() {
 
     gpstart -a
 
-    PSQL="$GPHOME"/bin/psql
     $PSQL -d postgres -p $PGPORT -c "DROP TABLE IF EXISTS test_pg_upgrade CASCADE;"
 }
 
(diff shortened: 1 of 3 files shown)


test/pg_upgrade.bats
```diff
@@ -119,3 +118,37 @@ setup_newmasterdir() {
...
+        --verbose \
+        --old-bindir "$GPHOME/bin" \
+        --new-bindir "$GPHOME/bin" \
+        --old-port "$PGPORT" \
+        --disk-free-ratio=0 3>&-
+    NEW_CLUSTER="$newmasterdir"
+
+    gpupgrade execute --verbose
+
+    local new_dbid_num=$(count_primary_gp_dbids 50432)
```
marksliva: Is there a reason we aren't being more specific here? We could assert that the dbids exactly match their corresponding old content ids:
`diff $datadir/qddir/demoDataDir-1/internal.auto.conf $datadir/qddir_upgrade/demoDataDir-1/internal.auto.conf`

Or to check that the new running primary db has the correct dbid, we could parse the old `internal.auto.conf` for the dbid and compare it to that.

marksliva: We talked about it, and decided that we do not need to assert that these dbids are equivalent.

kalensk: > We talked about it, and decided that we do not need to assert that these dbids are equivalent.

Just curious what the reasoning was for learning purposes, etc.?

dkrieger-pvtl: > > We talked about it, and decided that we do not need to assert that these dbids are equivalent.
> 
> Just curious what the reasoning was for learning purposes, etc.?

There is no _database_ requirement that the dbids match between the old and new clusters.  The only requirement is that the dbids are distinct amongst all segments(master, standby, primary, mirror) in a given cluster.  Hence, we did not want to have a test which pins an implementation-defined behavior.

PR #165, opened 2019-11-26
Render a 1-N pipeline to enable multiple 5-to-6 jobs

marksliva:
This automates the management of multiple 5-to-6 upgrade jobs. 6.x.y versions
listed in parse_template.go are used to create a 5-to-6.x.y job for each
one, and the changes are then updated and persisted to template.yml after
running `make set-pipeline`.

Parsing with text/template requires changing the '{{' to '((' in the template
yml file since everything contained in '{{' must be a function or a variable
that can be parsed by this package. Switching to this is fine, especially since
the concourse docs are documenting the '((' syntax these days.


Makefile
```diff
@@ -153,6 +153,8 @@ endif
 .PHONY: set-pipeline expose-pipeline
 # TODO: Keep this in sync with the README at github.com/greenplum-db/continuous-integration
 set-pipeline:
+	# Always regenerate pipeline.yml from the template to keep it up to date
+	go run ./ci/parse_template.go
```
jchampio: Can we make this part of `go generate` instead, to keep all of the generated file logic the same?

ci/pipeline.yml
```diff
@@ -1,3 +1,4 @@
+# This file was generated by parse_template.go
```
jchampio: If we use `go generate` (or even if we don't) this should become the [standard header](https://golang.org/cmd/go/#hdr-Generate_Go_files_by_processing_source):
> To convey to humans and machine tools that code is generated, generated source should have a line that matches the following regular expression (in Go syntax):
>
>        ^// Code generated .* DO NOT EDIT\.$
>
>The line may appear anywhere in the file, but is typically placed near the beginning so it is easy to find.

ci/parse_template.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+	// Duplicate version data here in order to simplify template logic
+	versionInfo := VersionInfo{
+		SourceVersions: sourceVersions,
+		TargetVersions: targetVersions,
+		CombinedVersions: combinedVersions,
+		LastTargetVersion: targetVersions[len(targetVersions) - 1],
+	}
+	// Create truncates the file if it already exists, and opens it for writing
+	pipelineFilename := "pipeline.yml"
```
jchampio: The hardcoded paths feel like they would be better specified up one level, by the caller of the utility.

ci/parse_template.go
```diff
@@ -0,0 +1,70 @@
...
+	"os"
+	"path"
+	"strings"
+	"text/template"
+)
+
+var sourceVersions = []string{"5"}
+var targetVersions = []string{"6.1.0"}
+
+type CombinedVersion struct {
```
jchampio: nit: `Pair` or `VersionPair`, maybe?

ci/parse_template.go
```diff
@@ -0,0 +1,70 @@
...
+	SourceVersion, TargetVersion string
+}
+type VersionInfo struct {
+	SourceVersions, TargetVersions []string
+	CombinedVersions []*CombinedVersion
+	LastTargetVersion string
+}
```
jchampio: nit: the `Version` stutter here could be cleaned up.

ci/parse_template.go
```diff
@@ -0,0 +1,70 @@
...
+
+func main() {
+	templateFuncs := template.FuncMap{"escapeVersion": func (version string) string {
+		return strings.Replace(version, ".", `\.`, -1)
+	}}
```
jchampio: It's not clear to me what this does; can you add an explanation?

ci/parse_template.go
```diff
@@ -0,0 +1,70 @@
...
+			})
+		}
+	}
+	// Duplicate version data here in order to simplify template logic
+	versionInfo := VersionInfo{
+		SourceVersions: sourceVersions,
+		TargetVersions: targetVersions,
+		CombinedVersions: combinedVersions,
+		LastTargetVersion: targetVersions[len(targetVersions) - 1],
+	}
```
jchampio: nit: this global constant setup might be better pulled into an `init` function to keep it separated from the actual work being done by the utility.

ci/pipeline.yml
marksliva: I like this idea. However, I ran into a parse error:
```
error: error unmarshaling JSON: while decoding JSON: json: cannot unmarshal string into Go value of type atc.Config
```

So we can use this part `Code generated .* DO NOT EDIT`, but then change the comment to start with ##

Makefile
marksliva: Done.

ci/parse_template.go
marksliva: Done.

ci/parse_template.go
marksliva: Sure. I added a comment explaining what it does.

ci/parse_template.go
marksliva: Good point. Pushed a change where this goes into an `Init()`.

ci/generate.go
```diff
@@ -0,0 +1,7 @@
+package ci
+
+// Compile parse_template
+//go:generate go build -o parser/parse_template parser/parse_template.go
+
+// Generate pipeline.yml from pipeline-template.yml using parse_template
+//go:generate ./parser/parse_template ./parser/pipeline-template.yml pipeline.yml
```
jchampio: You can combine these into a single `go run` call (and that way you don't have to add a `.gitignore` entry either):
```go
// Generate pipeline.yml from pipeline-template.yml using parse_template
//go:generate go run ./parser/parse_template.go ./parser/pipeline-template.yml pipeline.yml
```

ci/parser/parse_template.go
```diff
@@ -0,0 +1,77 @@
...
+	// This function is used to ensure that the gcs-resource concourse plugin regex matches the version
+	// correctly. As an example if we didn't do this, 60100 would match version 6.1.0
+	templateFuncs := template.FuncMap{"escapeVersion": func (version string) string {
+		return strings.Replace(version, ".", `\.`, -1)
+	}}
+
+	yamlTemplate, err := template.New("Pipeline Template").Funcs(templateFuncs).ParseFiles(templateFilepath)
+	if err != nil {
+		log.Fatalf("error parsing %s: %+v", templateFilepath, err)
+	}
```
jchampio: I think the `yamlTemplate` and path logic still belongs in `main()`; I just meant that a `VersionInfo` global could be populated in a literal `init()` function.

ci/parser/parse_template.go
```diff
@@ -0,0 +1,77 @@
...
+	SourceVersions, TargetVersions []string
+	CombinedVersions []*CombinedVersion
+	LastTargetVersion string
+}
+
+func Init(templateFilepath string, pipelineFilepath string) (*template.Template, VersionInfo, string, string) {
+	// This function is used to ensure that the gcs-resource concourse plugin regex matches the version
+	// correctly. As an example if we didn't do this, 60100 would match version 6.1.0
+	templateFuncs := template.FuncMap{"escapeVersion": func (version string) string {
+		return strings.Replace(version, ".", `\.`, -1)
```
jchampio: I think `strings.ReplaceAll()` is what you want.

Actually, now that I understand the usage here, you could call this function `escapeRegex` and replace the implementation with [`regexp.QuoteMeta()`](https://golang.org/pkg/regexp/#QuoteMeta); that way it would "just work" with any strings.

ci/parser/parse_template.go
```diff
@@ -0,0 +1,77 @@
...
+}
+type VersionInfo struct {
+	SourceVersions, TargetVersions []string
+	CombinedVersions []*CombinedVersion
+	LastTargetVersion string
+}
+
+func Init(templateFilepath string, pipelineFilepath string) (*template.Template, VersionInfo, string, string) {
+	// This function is used to ensure that the gcs-resource concourse plugin regex matches the version
+	// correctly. As an example if we didn't do this, 60100 would match version 6.1.0
```
jchampio: Thanks, this helps. As one further suggestion, `This function` could be `The escapeVersion function` to draw attention to what's actually being defined.

ci/parse_template.go
marksliva: I'm not sure what you mean. Is there something that should be renamed, or can something be combined that I'm not seeing? Some data is duplicated here so that the template does nothing more complicated beyond iterating one loop at a time. I tried using a nested loop, but ran into issues when the outer loop's elements were inaccessible to the inner loop.

ci/parse_template.go
marksliva: Renamed to VersionPair.

ci/parse_template.go
marksliva: Fixed.

ci/parser/parse_template.go
marksliva: Fixed.

ci/generate.go
marksliva: fixed.

ci/parser/parse_template.go
marksliva: Fixed.

ci/parser/parse_template.go
marksliva: Good find. This is a drop in replacement for what I was trying to do.

PR #166, opened 2019-12-09
Streaming fixes

jmcatamney:
See commit messages for details.

### DIFF
--- a/agent/services/execute_upgrade_primaries_sub_step.go
+++ b/agent/services/execute_upgrade_primaries_sub_step.go
@@ -9,11 +9,10 @@ import (
 	"strconv"
 	"sync"
 
-	"golang.org/x/xerrors"
-
 	"github.com/greenplum-db/gp-common-go-libs/gplog"
 	"github.com/hashicorp/go-multierror"
 	"github.com/pkg/errors"
+	"golang.org/x/xerrors"
 
 	"github.com/greenplum-db/gpupgrade/hub/upgradestatus"
 	"github.com/greenplum-db/gpupgrade/idl"
@@ -32,7 +31,8 @@ func (s *AgentServer) AgentExecuteUpgradePrimariesSubStep(ctx context.Context, i
 
 type Segment struct {
 	*idl.DataDirPair
-	UpgradeDir string
+
+	WorkDir string // the pg_upgrade working directory, where logs are stored
 }
 
 func UpgradePrimaries(sourceBinDir string, targetBinDir string, dataDirPairs []*idl.DataDirPair, stateDir string, checkOnly bool) error {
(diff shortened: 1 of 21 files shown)


agent/services/execute_upgrade_primaries_sub_step.go
```diff
@@ -45,18 +44,27 @@ func UpgradePrimaries(sourceBinDir string, targetBinDir string, dataDirPairs []*
 			return err
 		}
 
-		segments = append(segments, Segment{DataDirPair: dataPair, UpgradeDir: dataPair.NewDataDir})
+		workdir := utils.SegmentPGUpgradeDirectory(stateDir, int(dataPair.Content))
```
kalensk: I am getting a little lost between the differences between `dataPair.NewDataDir` and `workdir`. That is, `utils.System.MkdirAll(dataPair.NewDataDir, 0700)` and `workdir`. What is the reason why we can't just replace the above line with workdir and delete all references to `dataPair.NewDataDir`?

jchampio: `workdir` is the working directory of pg_upgrade; i.e. where all the logs get dropped. The new data directory is the datadir for the new segment.

agent/services/execute_upgrade_primaries_sub_step.go
```diff
@@ -102,45 +109,11 @@ func UpgradeSegments(sourceBinDir string, targetBinDir string, segments []Segmen
...
-						err = multierror.Append(err,
-							xerrors.Errorf("failed to close pg_upgrade_check_stderr log: %w", closeErr))
-					}
-				}()
-
-				cmd.Stdout = stdout
-				cmd.Stderr = stderr
-			}
-
-			err = cmd.Run()
```
kalensk: Don't we need `cmd.Run()`? What am I missing - how is the command actually being run?

jchampio: It's the line below this.
```go
_, err := cmd.Output()
```

kalensk:
Looks good overall. I would like to reflect on it a bit more and review it again. (Others can feel free to approve it though.)

kalensk:
Nit: In `execute_upgrade_primaries_sub_step.go` to make things consistent and clear can we 
 either 1) use the term workdir everywhere (ie: rename UpgradeDir in the Segment struct), _or_ 2) use the term upgradeDir everywhere (ie: rename workdir to UpgradeDir)?
```
type Segment struct {
	*idl.DataDirPair
	UpgradeDir string
}
```

jchampio:
> Nit: In `execute_upgrade_primaries_sub_step.go` to make things consistent and clear can we
> either 1) use the term workdir everywhere (ie: rename UpgradeDir in the Segment struct)]

@kalensk I think 1) is a good idea.

hub/services/execute_test.go
```diff
@@ -127,152 +92,52 @@ func TestStreaming(t *testing.T) {
...
-		mockStream.EXPECT().
-			Send(gomock.Any()).
-			AnyTimes()
-
-		// Write ten bytes each to stdout/err.
-		execCommand = exectest.NewCommand(TenByteMain)
-
-		var buf bytes.Buffer
-		err := pair.ConvertMaster(mockStream, &buf, "", false)
+		err := pair.ConvertMaster(stream, "", false)
```
marksliva: Why do we have this test file at all? Nothing in here tests `execute.go` code. Since these tests are covering only `ConvertMaster`, they should be moved to `hub/services/execute_upgrade_master_sub_step_test.go`.

jchampio: Looks like these came in in #138.

kalensk: If I recall correctly (and following correctly) that is because upgrade_master was the first task to do streaming and I thought those tests were more streaming related at the time than specific to upgrading the master. Now that we have more tests for each and putting the responsibilities of utilities in their respective places it probably makes sense to move the tests.

kalensk: If I recall correctly (and following correctly) that is because upgrade_master was the first task to do streaming and I thought those tests were more streaming related at the time than specific to upgrading the master. Now that we have more tests for each and putting the responsibilities of utilities in their respective places it probably makes sense to move the tests.

PR #169, opened 2019-12-17
Merge agent and hub binaries to gpupgrade binary

bhuvnesh2703:
This PR has primarily 2 commits and some minor refactors:
1. Merge gpupgrade_hub binary into gpupgrade binary
2. Merge gpupgrade_agent binary into gpupgrade binary

This reduces the size of the binary by ~38-40MB


Makefile
```diff
@@ -88,9 +86,7 @@ build_linux: OS := LINUX
 build_mac: OS := MAC
 build_linux build_mac: build
 
-agent-package install-agent-package: EXE_NAME := $(AGENT)
 cli-package install-cli-package: EXE_NAME := $(CLI)
```
kalensk: Since there is just a single binary I think it would make things more clear to not reference CLI in this context and rename it to something more generic to reference the single EXE_NAME? I don't have a good suggestion at the moment other than `gpupgrade` which might be an overloaded term in this context.

I guess I am envisioning something like:
```
package install-package: EXE_NAME := $(gpupgrade)
```

ci/scripts/upgrade-cluster.bash
```diff
@@ -73,7 +73,7 @@ sudo ln -s trampoline pg_upgrade
 sudo ln -s trampoline psql
 sudo ln -s trampoline vacuumdb
 
-sudo ln -s trampoline gpupgrade_agent # XXX this is silly
+//sudo ln -s trampoline gpupgrade_agent # XXX this is silly
```
kalensk: Curious as to why is this line not deleted? Maybe a comment explaining why?

cli/commands/commands.go
```diff
@@ -60,6 +62,8 @@ func BuildRootCommand() *cobra.Command {
 	root.AddCommand(finalize)
 	root.AddCommand(restartServices)
 	root.AddCommand(killServices)
+	root.AddCommand(hub.CLICommand())
+	root.AddCommand(agent.CLICommand())
```
kalensk: nit: I think I prefer the simplicity of `hub.Command()` and `agent.Command()` rather than CLI which seems a little confusing to me. Just a preference and I'm open to what other people think.

kalensk:
On a first glance, it looks good overall. I like this change and happy to see it finally taking place. Thanks!

cli/commands/commands.go
jchampio: I recommended `CLICommand` to avoid any confusion with the `exec.Command` and related test helpers, but if no one is actually confused then I'm all for the simpler `Command()` name.

agent/commands.go
```diff
@@ -1,19 +1,17 @@
-package main
+package agent
 
 import (
-	"fmt"
-	"os"
-
 	"github.com/greenplum-db/gp-common-go-libs/cluster"
 	"github.com/greenplum-db/gp-common-go-libs/gplog"
+	"github.com/spf13/cobra"
```
d: This change is strange. Does `gofmt` literally order the cobra import in between various greenplum imports??

jchampio: `github.com/greenplum-db/gp-common-go-libs` is external to this package; `github.com/greenplum-db/gpupgrade` is local. We could choose to change that as a team if we wanted to, and define `github.com/greenplum-db` as local, but I'd personally prefer not to.

jchampio: (`goimports -local github.com/greenplum-db/gpupgrade` is the formatter that's doing this.)

Makefile
jchampio: Agreed; we're working on further simplifying this (now overly generic) code.

ci/scripts/upgrade-cluster.bash
jchampio: Removed now.

bhuvnesh2703: i left it as it was a WIP PR to run it with Jacob, but now deleted.

Makefile
```diff
@@ -112,9 +99,7 @@ installcheck:
 
 clean:
 		# Build artifacts
-		rm -f $(AGENT)
 		rm -f $(CLI)
```
jchampio: `$(CLI)` should now be `gpupgrade` here.

hub/commands.go
```diff
@@ -23,27 +23,20 @@ import (
...
 
-	var RootCmd = &cobra.Command{
-		Use:   os.Args[0],
-		Short: "Start the gpupgrade_hub (blocks)",
-		Long:  `Start the gpupgrade_hub (blocks)`,
+	var cmd = &cobra.Command{
+		Use:   "hub",
+		Short: "Start the gpupgrade hub (blocks)",
+		Long:  `Start the gpupgrade hub (blocks)`,
 		Args:  cobra.MaximumNArgs(0), //no positional args allowed
```
jchampio: We should hide the `hub` and `agent` commands from help. We can do that by adding a
```go
Hidden: true,
```
member to the `Command`.

Makefile
jchampio: Done. The Makefile is much lighter now.

agent/commands.go
```diff
@@ -23,23 +21,16 @@ func main() {
 	//}
```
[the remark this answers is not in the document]
jchampio: Done.

cli/commands/commands.go
jchampio: Done; switched to plain old `Command()`.

marksliva:
We should remove these 2 lines from `.gitignore`:
```
/gpupgrade_hub
/gpupgrade_agent
```

Makefile
```diff
@@ -74,34 +71,24 @@ coverage:
...
-INSTALLS := $(addprefix install-,$(PACKAGES))
 BUILD_ENV = $($(OS)_ENV)
 EXTENSION = $($(OS)_EXTENSION)
 
-.PHONY: build build_linux build_mac $(PACKAGES)
+.PHONY: build build_linux build_mac
 
-build: $(PACKAGES)
+build: .Gopkg.updated
+	$(BUILD_ENV) go build -o gpupgrade$(EXTENSION) $(BUILD_FLAGS) github.com/greenplum-db/gpupgrade/cmd/gpupgrade
```
marksliva: As a side note, do we want to keep supporting platform specific builds? We do not currently test any of these builds in CI, although we could add a test for `make build_linux`.

Unfortunately, the `make build_mac` command will not be possible to test in concourse CI, since running macOS in a container is not currently feasible.

agent/commands.go
```diff
@@ -0,0 +1,51 @@
...
+	"github.com/greenplum-db/gpupgrade/utils"
+	"github.com/greenplum-db/gpupgrade/utils/daemon"
+	"github.com/greenplum-db/gpupgrade/utils/log"
+)
+
+func Command() *cobra.Command {
+	var logdir, statedir string
+	var shouldDaemonize bool
+
+	var cmd = &cobra.Command{
```
marksliva: Now that `agent` is a part of `gpupgrade`, should we prepend the doc string with `(INTERNAL)`? Or replace it with `'gpupgrade agent' is an internal command`?
Something like:
`"(INTERNAL) Start the Command Listener (blocks)"`
or
`'gpupgrade agent' is an internal command`

agent/commands.go
```diff
@@ -0,0 +1,51 @@
...
+		Short:  "Start the Command Listener (blocks)",
+		Long:   `Start the Command Listener (blocks)`,
+		Hidden: true,
+		Args:   cobra.MaximumNArgs(0), //no positional args allowed
+		RunE: func(cmd *cobra.Command, args []string) error {
+			gplog.InitializeLogging("gpupgrade agent", logdir)
+			defer log.WritePanics()
+
+			conf := services.AgentConfig{
+				Port:     6416,
```
marksliva: Side question: Shouldn't we check that this port is available on the host before trying to start the agent? And also allow users to override it in case they are already using this port for a different service?

jchampio:
Rebased onto latest master.

jchampio:
> We should remove these 2 lines from .gitignore:

@marksliva Good catch! `ag` skipped `.gitignore` and we missed those. Fixed.

Makefile
jchampio: > As a side note, do we want to keep supporting platform specific build commands?

I've used them semi-recently, for cross-building Linux, so I don't see a good reason to get rid of them.

> We do not currently test any of these builds in CI, although we could add a test for make build_linux.

Considering this is a developer convenience and the functionality we'd be testing is provided by Go itself, I don't think we should spend much (if any) time adding this to the CI.

agent/commands.go
jchampio: @bharath1028 We have a few internal-only commands at the moment that are hidden from the user; how would you prefer we document them when running `--help`?

(I'll add a chore for this, to not hold up this PR.)

agent/commands.go
jchampio: Yep, there have already been a couple stories/chores filed on that.

PR #171, opened 2019-12-18
Add an .editorconfig

jchampio:
...so that hopefully GitHub will display our PRs correctly. Also, possibly, some editors will stop doing annoying things, like leaving off newlines at EOF.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,25 @@
+root = true
+
+[*]
+end_of_line = lf
+insert_final_newline = true
+
+[*.go]
+indent_style = tab
+indent_size = 4
+
+[*.{bash,bats,sh}]
+indent_style = space
+indent_size = 4
+
+[Makefile]
+indent_style = tab
+
+[ci/**.yml]
+indent_style = space
+indent_size = 2
+
+[*.proto]
+# https://developers.google.com/protocol-buffers/docs/style
+indent_style = space
+indent_size = 2


dkrieger-pvtl:
The change itself LGTM; it appears to follow the standard coding guidelines for the respective languages.  

Questions:
1). Do you know how the `.editorconfig` interacts with the per-editor code-style settings?  For intelliJ, it appears to override them but I'm not sure about vi or emacs.
2). Oddly, in git, when I change the `.go` settings from 5 to 10 spaces, `git diff` shows nothing.  This surprises me.  Do you know what's going on?

jchampio:
@dkrieger-pvtl 
>1). Do you know how the .editorconfig interacts with the per-editor code-style settings? For intelliJ, it appears to override them but I'm not sure about vi or emacs.

That's up to the plugin.

> 2). Oddly, in git, when I change the .go settings from 5 to 10 spaces, git diff shows nothing. This surprises me. Do you know what's going on?

The characters in the file are still just tabs. For Go, `.editorconfig` is just going to tell your editor what the tab stops are.

.editorconfig
```diff
@@ -0,0 +1,25 @@
...
+[*.go]
+indent_style = tab
+indent_size = 4
+
+[*.{bash,bats,sh}]
+indent_style = space
+indent_size = 4
+
+[Makefile]
+indent_style = tab
```
marksliva: Shouldn't we specify an `indent_size` for `Makefile`s? It can diverge depending on which editor is being used:
```
If indent_size equals to tab, the indent_size will be set to the tab size, which should be tab_width if tab_width is specified, or the tab size set by editor if tab_width is not specified.
```
https://github.com/editorconfig/editorconfig/wiki/EditorConfig-Properties#indent_size

jchampio: > Shouldn't we specify an indent_size for Makefiles?

It won't affect the file contents. So if you want your Makefiles to be displayed with 3- or 8- or 42-column tab stops that's up to you.

The config specifies 4-column stops for Go because the default 8-column stops in GitHub make our diffs very wide. I don't think we currently have any problems with difficult-to-read Makefile diffs in GitHub, but if we do have those later it'd be an easy addition to the config.

PR #172, opened 2020-01-04
hub/agent: consolidate upgrade implementations

jchampio:
The first commit adds a new upgrade API for both the hub and agent to use, which looks a little like this in practice:
```go
segmentPair := upgrade.SegmentPair{
    Source: &upgrade.Segment{sourceBinDir, segment.OldDataDir, dbid, int(segment.OldPort)},
    Target: &upgrade.Segment{targetBinDir, segment.NewDataDir, dbid, int(segment.NewPort)},
}

err := upgrade.Run(segmentPair,
    upgrade.WithExecCommand(execCommand),
    upgrade.WithWorkDir(segment.WorkDir),
    upgrade.WithSegmentMode(),
)
```

The second commit makes use of this API throughout the code base and removes redundant tests. One business logic change is that master upgrade logs are now correctly `--retain`ed due to the consolidation of code.


upgrade/run.go
```diff
@@ -0,0 +1,149 @@
...
+	return cmd.Run()
+}
+
+// Option configures the way Run executes pg_upgrade.
+type Option func(*optionList)
+
+// WithWorkDir configures the working directory for pg_upgrade. Log files will
+// be retained in this directory. If this option is unset or empty, the calling
+// process's working directory will be used.
+func WithWorkDir(wd string) Option {
```
berlin-ab: Cool pattern with these builder-style options. Do we get any benefit from the delayed execution? Why not build up these options as state on a struct?

upgrade/run.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+}
+
+// optionList holds the combined result of all possible Options. Zero values
+// represent the default settings.
+type optionList struct {
+	Dir            string
+	CheckOnly      bool
+	ExecCommand    func(string, ...string) *exec.Cmd
+	ExecCommandSet bool // was ExecCommand explicitly set?
```
kalensk: Maybe I need to stare at the code longer, but why is this needed? Shouldn't simply setting the ExecCommand be enough? Also, why is ExecCommand even needed on the optionList, rather than simply doing something like this directly:
```

cmd := execCommand(path, args...)
```

upgrade/run.go
jchampio: I started to write up an explanation and then realized that [Dave Cheney summarizes it the best in the place where I got the idea](https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis).

upgrade/run.go
jchampio: Good questions, and maybe we need to add some explanation in the code. I'll answer the second one first.
> Also, why is ExecCommand even needed on the optionList, rather than simply doing something like this directly: [...]

Problem with that is that you can no longer stub out the dependency across packages. So if you're running Hub tests and one of the functions you're testing needs to call upgrade, you can't access `upgrade.execCommand` to be able to control it (i.e. prevent upgrade from actually running). So we borrowed the gRPC idea of allowing an option to inject a dependency, similar to `WithContextDialer()`.

> Shouldn't simply setting the ExecCommand be enough?

That was what I thought too. Then @bhuvnesh2703 asked "so what happens if you don't set up a mock command in the tests?" and I answered "it'll crash to let us know there's a problem, because we set it to `nil` at the beginning of the tests, see?"

And then of course it did _not_ crash, and instead accidentally executed `pg_upgrade`, because we took `nil` to mean "use the default". Having a separate boolean allows us to explicitly `nil` out the `MockCommand` implementation.

kalensk: Thanks for the explanation! 

Talking out loud here...It would be nice if the `CheckOnly` bool was not needed. That is, I was hoping for defaulting to null. And in the production code use the correct execCommand and for tests use the fake execCommand, and if its not set it will fail since the default is null. However, if one was to implement that there would probably be no way for the tests to "fail" and not use the real pg_upgrade if one forgets to set the fake pg_upgrade execCommand when writing a test. Unfortunate.

upgrade/run_test.go
```diff
@@ -0,0 +1,280 @@
...
+
+	t.Run("unsets PGPORT and PGHOST", func(t *testing.T) {
+		// Set our environment.
+		os.Setenv("PGPORT", "5432")
+		os.Setenv("PGHOST", "localhost")
+		defer func() {
+			os.Unsetenv("PGPORT")
+			os.Unsetenv("PGHOST")
+		}()
+
```
dkrieger-pvtl: It's a pedantic point, but you risk polluting other tests in this package by changing their `PGPORT` and `PGHOST` values to the unset state.  As far as I can quickly tell, `go test` tests all run in the same process, hence my concern.  Instead, use this idiom:
```
	oldVal, isSet := os.LookupEnv("VAR")
	defer func() {
		if isSet {
			os.Setenv("VAR", oldVal)
		}
	}()
	err = os.Setenv("VAR", "5432")
```

agent/upgrade_primaries.go
```diff
@@ -39,14 +37,8 @@ func UpgradePrimary(sourceBinDir string, targetBinDir string, dataDirPairs []*id
 	segments := make([]Segment, 0, len(dataDirPairs))
 
 	for _, dataPair := range dataDirPairs {
-		err := utils.System.MkdirAll(dataPair.NewDataDir, 0700)
```
dkrieger-pvtl: Where is the target segment data directory created now?  Or was it always unnecessary?

dkrieger-pvtl:
This needs to be rebased with origin/master(see cli/commanders/initialize.go).

upgrade/run.go
dkrieger-pvtl: This is a pretty cool pattern.  And, unlike the blog post, you don't expose internal implementation details but set an Options struct.

upgrade/run.go
dkrieger-pvtl: @kalensk why would you not need `checkOnly` as an option?  Sometimes we run

upgrade/run_test.go
bhuvnesh2703: applied feedback.

bhuvnesh2703:
Rebased and applied the feedback required by @dkrieger-pvtl

PR #173, opened 2020-01-09
make first two substeps of Initialize idempotent 

dkrieger-pvtl:
In preparation for skipping completed substeps and continuing when an error occurred, we make the first two substeps of Initialize idempotent.  In addition, we split the first substep into two substeps: creating the state directory and creating the initial cluster config files.

Note that these substeps are run in the cli(vs. the hub), so they do not use the checklist manager.

Note that there are two commits but, due to inadvertent rebasing issues, some of the code that should be in the second commit are actually in the first(namely, the startHub substep IsHubRunning being included in the first commit).

### DIFF
--- a/cli/commanders/initialize.go
+++ b/cli/commanders/initialize.go
@@ -89,7 +89,7 @@ func StartHub() (err error) {
 		return err
 	}
 	if running {
-		gplog.Debug("gpupgrade hub already running...")
+		gplog.Debug("gpupgrade hub already running...skipping.")
 		return nil
 	}
 
--- a/cli/commanders/initialize.go
+++ b/cli/commanders/initialize.go
@@ -19,23 +19,31 @@ import (
 var execCommandHubStart = exec.Command
 var execCommandHubCount = exec.Command
 
-// we create the state directory in the cli to ensure that at most one gpupgrade is occuring
+// we create the state directory in the cli to ensure that at most one gpupgrade is occurring
 // at the same time.
-func CreateStateDirAndClusterConfigs(sourceBinDir, targetBinDir string) (err error) {
-	s := Substep("Creating directories...")
+func CreateStateDir() (err error) {
+	s := Substep("Creating state directory...")
 	defer s.Finish(&err)
 
 	stateDir := utils.GetStateDir()
 	err = os.Mkdir(stateDir, 0700)
 	if os.IsExist(err) {
-		return fmt.Errorf("gpupgrade state dir (%s) already exists. Did you already run gpupgrade initialize?", stateDir)
-	} else if err != nil {
+		gplog.Debug("State directory %s already present...skipping", stateDir)
+		return nil
+	}
+	if err != nil {
+		gplog.Debug("State directory %s could not be created.", stateDir)
 		return err
 	}
 
-	// Create empty clusters in source and target so that gpupgrade hub can
-	// start without having replaced them with current values.
-	// TODO: implement a slicker scheme to allow this.
+	return nil
+}
+
+func CreateInitialClusterConfigs(sourceBinDir, targetBinDir string) (err error) {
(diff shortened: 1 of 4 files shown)


dkrieger-pvtl:
Note: our [fixup commit](https://github.com/greenplum-db/gpupgrade/pull/173/commits/2835d7f16ec946aebe6f580033d892391c547b43) removed a sequence of `t.Run()` to a single one so we could guarantee the separate parts of the test run sequentially.  We could not think of a way to have several `r.Run()` subtests to run sequentially.

jchampio:
> We could not think of a way to have several r.Run() subtests to run sequentially.

Subtests run sequentially by default. You have to explicitly mark them parallel-safe to get them to run concurrently.

cli/commanders/initialize.go
```diff
@@ -19,40 +19,57 @@ import (
...
+func CreateStateDir() (err error) {
+	s := Substep("Creating state directory...")
 	defer s.Finish(&err)
 
 	stateDir := utils.GetStateDir()
 	err = os.Mkdir(stateDir, 0700)
 	if os.IsExist(err) {
-		return fmt.Errorf("gpupgrade state dir (%s) already exists. Did you already run gpupgrade initialize?", stateDir)
-	} else if err != nil {
+		gplog.Debug("State directory already present...skipping.")
```
bhuvnesh2703: should we also log the path of the state directory so the user knows?

cli/commanders/initialize.go
```diff
@@ -19,40 +19,57 @@ import (
...
-	}
 
 	target := &utils.Cluster{
 		Cluster:    emptyCluster,
 		BinDir:     path.Clean(targetBinDir),
 		ConfigPath: filepath.Join(stateDir, utils.TARGET_CONFIG_FILENAME),
 	}
+
+	if source.Load() == nil && target.Load() == nil {
+		gplog.Debug("Initial cluster config files already present...skipping.")
```
bhuvnesh2703: the location of cluster configuration files should be logged.

cli/commands/commands.go
```diff
@@ -272,19 +272,14 @@ This step can be reverted.
 			fmt.Println("Initialization in progress.")
 			fmt.Println()
 
-			err = commanders.CreateStateDirAndClusterConfigs(oldBinDir, newBinDir)
+			err = commanders.CreateStateDir()
 			if err != nil {
-				return errors.Wrap(err, "tried to create state directory")
+				return errors.Wrap(err, "creating state directory")
```
bhuvnesh2703: location of the state directory should be logged.

dkrieger-pvtl:
> > We could not think of a way to have several r.Run() subtests to run sequentially.
> 
> Subtests run sequentially by default. You have to explicitly mark them parallel-safe to get them to run concurrently.

Is this documented somewhere?  I couldn't see anything on it in the [somewhat dated official documentation](https://blog.golang.org/subtests) but [non-normative references](https://stackoverflow.com/questions/31201858/how-to-run-golang-tests-sequentially/31204016#31204016) claim the order is non-deterministic between subtests.

That said, we decided to place use a single subtest, since the different subtests we had are in fact dependent on each other.  We didn't want to leave a trap for future developers(likely us) to place a test in between the current ones and perhaps cause a "spurious" test failure.

Besides, it appears that the different subtests are idiomatically not supposed to be dependent on each other.

cli/commanders/initialize.go
dkrieger-pvtl: Done

cli/commands/commands.go
dkrieger-pvtl: The state directory is not directly accessible here(technically it is via a global, but I'd rather start limited uses of it).  Instead, I've added a log statement to all error branches of the underlying function `CreateStateDir()`.

PR #175, opened 2020-01-11
make creation of the state directory a singleton

dkrieger-pvtl:
GetStateDir(): make this a singleton via sync.Once

This function uses the GPUPGRADE_HOME or HOME environment variables, so
we must latch the value on the first call of any process and keep it. 
Otherwise, havoc occurs: the state directory might end up being
inconsistent or the program might behave in unexpected ways.

Because the existing initialize_test.go file used GPUPGRADE_HOME twice
with different values, we split those tests into diffferent files.


cli/commanders/initialize_creater_cluster_test.go
```diff
@@ -0,0 +1,87 @@
+package commanders
```
marksliva: I have a strong preference for this unit test to stay in the existing `initialize_test.go` file. That will make it significantly easier to maintain both of these files, because then there will be just 1 conventional file location that contains all of the associated unit tests for `initialize.go`.

marksliva: This can be accomplished by reorganizing the two groups of tests into 2 different top level test contexts `func TestInitialize...`.

utils/sys_utils.go
```diff
@@ -113,10 +116,13 @@ func GetHost() (string, error) {
 }
 
 func GetStateDir() string {
-	stateDir := os.Getenv("GPUPGRADE_HOME")
-	if stateDir == "" {
-		stateDir = filepath.Join(os.Getenv("HOME"), ".gpupgrade")
-	}
+
+	setStateDirOnce.Do(func() {
```
marksliva: Should we be ignoring new user-specified values for $GPUPGRADE_HOME in all cases?

marksliva: It looks isolated for now and probably safe to do it this way, but just wondering if we have considered this case.

utils/sys_utils.go
```diff
@@ -113,10 +116,13 @@ func GetHost() (string, error) {
...
+		stateDir = os.Getenv("GPUPGRADE_HOME")
+		if stateDir == "" {
+			stateDir = filepath.Join(os.Getenv("HOME"), ".gpupgrade")
+		}
```
jmcatamney: There should be at few tests for this fallback-to-`HOME` behavior.  Maybe one where both variables start out set, one where `HOME` starts set and `GPUPGRADE_HOME` is set after calling `GetStateDir`, and so forth.

utils/sys_utils.go
jmcatamney: > Should we be ignoring new user-specified values for $GPUPGRADE_HOME in all cases?

I can't think of a case where a user could safely change `GPUPGRADE_HOME` after gpupgrade starts (changing from an existing state directory to an empty one would give invalid status information, accidentally pointing one run at a different run's state directory would make gpupgrade skip or repeat steps, and so on), and our BATS tests start a new process for each test already, so I'd say ignoring all new values makes sense.

dkrieger-pvtl: > Should we be ignoring new user-specified values for $GPUPGRADE_HOME in all cases?

I hope so...`$GPUPGRADE_HOME` determines the state directory and I cannot see why we would want to change that after starting a run of gpupgrade.  It's very tricky to allow that directory to change during a run, and I cannot think of a reason to do that.

As an aside, I initially think all the initial parameters(old/new bindir and old/new datadir) should be done as singletons.  But we have(or used to) have a way of changing some of those for testing, and we might have need for that for testing.

cli/commanders/initialize_creater_cluster_test.go
dkrieger-pvtl: I would too.  The issue I ran into is that it appears that each `_test.go` file runs in its own process.  Both `initialize_test.go::TestCreateStateDir()` and `initialize_creater_cluster_test.go::TestCreateInitialClusterConfigs()` reset their own `GPUPGRADE_HOME` via a call to CreateStateDir().  That is why they are in separate files.

utils/sys_utils_test.go
```diff
@@ -195,4 +196,40 @@ var _ = Describe("user utils", func() {
 			Expect(err).ToNot(HaveOccurred())
 		})
 	})
+
+	Describe("GetStateDir()", func() {
+		It("correctly uses sync.Once and latches the value on the first call", func() {
```
kalensk: I suppose I can't convince this PR to include a separate commit to port this file of tests over to go style :p

jchampio:
> This function uses the GPUPGRADE_HOME or HOME environment variables, so
we must latch the value on the first call of any process and keep it.

We are in full control over this. The user can't change our envvars from the outside once the process has started. So although I'm in full support of getting rid of the confusing caching layer and just using `GetStateDir()` directly, I'm lukewarm on introducing this change if it makes our tests more confusing, because I doubt we'll see a benefit in production.

berlin-ab:
It seems that rather than force `GetStateDir` to only execute once, we should rather only call it once during our `main`/`Command` execution, and then pass it to where needed.

cli/commanders/initialize_creater_cluster_test.go
marksliva: I see. That could work, but is a little weird. If we end up getting rid of caching according to the suggestion below by @jchampio, then we wouldn't need this work around.

dkrieger-pvtl:
For both comments [here](https://github.com/greenplum-db/gpupgrade/pull/175#issuecomment-574210676) and [other here](https://github.com/greenplum-db/gpupgrade/pull/175#issuecomment-574243789), the point of the PR is that the state directory _should_ be immutable once set.  Yes, if we change it is a bug, but it makes the code much easier to reason about if you know it cannot change by code design.  

Similarly, we could call it once and pass it around(either on the hub to receiver methods or directly as a function argument), but global _immutable_ state for configuration is a well-accepted pattern AFAIK.

It does slightly complicate one test, but to me it isn't that big a deal relative to knowing that the state directory is immutable.  As further evidence, we do actually call GetStateDir() in a few places that would lead to a changed state directory if we somehow modified the GPUPGRADE_HOME directory.

utils/sys_utils_test.go
dkrieger-pvtl: That'd be a different 10%...I'd really prefer to add this change without expanding the scope.

cli/commanders/initialize_creater_cluster_test.go
dkrieger-pvtl: Right, but "getting rid of caching" simply means trusting ourselves to only call GetStateDir() once.  From a robustness standpoint, making that true by design seems much better.  It does slightly complicate the design of one test, but I argue [below](https://github.com/greenplum-db/gpupgrade/pull/175#issuecomment-574254261) that this is worth the tradeoff.

dkrieger-pvtl: > I am open to ideas. I just pushed a commit to this PR that does place all tests in initialize_test.go. It uses a trick(table-driven tests) to force the order of the tests. It's not pretty but seems ok to me.

It's much easier to just call the functions in order ;).  The table-driven testing isn't needed.

utils/sys_utils.go
marksliva: Sure. I was thinking more of a case where the hub goes down, and then a user accidentally runs a gpupgrade command as a different user for example. In this case, initialize would be affected and would create a new target cluster, but at least execute and finalize would fail when the state directory is not found.

hub/commands.go
```diff
@@ -40,27 +40,26 @@ func Command() *cobra.Command {
 			conf := &Config{
 				CliToHubPort:   7527,
 				HubToAgentPort: 6416,
-				StateDir:       utils.GetStateDir(),
 				LogDir:         logdir,
 			}
 
-			finfo, err := os.Stat(conf.StateDir)
+			finfo, err := os.Stat(utils.GetStateDir())
```
berlin-ab: This is the only place that we validate that the given state dir is legit. We call `GetStateDir` in a lot of other places without validation. Are there any guarantees that this is the first invocation?

utils/sys_utils.go
dkrieger-pvtl: I just pushed a commit that does what I think we can without testing the Once functionality. One of the drawbacks of using Once is that if you have branches inside the Once function, you cannot test those.  However, I broke out the internals into a non-Once function.  @jmcatamney Let me know what you think.

hub/commands.go
dkrieger-pvtl: What sort of guarantee do you want?  It's called in the cobra command pretty early.  We could call it in a `PreE` function, perhaps.  With the current, non-Once design, you also have no such guarantee.  You just know, like in the PR, that all uses after this is called reference the same value _if_ they are passed in(as opposed to calling the function again).

dkrieger-pvtl:
> The desire to write a test for a private function is a signal that the design could be improved.

Perhaps...we have a branch inside a Once function, which is where the difficulty in testing arises.  I'm open to ideas that preserve the GetStateDir() as immutable once called the first time.

utils/sys_utils.go
dkrieger-pvtl: That's outside the scope of the current gpupgrade...we never intended to support this.

kalensk:
> the point of the PR is that the state directory _should_ be immutable once set. ... it makes the code much easier to reason about if you know it cannot change by code design.
> 
> It does slightly complicate one test, but to me it isn't that big a deal relative to knowing that the state directory is immutable.

+1 I think this change is a great step forward. Thank you for doing it.

hub/commands.go
dkrieger-pvtl: Besides, in some sense, the first caller in time(e.g. dynamically when the program is run) must necessarily be the first caller in space(e.g. in code order, the first time the function was called).  So I think the `first invocation` doesn't matter per se.

cli/commanders/initialize_test.go
```diff
@@ -156,122 +158,103 @@ func TestCreateStateDir(t *testing.T) {
...
-			}
+func RunTestCreateStateDir(t *testing.T, stateDir string) {
+	var infoOld os.FileInfo
 
-			if infoOld, err = os.Stat(home); err != nil {
-				t.Errorf("unexpected error %#v", err)
-			}
+	{ // creates state directory if none exist or fails
+		if _, err := os.Stat(stateDir); err == nil {
+			t.Errorf("stateDir exists")
```
bhuvnesh2703: can we put the path in the error message?

utils/sys_utils_test.go
```diff
@@ -195,4 +196,84 @@ var _ = Describe("user utils", func() {
...
+			var stateDir_1 string
+			{ // the first call to GetStateDir() sets the value to GPUPGRADE_HOME
+				home_1, err := ioutil.TempDir("", "GetStateDir_")
+				Expect(err).ToNot(HaveOccurred())
+
+				oldStateDir, isSet := os.LookupEnv("GPUGRADE_HOME")
+				defer func() {
+					if isSet {
+						os.Setenv("GPUPGRADE_HOME", oldStateDir)
+					}
```
bhuvnesh2703: planning for a else to unset it if not set?

utils/sys_utils_test.go
```diff
@@ -195,4 +196,84 @@ var _ = Describe("user utils", func() {
...
+
+			err := os.Unsetenv("GPUPGRADE_HOME")
+			Expect(err).ToNot(HaveOccurred())
+
+			expectedDir := filepath.Join(curHOME, ".gpupgrade")
+
+			actualDir := getStateDirInternal()
+			Expect(expectedDir).To(Equal(actualDir))
+
+		})
```
bhuvnesh2703: what happens if both HOME and GPUPGRADE_HOME is not set?

bhuvnesh2703:
This PR will require a rebase and minor modification in tests, as now we only have one cluster configuration file.

cli/commanders/initialize_test.go
dkrieger-pvtl: Done.

berlin-ab:
Let's rebase this and then wrap it up.

utils/sys_utils_test.go
dkrieger-pvtl: added.

utils/sys_utils_test.go
dkrieger-pvtl: `HOME` is [required](https://pubs.opengroup.org/onlinepubs/009695399/basedefs/xbd_chap08.html) to be set on a POSIX system.  We can revisit that later; this is a feature of the current code.  I've added a tracker story.

dkrieger-pvtl:
This generated enough discussion that, if we decide to address this, it will be a different implementation.

PR #176, opened 2020-01-14
Introduce a new package to run, track status, and skip substeps

kalensk:
We implement a new scheme to run all substeps. This allows us to skip completed substeps when re-running a step after an error. This eliminates existing boilerplate code to fail a step if one substep fails. This includes a mechanism to durably store substep status to disk.  We use a JSON file that is written atomically to disk, and all retrievals of substep status read directly from the disk. 

This PR maintains the current checklist manager and substep tracking mechanism as well as its use by Execute and Finalize.  This allows us to port substeps one at a time over to this new mechanism without breaking existing code.

We initially wanted to refactor the existing checklist manager, but decided to simply rewrite it due to the difficulty in maintaining that existing code.  Some issues we ran into:

- Having two sources of truth for the suSstep status leads to subtle bugs 
  - the in-memory representation on the hub
  - what is on disk. 
- Having a separate checklist containing a stateReader and stateWriter for each individual substep rather than one single reader & writer. 
 - Having to define all the steps upon initialization rather keeping the reader & writer generic. This also allows us to have a single source of truth-the code order of substeps.

This is a draft PR to gather feedback as we continue flushing out unit tests. Once this pattern and mechanism to track substep status is in place the work to make all steps idempotent can be done in parallel.

### DIFF
--- a/hub/initialize.go
+++ b/hub/initialize.go
@@ -11,112 +11,80 @@ import (
 	"github.com/greenplum-db/gp-common-go-libs/gplog"
 	"github.com/hashicorp/go-multierror"
 	"github.com/pkg/errors"
-	"golang.org/x/xerrors"
 
 	"github.com/greenplum-db/gpupgrade/db"
-	"github.com/greenplum-db/gpupgrade/hub/upgradestatus"
 	"github.com/greenplum-db/gpupgrade/idl"
+	"github.com/greenplum-db/gpupgrade/step"
 	"github.com/greenplum-db/gpupgrade/utils"
 )
 
 func (h *Hub) Initialize(in *idl.InitializeRequest, stream idl.CliToHub_InitializeServer) (err error) {
-	log, err := utils.System.OpenFile(
-		filepath.Join(utils.GetStateDir(), "initialize.log"),
-		os.O_WRONLY|os.O_CREATE,
-		0600,
-	)
+	s, err := BeginStep(h.conf.StateDir, "initialize", stream)
 	if err != nil {
 		return err
 	}
+
 	defer func() {
-		if closeErr := log.Close(); closeErr != nil {
-			err = multierror.Append(err,
-				xerrors.Errorf("failed to close initialize log: %w", closeErr))
+		if ferr := s.Finish(); ferr != nil {
+			err = multierror.Append(err, ferr).ErrorOrNil()
+		}
+
+		if err != nil {
+			gplog.Error(fmt.Sprintf("initialize: %s", err))
 		}
 	}()
 
-	initializeStream := newMultiplexedStream(stream, log)
+	s.Run(idl.UpgradeSteps_CONFIG, func(stream step.OutStreams) error {
+		return h.fillClusterConfigsSubStep(stream, in.OldBinDir, in.NewBinDir, int(in.OldPort))
+	})
 
-	_, err = log.WriteString("\nInitialize in progress.\n")
-	if err != nil {
-		return xerrors.Errorf("failed writing to initialize log: %w", err)
-	}
+	s.Run(idl.UpgradeSteps_START_AGENTS, func(stream step.OutStreams) error {
+		return h.startAgentsSubStep(stream)
+	})
 
-	err = h.Substep(initializeStream, upgradestatus.CONFIG,
-		func(stream OutStreams) error {
-			return h.fillClusterConfigsSubStep(stream, in.OldBinDir, in.NewBinDir, int(in.OldPort))
-		})
-	if err != nil {
-		return err
-	}
-
-	err = h.Substep(initializeStream, upgradestatus.START_AGENTS, h.startAgentsSubStep)
-	if err != nil {
-		return err
-	}
-
-	return nil
+	return s.Err()
 }
 
 func (h *Hub) InitializeCreateCluster(in *idl.InitializeCreateClusterRequest, stream idl.CliToHub_InitializeCreateClusterServer) (err error) {
-	log, err := utils.System.OpenFile(
-		filepath.Join(utils.GetStateDir(), "initialize.log"),
-		os.O_WRONLY|os.O_APPEND,
-		0600,
-	)
+	s, err := BeginStep(h.conf.StateDir, "initialize", stream)
 	if err != nil {
 		return err
 	}
+
 	defer func() {
-		if closeErr := log.Close(); closeErr != nil {
-			err = multierror.Append(err,
-				xerrors.Errorf("failed to close initialize log: %w", closeErr))
+		if ferr := s.Finish(); ferr != nil {
+			err = multierror.Append(err, ferr).ErrorOrNil()
+		}
+
+		if err != nil {
+			gplog.Error(fmt.Sprintf("initialize: %s", err))
 		}
 	}()
 
-	initializeStream := newMultiplexedStream(stream, log)
-
-	_, err = log.WriteString("\nInitialize hub in progress.\n")
-	if err != nil {
-		return xerrors.Errorf("failed writing to initialize log for hub: %w", err)
-	}
-
 	var targetMasterPort int
-	err = h.Substep(initializeStream, upgradestatus.CREATE_TARGET_CONFIG,
-		func(_ OutStreams) error {
-			var err error
-			targetMasterPort, err = h.GenerateInitsystemConfig(in.Ports)
-			return err
-		})
-	if err != nil {
+	s.Run(idl.UpgradeSteps_CREATE_TARGET_CONFIG, func(_ step.OutStreams) error {
+		var err error
+		targetMasterPort, err = h.GenerateInitsystemConfig(in.Ports)
 		return err
-	}
+	})
 
-	err = h.Substep(initializeStream, upgradestatus.SHUTDOWN_SOURCE_CLUSTER,
-		func(stream OutStreams) error {
-			return StopCluster(stream, h.source)
-		})
-	if err != nil {
-		return err
-	}
+	s.Run(idl.UpgradeSteps_SHUTDOWN_SOURCE_CLUSTER, func(stream step.OutStreams) error {
+		return StopCluster(stream, h.source)
+	})
 
-	err = h.Substep(initializeStream, upgradestatus.INIT_TARGET_CLUSTER,
-		func(stream OutStreams) error {
-			return h.CreateTargetCluster(stream, targetMasterPort)
-		})
-	if err != nil {
-		return err
-	}
+	s.Run(idl.UpgradeSteps_INIT_TARGET_CLUSTER, func(stream step.OutStreams) error {
+		return h.CreateTargetCluster(stream, targetMasterPort)
+	})
 
-	err = h.Substep(initializeStream, upgradestatus.SHUTDOWN_TARGET_CLUSTER,
-		func(stream OutStreams) error {
-			return h.ShutdownCluster(stream, false)
-		})
-	if err != nil {
-		return err
-	}
+	s.Run(idl.UpgradeSteps_SHUTDOWN_TARGET_CLUSTER, func(stream step.OutStreams) error {
+		return h.ShutdownCluster(stream, false)
+	})
 
-	return h.Substep(initializeStream, upgradestatus.CHECK_UPGRADE, h.CheckUpgrade)
+	s.Run(idl.UpgradeSteps_CHECK_UPGRADE, func(stream step.OutStreams) error {
+		return h.CheckUpgrade(stream)
+	})
+
+	return s.Err()
 }
 
 // create old/new clusters, write to disk and re-read from disk to make sure it is "durable"
--- a/hub/substeps_new.go
+++ b/hub/substeps_new.go
@@ -1,0 +1,63 @@
+package hub
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/hashicorp/go-multierror"
+	"golang.org/x/xerrors"
+
+	"github.com/greenplum-db/gpupgrade/step"
+)
+
+func BeginStep(stateDir string, name string, sender messageSender) (*step.Step, error) {
+	path := filepath.Join(stateDir, fmt.Sprintf("%s.log", name))
+	log, err := os.OpenFile(path, os.O_WRONLY|os.O_CREATE|os.O_APPEND, 0600)
+	if err != nil {
+		return nil, xerrors.Errorf(`step "%s": %w`, name, err)
+	}
+
+	_, err = fmt.Fprintf(log, "\n%s in progress.\n", strings.Title(name))
+	if err != nil {
+		log.Close()
+		return nil, xerrors.Errorf(`logging step "%s": %w`, name, err)
+	}
+
+	statusPath, err := getStatusFile(stateDir)
+	if err != nil {
+		return nil, xerrors.Errorf("step %q: %w", name, err)
+	}
+
+	streams := newMultiplexedStream(sender, log)
+	return step.New(name, sender, step.NewFileStore(statusPath), streams), nil
+}
+
+// Returns path to status file, and if one does not exist it creates an empty
+// JSON file.
+func getStatusFile(stateDir string) (path string, err error) {
+	path = filepath.Join(stateDir, "status.json")
+
+	f, err := os.OpenFile(path, os.O_EXCL|os.O_CREATE|os.O_WRONLY, 0600)
+	if os.IsExist(err) {
+		return path, nil
+	}
+	if err != nil {
+		return "", err
+	}
+
+	defer func() {
+		if cErr := f.Close(); cErr != nil {
+			err = multierror.Append(err, cErr).ErrorOrNil()
+		}
+	}()
+
+	// MarshallJSON requires a well-formed JSON file
+	_, err = f.WriteString("{}")
+	if err != nil {
+		return "", err
+	}
+
+	return path, nil
+}
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -53,6 +53,14 @@
   version = "v1.3.2"
 
 [[projects]]
+  digest = "1:c3388642e07731a240e14f4bc7207df59cfcc009447c657b9de87fec072d07e3"
+  name = "github.com/google/renameio"
+  packages = ["."]
+  pruneopts = "UT"
+  revision = "f0e32980c006571efd537032e5f9cd8c1a92819e"
+  version = "v0.1.0"
+
+[[projects]]
   branch = "master"
   digest = "1:b7388e8388cea56e5fd6b604eecfe56f5d02c0f41c49ad5a853da39b6dc197a0"
   name = "github.com/greenplum-db/gp-common-go-libs"
@@ -383,6 +391,7 @@
     "github.com/golang/mock/mockgen",
     "github.com/golang/protobuf/proto",
     "github.com/golang/protobuf/protoc-gen-go",
+    "github.com/google/renameio",
     "github.com/greenplum-db/gp-common-go-libs/cluster",
     "github.com/greenplum-db/gp-common-go-libs/dbconn",
     "github.com/greenplum-db/gp-common-go-libs/gplog",
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -101,3 +101,7 @@ required = [
 [[constraint]]
   branch = "master"
   name = "golang.org/x/xerrors"
+
+[[constraint]]
+  name = "github.com/google/renameio"
+  version = "0.1.0"
--- a/hub/substeps.go
+++ b/hub/substeps.go
@@ -81,6 +81,7 @@ type messageSender interface {
 	Send(*idl.Message) error // matches gRPC streaming Send()
 }
 
+// TODO: remove; this is part of step.Step now
 func sendStatus(stream messageSender, step idl.UpgradeSteps, status idl.StepStatus) {
 	// A stream is not guaranteed to remain connected during execution, so
 	// errors are explicitly ignored.
@@ -156,6 +157,16 @@ func (m *multiplexedStream) Stdout() io.Writer {
 
 func (m *multiplexedStream) Stderr() io.Writer {
 	return m.stderr
+}
+
+// Close closes the stream's io.Writer if that writer also provides a Close
+// method (i.e. it also implements io.WriteCloser). If not, Close is a no-op.
+func (m *multiplexedStream) Close() error {
+	if closer, ok := m.writer.(io.WriteCloser); ok {
+		return closer.Close()
+	}
+
+	return nil
 }
 
 type streamWriter struct {
--- a/hub/substeps_new_test.go
+++ b/hub/substeps_new_test.go
@@ -1,0 +1,66 @@
+package hub
+
+import (
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"testing"
+)
+
+func TestStatusFile(t *testing.T) {
+	stateDir, err := ioutil.TempDir("", "")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer func() {
+		if err := os.RemoveAll(stateDir); err != nil {
+			t.Errorf("removing temp directory: %v", err)
+		}
+	}()
+
+	path := filepath.Join(stateDir, "status.json")
+
+	t.Run("creates status file if it does not exist", func(t *testing.T) {
+		_, err := os.Open(path)
+		if !os.IsNotExist(err) {
+			t.Errorf("returned error %#v want ErrNotExist", err)
+		}
+
+		statusFile, err := getStatusFile(stateDir)
+		if err != nil {
+			t.Errorf("unexpected error %v", err)
+		}
+
+		actual, err := ioutil.ReadFile(statusFile)
+		if err != nil {
+			t.Fatalf("ReadFile(%q) returned error %#v", statusFile, err)
+		}
+
+		expected := "{}"
+		if string(actual) != expected {
+			t.Errorf("read %v want %v", string(actual), expected)
+		}
+	})
+
+	t.Run("does not create status file if it already exists", func(t *testing.T) {
+		expected := "1234"
+		err := ioutil.WriteFile(path, []byte(expected), 0600)
+		if err != nil {
+			t.Fatalf("unexpected error %v", err)
+		}
+
+		statusFile, err := getStatusFile(stateDir)
+		if err != nil {
+			t.Errorf("unexpected error %v", err)
+		}
+
+		actual, err := ioutil.ReadFile(statusFile)
+		if err != nil {
+			t.Errorf("ReadFile(%q) returned error %#v", statusFile, err)
+		}
+
+		if string(actual) != expected {
+			t.Errorf("read %v want %v", string(actual), expected)
+		}
+	})
+}
--- a/idl/cli_to_hub.proto
+++ b/idl/cli_to_hub.proto
@@ -34,11 +34,13 @@ message RestartAgentsReply {
 message StopServicesRequest {}
 message StopServicesReply {}
 
+// TODO: rename to SubstepStatus
 message UpgradeStepStatus {
     UpgradeSteps step = 1;
     StepStatus status = 2;
 }
 
+// TODO: rename to Substep
 enum UpgradeSteps {
     UNKNOWN_STEP = 0; // http://androiddevblog.com/protocol-buffers-pitfall-adding-enum-values/
     CONFIG = 1;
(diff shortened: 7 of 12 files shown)


hub/substep_status.go
```diff
@@ -0,0 +1,55 @@
...
+
+type SubStepsStore interface {
+	Read() (map[string]*step.SubStepData, error)
+	Write(step *step.SubStepData) (err error)
+}
+
+type HubSubStepStore struct {
+	store  SubStepsStore
+	stream messageSender
+}
```
kalensk: I could envision a fileStore and streamingStore to separate out the two implementations. Something like:

```
type HubSubStepStore struct {
	fileStore  FileStore
	streamingStore  streamingStore
	stream messageSender
}
```

Basically renaming `substep_store.go` to `substep_fileStore.go`, etc.

utils/sys_utils.go
```diff
@@ -146,3 +148,29 @@ func WriteJSONFile(fileName string, contents interface{}) error {
 
 	return nil
 }
+
+var createStatusFileOnce sync.Once
+
+func GetStatusFile() (statusFile string, err error) {
```
kalensk: I think this can be moved into `substep_store.go` as a private helper function since it is only used there. It can be extracted later if it ends up being needed at more places.

subStepStore/substep_store.go
```diff
@@ -0,0 +1,71 @@
...
+	if err != nil {
+		return err
+	}
+
+	statusFile, err := utils.GetStatusFile()
+	if err != nil {
+		return err
+	}
+
+	updatedStatusFile := statusFile + ".updated"
```
kalensk: Might be nice to make this a hidden file with something like:
`updatedStatusFile := filepath.Join(utils.GetStateDir(), ".status.json")`

utils/sys_utils.go
```diff
@@ -146,3 +148,29 @@ func WriteJSONFile(fileName string, contents interface{}) error {
...
+	statusFile = path.Join(GetStateDir(), "status.json")
```
kalensk: I think we should use `filepath.join` here instead of `path.join`.

hub/substeps_new.go
```diff
@@ -0,0 +1,99 @@
...
+
+func Err(s *StepData) error {
+	return s.errChain
+}
+
+type StepData struct {
+	name        string
+	multiStream *multiplexedStream
+	log         io.WriteCloser
+	dir         string
```
kalensk: I don't think the `dir` field is used anymore and can be deleted.

hub/substeps_new.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	store := HubSubStepStore{store: &subStepStore.Store{}, stream: s.multiStream.stream}
+	subStep, err := store.ReadSubStep(name)
+	if err != nil {
+		return
+	}
+
+	// Only re-run tasks that have failed, pending, or were running.
+	if subStep.Status == idl.StepStatus_COMPLETE {
```
berlin-ab: Don't need to do this now, but since `store` already is aware of the `Complete` status, we might want to ask `store.isComplete(subStep)`.

hub/substeps_new.go
```diff
@@ -0,0 +1,99 @@
...
+		store.MarkComplete(subStep)
+		return
+	}
+
+	_, err = fmt.Fprintf(s.multiStream.writer, "\nStarting %s...\n\n", name)
```
berlin-ab: This is probably a hold-over from what used to be here, but now it sticks out as unlike its surroundings.

Don't need to address this now, but let's keep it in mind.

hub/substeps_new.go
```diff
@@ -0,0 +1,99 @@
...
+	if err != nil {
+		return
+	}
+
+	err = store.MarkInProgress(subStep)
+	if err != nil {
+		return
+	}
+
+	err = f(s.multiStream)
```
berlin-ab: I'd prefer to have better names for `f` and `s` here, so I don't have to search for their meaning. Not urgent, but food for thought.

step/steps.go
```diff
@@ -0,0 +1,41 @@
...
+type SubStepData struct {
+	SubStep idl.UpgradeSteps
+	Status  idl.StepStatus
+}
+
+// Use a custom marshaller to correctly serialize enum values. This enables the
+// enum string to be used instead of the numeric value for ease of reading.
+// This uses a pointer receiver to marshal pointers. Create a non-pointer
+// receiver to marshal non-pointers.
+func (s *SubStepData) MarshalJSON() ([]byte, error) {
```
berlin-ab: We should consider moving the JSON serialization into the layer that needs to serialize. In this case, I think it is the file writer.  JSON serialization is not a core concern of upgrade, but rather an implementation detail of storing a file containing substeps on disk.

subStepStore/substep_store.go
```diff
@@ -0,0 +1,71 @@
+package subStepStore
```
berlin-ab: I was curious about the idiom in go for package names:

https://medium.com/rungo/everything-you-need-to-know-about-packages-in-go-b8bac62b74cc#ab0c

There's at least one person's opinion.

subStepStore/substep_store.go
```diff
@@ -0,0 +1,71 @@
...
+		return nil, err
+	}
+
+	return subSteps, nil
+}
+
+// First read the latest values from the filesystem, rather than storing
+// in-memory on a struct. This avoids having two sources of truth.
+// We write to a temporary file and use os.Rename to atomically update the step.
+func (s *Store) Write(subStep *step.SubStepData) (err error) {
```
berlin-ab: Do we do atomic storage of JSON for the state directory elsewhere? This seems likely to be duplicate code mixed in with substep specific needs.

utils/sys_utils.go
berlin-ab: what are the benefits of `filepath.join`?

subStepStore/substep_store.go
```diff
@@ -0,0 +1,71 @@
...
+		return err
+	}
+
+	statusFile, err := utils.GetStatusFile()
+	if err != nil {
+		return err
+	}
+
+	updatedStatusFile := statusFile + ".updated"
+	err = ioutil.WriteFile(updatedStatusFile, data, 0644)
```
berlin-ab: I wouldn't know by looking at this code that we're renaming this file in order to get an atomic write. Maybe an intention-explaining function would be useful to clarify.

hub/substeps_new.go
```diff
@@ -0,0 +1,99 @@
...
+
+	"golang.org/x/xerrors"
+
+	"github.com/greenplum-db/gpupgrade/idl"
+	"github.com/greenplum-db/gpupgrade/subStepStore"
+	"github.com/greenplum-db/gpupgrade/utils"
+)
+
+func BeginStep(stateDir string, name string, stream messageSender) (*StepData, error) {
+	path := filepath.Join(stateDir, fmt.Sprintf("%s.log", name))
```
berlin-ab: In the future we should consider pushing UI concerns up a layer into the caller.

berlin-ab: This includes log file creation, writing, streaming, and knowledge of `idl`.

utils/sys_utils.go
kalensk: https://stackoverflow.com/a/39182128 ?

step/steps.go
kalensk: Do we _really_ need another layer? What does that give us?

hub/substeps_new.go
kalensk: fwiw, I don't mind the `f` and `s` names here.

subStepStore/substep_store.go
kalensk: There is a function level comment explaining this.

hub/substeps_new.go
kalensk: I am not sure I agree with this. This encapsulates common functionality, and probably a few other advantages that I can't articulate at this minute.

step/steps.go
berlin-ab: The other layer already exists. The new Read/Write functions.

subStepStore/substep_store.go
berlin-ab: Ah, I missed that. Something in context might still be useful, but the existing comment is helpful.

hub/substeps_new.go
berlin-ab: I'm saying it's something to consider, not something to act on yet.

utils/sys_utils.go
berlin-ab: ah, that is useful.

hub/substeps_new.go
dkrieger-pvtl: Are you suggesting we have each caller handle its UI?  Can you provide an example in our code(gpdb, say) where we do this?

hub/substeps_new.go
dkrieger-pvtl: What do you mean?  This can be useful information.  What do you prefer instead?

hub/substeps_new.go
dkrieger-pvtl: The nice thing about `f` is that it is idiomatic.  You know it's a function pointer; you don't care what it is per se.  For `s`, it's becoming in our code too.

subStepStore/substep_store.go
dkrieger-pvtl: Technically, no: we have the new scheme which uses the JSON file and the old scheme using per-directory status.  But they don't overlap.  There's no other place that does this JSON writing.

Where do you prefer it?

berlin-ab: Nowhere for now. Just wondering.

hub/substeps_new.go
berlin-ab: I'll get used to the idiom then.

hub/substeps_new.go
berlin-ab: It would be a consequence of applying the Dependency Inversion Principle - 

(from https://en.wikipedia.org/wiki/Dependency_inversion_principle)

- High-level modules should not depend on low-level modules. Both should depend on abstractions (e.g. interfaces).
- Abstractions should not depend on details. Details (concrete implementations) should depend on abstractions.

step/steps.go
```diff
@@ -0,0 +1,41 @@
...
+	return json.Marshal(&struct {
```
marksliva: This stuct is repeated in marshal and unmarshal. We can define it once like `type SubstepJSON struct` to keep them in sync.

kalensk: Right. Thanks for catching that as that is a hold over from implementing what is discussed at the end of this blog post: http://choly.ca/post/go-json-marshalling/

marksliva: I made this change and `make check` still passes.

subStepStore/substep_store.go
marksliva: I don't like this Go idiom, but I guess it fits with the Go philosophy. Looking at Go's standard libraries, they cheat a little using names like `iotest` `jsonrpc` `suffixarray` etc. Another one of their workarounds is nested libraries:
```
net
    http
        cgi
```
https://golang.org/pkg/
So far, our codebase has mostly stuck to that convention though so I guess we can continue doing that.

step/steps.go
kalensk: Right. We don't have tests for that yet. I believe a quick manual verification of seeing the `status.json` file in the state directory will show the numbers instead of strings when removing it.

utils/sys_utils.go
jchampio: +1. Let's stop feeding the `utils` package.

utils/sys_utils.go
```diff
@@ -146,3 +147,26 @@ func WriteJSONFile(fileName string, contents interface{}) error {
...
+	createStatusFileOnce.Do(func() {
+		func() { err = createStatusFile(statusFile) }()
+	})
+
+	return statusFile, err
+}
+
+// Creates an empty JSON file if none exists
+func createStatusFile(statusFile string) error {
+	f, err := os.OpenFile(statusFile, os.O_EXCL|os.O_RDWR, 0644)
```
jchampio: `O_EXCL` without `O_CREAT`?

utils/sys_utils.go
```diff
@@ -146,3 +147,26 @@ func WriteJSONFile(fileName string, contents interface{}) error {
...
+	if err != nil {
+		// MarshallJSON requires a well-formed JSON file
+		return ioutil.WriteFile(statusFile, []byte("{}"), 0644)
```
jchampio: We don't use the file handle (`f`) we just opened here?

Wait, we rewrite the status file to empty every time the hub starts? Why?

utils/sys_utils.go
```diff
@@ -146,3 +147,26 @@ func WriteJSONFile(fileName string, contents interface{}) error {
...
+	}
+
+	return f.Close()
```
jchampio: `f.Close()` belongs in a defer statement so that we clean up our file handle even on errors.

step/steps.go
dkrieger-pvtl: We did it in our editor, and it looks less clean.

Original:
```go
func (s *SubStepData) MarshalJSON() ([]byte, error) {
	return json.Marshal(&struct {
		Name   string
		Status string
	}{
		Name:   s.SubStep.String(),
		Status: s.Status.String(),
	})
}

func (s *SubStepData) UnmarshalJSON(data []byte) error {
	subStep := &struct {
		Name   string
		Status string
	}{}

	if err := json.Unmarshal(data, &subStep); err != nil {
		return err
	}

	s.SubStep = idl.UpgradeSteps(idl.UpgradeSteps_value[subStep.Name])
	s.Status = idl.StepStatus(idl.StepStatus_value[subStep.Status])
	return nil
}
```

New:
```go
type SubStepJSONData struct {
	Name   string
	Status string
}
func (s *SubStepData) MarshalJSON() ([]byte, error) {
	return json.Marshal(&SubStepJSONData{
		Name:   s.SubStep.String(),
		Status: s.Status.String(),
	})
}
func (s *SubStepData) UnmarshalJSON(data []byte) error {
	subStep := &SubStepJSONData{}
	if err := json.Unmarshal(data, &subStep); err != nil {
		return err
	}

	s.SubStep = idl.UpgradeSteps(idl.UpgradeSteps_value[subStep.Name])
	s.Status = idl.StepStatus(idl.StepStatus_value[subStep.Status])
	return nil
}

```

utils/sys_utils.go
dkrieger-pvtl: Yes...this is our way of checking whether `statusFile` was previously created.  But I am rethinking this as apparently `O_EXCL` without `O_CREAT` isn't well-defined.  We'll think of a better scheme.

utils/sys_utils.go
dkrieger-pvtl: No...we couldn't think of a cleaner way to do this, but here's the code with more explanation

```go
         // OpenFile will return nil if the statusFile exists, in which case we have no work to do
	f, err := os.OpenFile(statusFile, os.O_EXCL|os.O_RDWR, 0644)

         // otherwise, WriteFile creates the file and then writes "{}" to it.
	if err != nil {
		// MarshallJSON requires a well-formed JSON file
		return ioutil.WriteFile(statusFile, []byte("{}"), 0644)
	}
```

dkrieger-pvtl:
@jchampio  What code or design change are you suggesting by the [dependency injection comment](https://github.com/greenplum-db/gpupgrade/pull/176#discussion_r366644244)?

subStepStore/substep_store.go
```diff
@@ -0,0 +1,71 @@
...
+	err = json.Unmarshal(data, &subSteps)
+	if err != nil {
+		return nil, err
+	}
+
+	return subSteps, nil
+}
+
+// First read the latest values from the filesystem, rather than storing
+// in-memory on a struct. This avoids having two sources of truth.
```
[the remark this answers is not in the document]
dkrieger-pvtl: We have two sources of truth: the status as written to disk and the status as streamed.  We want to make sure that, if we stream out the new status, that we have committed it to disk first.

There's no bug per se, just trying to make sure the step status is consistent.

subStepStore/substep_store.go
```diff
@@ -0,0 +1,71 @@
...
+// We write to a temporary file and use os.Rename to atomically update the step.
```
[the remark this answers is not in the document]
dkrieger-pvtl: Why not?  os.Rename() calls the POSIX API, which is atomic.

hub/substeps_new.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	return nil
+}
+
+func Err(s *StepData) error {
+	return s.errChain
+}
+
+type StepData struct {
```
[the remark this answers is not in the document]
kalensk: I am running into trouble with an import cycle on the hub when doing this as this seems to pull in `substeps_new.go` for the methods on the struct.

hub/substeps_new.go
```diff
@@ -0,0 +1,99 @@
...
+		if err != nil {
+			s.errChain = xerrors.Errorf(`substep "%s": %w`, name, err)
+		}
+	}()
+
+	if s.errChain != nil {
+		return
+	}
+
+	store := HubSubStepStore{store: &subStepStore.Store{}, stream: s.multiStream.stream}
```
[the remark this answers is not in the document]
kalensk: It seemed useful to help with testing were we can pass in a fake store, but maybe it is not needed.

step/steps.go
```diff
@@ -0,0 +1,41 @@
+package step
+
+import (
+	"encoding/json"
+
+	"github.com/greenplum-db/gpupgrade/idl"
+)
+
+type SubStepData struct {
```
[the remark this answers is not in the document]
kalensk: Didn't realize we had that! We'll see if we can use that.

step/steps.go
```diff
@@ -0,0 +1,41 @@
...
+	SubStep idl.UpgradeSteps
+	Status  idl.StepStatus
+}
+
+// Use a custom marshaller to correctly serialize enum values. This enables the
+// enum string to be used instead of the numeric value for ease of reading.
+// This uses a pointer receiver to marshal pointers. Create a non-pointer
+// receiver to marshal non-pointers.
```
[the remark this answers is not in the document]
kalensk: I tried serializing step pointers using `func (s SubstepData) MarshalJSON()` and that method was never being called until I made it a pointer receiver (ie: `func (s *SubstepData) MarshalJSON()`

hub/substeps_new.go
berlin-ab: Ah, I didn't realize that Streaming was ours.

hub/substeps_new.go
berlin-ab: I'm talking about the level of abstraction and the difference between `store.MarkComplete` and the low-level formatting decisions made by `fmt.Fprintf`.

jchampio:
> @jchampio What code or design change are you suggesting by the dependency injection comment?

@dkrieger-pvtl It was meant to be a response to the ongoing DI thread. Put another way, I want to invert dependencies when it makes sense, but I'm not sure that particular location is a good place to start.

utils/sys_utils.go
dkrieger-pvtl: We want to place it in a different package from its usage for testing.  The best pattern we can think of to test, since the Do function must accept 0 arguments:

```go
   func ExportedFunc() {  
       once.Do(func() { //implementing code that accepts a parameter} ()
   }
```
is to separate this out the implementing code:

```go
   
   func ExportedFunc() {  
       once.Do(func() { packagePrivateFunc(parameter) } ()
   }

   func packagePrivateFunc(parameter type) {
       //implementing code that accepts a parameter
   }
```

and then we test `packagePrivateFunc`.

hub/substeps_new.go
```diff
@@ -0,0 +1,99 @@
...
+	subStep, err := store.ReadSubStep(name)
+	if err != nil {
+		return
+	}
+
+	// Only re-run tasks that have failed, pending, or were running.
```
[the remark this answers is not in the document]
dkrieger-pvtl: Yes, that's a good point.  The current implementation does that also.  We can add a check to fail-fast if we attempt to run a substep that's already running according to the store.

utils/sys_utils.go
kalensk: Right, just to emphasize that putting this in a separate package forces callers to not accidentally use the privateFunction needed for testing.

hub/substep_status.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/greenplum-db/gpupgrade/idl"
+	"github.com/greenplum-db/gpupgrade/step"
+)
+
+type SubStepsStore interface {
+	Read() (map[string]*step.SubStepData, error)
+	Write(step *step.SubStepData) (err error)
+}
+
+type HubSubStepStore struct {
```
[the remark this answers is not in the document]
kalensk: Addressed in a followup [squash commit](https://github.com/greenplum-db/gpupgrade/pull/176/commits/d53605a1e82a81c2ac0dcfa5874ad131b5435742).

hub/substep_status.go
```diff
@@ -0,0 +1,54 @@
+package hub
+
+import (
+	"github.com/greenplum-db/gpupgrade/idl"
+	"github.com/greenplum-db/gpupgrade/step"
+)
+
+type SubStepsStore interface {
```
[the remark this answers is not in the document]
kalensk: Addressed in a followup [squash commit](https://github.com/greenplum-db/gpupgrade/pull/176/commits/76f87257ad8cab6c0aad1f4f2fb6e61688637b31).

hub/substeps_new.go
```diff
@@ -0,0 +1,99 @@
...
+
+	return &StepData{
+		name:        name,
+		multiStream: newMultiplexedStream(stream, log),
+		log:         log,
+		dir:         stateDir,
+	}, nil
+}
+
+func FinishStep(s *StepData) error {
```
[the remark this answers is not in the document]
kalensk: Addressed in a followup [squash commit](https://github.com/greenplum-db/gpupgrade/pull/176/commits/e7b18d0ca3650aaefa085277f7c73d9b4ccc0dc9).

hub/substeps_new.go
```diff
@@ -0,0 +1,99 @@
...
+	if err := s.log.Close(); err != nil {
+		return xerrors.Errorf(`step "%s": %w`, s.name, err)
+	}
+
+	return nil
+}
+
+func Err(s *StepData) error {
```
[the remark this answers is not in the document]
kalensk: Addressed in a followup [squash commit](https://github.com/greenplum-db/gpupgrade/pull/176/commits/18e4b6fb7ed89210c0b63dfd087e8b7e7a68c3d0).

hub/substeps_new.go
```diff
@@ -0,0 +1,99 @@
...
+	return s.errChain
+}
+
+type StepData struct {
+	name        string
+	multiStream *multiplexedStream
+	log         io.WriteCloser
+	dir         string
+	errChain    error
```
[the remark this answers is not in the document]
kalensk: Addressed in a followup [squash commit](https://github.com/greenplum-db/gpupgrade/pull/176/commits/18e4b6fb7ed89210c0b63dfd087e8b7e7a68c3d0).

hub/substeps_new.go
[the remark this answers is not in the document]
kalensk: Addressed in a followup [squash commit](https://github.com/greenplum-db/gpupgrade/pull/176/commits/e619d0aa2175d0d387b0fb562277f0888c9959c8). Error message wording could probably use some team input to help finalize it.

subStepStore/substep_store.go
kalensk: Addressed in a followup [squash commit](https://github.com/greenplum-db/gpupgrade/pull/176/commits/2e0864626237ee891223ea32f1ff95490101c6c7).

subStepStore/substep_store.go
```diff
@@ -0,0 +1,71 @@
...
+	updatedStatusFile := statusFile + ".updated"
+	err = ioutil.WriteFile(updatedStatusFile, data, 0644)
+	if err != nil {
+		return err
+	}
+
+	err = os.Rename(updatedStatusFile, statusFile)
+	if err != nil {
+		rErr := os.Remove(updatedStatusFile)
+		err := multierror.Append(err, rErr)
```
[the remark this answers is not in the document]
kalensk: Addressed in a followup [squash commit](https://github.com/greenplum-db/gpupgrade/pull/176/commits/7403a772718d36d90e0f89ed90bb30a65810373a).

step/steps.go
berlin-ab: I'm thinking of an analogy to the repository pattern:

`Model -> Repository Interface <- Repository implementation -> Database table layout`

`Model -> Reader/Writer Interface <- Reader/Writer implementation -> Json Format`

If the on-disk layout needs to change, the reader/writer implementation has the freedom to change without effecting the rest of the system. If the model changes, the reader/writer implementation has the opportunity to react to the change gracefully without needing to change the data format. (for example, adding new fields, changing data types of fields).

I don't feel strongly that this change needs to happen, but that's what the principles would have us do.

utils/sys_utils.go
kalensk: Since close returns an error, I felt this was cleaner rather than having a defer and capturing the error. I'm open to putting the close in a defer if that's _really_ desired.

utils/sys_utils.go
jchampio: It sounds like you're trying to protect a package from misusing its own implementation details. Is that correct?

kalensk: Yes, since we only want it to be called once. We are separating the `once.Do` semantics from the function being called for testing, but don't want other things to call the function. Do you know of a better way of ensuring the function only gets called once and still test it?

subStepStore/substep_store.go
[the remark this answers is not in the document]
jchampio: OS-level atomicity doesn't guarantee durability (hopefully I'm getting the terminology right). You need to `fsync` at the proper locations at the very least. See the link.

step/steps.go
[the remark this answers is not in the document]
jchampio: The JSON unmarshalling rules basically feed back into themselves. When unmarshalling a pointer, if the thing you're unmarshalling into isn't allocated yet, it will be allocated and then the value will be unmarshalled using _its_ rules. So you shouldn't need an additional method.

step/steps.go
```diff
@@ -0,0 +1,41 @@
...
+		Name   string
+		Status string
+	}{}
+
+	if err := json.Unmarshal(data, &subStep); err != nil {
+		return err
+	}
+
+	s.SubStep = idl.UpgradeSteps(idl.UpgradeSteps_value[subStep.Name])
+	s.Status = idl.StepStatus(idl.StepStatus_value[subStep.Status])
```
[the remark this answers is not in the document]
jchampio: For the record, I'm wrong again. I keep forgetting that map lookup doesn't panic in Go, even though out-of-bounds slice access does.

subStepStore/substep_store.go
kalensk: Not needed since the state directory is already hidden.

hub/initialize.go
```diff
@@ -11,112 +11,80 @@ import (
...
 	"github.com/greenplum-db/gpupgrade/utils"
 )
 
 func (h *Hub) Initialize(in *idl.InitializeRequest, stream idl.CliToHub_InitializeServer) (err error) {
-	log, err := utils.System.OpenFile(
-		filepath.Join(utils.GetStateDir(), "initialize.log"),
-		os.O_WRONLY|os.O_CREATE,
-		0600,
-	)
+	s, err := BeginStep(h.conf.StateDir, "initialize", stream)
```
berlin-ab: It looks like we're getting really close to an abstraction that is meta-data map of the substeps to run, then we can iterate over that list and call s.Run(type, func)

hub/substeps_new.go
```diff
@@ -0,0 +1,63 @@
...
+		return nil, xerrors.Errorf("step %q: %w", name, err)
+	}
+
+	streams := newMultiplexedStream(sender, log)
+	return step.New(name, sender, step.NewFileStore(statusPath), streams), nil
+}
+
+// Returns path to status file, and if one does not exist it creates an empty
+// JSON file.
+func getStatusFile(stateDir string) (path string, err error) {
```
berlin-ab: This function seems like it belongs elsewhere. It is not a part of the high-level policy of BeginStep()

kalensk: So far it is only called here in `substeps_new.go`. We can consider moving it _if_ it ends up being called in more than one place.

berlin-ab:
Looks good. Ready to ship.

PR #177, opened 2020-01-16
Add --link flag in gpupgrade initialize to allow passing it to pg_upgrade

bhuvnesh2703:
The PR contains several commits, please refer to the descriptions for details of the commits.

Before we the commit which adds --link flag, we performed bunch of refactoring and consolidation of configuration.

### DIFF
--- a/agent/upgrade_primaries.go
+++ b/agent/upgrade_primaries.go
@@ -23,7 +23,7 @@ var execCommand = exec.Command
 func (s *Server) UpgradePrimaries(ctx context.Context, in *idl.UpgradePrimariesRequest) (*idl.UpgradePrimariesReply, error) {
 	gplog.Info("agent starting %s", upgradestatus.UPGRADE_PRIMARIES)
 
-	err := UpgradePrimary(in.OldBinDir, in.NewBinDir, in.DataDirPairs, s.conf.StateDir, in.CheckOnly)
+	err := UpgradePrimary(in.OldBinDir, in.NewBinDir, in.DataDirPairs, s.conf.StateDir, in.CheckOnly, in.UseLinkMode)
 	return &idl.UpgradePrimariesReply{}, err
 }
 
@@ -33,7 +33,7 @@ type Segment struct {
 	WorkDir string // the pg_upgrade working directory, where logs are stored
 }
 
-func UpgradePrimary(sourceBinDir string, targetBinDir string, dataDirPairs []*idl.DataDirPair, stateDir string, checkOnly bool) error {
+func UpgradePrimary(sourceBinDir string, targetBinDir string, dataDirPairs []*idl.DataDirPair, stateDir string, checkOnly bool, useLinkMode bool) error {
 	segments := make([]Segment, 0, len(dataDirPairs))
 
 	for _, dataPair := range dataDirPairs {
@@ -49,7 +49,7 @@ func UpgradePrimary(sourceBinDir string, targetBinDir string, dataDirPairs []*id
 		})
 	}
 
-	err := UpgradeSegments(sourceBinDir, targetBinDir, segments, checkOnly)
+	err := UpgradeSegments(sourceBinDir, targetBinDir, segments, checkOnly, useLinkMode)
 	if err != nil {
 		return errors.Wrap(err, "failed to upgrade segments")
 	}
@@ -57,7 +57,7 @@ func UpgradePrimary(sourceBinDir string, targetBinDir string, dataDirPairs []*id
 	return nil
 }
 
-func UpgradeSegments(sourceBinDir string, targetBinDir string, segments []Segment, checkOnly bool) (err error) {
+func UpgradeSegments(sourceBinDir string, targetBinDir string, segments []Segment, checkOnly bool, useLinkMode bool) (err error) {
 	host, err := os.Hostname()
 	if err != nil {
 		return err
@@ -80,6 +80,10 @@ func UpgradeSegments(sourceBinDir string, targetBinDir string, segments []Segmen
 		}
 		if checkOnly {
 			options = append(options, upgrade.WithCheckOnly())
+		}
+
+		if useLinkMode {
+			options = append(options, upgrade.WithLinkMode())
 		}
 
 		content := segment.Content
--- a/agent/upgrade_primaries_test.go
+++ b/agent/upgrade_primaries_test.go
@@ -81,7 +81,7 @@ func TestUpgradePrimary(t *testing.T) {
 		execCommand = exectest.NewCommand(FailedMain)
 		defer func() { execCommand = nil }()
 
-		err := UpgradePrimary("/old/bin", "/new/bin", pairs, tempDir, true)
+		err := UpgradePrimary("/old/bin", "/new/bin", pairs, tempDir, true, false)
 		if err == nil {
 			t.Fatal("UpgradeSegments() returned no error")
 		}
@@ -100,7 +100,7 @@ func TestUpgradePrimary(t *testing.T) {
 		execCommand = exectest.NewCommand(FailedMain)
 		defer func() { execCommand = nil }()
 
-		err := UpgradePrimary("/old/bin", "/new/bin", pairs, tempDir, false)
+		err := UpgradePrimary("/old/bin", "/new/bin", pairs, tempDir, false, false)
 		if err == nil {
 			t.Fatal("UpgradeSegments() returned no error")
 		}
--- a/cli/commanders/initialize.go
+++ b/cli/commanders/initialize.go
@@ -4,14 +4,11 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
-	"path"
 	"path/filepath"
 
-	"github.com/pkg/errors"
-
-	"github.com/greenplum-db/gp-common-go-libs/cluster"
 	"github.com/greenplum-db/gp-common-go-libs/gplog"
 
+	"github.com/greenplum-db/gpupgrade/hub"
 	"github.com/greenplum-db/gpupgrade/utils"
 )
 
@@ -39,43 +36,26 @@ func CreateStateDir() (err error) {
 	return nil
 }
 
-func CreateInitialClusterConfigs(sourceBinDir, targetBinDir string) (err error) {
+func CreateInitialClusterConfigs() (err error) {
 	s := Substep("Creating initial cluster config files...")
 	defer s.Finish(&err)
 
-	stateDir := utils.GetStateDir()
-	emptyCluster := cluster.NewCluster([]cluster.SegConfig{})
+	// if empty json configuration file exists, skip recreating it
+	filename := filepath.Join(utils.GetStateDir(), hub.ConfigFileName)
+	_, err = os.Stat(filename)
 
-	source := &utils.Cluster{
-		Cluster:    emptyCluster,
-		BinDir:     path.Clean(sourceBinDir),
-		ConfigPath: filepath.Join(stateDir, utils.SOURCE_CONFIG_FILENAME),
-	}
-
-	target := &utils.Cluster{
-		Cluster:    emptyCluster,
-		BinDir:     path.Clean(targetBinDir),
-		ConfigPath: filepath.Join(stateDir, utils.TARGET_CONFIG_FILENAME),
-	}
-
-	if source.Load() == nil && target.Load() == nil {
-		gplog.Debug("Initial cluster config files(%s or %s) already present...skipping.", source.ConfigPath, target.ConfigPath)
+	if err == nil {
+		gplog.Debug("Initial cluster configuration file %s already present...skipping", filename)
 		return nil
 	}
 
-	// Create empty clusters in source and target so that gpupgrade hub can
-	// start without having replaced them with current values.
-	// TODO: implement a slicker scheme to allow this.
-	err = source.Commit()
+	file, err := os.Create(filename)
 	if err != nil {
-		return errors.Wrap(err, "Unable to save empty source cluster configuration")
+		return err
 	}
+	defer file.Close()
 
-	err = target.Commit()
-	if err != nil {
-		return errors.Wrap(err, "Unable to save empty target cluster configuration")
-	}
-
+	fmt.Fprintf(file, "{}") // the hub will fill this in during initialization
 	return nil
 }
 
--- a/cli/commanders/initialize.go
+++ b/cli/commanders/initialize.go
@@ -44,9 +44,17 @@ func CreateInitialClusterConfigs() (err error) {
 	filename := filepath.Join(utils.GetStateDir(), hub.ConfigFileName)
 	_, err = os.Stat(filename)
 
-	if err == nil {
+	// if the file exists, there will be no error or if there is an error it might
+	// also indicate that the file exists, in either case don't overwrite the file
+	if err == nil || os.IsExist(err) {
 		gplog.Debug("Initial cluster configuration file %s already present...skipping", filename)
 		return nil
+	}
+
+	// if the err is anything other than file does not exist, error out
+	if !os.IsNotExist(err) {
+		gplog.Debug("Check to find presence of initial cluster configuration file %s failed", filename)
+		return err
 	}
 
 	file, err := os.Create(filename)
--- a/hub/commands.go
+++ b/hub/commands.go
@@ -6,10 +6,9 @@ import (
 	"path/filepath"
 	"runtime/debug"
 
-	"github.com/pkg/errors"
-
 	"github.com/greenplum-db/gp-common-go-libs/gplog"
 	"github.com/spf13/cobra"
+	"golang.org/x/xerrors"
 	"google.golang.org/grpc"
 
 	"github.com/greenplum-db/gpupgrade/hub/upgradestatus"
@@ -18,6 +17,8 @@ import (
 	"github.com/greenplum-db/gpupgrade/utils/daemon"
 	"github.com/greenplum-db/gpupgrade/utils/log"
 )
+
+const ConfigFileName = "config"
 
 // This directory to have the implementation code for the gRPC server to serve
 // Minimal CLI command parsing to embrace that booting this binary to run the hub might have some flags like a log dir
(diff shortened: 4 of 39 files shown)


jchampio:
> TODO talk about PersistedConfig

This commit message needs to be fleshed out 😄

agent/upgrade_primaries.go
```diff
@@ -23,7 +23,7 @@ var execCommand = exec.Command
 func (s *Server) UpgradePrimaries(ctx context.Context, in *idl.UpgradePrimariesRequest) (*idl.UpgradePrimariesReply, error) {
 	gplog.Info("agent starting %s", upgradestatus.UPGRADE_PRIMARIES)
 
-	err := UpgradePrimary(in.OldBinDir, in.NewBinDir, in.DataDirPairs, s.conf.StateDir, in.CheckOnly)
+	err := UpgradePrimary(in.OldBinDir, in.NewBinDir, in.DataDirPairs, s.conf.StateDir, in.CheckOnly, in.UseLinkMode)
```
berlin-ab: Something to consider for the future: 

We should adopt the option-style arguments, instead of continuing to add parameters here.

jchampio: +1, or possibly allow the `idl.UpgradePrimariesRequest` to leak down a layer, since experience seems to show that they are tightly coupled anyway.

upgrade/run.go
```diff
@@ -118,6 +123,13 @@ func WithCheckOnly() Option {
 	}
 }
 
+// WithLinkMode allows pg_upgrade to run upgrade with/without --link mode
+func WithLinkMode(value bool) Option {
```
jchampio: nit: can we match the `WithCheckOnly()` pattern above, where we would simply not pass the `WithLinkMode()` option if we don't want it? Or at least discuss why we might want them to differ. My opinion is that passing `WithLinkMode(false)` doesn't make much sense; that implies that the option isn't optional.

agent/upgrade_primaries_test.go
```diff
@@ -96,11 +96,27 @@ func TestUpgradePrimary(t *testing.T) {
 		}
 	})
 
+	t.Run("when pg_upgrade with link mode fails it returns an error", func(t *testing.T) {
```
jchampio: Does this add any coverage? If not, let's not add a test here; I think the interesting testable components are a level below this.

cli/commanders/initialize.go
```diff
@@ -39,43 +36,26 @@ func CreateStateDir() (err error) {
...
-		ConfigPath: filepath.Join(stateDir, utils.TARGET_CONFIG_FILENAME),
-	}
+	// if empty json configuration file exists, skip recreating it
+	filename := filepath.Join(utils.GetStateDir(), hub.ConfigFileName)
+	_, err = os.Stat(filename)
 
-	if source.Load() == nil && target.Load() == nil {
-		gplog.Debug("Initial cluster config files(%s or %s) already present...skipping.", source.ConfigPath, target.ConfigPath)
+	if err == nil {
+		gplog.Debug("Initial cluster configuration file %s already present...skipping", filename)
```
kalensk: Don't we need to differentiate between the file not existing, and other types of errors from `os.Stat`? Shouldn't we be using [`os.isExist`](https://golang.org/pkg/os/#IsExist)?

Not for this PR, but we really need to fix `utils.GetStateDir()`. That is, we have two source of truth `hub.conf.stateDir` and `utils.GetStateDir()`. We also need to ensure that only one state directory gets created if it does not exist and works across process restarts. This is related to PR https://github.com/greenplum-db/gpupgrade/pull/175

hub/hub.go
```diff
@@ -359,3 +354,51 @@ func (h *Hub) closeAgentConns() {
...
+	// happens if Config.Save() panics: we'll have truncated the file
+	// on disk and the hub will be unable to recover. For now, since we normally
+	// only save the configuration during initialize and any configuration
+	// errors could be fixed by reinitializing, the risk seems small.
+	file, err := utils.System.Create(filepath.Join(h.StateDir, ConfigFileName))
+	if err != nil {
+		return err
+	}
+	defer func() {
+		if cerr := file.Close(); cerr != nil {
```
kalensk: Maybe I can add a chore for the following, but it seems like this should be extracted into this common pattern we have where we check if we need a file or directory exists, if not create it. And ensure all callers call this one function.

It seems like we have this requirement for the state directory, status.json, config, etc.

hub/commands.go
```diff
@@ -19,6 +18,8 @@ import (
 	"github.com/greenplum-db/gpupgrade/utils/log"
 )
 
+const ConfigFileName = "config"
```
kalensk: nit: I think it would make a bit more sense to call this `config.json`.

hub/hub_internal_test.go
```diff
@@ -0,0 +1,43 @@
...
+	"reflect"
+	"testing"
+
+	"github.com/greenplum-db/gp-common-go-libs/cluster"
+
+	"github.com/greenplum-db/gpupgrade/testutils"
+)
+
+func TestConfig(t *testing.T) {
+	t.Run("saves itself to the provided stream", func(t *testing.T) {
```
kalensk: This test seems a tad off to me. Wouldn't it make more sense to have a test that "loads the saved config" which tests loading and saving?

_If_ we do keep this test the description doesn't seem correct to me. Do we mean to say something like "saves the same config"?

hub/hub_test.go
```diff
@@ -298,3 +266,90 @@ var _ = Describe("Hub", func() {
...
+		}()
+
+		// Write the hub's configuration to the pipe.
+		if err := h.SaveConfig(); err != nil {
+			t.Errorf("SaveConfig() returned error %+v", err)
+		}
+
+		// Reload the configuration from the read side of the pipe and ensure the
+		// contents are the same.
+		result := new(hub.Config)
```
kalensk: super minor nit: Rename result to actual, to follow the expected/actual pattern.

hub/hub_test.go
```diff
@@ -298,3 +266,90 @@ var _ = Describe("Hub", func() {
...
+		defer func() {
+			utils.System = utils.InitializeSystemFunctions()
+		}()
+
+		err := h.SaveConfig()
+
+		var merr *multierror.Error
+		if !xerrors.As(err, &merr) {
+			t.Fatalf("returned %#v, want error type %T", err, merr)
+		}
```
kalensk: Is this check needed given the assertion for `os.ErrInvalid` below?

hub/hub.go
```diff
@@ -359,3 +354,51 @@ func (h *Hub) closeAgentConns() {
...
+}
+
+func (c *Config) Save(w io.Writer) error {
+	enc := json.NewEncoder(w)
+	enc.SetIndent("", "  ")
+	return enc.Encode(c)
+}
+
+// SaveConfig persists the hub's configuration to disk.
+func (h *Hub) SaveConfig() (err error) {
```
kalensk: Not sure if this is for this PR or another one. But it is not super obvious what the difference is between `Save` and `SaveConfig`. Is there a way to clean this up or make it more clear? My naive desire is to want the external API for callers to be `read` and `write` the config, and I don't easily see how `saveConfig` fits into that desire.

testutils/insert_target_config/main.go
```diff
@@ -10,11 +10,13 @@
 package main
 
 import (
+	"io"
```
kalensk: What was the reason why `dump_config` got renamed to `insert_target_config`? Why wouldn't we keep it generic and have it dump the config from any cluster we point it to?

testutils/insert_target_config/main.go
```diff
@@ -25,14 +27,36 @@ func main() {
...
+	}
 
-	cluster, err := utils.ClusterFromDB(conn, binDir, configPath)
+	err = file.Truncate(0)
 	if err != nil {
 		log.Fatal(err)
 	}
 
-	err = cluster.Commit()
+	err = config.Save(file)
```
kalensk: Can you explain these changes with the `io.SeekStart` and `file.Truncate`? If needed, maybe a comment perhaps?

upgrade/run.go
kalensk: I believe I would like to see this change before merging. (However, I could probably be convinced otherwise.)

utils/cluster.go
```diff
@@ -124,6 +78,62 @@ func (c *Cluster) PrimaryHostnames() []string {
 	return list
 }
 
+// serializableCluster contains all of the members of utils.Cluster that can be
+// serialized to disk.
+//
+// Ideally, utils.Cluster would be serializable itself, but unfortunately the
+// Executor member cannot be put through a JSON marshal/unmarshal round trip. We
+// exclude it here.
```
kalensk: Can't we use the [tag `json:"-"`](https://golang.org/pkg/encoding/json/) to emit the executor and other fields? Would this remove the need for `serializableCluster`?

utils/cluster.go
```diff
@@ -124,6 +78,62 @@ func (c *Cluster) PrimaryHostnames() []string {
...
+type serializableCluster struct {
```
kalensk: Can't we use the [tag `json:"-"`](https://golang.org/pkg/encoding/json/) to emit the executor and other fields? Would this remove the need for `serializableCluster`?

kalensk:
@bhuvnesh2703 
Do we have a full test pipeline running? I didn't see one.

utils/cluster.go
jchampio: We could if we controlled `cluster.Cluster`, but it's part of the common-libs.

testutils/insert_target_config/main.go
jchampio: Because it doesn't dump a config anymore. It inserts the target configuration into the overall `config` file.

hub/hub.go
jchampio: `Config` and its methods are the easily unit-tested API. `hub.SaveConfig` is the integration that uses that API to actually store the configuration to disk.

hub/commands.go
jchampio: Does the extension give you anything? This is hub-internal.

hub/hub_test.go
jchampio: Yes. We wouldn't have `merr.Errors` to range over if we didn't pull out a `multierror.Error`.

testutils/insert_target_config/main.go
kalensk: Is the comment at the top of the file still accurate?

```
// The dump_config utility dumps the configuration of a running GPDB
// cluster into the specified <configPath> file.
// .... etc
```

hub/commands.go
kalensk: I believe it quickly helps my text editor syntax highlight the file. And it makes it consistent with `status.json`. I agree its a minor thing.

hub/hub_test.go
jchampio: To be more specific, a `multierror.Error` that contains `os.ErrInvalid` is not itself an instance of `os.ErrInvalid`. We have to unpack it to check the contents.

hub/hub_internal_test.go
jchampio: Both are correct. This tests the round-trip.

hub/hub.go
kalensk: Thanks for the explanation, I wonder how we can make that more apparent in the code.

utils/cluster.go
kalensk: Can we add a chore to see if we can make this change in common-libs and update our code?

hub/commands.go
jchampio: > I believe it quickly helps my text editor syntax highlight the file.

Ah. Yeah, I think that's probably valuable.

hub/hub_internal_test.go
kalensk: Can you clarify the phrase "provided stream"? I guess I don't understand that part.

testutils/insert_target_config/main.go
jchampio: Ha, no. Good catch.

utils/cluster.go
jchampio: It seems... unfair... to file that when we're trying so hard to get rid of our dependency on common-libs anyway. But we definitely could. @jmcatamney thoughts?

kalensk: Oh, if we can get rid of the dependency of common-libs even better.

hub/hub_internal_test.go
jchampio: "stream" refers to the `io.Writer`/`Reader` interfaces. They're similar to the Java `InputStream`/`OutputStream`, or C++ `std::istream`/`ostream`.

Basically, when trying to invert a dependency on something where we're writing or reading data (from file, network, etc.), you most likely want to inject a stream via the `io` package interfaces.

cli/commanders/initialize.go
jchampio: > Don't we need to differentiate between the file not existing, and other types of errors from os.Stat? Shouldn't we be using os.isExist?

Yes, good catch.

testutils/insert_target_config/main.go
jchampio: Yeah, I think some roadmap comments would be good.

The seek/truncate operations let us modify the current file in-place, without having to close/remove/recreate/write a new file.

agent/upgrade_primaries.go
berlin-ab: In this case, what do you think about bringing `UpgradePrimariesRequest` into the agent package? If I understand its purpose correctly, its reasons for change are going to be caused by a change in the agent domain logic.

(again, only something to consider in the future)

hub/hub_test.go
```diff
@@ -298,3 +266,90 @@ var _ = Describe("Hub", func() {
...
+func TestHubSaveConfig(t *testing.T) {
+	source, target := testutils.CreateMultinodeSampleClusterPair("/tmp")
+	source.Executor = new(cluster.GPDBExecutor)
+	target.Executor = new(cluster.GPDBExecutor)
+	useLinkMode := false
+	conf := &hub.Config{source, target, 12345, 54321, useLinkMode}
+
+	h := hub.New(conf, nil, "", nil)
+
+	t.Run("saves correct contents to disk", func(t *testing.T) {
```
berlin-ab: nit: this title should call out what "correct" means.

hub/hub_test.go
```diff
@@ -298,3 +266,90 @@ var _ = Describe("Hub", func() {
...
+		if err := result.Load(read); err != nil {
+			t.Errorf("loading configuration results: %+v", err)
+		}
+
+		if !reflect.DeepEqual(h.Config, result) {
+			t.Errorf("wrote config %#v, want %#v", result, h.Config)
+		}
+	})
+
+	t.Run("correctly bubbles up file creation errors", func(t *testing.T) {
```
berlin-ab: nit: could probably drop "correctly" from each of these tests.

agent/upgrade_primaries.go
jchampio: `idl.UpgradePrimariesRequest` is machine-generated. The method of changing it will remain the same whether we call the package `idl` or `agent`.

upgrade/run.go
bhuvnesh2703: fixed..will use WithLinkMode()

agent/upgrade_primaries_test.go
bhuvnesh2703: thinking it again, i think the failure is capture irrespective of the link mode, so we can remove this test. removed.

cli/commanders/initialize.go
bhuvnesh2703: in case a file exists and there are no errors, os.IsExist(err) will return false, so if there is no error or if there is an error but it indicates files are existing, we can continue else we should fail, changed the code to reflect that.

```
	if err == nil || os.IsExist(err) {
		gplog.Debug("Initial cluster configuration file %s already present...skipping", filename)
		return nil
	}

	if err != nil {
		gplog.Debug("Check to find presence of initial cluster configuration file %s failed", filename)
		return err
	}
```

hub/hub.go
bhuvnesh2703: thanks..

hub/commands.go
bhuvnesh2703: added json suffix

hub/hub_internal_test.go
bhuvnesh2703: added a comment : "stream" refers to the io.Writer/Reader interfaces. "

hub/hub_test.go
bhuvnesh2703: renamed

hub/hub_test.go
bhuvnesh2703: added comment

testutils/insert_target_config/main.go
bhuvnesh2703: fixed the comment..

testutils/insert_target_config/main.go
bhuvnesh2703: added comments..

hub/hub_test.go
bhuvnesh2703: updated the description

hub/hub_test.go
bhuvnesh2703: fixed

hub/hub.go
bhuvnesh2703: added comments to explain

cli/commanders/initialize.go
bhuvnesh2703: in case a file exists and there are no errors, os.IsExist(err) will return false, so if there is no error or if there is an error but it indicates files are existing, we can continue else we should fail when the error is anything other than file not existing, changed the code to reflect that.
	// if the file exists, there will be no error or if there is an error it might
	// also indicate that the file exists, in either case don't overwrite the file
	if err == nil || os.IsExist(err) {
		gplog.Debug("Initial cluster configuration file %s already present...skipping", filename)
		return nil
	}

	// if the err is anything other than file does not exist, error out
	if !os.IsNotExist(err) {
		gplog.Debug("Check to find presence of initial cluster configuration file %s failed", filename)
		return err
	}
```

bhuvnesh2703:
Applied Feedback, please review.

cli/commanders/initialize.go
jchampio: I have a feeling the second part of that conditional (`|| os.IsExist(err)`) will never be hit in practice. Can we add a chore to make this more of an [EAFP implementation](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use#Preventing_TOCTOU)? I don't think it needs to be made perfect in this PR.

bhuvnesh2703: sure will do.

PR #179, opened 2020-01-22
initialize: add basic test for substep idempotence

jchampio:
This (draft) PR tests a very basic property of each `initialize` substep: if a substep completes its work and then crashes immediately after, can it be restarted and still succeed? To do this, we manually mark every completed substep as `FAILED` in the status file and re-run. It doesn't work yet, unsurprisingly, as several substeps don't have this property yet.

The first two commits are prefactors that made writing/debugging the new test easier.

### DIFF
--- a/cli/commands/commands.go
+++ b/cli/commands/commands.go
@@ -214,6 +214,7 @@ func createConfigShowSubcommand() *cobra.Command {
 
 	subShow.Flags().Bool("old-bindir", false, "show install directory for old gpdb version")
 	subShow.Flags().Bool("new-bindir", false, "show install directory for new gpdb version")
+	subShow.Flags().Bool("new-datadir", false, "show temporary data directory for new gpdb cluster")
 
 	return subShow
 }
--- a/hub/config.go
+++ b/hub/config.go
@@ -38,6 +38,8 @@ func (h *Hub) GetConfig(ctx context.Context, in *idl.GetConfigRequest) (*idl.Get
 		resp.Value = h.Source.BinDir
 	case "new-bindir":
 		resp.Value = h.Target.BinDir
+	case "new-datadir":
+		resp.Value = h.Target.MasterDataDir()
 	default:
 		return nil, status.Errorf(codes.NotFound, "%s is not a valid configuration key", in.Name)
 	}
--- a/test/config.bats
+++ b/test/config.bats
@@ -62,7 +62,8 @@ teardown() {
     run gpupgrade config show
     [ "$status" -eq 0 ]
     [ "${lines[0]}" = "new-bindir - /my/new/bin/dir" ]
-    [ "${lines[1]}" = "old-bindir - /my/old/bin/dir" ]
+    [ "${lines[1]}" = "new-datadir - " ] # This isn't populated until cluster creation, but it's still displayed here
+    [ "${lines[2]}" = "old-bindir - /my/old/bin/dir" ]
 }
 
 @test "multiple configuration values can be set at once" {
@@ -71,5 +72,5 @@ teardown() {
     run gpupgrade config show
     [ "$status" -eq 0 ]
     [ "${lines[0]}" = "new-bindir - /my/new/bin/dir" ]
-    [ "${lines[1]}" = "old-bindir - /my/old/bin/dir" ]
+    [ "${lines[2]}" = "old-bindir - /my/old/bin/dir" ]
 }
--- a/test/gpinitsystem.bats
+++ b/test/gpinitsystem.bats
@@ -27,6 +27,25 @@ teardown() {
     fi
 }
 
+# Takes an old datadir and echoes the expected new datadir path.
+#
+# NOTE for devs: this is just for getting the expected data directories, which
+# is an implementation detail. If you want the actual location of the new master
+# data directory after an initialization, you can just ask the hub with
+#
+#    gpupgrade config show --new-datadir
+#
+expected_datadir() {
+    local base="$(basename $1)"
+    local dir="$(dirname $1)_upgrade"
+
+    # Sanity check.
+    [ -n "$base" ]
+    [ -n "$dir" ]
+
+    echo "$dir/$base"
+}
+
 @test "initialize runs gpinitsystem based on the source cluster" {
     # Store the data directories for each source segment by port.
     run $PSQL -AtF$'\t' -p $PGPORT postgres -c "select port, datadir from gp_segment_configuration where role = 'p'"
@@ -39,7 +58,6 @@ teardown() {
 
     local masterdir="${olddirs[$PGPORT]}"
     local newport=50432
-    local newmasterdir="$(upgrade_datadir $masterdir)"
 
     gpupgrade initialize \
         --verbose \
@@ -49,7 +67,11 @@ teardown() {
         --disk-free-ratio 0 3>&-
 
     # Make sure we clean up during teardown().
-    NEW_CLUSTER="$newmasterdir"
+    local newmasterdir="$(gpupgrade config show --new-datadir)"
+    NEW_CLUSTER="${newmasterdir}"
+
+    # Sanity check the newly created master's location.
+    [ "$newmasterdir" = $(expected_datadir "$masterdir") ]
 
     PGPORT=$newport gpstart -a -d "$newmasterdir"
 
@@ -70,16 +92,13 @@ teardown() {
         (( newport++ ))
 
         [ -n "$newdir" ] || fail "could not find upgraded segment on expected port $newport"
-        [ "$newdir" = $(upgrade_datadir "$olddir") ]
+        [ "$newdir" = $(expected_datadir "$olddir") ]
     done
 }
 
 @test "initialize accepts a port range" {
     local expected_ports="15432,15433,15434,15435"
     local newport=15432
-
-    local masterdir="$($PSQL -At postgres -c "select datadir from gp_segment_configuration where content = -1 and role = 'p'")"
-    local newmasterdir="$(upgrade_datadir $masterdir)"
 
     gpupgrade initialize \
         --ports $expected_ports \
@@ -90,7 +109,8 @@ teardown() {
         --disk-free-ratio 0 3>&-
 
     # Make sure we clean up during teardown().
-    NEW_CLUSTER="$newmasterdir"
+    local newmasterdir="$(gpupgrade config show --new-datadir)"
+    NEW_CLUSTER="${newmasterdir}"
 
     PGPORT=$newport gpstart -a -d "$newmasterdir"
 
--- a/test/helpers.bash
+++ b/test/helpers.bash
@@ -62,18 +62,6 @@ delete_target_datadirs() {
     rm -rf "${datadir}"/*_upgrade
 }
 
-# Takes an old datadir and echoes the expected new datadir path.
-upgrade_datadir() {
-    local base="$(basename $1)"
-    local dir="$(dirname $1)_upgrade"
-
-    # Sanity check.
-    [ -n "$base" ]
-    [ -n "$dir" ]
-
-    echo "$dir/$base"
-}
-
 # require_gnu_stat tries to find a GNU stat program. If one is found, it will be
 # assigned to the STAT global variable; otherwise the current test is skipped.
 require_gnu_stat() {
--- a/test/initialize.bats
+++ b/test/initialize.bats
@@ -41,8 +41,7 @@ teardown() {
 }
 
 set_target_cluster_var_for_teardown() {
-    local newmasterdir="$(upgrade_datadir $MASTER_DATA_DIRECTORY)"
-    TARGET_CLUSTER="${newmasterdir}"
+    TARGET_CLUSTER="$(gpupgrade config show --new-datadir)"
 }
 
 teardown_target_cluster() {
@@ -247,14 +246,14 @@ wait_for_port_change() {
 }
 
 @test "the check_upgrade substep always runs" {
-    set_target_cluster_var_for_teardown
-    TEARDOWN_FUNCTIONS+=( teardown_target_cluster )
-
     gpupgrade initialize \
         --old-bindir="$GPHOME/bin" \
         --new-bindir="$GPHOME/bin" \
         --old-port="${PGPORT}" \
         --disk-free-ratio 0 3>&-
+
+    set_target_cluster_var_for_teardown
+    TEARDOWN_FUNCTIONS+=( teardown_target_cluster )
 
     setup_check_upgrade_to_fail
     TEARDOWN_FUNCTIONS+=( teardown_check_upgrade_failure )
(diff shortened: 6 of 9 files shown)


test/helpers.bash
```diff
@@ -24,6 +24,7 @@ abort() {
 skip_if_no_gpdb() {
     [ -n "${GPHOME}" ] || skip "this test requires an active GPDB cluster (set GPHOME)"
     [ -n "${PGPORT}" ] || skip "this test requires an active GPDB cluster (set PGPORT)"
+    pg_isready -q || skip "this test requires an active GPDB cluster (run gpstart)"
```
dkrieger-pvtl: Note that `pg_isready` is only available in GPDB6 and later(it is not in GPDB-5).  We do plan to have the bats tests run on 5 eventually.   I think using instead:

`psql -d postgres -c "select version()"`

is better since this works on 5 and 6.  That also makes sure there is a cluster up and accepting connections.

jchampio: We don't support 5->5 upgrades, so a version of `pg_isready` is always guaranteed to exist.

dkrieger-pvtl: Yes, but then don't you want to specify the target cluster's path?  `GPHOME` for a 5->6 upgrade will point to the GPDB-5 cluster's binary directory, and then `pg_isready` won't be found.

hub/config.go
```diff
@@ -36,6 +36,8 @@ func (h *Hub) GetConfig(ctx context.Context, in *idl.GetConfigRequest) (*idl.Get
 		resp.Value = h.Source.BinDir
 	case "new-bindir":
 		resp.Value = h.Target.BinDir
+	case "new-datadir":
```
berlin-ab: nit: probably an aside, but it seems strange that we request for a config value one at a time.

test/helpers.bash
jchampio: > GPHOME for a 5->6 upgrade will point to the GPDB-5 cluster's binary directory

You're gonna have to specify the target cluster's `GPHOME` somehow when testing. We can solve this in exactly the same way when we implement that, right?

hub/config.go
jchampio: Not sure what you mean by this.

berlin-ab: Can the user specify multiple options at once when requesting configuration values?

cli/commands/commands.go
```diff
@@ -217,6 +217,7 @@ func createConfigShowSubcommand() *cobra.Command {
 
 	subShow.Flags().Bool("old-bindir", false, "show install directory for old gpdb version")
 	subShow.Flags().Bool("new-bindir", false, "show install directory for new gpdb version")
+	subShow.Flags().Bool("new-datadir", false, "show temporary data directory for new gpdb cluster")
```
berlin-ab: nit: seems like there's some duplication of knowledge about parameter names. There's likely some PoGo modeling we could do to reduce some of that.

hub/config.go
jchampio: Yes.

kalensk:
Can you explain a bit more about the need for `gpupgrade config show --new-datadir`? I thought we could derive this information since it gets created in a well known location?

jchampio:
> Can you explain a bit more about the need for gpupgrade config show --new-datadir? I thought we could derive this information since it gets created in a well known location?

We can, but different tests are doing it in different ways. This takes the guesswork out and also decouples most tests from the implementation detail. (There will be one remaining test that duplicates the logic, for comparing expected segment data directories.)

kalensk:
> We can, but different tests are doing it in different ways. This takes the guesswork out and also decouples most tests from the implementation detail. (There will be one remaining test that duplicates the logic, for comparing expected segment data directories.)

Thanks for the info. I like it!

test/helpers.bash
dkrieger-pvtl: Sure.  Why do you want to use `pg_isready` instead of a `psql` query to determine if the database is available?

kalensk: > Sure. Why do you want to use `pg_isready` instead of a `psql` query to determine if the database is available?

Its a short single command that is self-documenting and gives a proper exit status to allow us to chain with the skip?

jchampio: > Its a short single command that is self-documenting and gives a proper exit status to allow us to chain with the skip?

Yep. It does What We Want.

kalensk:
A friendly reminder to remove the test "start agents idempotent" in [initialize.bats](https://github.com/greenplum-db/gpupgrade/blob/c7d2cc2a91f479d54538fd89bdadab557bab2044/test/initialize.bats#L179). As this PR subsumes that test.

jchampio:
@kalensk Yep, see the rebased PR contents.

test/initialize.bats
```diff
@@ -270,3 +248,30 @@ wait_for_port_change() {
     # for check_upgrade, we always run it.
     [ "$status" -eq 1 ] || fail "$output"
 }
+
+# This is a very simple way to flush out the most obvious idempotence bugs. It
+# replicates what would happen if every substep failed/crashed right after
+# completing its work but before completion was signalled back to the hub.
+@test "all substeps can be re-run after completion" {
```
berlin-ab: suggestion: I like what this test is driving out. Would it make more sense at the Gotest level as it depends on implementation details of the status.json file?

test/initialize.bats
```diff
@@ -248,15 +226,15 @@ wait_for_port_change() {
...
-    set_target_cluster_var_for_teardown
-    TEARDOWN_FUNCTIONS+=( teardown_target_cluster )
-
     gpupgrade initialize \
         --old-bindir="$GPHOME/bin" \
         --new-bindir="$GPHOME/bin" \
         --old-port="${PGPORT}" \
         --disk-free-ratio 0 3>&-
 
+    set_target_cluster_var_for_teardown
```
berlin-ab: suggestion: would it make sense to move the teardown function setting into the function that creates its need?

test/initialize.bats
jchampio: > Would it make more sense at the Gotest level

I don't think so. I think that'd be a very costly approach, requiring us to spin up the world in an integration environment.

hub/config.go
berlin-ab: was thinking about `cli/commands/commands.go:200`, and how it seems unnecessary to make N+1 over the network queries for configuration when we already have the full list. No need to act, just an observation.

jchampio: Gotcha. Yeah, seemed like we could defer a more efficient approach to cases in which N is regularly greater than 1. I'd say that's paid off so far. 😄

test/initialize.bats
```diff
@@ -270,3 +248,30 @@ wait_for_port_change() {
...
+        --old-port="${PGPORT}"\
+        --disk-free-ratio 0 3>&-
+
+    set_target_cluster_var_for_teardown
+    TEARDOWN_FUNCTIONS+=( teardown_target_cluster )
+
+    # Mark every substep in the status file as failed. Then restart the cluster
+    # and re-initialize.
+    sed -i .bak -e 's/"COMPLETE"/"FAILED"/g' "$GPUPGRADE_HOME/status.json"
+    gpstart -a
```
dkrieger-pvtl: Note that, even though we've decided to defer the development required to make STOP_SOURCE_CLUSTER idempotent, the `gpstart -a` here essentially resets the state of the cluster as far as that substep is concerned, so it works upon re-run nonetheless.

jchampio: Regardless of the stop-cluster implementation, the substep that does the configuration gathering currently requires the cluster to be running.

jchampio:
Test pipeline hung during `install` tests, so I'm waiting for that to go green to make sure I didn't introduce a regression.

jchampio:
Looks like there _is_ a regression that causes a hang. I'll need to debug.

test/gpinitsystem.bats
```diff
@@ -29,6 +29,25 @@ teardown() {
...
+    local base="$(basename $1)"
+    local dir="$(dirname $1)_upgrade"
+
+    # Sanity check.
+    [ -n "$base" ]
+    [ -n "$dir" ]
+
+    echo "$dir/$base"
+}
+
```
kalensk: I might be missing something, but can we remove this function in favor of `gpupgrade config show --new-datadir`? 

Seems like there is only one usage left, could this be inlined as this function appears to be old code in some sense?

jchampio: From the comment:
```
# NOTE for devs: this is just for getting the segment data directories, which is
# an implementation detail. If you want the location new master data directory
# after an initialization, you can just ask the hub with
#
#    gpupgrade config show --new-datadir
#
```

kalensk: Right. I read that comment and I didn't quite understand it at first.

Maybe renaming the function to quickly show that it is only for verifying or getting the segment data directory?  `expected_upgrade_dir`? I am not sure.

jchampio:
Some additional changes to clean up some embarrassing oversights, sorry:
- Add test changes that should have been added to the original commits but weren't.
- Fix in-place `sed` invocation to handle GNU systems too.

Some changes crept in after the rebase too:
- Revert the `pg_isready` check, as some of the new execute tests cause that to fail out in surprising ways. More investigation is needed.
- Update the new implementations to handle changes to master since this PR was opened (new tests and the moving of stop-source-cluster to execute).

Each is in its own squash commit, which I'll combine tomorrow.

test/config.bats
```diff
@@ -71,5 +72,5 @@ teardown() {
     run gpupgrade config show
     [ "$status" -eq 0 ]
     [ "${lines[0]}" = "new-bindir - /my/new/bin/dir" ]
-    [ "${lines[1]}" = "old-bindir - /my/old/bin/dir" ]
+    [ "${lines[2]}" = "old-bindir - /my/old/bin/dir" ]
```
dkrieger-pvtl: I think you want these checks to match the above, that is:
```
    [ "${lines[0]}" = "new-bindir - /my/new/bin/dir" ]
    [ "${lines[1]}" = "new-datadir - " ] # This isn't populated until cluster creation, but it's still displayed here
    [ "${lines[2]}" = "old-bindir - /my/old/bin/dir" ]
```

Here, you left out the middle check.

hub/config.go
```diff
@@ -38,6 +38,8 @@ func (h *Hub) GetConfig(ctx context.Context, in *idl.GetConfigRequest) (*idl.Get
...
```
dkrieger-pvtl: I do not see this case tested anywhere(other than that it is empty before the target cluster is created).  Given we use this to delete the target cluster in many bats tests, I'd like a test that this is set properly.

test/config.bats
jchampio: The middle check is uninteresting/don't-care for this particular test.

hub/config.go
jchampio: > Given we use this to delete the target cluster in many bats tests, I'd like a test that this is set properly.

If it stops working, the test cleanup steps will all fail. We can add an additional test if you'd really like, but keep in mind that any failures in that test will be buried in all the other failures that happen at the same time.

test/config.bats
dkrieger-pvtl: Yeah, it's a nit/preference, but I find it easier to read this test if it's identical to the one above.

hub/config.go
dkrieger-pvtl: My concern is that this could fail in such a way as to return "" or "/" and a developer might decide to `rm -fr newdatadir`.  Some checks actually sanity check(like `delete cluster`) but it is possible to use this new config return value without any sanity checking.

Yes, it is currently tested implicitly via cleanup, but I'd still prefer an explicit test since an implicit test means it can be changed without knowing that you're removing that implicit test.

jchampio: > Some checks actually sanity check(like delete cluster) but it is possible to use this new config return value without any sanity checking.

Since we can't predicate running cleanups on the results of a test, any failures of the sort you're talking about are going to end up blowing away your system regardless of whether the test fails.

I'll add a test -- it's cheap -- but I'm not sure it's giving you the advantage you're hoping for.

test/config.bats
jchampio: > I find it easier to read this test if it's identical to the one above.

...but they're not identical tests? Why add unneeded logic to make them seem like they are?

hub/config.go
dkrieger-pvtl: I was hoping the test would fail in such a way to not cleanup the datadir...like not setting `NEWDATADIR`...

test/config.bats
dkrieger-pvtl: The semantic difference between the tests are that the first sets the values separately and the second sets them in a single config command, so I'd expect the test checks to be identical.

jchampio: The difference between `configuration can be dumped as a whole` and `multiple configuration values can be set at once` is that the first test is testing a bare `config show` (where the two `config set` calls are simply setup of expected values), and the second test is testing `config set` with multiple arguments (where the `config show` is simply a convenient way to verify both at once).

I see the confusion, though -- any way to easily clarify the situation in your eyes?

hub/config.go
jchampio: > I was hoping the test would fail in such a way to not cleanup the datadir...like not setting NEWDATADIR if that's the case...

We can do anything we want for _that_ test. The _other_ tests will run unabated.

berlin-ab: I do like the idea of having an explicit acceptance test for the feature if it is exposed to the user, even if we do use it for other tests.

test/config.bats
dkrieger-pvtl: Maybe I'm missing something...I think I understand your explanation, but if you look at the first test, if you simply fold the second `config set` into the first, you get the second test without the middle line check.

dkrieger-pvtl: We resolved this by you adding a chore to make the intent in the test cleaner.

test/gpinitsystem.bats
jchampio: Let's try `expected_segment_datadir`?

test/gpinitsystem.bats
```diff
@@ -49,7 +67,11 @@ teardown() {
         --disk-free-ratio 0 3>&-
 
     # Make sure we clean up during teardown().
-    NEW_CLUSTER="$newmasterdir"
+    local newmasterdir="$(gpupgrade config show --new-datadir)"
+    NEW_CLUSTER="${newmasterdir}"
+
+    # Sanity check the newly created master's location.
+    [ "$newmasterdir" = $(expected_datadir "$masterdir") ]
```
dkrieger-pvtl: thanks for adding this low-impact test of `gpupgrade config show --new-datadir`.

hub/config.go
dkrieger-pvtl: Thanks for adding the new low-impact test in `gpinitsystem.bats`.

test/gpinitsystem.bats
jchampio: Switched to `expected_datadir`.

PR #180, opened 2020-01-24
Pass the request object instead of individual parameters

bhuvnesh2703:
As the number of parameters are increased in the initialize command,
instead of adding every parameter the request object should be passed.
This is more scalable and little modification is required for every new
parameter addition

### DIFF
--- a/agent/upgrade_primaries.go
+++ b/agent/upgrade_primaries.go
@@ -19,10 +19,10 @@ import (
 // Allow exec.Command to be mocked out by exectest.NewCommand.
 var execCommand = exec.Command
 
-func (s *Server) UpgradePrimaries(ctx context.Context, in *idl.UpgradePrimariesRequest) (*idl.UpgradePrimariesReply, error) {
+func (s *Server) UpgradePrimaries(ctx context.Context, request *idl.UpgradePrimariesRequest) (*idl.UpgradePrimariesReply, error) {
 	gplog.Info("agent starting %s", idl.Substep_UPGRADE_PRIMARIES)
 
-	err := UpgradePrimary(in.OldBinDir, in.NewBinDir, in.DataDirPairs, s.conf.StateDir, in.CheckOnly, in.UseLinkMode)
+	err := UpgradePrimary(s.conf.StateDir, request)
 	return &idl.UpgradePrimariesReply{}, err
 }
 
@@ -32,10 +32,10 @@ type Segment struct {
 	WorkDir string // the pg_upgrade working directory, where logs are stored
 }
 
-func UpgradePrimary(sourceBinDir string, targetBinDir string, dataDirPairs []*idl.DataDirPair, stateDir string, checkOnly bool, useLinkMode bool) error {
-	segments := make([]Segment, 0, len(dataDirPairs))
+func UpgradePrimary(stateDir string, request *idl.UpgradePrimariesRequest) error {
+	segments := make([]Segment, 0, len(request.DataDirPairs))
 
-	for _, dataPair := range dataDirPairs {
+	for _, dataPair := range request.DataDirPairs {
 		workdir := utils.SegmentPGUpgradeDirectory(stateDir, int(dataPair.Content))
 		err := utils.System.MkdirAll(workdir, 0700)
 		if err != nil {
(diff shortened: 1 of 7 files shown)


agent/upgrade_primaries.go
```diff
@@ -23,7 +23,7 @@ var execCommand = exec.Command
 func (s *Server) UpgradePrimaries(ctx context.Context, in *idl.UpgradePrimariesRequest) (*idl.UpgradePrimariesReply, error) {
 	gplog.Info("agent starting %s", upgradestatus.UPGRADE_PRIMARIES)
 
-	err := UpgradePrimary(in.OldBinDir, in.NewBinDir, in.DataDirPairs, s.conf.StateDir, in.CheckOnly, in.UseLinkMode)
+	err := UpgradePrimary(s.conf.StateDir, in)
```
kalensk: nit: I find renaming `in` to `request` to be more readable

agent/upgrade_primaries.go
```diff
@@ -49,15 +49,15 @@ func UpgradePrimary(sourceBinDir string, targetBinDir string, dataDirPairs []*id
 		})
 	}
 
-	err := UpgradeSegments(sourceBinDir, targetBinDir, segments, checkOnly, useLinkMode)
+	err := UpgradeSegments(segments, in)
```
kalensk: This is perhaps a good example where I would find `request` to be more readable than `in`.

agent/upgrade_primaries.go
```diff
@@ -69,20 +69,20 @@ func UpgradeSegments(sourceBinDir string, targetBinDir string, segments []Segmen
 	for _, segment := range segments {
 		dbid := int(segment.DBID)
 		segmentPair := upgrade.SegmentPair{
-			Source: &upgrade.Segment{sourceBinDir, segment.OldDataDir, dbid, int(segment.OldPort)},
-			Target: &upgrade.Segment{targetBinDir, segment.NewDataDir, dbid, int(segment.NewPort)},
+			Source: &upgrade.Segment{in.OldBinDir, segment.OldDataDir, dbid, int(segment.OldPort)},
+			Target: &upgrade.Segment{in.NewBinDir, segment.NewDataDir, dbid, int(segment.NewPort)},
```
kalensk: nit: It might be useful to 1) rename this is a followup commit in this PR, or 2) add a code todo to rename `OldBinDir` to `SourceBinDir`, and `NewBindDir` to `TargetBinDir` on the `UpgradePrimariesRequest` object since this is our convention throughout the codebase.

hub/initialize.go
```diff
@@ -86,18 +86,18 @@ func (h *Hub) InitializeCreateCluster(in *idl.InitializeCreateClusterRequest, st
 }
 
 // create old/new clusters, write to disk and re-read from disk to make sure it is "durable"
-func (h *Hub) fillClusterConfigsSubStep(_ OutStreams, oldBinDir, newBinDir string, oldPort int, linkMode bool) error {
-	conn := db.NewDBConn("localhost", oldPort, "template1")
+func (h *Hub) fillClusterConfigsSubStep(_ OutStreams, request *idl.InitializeRequest) error {
+	conn := db.NewDBConn("localhost", int(request.OldPort), "template1")
```
kalensk: Similarly, we should consider renaming this to `sourcePort`.

kalensk:
Are we running a full dev pipeline?

agent/upgrade_primaries_test.go
```diff
@@ -100,7 +106,13 @@ func TestUpgradePrimary(t *testing.T) {
 		execCommand = exectest.NewCommand(FailedMain)
 		defer func() { execCommand = nil }()
 
-		err := UpgradePrimary("/old/bin", "/new/bin", pairs, tempDir, false, false)
+		request := &idl.UpgradePrimariesRequest{
+			OldBinDir:    "/old/bin",
+			NewBinDir:    "/new/bin",
+			DataDirPairs: pairs,
+			CheckOnly:    false,
+			UseLinkMode:  false}
```
jchampio: super minor nit:
```go
    UseLinkMode: false,
}
```
is a more common indentation pattern.

agent/upgrade_primaries.go
bhuvnesh2703: will do that as a separate PR.

hub/initialize.go
bhuvnesh2703: in a separate PR

agent/upgrade_primaries_test.go
bhuvnesh2703: changed.

PR #181, opened 2020-01-24
StopCluster: make idempotent by handling any starting cluster state

jchampio:
This substep now runs through these steps on each call, regardless of the current state of the cluster:
1) pg_ctl stop the master
2) gpstart the cluster, putting it in a known good state
3) gpstop the cluster

We tested this via manually injecting a fault using the debugger.

### DIFF
--- a/hub/start_or_stop_cluster.go
+++ b/hub/start_or_stop_cluster.go
@@ -5,6 +5,7 @@ import (
 	"os/exec"
 	"path/filepath"
 
+	"github.com/blang/semver"
 	"github.com/pkg/errors"
 
 	"github.com/greenplum-db/gpupgrade/testutils/exectest"
@@ -53,6 +54,17 @@ func runGPCommand(cmdFunc exectest.Command, path, args string, stream OutStreams
 
 	cmd.Stdout = stream.Stdout()
 	cmd.Stderr = stream.Stderr()
+
+	// Don't leak the hub environment to the subprocesses.
+	cmd.Env = []string{}
+
+	// XXX If this is a 5X cluster, gpstart and gpstop need the
+	// MASTER_DATA_DIRECTORY environment variable to be set due to a bug in the
+	// -d option's implementation.
+	if cluster.Version.SemVer.LT(semver.MustParse("6.0.0")) {
+		mdd := fmt.Sprintf("MASTER_DATA_DIRECTORY=%s", cluster.MasterDataDir())
+		cmd.Env = append(cmd.Env, mdd)
+	}
 
 	fmt.Fprintf(stream.Stdout(), "executing %q %q\n", cmd.Path, cmd.Args)
 	return cmd.Run()
--- a/hub/start_or_stop_cluster.go
+++ b/hub/start_or_stop_cluster.go
@@ -3,14 +3,23 @@ package hub
 import (
 	"fmt"
 	"os/exec"
+	"path/filepath"
 
 	"github.com/pkg/errors"
 
+	"github.com/greenplum-db/gpupgrade/testutils/exectest"
 	"github.com/greenplum-db/gpupgrade/utils"
 )
 
-var isPostmasterRunningCmd = exec.Command
-var startStopClusterCmd = exec.Command
+// These variables provide injection points for exectest.Command.
+// XXX we badly need a way to combine these; they're proliferating
+var (
+	isPostmasterRunningCmd = exec.Command
+	startStopClusterCmd    = exec.Command
+	pgCtlCmd               = exec.Command
+	gpstopCmd              = exec.Command
+	gpstartCmd             = exec.Command
+)
 
 func (h *Hub) ShutdownCluster(stream OutStreams, isSource bool) error {
 	if isSource {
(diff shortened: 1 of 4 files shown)


hub/start_or_stop_cluster_test.go
```diff
@@ -70,40 +72,106 @@ func TestStartOrStopCluster(t *testing.T) {
...
+					t.Errorf("first argument to %q must be -c; got %q", path, args[0])
+				}
+
+				expected = fmt.Sprintf("source /source/bindir/../greenplum_path.sh && %s", expected)
+				if args[1] != expected {
+					t.Errorf("bash command was %q, want %q", args[1], expected)
+				}
+			}
+
+			pgCtlCmd = exectest.NewCommandWithVerifier(Success,
```
berlin-ab: We should consider pushing this logic behind an abstraction, so we don't need to be concerned with the specific options passed to a specific binary, but rather that `pgCtlStop` was called for a given segment.

hub/start_or_stop_cluster_test.go
```diff
@@ -70,40 +72,106 @@ func TestStartOrStopCluster(t *testing.T) {
...
+
+		t.Run("fails if gpstart fails", func(t *testing.T) {
+			setup(t)
+			defer teardown()
+
+			gpstartCmd = exectest.NewCommand(Failure)
+			err := StopCluster(DevNull, source)
+
+			var actual *exec.ExitError
+			if !xerrors.As(err, &actual) {
```
berlin-ab: I cannot understand what assertion is being made here.

hub/start_or_stop_cluster_test.go
jchampio: Agreed. We discussed ways to do that and ran into walls quickly. Our reliance on `greenplum_path.sh` makes it more costly to pull up and abstract than it probably should be.

hub/start_or_stop_cluster_test.go
jchampio: `[x]errors.As()` is an interesting beast. If it returns true, then either `err` or an error that `err` wraps is an `*exec.ExitError`. It's essentially a type assertion on the entire error chain as opposed to the tip of it, which allows us to add debugging information later without having to change this test.

jchampio:
> Would it be possible to introduce an integration test against StopCluster? My future self would like a test to fail if I change the behavior in a way that disables the benefit we're trying to achieve with this design.

We talked about this too! Initial frontrunners that came up were a fault injection system or mocking out the actual binaries that are called end-to-end. More cost than we wanted to take on here in this PR.

> It seems like the steps we take in StopCluster are specific to the recovery situation, but not generally useful. Should we separate the normal behavior from the exceptional behavior?

That's an anti-goal. The current plan is to make all substeps [crash-only](https://en.wikipedia.org/wiki/Crash-only_software), to minimize the number of combinatorial states in the code that a user could have stepped through to get to a specific situation. One downside is that the common case becomes slightly less obvious. 😄

hub/start_or_stop_cluster_test.go
jchampio: [More reading on this feature of Go.](https://github.com/golang/go/wiki/ErrorValueFAQ)

dkrieger-pvtl:
> * Would it be possible to introduce an integration test against `StopCluster`? My future self would like a test to fail if I change the behavior in a way that disables the benefit we're trying to achieve with this design.

I really want to have an integration test that causes each substep to return FAILURE somewhere reasonable in its implementation the first time, and then rerun the containing step(i.e. re-run `gpupgrade initialize` if the substep START_AGENTS fails) and have it pass to test idempotence for each step.  But that's beyond the scope of this PR.

jchampio:
Totally in the scope of [this PR](https://github.com/greenplum-db/gpupgrade/pull/179), though.

hub/start_or_stop_cluster.go
```diff
@@ -28,9 +37,55 @@ func (h *Hub) ShutdownCluster(stream OutStreams, isSource bool) error {
 	return nil
 }
 
+// runGPCommand returns the results of an exec.Cmd run for the given GPDB
+// utility invocation.
+//
+// XXX There are way too many inputs to this.
+func runGPCommand(cmdFunc exectest.Command, path, args string, stream OutStreams, cluster *utils.Cluster) error {
```
kalensk: Not for this PR, but I could see `cmdFunc exectest.Command, path, args string` being rolled into an object with this as a `Run` method.

hub/start_or_stop_cluster.go
```diff
@@ -28,9 +37,55 @@ func (h *Hub) ShutdownCluster(stream OutStreams, isSource bool) error {
...
+	err := runGPCommand(pgCtlCmd,
+		"pg_ctl", fmt.Sprintf("stop -m fast -w -D %s", cluster.MasterDataDir()),
+		stream, cluster)
+	if err != nil {
+		// Because `pg_ctl stop` can fail because the master is already stopped,
+		// but it does not give us a programmatic distinction between a hard
+		// failure and an already-stopped failure, we explicitly ignore non-zero
+		// exit codes here and rely on the following gpstart to catch any
+		// problems.
+		fmt.Fprintf(stream.Stderr(), "ignoring pg_ctl failure: %+v\n", err)
```
kalensk: Do we want to give more context when returning this? That is, explain what the comment is saying in the returned string?

hub/start_or_stop_cluster.go
```diff
@@ -28,9 +37,55 @@ func (h *Hub) ShutdownCluster(stream OutStreams, isSource bool) error {
...
 
+// runGPCommand returns the results of an exec.Cmd run for the given GPDB
+// utility invocation.
+//
+// XXX There are way too many inputs to this.
+func runGPCommand(cmdFunc exectest.Command, path, args string, stream OutStreams, cluster *utils.Cluster) error {
+	path = filepath.Join(cluster.BinDir, path)
+
+	cmd := cmdFunc("bash", "-c",
+		fmt.Sprintf("source %s/../greenplum_path.sh && %s %s",
```
marksliva: We should do a `filepath.Join` here as well.

hub/start_or_stop_cluster.go
jchampio: I could see adding some instructions to the debugging developer, but I don't think we should push that comment into this log line.

kalensk: Yeah, I was thinking it would be good to add a bit of a hint to the debugging developer.

hub/start_or_stop_cluster.go
```diff
@@ -28,9 +37,55 @@ func (h *Hub) ShutdownCluster(stream OutStreams, isSource bool) error {
...
+		// but it does not give us a programmatic distinction between a hard
+		// failure and an already-stopped failure, we explicitly ignore non-zero
+		// exit codes here and rely on the following gpstart to catch any
+		// problems.
+		fmt.Fprintf(stream.Stderr(), "ignoring pg_ctl failure: %+v\n", err)
+	}
+
+	// gpstart the cluster to fully bring up the cluster
+	err = runGPCommand(gpstartCmd,
+		"gpstart", fmt.Sprintf("-a -d %s", cluster.MasterDataDir()),
```
jchampio: 5 to 6 is failing because the `-d` option is buggy in 5X. We have to set `MASTER_DATA_DIRECTORY` explicitly. 🙄

hub/start_or_stop_cluster.go
jchampio: Done.

hub/start_or_stop_cluster.go
jchampio: Done.

jchampio:
Closing due to the above critical issues, which we currently have no proposed fix for...

hub/start_or_stop_cluster.go
jchampio: Fixed.

PR #183, opened 2020-01-24
Rename proto name's for clarity and consistency

kalensk:
Rename the following:
idl.UpgradeSteps => idl.Substep
idl.StepStatus => idl.Status
idl.UpgradeStepStatus => idl.SubstepStatus

Test pipeline is running...


idl/cli_to_hub.proto
```diff
@@ -35,14 +35,12 @@ message RestartAgentsReply {
...
-    StepStatus status = 2;
+message SubstepStatus {
+  Substep step = 1;
+  Status status = 2;
 }
 
-// TODO: rename to Substep
-enum UpgradeSteps {
+enum Substep {
     UNKNOWN_STEP = 0; // http://androiddevblog.com/protocol-buffers-pitfall-adding-enum-values/
```
jchampio: Can this be just `UNKNOWN`...

idl/cli_to_hub.proto
```diff
@@ -58,8 +56,7 @@ enum UpgradeSteps {
     RECONFIGURE_PORTS = 12;
 }
 
-// TODO: rename to Status
-enum StepStatus {
+enum Status {
     UNKNOWN_STATUS = 0; // http://androiddevblog.com/protocol-buffers-pitfall-adding-enum-values/
```
jchampio: ...along with this?

idl/cli_to_hub.proto
kalensk: I wish...doing so resulted in the following  :(
```
cli_to_hub.proto:60:3: "UNKNOWN" is already defined in "idl".
cli_to_hub.proto:60:3: Note that enum values use C++ scoping rules, meaning that enum values are siblings of their type, not children of it.  Therefore, "UNKNOWN" must be unique within "idl", not just within "Status".
```

jchampio: _twitch_

Thanks for trying.

PR #185, opened 2020-01-25
hub accepts a --state-directory flag

dkrieger-pvtl:
For a given cluster upgrade, the CLI and Hub must agree on the
state directory used as the CLI and Hub are both required to run
on the master host.  Currently, we also want to Hub and Agent
state directory to match, even though the segment hosts need not
coincide with the master host.

To keep this consistent, we have the CLI pass its state directory
to the Hub via this new command flag.  The Hub alread does the same
thing with each agent.

### DIFF
--- a/cli/commanders/initialize.go
+++ b/cli/commanders/initialize.go
@@ -18,35 +18,35 @@ var execCommandHubCount = exec.Command
 
 // we create the state directory in the cli to ensure that at most one gpupgrade is occurring
 // at the same time.
-func CreateStateDir() (err error) {
+func CreateStateDir() (stateDir string, err error) {
 	s := Substep("Creating state directory...")
 	defer s.Finish(&err)
 
-	stateDir := utils.GetStateDir()
+	stateDir = utils.GetStateDir()
 	err = os.Mkdir(stateDir, 0700)
 	if os.IsExist(err) {
 		gplog.Debug("State directory %s already present...skipping", stateDir)
-		return nil
+		return stateDir, nil
 	}
 	if err != nil {
 		gplog.Debug("State directory %s could not be created.", stateDir)
-		return err
+		return "", err
 	}
 
-	return nil
+	return stateDir, nil
 }
 
-func CreateInitialClusterConfigs() (err error) {
+func CreateInitialClusterConfigs(stateDir string) (err error) {
 	s := Substep("Creating initial cluster config files...")
 	defer s.Finish(&err)
 
 	// if empty json configuration file exists, skip recreating it
-	filename := filepath.Join(utils.GetStateDir(), hub.ConfigFileName)
+	filename := filepath.Join(stateDir, hub.ConfigFileName)
 	_, err = os.Stat(filename)
 
 	// if the file exists, there will be no error or if there is an error it might
 	// also indicate that the file exists, in either case don't overwrite the file
-	if err == nil || os.IsExist(err) {
+	if err == nil {
 		gplog.Debug("Initial cluster configuration file %s already present...skipping", filename)
 		return nil
 	}
@@ -67,7 +67,7 @@ func CreateInitialClusterConfigs() (err error) {
 	return nil
 }
 
-func StartHub() (err error) {
+func StartHub(stateDir string) (err error) {
 	s := Substep("Starting hub...")
 	defer s.Finish(&err)
 
@@ -81,7 +81,7 @@ func StartHub() (err error) {
 		return nil
 	}
 
-	cmd := execCommandHubStart("gpupgrade", "hub", "--daemonize")
+	cmd := execCommandHubStart("gpupgrade", "hub", "--daemonize", "--state-directory", stateDir)
 	stdout, cmdErr := cmd.Output()
 	if cmdErr != nil {
 		err := fmt.Errorf("failed to start hub (%s)", cmdErr)
--- a/cli/commanders/initialize_test.go
+++ b/cli/commanders/initialize_test.go
@@ -103,7 +103,7 @@ func TestStartHub_Succeeds(t *testing.T) {
 
 	execCommandHubCount = exectest.NewCommand(IsHubRunning_False)
 	execCommandHubStart = exectest.NewCommand(GpupgradeHub_good_Main)
-	err := StartHub()
+	err := StartHub("")
 	g.Expect(err).To(BeNil())
 }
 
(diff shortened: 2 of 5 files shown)


cli/commanders/initialize.go
```diff
@@ -18,35 +18,35 @@ var execCommandHubCount = exec.Command
...
 
 	// if empty json configuration file exists, skip recreating it
-	filename := filepath.Join(utils.GetStateDir(), hub.ConfigFileName)
+	filename := filepath.Join(stateDir, hub.ConfigFileName)
 	_, err = os.Stat(filename)
 
 	// if the file exists, there will be no error or if there is an error it might
 	// also indicate that the file exists, in either case don't overwrite the file
-	if err == nil || os.IsExist(err) {
+	if err == nil {
```
dkrieger-pvtl: Note: the `os.IsExist(err)` can never be reached. `os.Stat(filename)` will not return an error if `filename` exists; it will simply return `nil` as the error.

hub/commands.go
```diff
@@ -101,6 +100,7 @@ func Command() *cobra.Command {
 	}
 
 	cmd.PersistentFlags().StringVar(&logdir, "log-directory", "", "gpupgrade hub log directory")
+	cmd.Flags().StringVar(&stateDir, "state-directory", utils.GetStateDir(), "Agent state directory")
```
dkrieger-pvtl: The flag description `Agent state directory` needs to be modified to read `Hub state directory`.

cli/commanders/initialize_test.go
```diff
@@ -164,21 +164,27 @@ func TestCreateStateDir(t *testing.T) {
 				t.Errorf("stateDir exists")
 			}
 
-			err = CreateStateDir()
+			result, err := CreateStateDir()
```
kalensk: minor nit: rename to `actual`. It seems to usually be expected and actual.

cli/commands/commands.go
```diff
@@ -273,17 +273,17 @@ This step can be reverted.
 			fmt.Println("Initialization in progress.")
 			fmt.Println()
 
-			err = commanders.CreateStateDir()
+			stateDir, err := commanders.CreateStateDir()
```
kalensk: _Potential Blocker_:  Seeing the usage here is wondering why we don't abstract `CreateStateDir` inside of `utils.GetStateDir`. Then one simply calls `utils.GetStateDir` everywhere and the implementation takes care of creating the directory or not.

It seems weird to me to see creating the stateDir and passing it around in this method, but then in other cases we call `utils.GetStateDir`. This is inconsistent and confusing to me, and potentially prone to difficulty maintainability/etc.

2) FWIW, I don't mind if this PR simply passes `--create-statedir` when starting the hub.

dkrieger-pvtl:
> _Potential Blocker_: Seeing the usage in [commands.go](https://github.com/greenplum-db/gpupgrade/pull/185/files#diff-5ebe1d0a554c9fe397d8b44fa02c7d08R276) I'm wondering why we don't abstract `CreateStateDir` inside of `utils.GetStateDir`. Then one simply calls `utils.GetStateDir` everywhere and the implementation takes care of creating the directory or not.

From a semantic standpoint, I'd expect a function called `GetStateDir` to return the directory, not to create it also.  That's why I think it makes sense to have `CreateStateDir()` call `GetStateDir` and not vice versa.  

As far as calling utils.GetStatDir "everywhere", I had done that in #175 by making it a Once function(e.g. singleton), but the strong objection was that it made testing hard by not allowing simple dependency injection(e.g. passing h.stateDir around).  So the idea here was to carry that idea forward by passing in the `stateDir` as a function parameter instead of calling `utils.GetStateDir()` internally.

As it is, that function technically must be called only once per process or we risk using multiple stateDirs, creating havoc.  I plan to handle this in a follow up PR, but for now, these changes align with the current "call-as-many-times-as-you-want-for-testing-but-only-once-in-production) state of the current code.
 
> It seems weird to me to see creating the stateDir and passing it around in this method, but then in other cases we call `utils.GetStateDir`. This is inconsistent and confusing to me, and potentially prone to difficulty maintainability/etc.

In this PR, we do only call `GetStateDir` once per process except for creating of the `execute.log` and `finalize.log` files.  I plan to adjust that in a future PR, but this PR need not handle that.

> 
> 1. FWIW, I don't mind if this PR simply passes `--create-statedir` when starting the hub.

I can nix all changes BUT the ones required to pass `--create-statedir` to the Hub, if you find that cleaner.  Then we can assess the uses of `GetStateDir()` to a future PR.

kalensk:
I am assuming whatever stateDir we use for the CLI we want to use for the Hub and Agent, so all components are in sync in terms of stateDir.  

I can see two main approaches so far (and maybe there are others):
**1)** Use `utils.GetStateDir()` everywhere and create the stateDir if it does not exists by first trying `GPUPGRADE_HOME`, and then falling back to `~/.gpupgrade`. This has the edge case where the cli, hub, and agent could _potentially_ have stateDir's out of sync. Hence, the desire to have a `--state-directory` argument for the hub and agent. Again I think is an _edge case_ that I am ok with leaving as is.

**2)** The desire to have a `--state-directory` argument for the hub and agent to keep the state directories in sync among all components. The hub and  agent have a required parameter `--state-directory`. Early in the startup lifecycle for the hub and agent it checks if the directory exists and if not, creates it.  The tricky part for me when implementing approach 2 is that `utils.GetStateDir` can be overloaded and confusing, especially when it comes to dealing with the hub and agent. _If_ we choose this second approach then something like the following:
 - Get rid of `utils.GetStateDir()`
 - For the CLI simply create the stateDir once in commands.go and use it everywhere. This creation would check if it already exists and if so, returns it.
- For the hub and agent, pass in `--state-directory` which is a **_required parameter_**, and save it on the `h.conf.stateDir` or `a.conf.stateDir` and very early in the lifecycle check if the directory exists and if not create it. By having it as a required parameter there is no need to check GPUPGRADE_HOME or ~/.gpupgrade. And hopefully for testing we can override `h.conf.stateDir`?

**Ask:** Can we as a team be very clear what our goals are in terms of the state directory is, and what approach we want to take? I am "ok" with either (with a preference for the first approach). But it is not clear to me if we have made a conscious choice towards which approach.

dkrieger-pvtl:
As this was done as a self-directed activity, we can table this for a while.  It is not a customer request, but a chore I thought would be straightforward.  Since it generated so much discussion, however, we should discuss as a team what our intent for the state directory is, and how we want to handle it.

PR #186, opened 2020-01-26
Idempotent init create cluster

bhuvnesh2703:
This PR contains 2 commits

1. Persist the New master database port
2. create directories along with marker files, and recreate them in the next run if directories exist with marker file else error out.


hub/initialize.go
```diff
@@ -71,7 +79,7 @@ func (h *Hub) InitializeCreateCluster(in *idl.InitializeCreateClusterRequest, st
 	})
 
 	s.Run(idl.Substep_INIT_TARGET_CLUSTER, func(stream step.OutStreams) error {
-		return h.CreateTargetCluster(stream, targetMasterPort)
+		return h.CreateTargetCluster(stream, int(h.Config.TargetMasterPort))
```
berlin-ab: We should probably transform TargetMasterPort into an integer at the earliest point that we can.

utils/sys_utils.go
```diff
@@ -146,3 +147,46 @@ func WriteJSONFile(fileName string, contents interface{}) error {
...
+	f, err := System.Create(markerFile)
+	if err != nil {
+		return xerrors.Errorf("create gpupgrade marker file %s: %w", markerFile, err)
+	}
+	defer f.Close()
+
+	return nil
+}
+
+func CreateDataDirectories(dataDir string) error {
```
berlin-ab: It would be nice to see the marking of directories be an abstraction instead of a concretion here. We should make the implementation pluggable to make the cost of changing our minds on the implementation lower.

utils/sys_utils.go
```diff
@@ -146,3 +147,46 @@ func WriteJSONFile(fileName string, contents interface{}) error {
...
+	markerFile := filepath.Join(dataDir, ".gpupgrade")
```
berlin-ab: If we're hiding the implementation of CreateMarkerFile, do we need to allow for the name of the marker file to be configurable for each caller?

utils/sys_utils.go
```diff
@@ -146,3 +147,46 @@ func WriteJSONFile(fileName string, contents interface{}) error {
...
+
+	} else if err == nil { // target data directory exists, check for the marker file existence
+		_, err = System.Stat(markerFile)
+		if os.IsNotExist(err) {
+			// marker file does not exist, can't delete the target data directory
+			return xerrors.Errorf("marker file %s doesn't exist, remove master data directory %s", markerFile, dataDir)
+		} else if err == nil {
+			// marker file exists, remove the target data directory and it's content
+			err = System.RemoveAll(dataDir)
+			if err != nil {
```
berlin-ab: Let's work on simplifying these nested conditionals.

hub/init_target_cluster_test.go
```diff
@@ -263,6 +263,12 @@ func TestCreateAllDataDirectories(t *testing.T) {
 			return nil
 		}
 
+		createCalls := []string{}
+		utils.System.Create = func(path string) (*os.File, error) {
```
berlin-ab: It would be nice to be able to inject the creation mechanism rather than patch a lower level implementation decision.

hub/initialize.go
jchampio: +1

utils/sys_utils.go
jchampio: ...maybe? I'm not seeing a huge win for this concrete example -- the integration is "call this method in a for loop", and it's used in two places. So the layer above this call is not a very helpful unit to test, and we don't get a significant speedup on future maybe-refactors.

jchampio:
I'm less interested in injecting this as an abstraction until it's shown that it helps us compose and test -- but I'm not really seeing much in the way of test coverage, which for me is a blocker given the corner cases being handled in the implementation.

utils/sys_utils.go
kalensk: I don't think so. What might be the _current_ need or use case for doing so?

utils/sys_utils.go
kalensk: A big plus one to this! This is a blocker for me.

I am hoping this code can get much more readable/maintainable with some time and thought.

utils/sys_utils.go
```diff
@@ -146,3 +147,46 @@ func WriteJSONFile(fileName string, contents interface{}) error {
...
+	defer f.Close()
+
+	return nil
+}
+
+func CreateDataDirectories(dataDir string) error {
+	markerFile := filepath.Join(dataDir, ".gpupgrade")
+	_, err := System.Stat(dataDir)
+	if os.IsNotExist(err) {
+		return CreateDirAndMarkerFile(dataDir, markerFile)
```
kalensk: 1) Why are we passing in the marker file? Isn't that an implementation detail that we can safely "hardcode"? I don't really see that changing anytime soon. 

2) To me it makes sense to just call it `CreateDataDirectories` or similar. Keep it simple.

hub/initialize.go
```diff
@@ -59,10 +59,18 @@ func (h *Hub) InitializeCreateCluster(in *idl.InitializeCreateClusterRequest, st
...
-		targetMasterPort, err = h.GenerateInitsystemConfig(in.Ports)
+		targetMasterPort, err := h.GenerateInitsystemConfig(in.Ports)
+		if err != nil {
+			return err
+		}
+
+		h.Config.TargetMasterPort = targetMasterPort
+		if err := h.SaveConfig(); err != nil {
+			return err
+		}
```
kalensk: I think a brief comment explaining why we need to persist the TargetMasterPort would be super useful.

kalensk:
> but I'm not really seeing much in the way of test coverage, which for me is a blocker given the corner cases being handled in the implementation.

+1. There are a couple of blockers here for me including this.

utils/sys_utils.go
bhuvnesh2703: sure

hub/initialize.go
bhuvnesh2703: added

hub/initialize.go
bhuvnesh2703: it was unnecessary cast here, it's int. thanks for catching.

utils/sys_utils.go
bhuvnesh2703: agree, it does not need to be configurable.

utils/sys_utils.go
bhuvnesh2703: refactored.

utils/sys_utils.go
berlin-ab: I'm seeing `CreateDataDirectories` as the owner of the business-logic here. It's saying: 

* we need to create this directory
* we need to remember this directory so we can delete it later
* we need to report failures for either of those steps

The current implementation:
- ties those responsibilities together in another function
- spreads the error reporting into several functions
- chooses a concrete implementation instead of an abstraction for "remembering a file for later"

The reason for my suggestion was, if implemented, this function would only need to change later if our business logic changes, rather than needing to change if our strategy for remembering files changes.

utils/sys_utils.go
```diff
@@ -146,3 +147,57 @@ func WriteJSONFile(fileName string, contents interface{}) error {
...
+	}
+	defer f.Close()
+
+	return nil
+}
+
+func CreateDataDirectories(dataDir string) error {
+	_, err := System.Stat(dataDir)
+
+	// create data directory if not existing
```
kalensk: This comment is not needed.

utils/sys_utils.go
```diff
@@ -146,3 +147,57 @@ func WriteJSONFile(fileName string, contents interface{}) error {
...
+	if os.IsNotExist(err) {
+		return CreateDir(dataDir)
+	}
+
+	// if there is an error, bubble up the error
```
kalensk: This comment is not needed.

bhuvnesh2703:
> I'm less interested in injecting this as an abstraction until it's shown that it helps us compose and test -- but I'm not really seeing much in the way of test coverage, which for me is a blocker given the corner cases being handled in the implementation.

added more tests
1. remove existing directories and create new directories and marker file (new test)
2. error out if directory exist and can't stat the marker file (new test)
3. create all directories and marker file on a fresh cluster (added in old test)

bhuvnesh2703:
feedback applied, pls review.

utils/sys_utils.go
```diff
@@ -12,12 +12,15 @@ import (
...
 	"github.com/greenplum-db/gp-common-go-libs/gplog"
 	"github.com/pkg/errors"
+	"golang.org/x/xerrors"
 )
 
 var (
 	System = InitializeSystemFunctions()
 )
 
+const markerFile = ".gpupgrade"
```
dkrieger-pvtl: nit: `.gpugrade` is also the default state directory name.  I'd prefer this be changed to `.gpupgrade_created` or anything obvious other than `.gpupgrade`.

utils/sys_utils.go
```diff
@@ -146,3 +149,53 @@ func WriteJSONFile(fileName string, contents interface{}) error {
...
+		return xerrors.Errorf("create data directory %s: %w", dataDir, err)
+	}
+
+	file := filepath.Join(dataDir, markerFile)
+	gplog.Info("creating marker file %s", file)
+	f, err := System.Create(file)
+	if err != nil {
+		return xerrors.Errorf("create gpupgrade marker file %s: %w", file, err)
+	}
+	defer f.Close()
```
dkrieger-pvtl: `Close()` can return an error.  I think you want to use instead:

```go
defer func() {
   err = f.Close()
   return xerrors.Errorf("closing marker file %s: %w, file, err)
}
```
I think this also requires changing the function signature to a named error parameter, e.g. 
```go
func createDir(dataDir string) (err error)
```

```

**Comment by dkrieger-pvtl** [comment]
It's not worth the effort to eliminate the various race conditions, etc, but there are race conditions between checking that the requested `dataDir` exists and its creation(along with creating the marker file).  There might be a good way to atomically create the directory only if it doesn't exist.  But hitting this would require the DBA to be creating our exact `dataDir` when we do, which isn't likely.

File: utils/sys_utils.go (hunk 10 above)
Comment (jchampio) [reply utils/sys_utils.go]: I don't think we're likely to confuse a file with a directory, do you?

Comment (marksliva) [inline utils/sys_utils.go]: One thing I'm wondering about the `markerFile` is why we don't delete or rename it during the finalize step. Since we plan to support upgrades such as 6-to-7 in the future, this file could exist on source cluster data directories.
This would occur when a user completes a 5-to-6 upgrade and then later starts a 6-to-7 upgrade with the previously upgraded 6 cluster.

File: utils/sys_utils.go (hunk 11 above)
Comment (jchampio) [reply utils/sys_utils.go]: That overwrites `err` with whatever `Close()` returns, which could be `nil`. We currently use `multierror` combination to deal with that.

File: utils/sys_utils.go (hunk 10 above)
Comment (jchampio) [reply utils/sys_utils.go]: Good catch!

File: hub/initialize.go [hunk 12]
```diff
@@ -59,10 +59,18 @@ func (h *Hub) InitializeCreateCluster(in *idl.InitializeCreateClusterRequest, st
 		}
 	}()
 
-	var targetMasterPort int
 	s.Run(idl.Substep_CREATE_TARGET_CONFIG, func(_ step.OutStreams) error {
 		var err error
-		targetMasterPort, err = h.GenerateInitsystemConfig(in.Ports)
+		targetMasterPort, err := h.GenerateInitsystemConfig(in.Ports)
```
jchampio: Not a blocker at this stage, but: once PMs decide how we want to handle the case where users pass arguments during a re-run of `initialize`, I think the persistence of ports might eventually need to move into the `CONFIG` substep.

jchampio:
> There might be a good way to atomically create the directory only if it doesn't exist.

That's the default behavior of `os.Mkdir()`, so I don't think that's a difficult ask. I'm still trying to figure out if there's a way to perform the check-marker-and-create operation in a way that's either atomic or doesn't race against other observers...

utils/sys_utils.go
bhuvnesh2703: thanks, modified.

utils/sys_utils.go
bhuvnesh2703: i will create a story for it.

bhuvnesh2703: i don't mind either .gpupgrade or .gpupgrade_created, but i agree one is a file and other one is a directory and they are both at a different location, so it should be fine. So, for now i will keep it as .gpupgrade and can change in future if we see a need.

jchampio:
> I'm still trying to figure out if there's a way to perform the check-marker-and-create operation in a way that's either atomic or doesn't race against other observers...

Not able to find a way to do that, but I do think I can simplify the current logic, which is
```
1. check /data/dir existence
    a. if it doesn't exist, create /data/dir
    b. touch /data/dir/.gpupgrade. done
2. check /data/dir/.gpupgrade existence
    a. if it doesn't exist, error. done
3. delete /data/dir
4. create /data/dir
5. touch /data/dir/.gpupgrade
```

I think this is equivalent to
```
1. check /data/dir/.gpupgrade existence
    a. if it doesn't exist, remove /data/dir
2. mkdir (atomic) /data/dir
3. touch /data/dir/.gpupgrade
```

jmcatamney:
> ```
> 1. check /data/dir/.gpupgrade existence
>     a. if it doesn't exist, remove /data/dir
> ```

Shouldn't (a) be to remove `/data/dir` if the marker file _does_ exist, and error if it doesn't?

jchampio:
@jmcatamney Oops, fixed.

utils/sys_utils.go
dkrieger-pvtl: @marksliva I agree that, for general code hygiene, we should remove the marker file during `gpupgrade finalize`.  However, I don't see the danger: we always try to create(and hence possibly delete during a retry) a _newly_ created directory.  We never try to create a directory for the _source_ cluster.

utils/sys_utils.go
```diff
@@ -146,3 +151,32 @@ func WriteJSONFile(fileName string, contents interface{}) error {
...
+
+	gplog.Info("creating directory %s", dataDir)
+	err = System.Mkdir(dataDir, 0755)
+	if err != nil {
+		return xerrors.Errorf("create data directory %s: %w", dataDir, err)
+	}
+
+	mFile := filepath.Join(dataDir, markerFile)
+	gplog.Info("creating marker file %s", mFile)
+	err = System.WriteFile(mFile, []byte{}, 0755)
```
jchampio: blocker: this should be `0644`.

utils/sys_utils_test.go
```diff
@@ -196,3 +200,167 @@ var _ = Describe("user utils", func() {
...
+		if err != nil {
+			t.Errorf("returned error %+v", err)
+		}
+
+		if !fileWritten {
+			t.Error("marker file was not created")
+		}
+	})
+
+	t.Run("cannot stat the master data directory", func(t *testing.T) {
```
jchampio: For this test, do we want to assert that modifications are _not_ made to the filesystem in this case?

utils/sys_utils_test.go
```diff
@@ -196,3 +200,167 @@ var _ = Describe("user utils", func() {
...
+			return expected
+		}
+
+		err := CreateDataDirectories(dataDir)
+		if !xerrors.Is(err, expected) {
+			t.Errorf("got %#v, want %#v", err, expected)
+		}
+	})
+
+	t.Run("data directory exist but without marker file .gpupgrade", func(t *testing.T) {
```
jchampio: Similarly, do we want this to test that the data directory is not removed in this case?

utils/sys_utils_test.go
```diff
@@ -196,3 +200,167 @@ var _ = Describe("user utils", func() {
 		})
 	})
 })
+
+func TestCreateAllDataDirectories(t *testing.T) {
```
jchampio: If there is ever additional work done to this function, it'd be nice to deduplicate some of the common test logic, but it's difficult for me to see good seams right now. Not something to do right now, IMO.

utils/sys_utils_test.go
bhuvnesh2703: sure, added check

utils/sys_utils_test.go
bhuvnesh2703: sure, added check

utils/sys_utils.go
marksliva: This function creates only one directory, so it would be better to call it `CreateDataDirectory`.

bhuvnesh2703: renamed

PR #187, opened 2020-01-27
Always run upgrade_check

marksliva:
Always run upgrade_check since the source cluster may change between re-runs of
initialize.

We are considering adding an integration test for this to initialize.bats after https://github.com/greenplum-db/gpupgrade/pull/179 is merged, but for now we think that the unit test we added is sufficient.

### DIFF
--- a/step/step.go
+++ b/step/step.go
@@ -54,7 +54,15 @@ func (s *Step) Err() error {
 	return s.err
 }
 
+func (s *Step) AlwaysRun(substep idl.Substep, f func(OutStreams) error) {
+	s.run(substep, f, true)
+}
+
 func (s *Step) Run(substep idl.Substep, f func(OutStreams) error) {
+	s.run(substep, f, false)
+}
+
+func (s *Step) run(substep idl.Substep, f func(OutStreams) error, alwaysRun bool) {
 	var err error
 	defer func() {
 		if err != nil {
@@ -78,8 +86,8 @@ func (s *Step) Run(substep idl.Substep, f func(OutStreams) error) {
 		return
 	}
 
-	// Only re-run subteps that have failed or pending
-	if status == idl.Status_COMPLETE {
+	// Only re-run substeps that are failed or pending. Do not skip substeps that must always be run.
+	if status == idl.Status_COMPLETE && !alwaysRun {
 		// Only send the status back to the UI; don't re-persist to the store
 		s.sendStatus(substep, idl.Status_COMPLETE)
 		return
--- a/step/step_test.go
+++ b/step/step_test.go
@@ -44,6 +44,36 @@ func TestStepRun(t *testing.T) {
 		}
 	})
 
+	t.Run("re-runs a completed substep when it must always be run", func(t *testing.T) {
+		ctrl := gomock.NewController(t)
+		defer ctrl.Finish()
+
+		server := mock_idl.NewMockCliToHub_ExecuteServer(ctrl)
+		server.EXPECT().
+			Send(&idl.Message{Contents: &idl.Message_Status{&idl.SubstepStatus{
+				Step:   idl.Substep_CHECK_UPGRADE,
+				Status: idl.Status_RUNNING,
+			}}})
+		server.EXPECT().
+			Send(&idl.Message{Contents: &idl.Message_Status{&idl.SubstepStatus{
+				Step:   idl.Substep_CHECK_UPGRADE,
+				Status: idl.Status_COMPLETE,
+			}}})
+
+		store := &TestStore{Status: idl.Status_COMPLETE}
+		s := step.New("Initialize", server, store, DevNull)
+
+		var called bool
+		s.AlwaysRun(idl.Substep_CHECK_UPGRADE, func(streams step.OutStreams) error {
+			called = true
+			return nil
+		})
+
+		if !called {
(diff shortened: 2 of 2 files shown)


step/step.go
```diff
@@ -18,6 +18,8 @@ type Step struct {
 	err     error
 }
 
+var forceRun = map[idl.Substep]bool{idl.Substep_CHECK_UPGRADE: true}
```
jchampio: This feels like it's in the wrong place. The substep logic should be controlled by the owner of the step system. We could do `s.AlwaysRun()`, or pass a `mustRun bool`, or...

jmcatamney: Agreed.  I'd lean toward passing in a bool, particularly if we want to be able to toggle things with a config file during testing.

jchampio: Toggling during testing should be possible with both.

jchampio: (Thinking about it more, since a boolean would pollute all callers, I think I prefer a separate entry point.)

marksliva: Done.

step/step_test.go
```diff
@@ -104,15 +134,15 @@ func TestStepRun(t *testing.T) {
...
+				Step:   idl.Substep_CONFIG,
 				Status: idl.Status_COMPLETE,
 			}}})
 
 		store := &TestStore{Status: idl.Status_COMPLETE}
 		s := step.New("Initialize", server, store, DevNull)
 
 		var called bool
-		s.Run(idl.Substep_CHECK_UPGRADE, func(streams step.OutStreams) error {
+		s.Run(idl.Substep_CONFIG, func(streams step.OutStreams) error {
```
jchampio: These diffs can go away.

step/step.go
```diff
@@ -54,7 +54,14 @@ func (s *Step) Err() error {
 	return s.err
 }
 
+func (s *Step) AlwaysRun(substep idl.Substep, f func(OutStreams) error) {
+	s.run(substep, f, true)
+}
```
jchampio: nit: need a newline here

kalensk: fixed!

step/step_test.go
kalensk: Good catch. They can for sure be removed.

dkrieger-pvtl:
(I see you just merged this which is fine...I realize going from COMPLETE to RUNNING is no different from going from RUNNING to COMPLETE but it's conceptually different so I wanted a test for this...).

You are relying on the step.store to update an existing step's status if it is COMPLETE.  That is, for CHECK_UPGRADE in the scenario discussed here, the status in the json file goes from COMPLETE to RUNNING to (FAILED or COMPLETE).  The implementation of `file_store.go` in fact does this now, but it's not clear to me that this is an intentional part of the design.

So my request is that you add a test for this in `file_store_test.go` that pins this...for instance, add in a test that's like:

```
	t.Run("updates a step status from COMPLETE to RUNNING", func(t *testing.T) {
		substep := idl.Substep_CHECK_UPGRADE
		expected := idl.Status_COMPLETE

		err := fs.Write(substep, expected)
		if err != nil {
			t.Fatalf("Write() returned error %#v", err)
		}

		status, err := fs.Read(substep)
		if err != nil {
			t.Errorf("Read() returned error %#v", err)
		}
		if status != expected {
			t.Errorf("read %v, want %v", status, expected)
		}

		substep = idl.Substep_CHECK_UPGRADE
		expected = idl.Status_RUNNING

		err = fs.Write(substep, expected)
		if err != nil {
			t.Fatalf("Write() returned error %#v", err)
		}

		status, err = fs.Read(substep)
		if err != nil {
			t.Errorf("Read() returned error %#v", err)
		}
		if status != expected {
			t.Errorf("read %v, want %v", status, expected)
		}

	})
```

kalensk:
@dkrieger-pvtl Is this the test you had in mind https://github.com/greenplum-db/gpupgrade/pull/188?

To me it makes more sense to test it at the substep level instead of the file store level since the store is only concerned about reading & writing. It doesn't care what it is reading and writing as that logic is up a layer.

dkrieger-pvtl:
@kalensk Both tests are relevant.  The file_store is responsible for replacing any existing status with the most recently written update(e.g. testing that `fs.Read` obtains the most recently written value), and the step logic is responsible for updating the status depending on the current state of the substep(e.g. that `s.Run` or `s.AlwaysRun` results in the correct state).  

These are different concepts.

jchampio:
@dkrieger-pvtl `FileStore` already has a test to make sure that the last thing you put in is what comes back out:
```go
t.Run("reads the same status that was written", ...)
```

dkrieger-pvtl:
@jchampio that test only does one `f.Write()`.  I am asking for a test that calls `f.Write()` at least twice with different values that validates that the most recent value is what is returned by `f.Read()`.

kalensk:
> I am asking for a test that calls `f.Write()` at least twice with different values that validates that the most recent value is what is returned by `f.Read()`.

Why twice? Shouldn't read should simply return what is written since that is the interface contract we want.

dkrieger-pvtl:
> Why twice? Shouldn't read should simply return what is written since that is the interface contract we want.

Yes, it should.  That's precisely why I want a test for it.  We do currently write status multiple times per substep(RUNNING->COMPLETE) in production but a unit test would be great.

PR #188, opened 2020-01-28
Verify substep status

kalensk:
Verify substep status

### DIFF
--- a/step/step_test.go
+++ b/step/step_test.go
@@ -15,7 +15,7 @@ import (
 )
 
 func TestStepRun(t *testing.T) {
-	t.Run("marks a successful substep run as complete", func(t *testing.T) {
+	t.Run("marks a successful substep as complete", func(t *testing.T) {
 		ctrl := gomock.NewController(t)
 		defer ctrl.Finish()
 
@@ -81,7 +81,7 @@ func TestStepRun(t *testing.T) {
 		}
 	})
 
-	t.Run("re-runs a completed substep when it must always be run", func(t *testing.T) {
+	t.Run("AlwaysRun re-runs a completed substep", func(t *testing.T) {
 		ctrl := gomock.NewController(t)
 		defer ctrl.Finish()
 
@@ -111,7 +111,7 @@ func TestStepRun(t *testing.T) {
 		}
 	})
 
-	t.Run("marks a failed substep run as failed", func(t *testing.T) {
+	t.Run("marks a failed substep as failed", func(t *testing.T) {
 		ctrl := gomock.NewController(t)
 		defer ctrl.Finish()
 
--- a/step/step_test.go
+++ b/step/step_test.go
@@ -44,6 +44,43 @@ func TestStepRun(t *testing.T) {
 		}
 	})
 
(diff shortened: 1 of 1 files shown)


step/step_test.go
```diff
@@ -44,6 +44,53 @@ func TestStepRun(t *testing.T) {
 		}
 	})
 
+	t.Run("correctly sets substep status", func(t *testing.T) {
```
dkrieger-pvtl: The test looks good as far as it goes.  As mentioned in #187, I'd also like a check for an `AlwaysRun()` substep that goes from RUNNING to COMPLETE to RUNNING to COMPLETE to show that the substep status correctly cycles in this case.  

I think it might be cleaner with a slice for the expected status(or a map with symbolic names for each step) instead of inlining it, but that might not turn out to be the case.

step/step_test.go
```diff
@@ -44,6 +44,53 @@ func TestStepRun(t *testing.T) {
...
+		var status idl.Status
+		var err error
+		s.Run(idl.Substep_CONFIG, func(streams step.OutStreams) error {
+			// save off status to verify that it is running
+			status, err = store.Read(idl.Substep_CONFIG)
+			if err != nil {
+				t.Errorf("unexpected error %#v", err)
+			}
+			return nil
+		})
```
jchampio: The use of the `TestStore` interface definitions here is, IMO, obscuring the fact that it's a test stub. That interface is only there to satisfy the unit under test, and it's not fully functional.

To put it another way, the current code implies that 1) we're actually reading the `Substep_CONFIG` stored value from the `TestStore` (we're not) and 2) the operation can fail (it can't). Just use the test stub directly:

```go
s.Run(idl.Substep_CONFIG, func(streams step.OutStreams) error {
    // save off status to verify that it is running
    status = store.Status
    return nil
})
```

step/step_test.go
```diff
@@ -44,6 +44,53 @@ func TestStepRun(t *testing.T) {
...
+		}
+
+		// verify status is complete after running the substep
+		status, err = store.Read(idl.Substep_CONFIG)
+		if err != nil {
+			t.Errorf("unexpected error %#v", err)
+		}
+
+		expected = idl.Status_COMPLETE
+		if status != expected {
```
jchampio: Same comment here:
```go
// verify status is complete after running the substep
expected = idl.Status_COMPLETE
if store.Status != expected {
    ...
}
```

step/step_test.go
jchampio: > I'd also like a check for an AlwaysRun() substep that goes from RUNNING to COMPLETE to RUNNING to COMPLETE to show that the substep status correctly cycles in this case.

We have a test just below this that cycles COMPLETE -> RUNNING -> COMPLETE; all we need to do is add a check for the `TestStore` contents.

step/step_test.go
kalensk: Fixed.

step/step_test.go
kalensk: Fixed.

kalensk:
Tweaked some minor wording on other test descriptions in the file for clarity.

PR #189, opened 2020-01-28
Rename old to source and new to target

kalensk:
For consistency use the terms source and target to follow our internal convention.

Test pipeline is running...


kalensk:
Yeah. I'll try and do a followup PR in the future to prune even more of the "old" and "new" references. This was the low-hanging fruit that gets us one step closer.

Thanks for the review!

PR #190, opened 2020-01-29
initialize: fix upgrade_check

marksliva:
Add an integration test which asserts that upgrade_check
always runs during the initialize step.

### DIFF
--- a/test/initialize.bats
+++ b/test/initialize.bats
@@ -5,21 +5,23 @@ load helpers
 setup() {
     skip_if_no_gpdb
 
-    STATE_DIR=`mktemp -d`
+    STATE_DIR=`mktemp -d /tmp/gpupgrade.XXXXXX`
     export GPUPGRADE_HOME="${STATE_DIR}/gpupgrade"
 
-    gpupgrade kill-services
+    # If this variable is set (to a master data directory), teardown() will call
+    # gpdeletesystem on this cluster.
+    TARGET_CLUSTER=
 
-    # XXX We use $PWD here instead of a real binary directory because
-    # `make check` is expected to test the locally built binaries, not the
-    # installation. This causes problems for tests that need to call GPDB
-    # executables...
+    gpupgrade kill-services
     gpupgrade initialize \
-        --old-bindir="$PWD" \
-        --new-bindir="$PWD" \
+        --old-bindir="${GPHOME}/bin" \
+        --new-bindir="${GPHOME}/bin" \
         --old-port="${PGPORT}"\
         --stop-before-cluster-creation \
         --disk-free-ratio 0 3>&-
+
+    PSQL="$GPHOME"/bin/psql
+    TEARDOWN_FUNCTIONS=()
 }
 
 teardown() {
@@ -28,6 +30,29 @@ teardown() {
         gpupgrade kill-services
         rm -r "$STATE_DIR"
     fi
+
+    for FUNCTION in "${TEARDOWN_FUNCTIONS[@]}"; do
+        $FUNCTION
+    done
+}
+
+set_target_cluster_var_for_teardown() {
+    local newmasterdir="$(upgrade_datadir $MASTER_DATA_DIRECTORY)"
+    TARGET_CLUSTER="${newmasterdir}"
+}
+
+teardown_target_cluster() {
+    delete_target_datadirs $TARGET_CLUSTER
+}
+
+setup_check_upgrade_to_fail() {
+    gpstart -a
+    $PSQL -d postgres -p $PGPORT -c "CREATE TABLE test_pg_upgrade(a int) DISTRIBUTED BY (a) PARTITION BY RANGE (a)(start (1) end(4) every(1));"
+    $PSQL -d postgres -p $PGPORT -c "CREATE UNIQUE INDEX fomo ON test_pg_upgrade (a);"
+}
+
+teardown_check_upgrade_failure() {
+    $PSQL -d postgres -p $PGPORT -c "DROP TABLE IF EXISTS test_pg_upgrade CASCADE;"
 }
 
 @test "hub daemonizes and prints the PID when passed the --daemonize option" {
@@ -52,12 +77,6 @@ teardown() {
     [ "$status" -eq 1 ]
 
     [[ "$output" = *"config.json: no such file or directory"* ]]
-}
-
-@test "initialize returns an error when it is ran twice" {
-    # second start should return an error
-    ! gpupgrade initialize --old-bindir="${GPHOME}/bin" --new-bindir="${GPHOME}/bin" --old-port="${PGPORT}"
-    # TODO: check for a useful error message
 }
 
 @test "hub does not return an error if an unrelated process has gpupgrade hub in its name" {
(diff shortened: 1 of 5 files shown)


test/initialize.bats
```diff
@@ -144,3 +133,24 @@ outputContains() {
...
+    NEW_CLUSTER="${newmasterdir}"
+
+    gpupgrade initialize \
+        --old-bindir="$GPHOME/bin" \
+        --new-bindir="$GPHOME/bin" \
+        --old-port="${PGPORT}" \
+        --disk-free-ratio 0 3>&-
+
+    # Check that the initialize log contains 2 runs of check_upgrade
+    local number_of_check_runs=$(grep 'Starting CHECK_UPGRADE...' $GPUPGRADE_HOME/initialize.log -c)
```
berlin-ab: Nice trick.

I was expecting a test that noted why this functionality might be useful. Maybe a run of checks that is clean, a modification to the source cluster, and a run of checks that fails.

berlin-ab: At the very least, the test description should include the motivation.

test/initialize.bats
```diff
@@ -54,12 +49,6 @@ teardown() {
     [[ "$output" = *"config.json: no such file or directory"* ]]
 }
 
-@test "initialize returns an error when it is ran twice" {
-    # second start should return an error
-    ! gpupgrade initialize --old-bindir="${GPHOME}/bin" --new-bindir="${GPHOME}/bin" --old-port="${PGPORT}"
-    # TODO: check for a useful error message
-}
-
```
jchampio: Why is a seemingly unrelated test being removed?

test/initialize.bats
jchampio: Yeah, this seems like a strange place to take on debt right now... someone else is going to have to rewrite this test as soon as that log message changes.

test/initialize.bats
marksliva: Now that all substeps in the initialize step are idempotent, this test is no longer relevant.

marksliva: Also, the change to the shared `setup()` caused it to fail. Before when we specified `pwd` for the old and new bin dirs in the `setup()` function, it persisted these bindirs to the config on disk, which inadvertently caused a rerun to fail when it could not find the `gpstart` binary in `pwd`. With the fix to `setup()`, this test started failing so it seemed a good time to remove it.

berlin-ab: Makes sense.

test/initialize.bats
berlin-ab: Thanks for taking a second pass at this.

test/pg_upgrade.bats
```diff
@@ -14,7 +14,7 @@ setup() {
     NEW_CLUSTER=
     KEEP_STATE_DIR=1
 
-    PSQL="$GPHOME"/bin/psql
+    define_psql
```
jchampio: nit: since the implementation we're hiding is one line, I don't think the `define_psql()` helper buys us enough to balance the magic behavior it's concealing.

test/initialize.bats
```diff
@@ -28,6 +30,29 @@ teardown() {
...
     fi
+
+    for FUNCTION in "${TEARDOWN_FUNCTIONS[@]}"; do
+        $FUNCTION
+    done
+}
+
+teardown_check_upgrade_failure() {
+    $PSQL -d postgres -p $PGPORT -c "DROP TABLE IF EXISTS test_pg_upgrade CASCADE;"
+}
```
jchampio: super minor: can we move the teardown below its corresponding setup?

test/initialize.bats
```diff
@@ -144,3 +163,27 @@ outputContains() {
...
+
+    run gpupgrade initialize \
+        --old-bindir="$GPHOME/bin" \
+        --new-bindir="$GPHOME/bin" \
+        --old-port="${PGPORT}" \
+        --disk-free-ratio 0 3>&-
+
+    # Other substeps are skipped when marked completed in the state dir,
+    # for check_upgrade, we always run it.
+    [ "$status" -eq 1 ] || fail "expected status: 1 got: $status"
```
jchampio: If I saw

     expected status: 1 got: 3

I don't think I would have enough information to debug. Can we print `$output` instead?

test/pg_upgrade.bats
berlin-ab: We were doing it a few times the same way. Might as well codify it.

jchampio: > Might as well codify it.

That's an antipattern. Abstraction and implementation-hiding is _costly_, especially in tests (where getting up to speed with unfamiliar code is something we need to optimize for), so it needs to provide enough benefit to outweigh that cost.

But if there's consensus that the cost/benefit is low enough, I won't stand in the way here.

kalensk: FWIW I prefer to be explicit here, and don't see the benefit of adding another abstraction (in fact I think it slows us down).

berlin-ab: Except for when an abstraction makes the cost of new requirements cheap. Imagine we need to provide a parameter to psql every single time we call it, no matter from where. Now we're in a position to make that change once, instead of N times. 

Also, we're reducing the possibility of implementation drift which would make it harder to reduce implementations into an abstraction later.

For each "eh, it's not that bad of a smell" (in this case duplicate code / DRY) we're lowering the bar of quality and raising the cost of maintenance. Keep the bar high.  If there's an argument for a better way to reduce the smell, I'd be willing to implement it.

kalensk: This seems like a classic case of premature optimization.

jchampio: > Imagine we need to provide a parameter to psql every single time we call it, no matter from where. Now we're in a position to make that change once, instead of N times.

I can't imagine a situation in which this _particular_ abstraction makes your described use case easier. We're not invoking `psql` in this abstraction; if we were, I would be in favor of deduplicating.

> Also, we're reducing the possibility of implementation drift which would make it harder to reduce implementations into an abstraction later.

I suspect the probability of implementation drift in this _particular_ case approaches zero; I'd be more willing to abstract when I actually saw implementation drift in practice. Premature abstraction leads to lock-in and worse maintenance characteristics.

["Duplication is far cheaper than the wrong abstraction."](https://www.sandimetz.com/blog/2016/1/20/the-wrong-abstraction)

berlin-ab: > I suspect the probability of implementation drift in this particular case approaches zero; I'd be more willing to abstract when I actually saw implementation drift in practice. Premature abstraction leads to lock-in and worse maintenance characteristics.

I agree with "prefer duplication over the wrong abstraction", but the argument is not "don't abstract". 

When the abstraction that we have introduced needs to vary is where the "wrong abstraction" part is introduced. The abstraction we have currently is the right abstraction, it solves the need of three call sites once.

> I can't imagine a situation in which this particular abstraction makes your described use case easier. We're not invoking psql in this abstraction; if we were, I would be in favor of deduplicating.

Maybe that's a better abstraction and we should have a function for executing `psql` and not a variable.

jchampio: > but the argument is not "don't abstract".

I wasn't trying to suggest that as an argument. My argument is that _this_ abstraction is not the one to do.

> The abstraction we have currently is the right abstraction, it solves the need of three call sites once.

I don't think that correctness is the only determinant of what makes a "right" or "wrong" abstraction. I suspect we just disagree on the cost of this approach.

marksliva: Looking closer, this seems like YAGNI. If it ever gets complex, we can refactor it into a function later. Removed.

test/initialize.bats
marksliva: Done.

test/initialize.bats
marksliva: Done.

PR #194, opened 2020-01-30
execute: add an end-to-end --link mode test

marksliva:


test/helpers.bash
```diff
@@ -51,3 +51,19 @@ delete_cluster() {
     # data directory to segments with "file exists".
     rm -rf "${datadir}"/*_upgrade
 }
+
+# require_gnu_stat tries to find a GNU stat program. If one is found, it will be
+# assigned to the STAT global variable; otherwise the current test is skipped.
+require_gnu_stat() {
```
jchampio: I think there's a second copy of this in `checks.bats` that should be removed.

test/execute.bats
```diff
@@ -0,0 +1,102 @@
...
+    ")
+
+    for relfilenode in "${relfilenodes[@]}"; do
+        local number_of_hardlinks=$($STAT --format "%h" "${relfilenode}")
+        [ $number_of_hardlinks -eq $expected_number_of_hardlinks ] \
+            || fail "expected $expected_number_of_hardlinks hardlinks to $relfilenode; found $number_of_hardlinks"
+    done
+}
+
+set_primary_datadirs() {
```
jchampio: nit: `primary` makes it seem like we're excluding the master, but we're not.

test/execute.bats
```diff
@@ -0,0 +1,102 @@
...
+        --verbose
+
+    local datadir=$(dirname $(dirname "${MASTER_DATA_DIRECTORY}"))
+    NEW_CLUSTER="${datadir}/qddir_upgrade/demoDataDir-1"
+
+    gpupgrade execute --verbose
+
+    PGPORT=50432 ensure_hardlinks_for_relfilenode_on_master_and_segments 'test_linking' 2
+
+    reset_primary_pg_control_files
```
jchampio: nit: any way to move this into the `teardown` somehow? That way we still have a working source cluster if the test fails.

berlin-ab: The `TEARDOWN_FUNCTIONS`  pattern we introduced might be useful here.

marksliva: Done.

test/execute.bats
marksliva: Changed it to include `master`.

test/helpers.bash
marksliva: Done.

marksliva:
> I think my only blocking concern is the second copy of `require_gnu_stat` that needs to be removed.

Done.

PR #195, opened 2020-01-31
Make Upgrade Master step idempotent

bhuvnesh2703:
In order to make the upgrade master step idempotent, this commit does
the following
1. As part of initialize, creates a master.bak directory and populates
it with the contents of the newly created clusters master data directory
2. As part of execute, uses the contents of the master.bak directory and
rsyncs it to the newly created cluster master data directory

Co-authored-by: Jacob Champion <pchampion@pivotal.io>


idl/hub_to_agent.pb.go
```diff
@@ -21,15 +22,15 @@ var _ = math.Inf
 // is compatible with the proto package it is being compiled against.
 // A compilation error at this line likely means your copy of the
 // proto package needs to be updated.
-const _ = proto.ProtoPackageIsVersion2 // please upgrade the proto package
+const _ = proto.ProtoPackageIsVersion3 // please upgrade the proto package
```
bhuvnesh2703: folks any specific issues because of ProtoPackageIsVersion3?

hub/initialize.go
```diff
@@ -78,6 +78,12 @@ func (h *Hub) InitializeCreateCluster(in *idl.InitializeCreateClusterRequest, st
 		return h.ShutdownCluster(stream, false)
 	})
 
+	s.Run(idl.Substep_BACKUP_TARGET_MASTER, func(stream step.OutStreams) error {
+		sourceDir := h.Target.MasterDataDir() + "/"
+		targetDir := filepath.Join(h.StateDir, backupTargetDirName)
```
berlin-ab: Where does backupTargetDirName come from?

hub/upgrade_master.go
```diff
@@ -21,6 +27,12 @@ func UpgradeMaster(source, target *utils.Cluster, stateDir string, stream step.O
 		return err
 	}
 
+	sourceDir := filepath.Join(stateDir, backupTargetDirName) + "/"
```
berlin-ab: Seems to be a duplicate of the `Substep_BACKUP_TARGET_MASTER`

hub/upgrade_master.go
```diff
@@ -50,3 +62,18 @@ func masterSegmentFromCluster(cluster *utils.Cluster) *upgrade.Segment {
 		Port:    cluster.MasterPort(),
 	}
 }
+
+func RsyncMasterDataDir(stream step.OutStreams, sourceDir, targetDir string) error {
+	// TODO: Once we have removed the need for duplication of execCommand injection
+	// update it with the execCommand call
+	cmd := execCommandRsync("rsync", "--archive", "--exclude=pg_log/*", sourceDir, targetDir)
```
berlin-ab: This is a repeated pattern throughout the codebase:

```
cmd := execCommand(...)
cmd.Stdout = stream.Stdout()
cmd.Stderr = stream.Stderr()
```

Seems like a missing abstraction.

hub/upgrade_master_test.go
```diff
@@ -174,6 +177,9 @@ func TestUpgradeMaster(t *testing.T) {
 		SetExecCommand(exectest.NewCommand(StreamingMain))
 		defer ResetExecCommand()
 
+		SetRsyncExecCommand(exectest.NewCommand(Success))
+		defer ResetRsyncExecCommand()
```
berlin-ab: Seems like it would be nice to have a `beforeEachRun` for this kind of resetting.

test/initialize.bats
```diff
@@ -144,3 +144,22 @@ outputContains() {
...
 }
+
+@test "initialize backs up the new master data directory" {
+    gpupgrade kill-services
+    STATE_DIR="/tmp/gpupgrade-"$RANDOM
+    mkdir -p "$STATE_DIR"
+    export GPUPGRADE_HOME="${STATE_DIR}/gpupgrade_backup"
+
+    rm -rf  "$GPUPGRADE_HOME"/master.bak
+    gpupgrade initialize \
```
berlin-ab: We repeat this statement many times throughout the bats suite with minor variation

test/initialize.bats
```diff
@@ -144,3 +144,22 @@ outputContains() {
         [[ $output = *'invalid argument '*' for "--disk-free-ratio" flag:'* ]] || fail
     done
 }
+
+@test "initialize backs up the new master data directory" {
```
berlin-ab: This test will be more useful once we have the "why" behind the backup. Right now it is not clear why this is useful from an acceptance test perspective.  Maybe that's a signal that this test would be more appropriate as in integration test in Go.

test/execute.bats
```diff
@@ -104,3 +110,29 @@ reset_master_and_primary_pg_control_files() {
 
     PGPORT=50432 ensure_hardlinks_for_relfilenode_on_master_and_segments 'test_linking' 2
 }
+
+@test "gpupgrade execute step to upgrade master should always rsync the master data dir from backup" {
+    require_gnu_stat
+    set_master_and_primary_datadirs
+
+    local datadir=$(dirname $(dirname "${MASTER_DATA_DIRECTORY}"))
```
berlin-ab: IIRC, there's a helper function for doing this.

berlin-ab: Ah, it's in another current PR. 

There is similar work being done elsewhere in the `delete_cluster` helper function.

test/initialize.bats
bhuvnesh2703: i have now removed this test, because i added a test in https://github.com/greenplum-db/gpupgrade/pull/195/files#diff-8c97fb56d8f26cf0ecccfd583252564bR114-R138 which tests end to end, creation of the backup directory and then resyncing it to the target master data dir.

test/execute.bats
bhuvnesh2703: ok, if that PR goes in before this, i will take that.

test/initialize.bats
bhuvnesh2703: removed this test.

hub/upgrade_master.go
bhuvnesh2703: Substep_BACKUP_TARGET_MASTER is a step and backupTargetDirName is a string = master.bak..

hub/initialize.go
bhuvnesh2703: defined in hub/upgrade_master.go `18:const backupTargetDirName = "master.bak"`

hub/initialize.go
```diff
@@ -78,6 +78,12 @@ func (h *Hub) InitializeCreateCluster(in *idl.InitializeCreateClusterRequest, st
 		return h.ShutdownCluster(stream, false)
 	})
 
+	s.Run(idl.Substep_BACKUP_TARGET_MASTER, func(stream step.OutStreams) error {
+		sourceDir := h.Target.MasterDataDir() + "/"
```
marksliva: We can replace `"/"` with `os.PathSeparator`.

idl/hub_to_agent.pb.go
jmcatamney: As far as I know we're not relying on any version-specific features, so this should be fine.

hub/initialize.go
kalensk: I don't think the trailing slash get preserved when calling filepath.Join? Seems like it is not needed, and we can inline both variables.

test/execute.bats
```diff
@@ -80,10 +80,21 @@ reset_master_and_primary_pg_control_files() {
     done
 }
 
+delete_datadirs() {
```
berlin-ab: nit: can we give this function context? we're deleting the target data dirs, so delete_target_datadirs might be an improvement.

Also, I think a function like that might belong in the helpers - and might already exist there.

hub/initialize.go
kalensk: Yeah, and if so it will have to be cast to a string since os.PathSeparator is a rune.

hub/upgrade_master.go
```diff
@@ -10,6 +13,9 @@ import (
 
 // Allow exec.Command to be mocked out by exectest.NewCommand.
 var execCommand = exec.Command
+var execCommandRsync = exec.Command
+
+const backupTargetDirName = "master.bak"
```
kalensk: When reading the code where it is used, I am finding this name confusing. I am still trying to wrap my head around everything, but does simply renaming `MasterBackup` work? I think that is clearer when reading its usage?

hub/upgrade_master.go
```diff
@@ -50,3 +62,18 @@ func masterSegmentFromCluster(cluster *utils.Cluster) *upgrade.Segment {
...
+	// TODO: Once we have removed the need for duplication of execCommand injection
+	// update it with the execCommand call
+	cmd := execCommandRsync("rsync", "--archive", "--exclude=pg_log/*", sourceDir, targetDir)
+
+	cmd.Stdout = stream.Stdout()
+	cmd.Stderr = stream.Stderr()
+
+	err := cmd.Run()
+	if err != nil {
+		return xerrors.Errorf("rsync source directory %q, target directory %q: %w", sourceDir, targetDir, err)
```
kalensk: super minor nit: This error and the one above are the same, but have different error strings. Consider renaming this to match the above with `"rsync %q to %q: %w"`

hub/upgrade_master_test.go
kalensk: I kind of like seeing the cleanup next to the allocation. Ideally it would be nice if allocating would automatically cleanup the resource kind of like C++'s RAII.

cli/commanders/steps.go
```diff
@@ -24,6 +24,7 @@ var lines = map[idl.Substep]string{
 	idl.Substep_SHUTDOWN_SOURCE_CLUSTER: "Stopping old cluster...",
 	idl.Substep_INIT_TARGET_CLUSTER:     "Creating new cluster...",
 	idl.Substep_SHUTDOWN_TARGET_CLUSTER: "Stopping new cluster...",
+	idl.Substep_BACKUP_TARGET_MASTER:    "Backing up new master...",
```
berlin-ab: nit: it seems strange that all substeps are bucketed into one enum and not organized by high-level step. not something we need to resolve immediately, but we should consider organizing at some point.

test/execute.bats
marksliva: It is in master now. The function is `delete_target_datadirs()`.

hub/upgrade_master.go
```diff
@@ -50,3 +62,18 @@ func masterSegmentFromCluster(cluster *utils.Cluster) *upgrade.Segment {
...
+func RsyncMasterDataDir(stream step.OutStreams, sourceDir, targetDir string) error {
+	// TODO: Once we have removed the need for duplication of execCommand injection
+	// update it with the execCommand call
+	cmd := execCommandRsync("rsync", "--archive", "--exclude=pg_log/*", sourceDir, targetDir)
+
+	cmd.Stdout = stream.Stdout()
+	cmd.Stderr = stream.Stderr()
+
+	err := cmd.Run()
+	if err != nil {
```
marksliva: nit: missing a unit test for this error

hub/upgrade_master.go
```diff
@@ -21,6 +27,12 @@ func UpgradeMaster(source, target *utils.Cluster, stateDir string, stream step.O
 		return err
 	}
 
+	sourceDir := filepath.Join(stateDir, backupTargetDirName) + "/"
+	err = RsyncMasterDataDir(stream, sourceDir, target.MasterDataDir())
+	if err != nil {
+		return xerrors.Errorf("rsync %q to %q: %w", sourceDir, target.MasterDataDir(), err)
```
marksliva: nit: missing a unit test for this failure

cli/commanders/steps.go
kalensk: I''m curious what challenge that would solve? I am not seeing the need for the added complexity?

berlin-ab: Ah, I wasn't looking at it in context in this file.

dkrieger-pvtl: I think it makes sense to have step-scoped namespaces for the substeps, but it is not something this PR should address.  To me, it makes the code more readable, but I know that's debatable.

kalensk:
I am curious as to why we chose to expose the backup master substep to the user rather than an implementation detail of upgrading the master (like most of our other idempotent substeps)? So far I don't have any opinions either way.

hub/initialize.go
dkrieger-pvtl: I don't think the caller of `RsyncMasterDataDir` should be formatted the input data directories as per the requirements of `rsync`...any suffix should be applied in the `RsyncMasterDataDir` function.

hub/upgrade_master.go
kalensk: I think this is a bit of a blocker for me, unless other people don't find it a big deal.

jchampio:
> I am curious as to why we chose to expose the backup master substep to the user rather than an implementation detail of upgrading the master (like most of our other idempotent substeps)?

We need separate substeps because once you start the master upgrade, you can't make a clean backup anymore. We could definitely hide that from the UI, though I don't think there's any prior art for doing a "hidden" substep at the moment.

hub/initialize.go
jchampio: > I don't think the caller of RsyncMasterDataDir should be formatted the input data directories as per the requirements of rsync

Big +1.

hub/upgrade_master.go
```diff
@@ -50,3 +62,18 @@ func masterSegmentFromCluster(cluster *utils.Cluster) *upgrade.Segment {
 		Port:    cluster.MasterPort(),
 	}
 }
+
+func RsyncMasterDataDir(stream step.OutStreams, sourceDir, targetDir string) error {
+	// TODO: Once we have removed the need for duplication of execCommand injection
+	// update it with the execCommand call
```
dkrieger-pvtl: nit: I know what this comment means but I am not sure it's needed here...replacing all our `exec.Command` variables is a higher level task that we know about, so there's no need to have this TODO.

cli/commanders/steps.go
kalensk: Just so I am on the same page, are we talking about renaming to something like: `idl.Initialize_BACKUP_TARGET_MASTER` and `idl.Execute_UPGRADE_MASTER`?

hub/upgrade_master.go
dkrieger-pvtl: nit: similarly, the error case in `RsyncMasterDataDir` is also untested.

hub/upgrade_master.go
bhuvnesh2703: Sure, masterBackup is more clean. Renamed it.

hub/upgrade_master.go
bhuvnesh2703: changed it..

hub/upgrade_master.go
bhuvnesh2703: added

hub/upgrade_master.go
bhuvnesh2703: added

hub/upgrade_master.go
bhuvnesh2703: removed it.

test/execute.bats
bhuvnesh2703: used it now.

hub/initialize.go
bhuvnesh2703: changed.

bhuvnesh2703: used os.PathSeparator inside the RsyncMasterDataDir function..

test/execute.bats
bhuvnesh2703: used the other helper function

cli/commanders/steps.go
dkrieger-pvtl: This is OT for this PR, but since we're here...I was thinking of using the proto spec:
```proto
enum InitializeSubstep {
    UNKNOWN_STEP = 0;  
    CONFIG = 1;
    START_AGENTS = 2;
    ...
}

enum ExecuteSubstep {
    UNKNOWN_STEP = 0;  
    UPGRADE_MASTER = 1;
    COPY_MASTER =     2;
    ...
}
```
but now that I write it out, it is not maintainable since we cannot have enum hierarchies in protobuf: they are just `int32`.  So we'd need different `Run()` functions for each step.

So having a prefix as you suggest above might be a bit cleaner than the current code.

hub/upgrade_master.go
```diff
@@ -10,6 +14,9 @@ import (
 
 // Allow exec.Command to be mocked out by exectest.NewCommand.
 var execCommand = exec.Command
+var execCommandRsync = exec.Command
+
+const backupTargetDirName = "masterBackup"
```
kalensk: I think this forgot to be renamed to `MasterBackup work` or the like.

hub/upgrade_master_test.go
```diff
@@ -197,12 +207,56 @@ func TestUpgradeMaster(t *testing.T) {
...
+
+		SetRsyncExecCommand(exectest.NewCommand(Failure))
+		defer ResetRsyncExecCommand()
+
+		stream := new(bufferedStreams)
+
+		err := UpgradeMaster(source, target, tempDir, stream, false, false)
+		if err == nil {
+			t.Errorf("expected error, returned nil")
+		}
```
kalensk: Would it make more sense to use `xerrors.Is`?

```
expected := errors.New("oops!")
    t.Errorf("returned error %#v, want %#v", err, expected)
}
```

hub/upgrade_master.go
dkrieger-pvtl: the role of the source and target are swapped in these two cases.

hub/upgrade_master.go
dkrieger-pvtl: oops...we renamed the directory and not the variable name...it now reads:
`const masterBackup = "master.bak"`

hub/upgrade_master_test.go
dkrieger-pvtl: hmmm...maybe.  Since we tested the exact same case but with the RsyncExecCommand set to Success, this seems unnecessary here.

kalensk: I am not seeing how testing with the same case, but set to Success helps with question at hand. This is a rather minor thing and definitely not worth holding up this PR.

PR #196, opened 2020-02-01
Make start agents idempotent

kalensk:
To make the start agents substep idempotent we reuse the RestartAgents implementation. We will need to revisit starting and stopping of the hub and agents as our current implementation has some drawbacks. Namely, we have no way of stopping spurious agents that the hub is not aware of. PID files seems like the way to go.

This introduces a test helper program port_listener/main.go to listen on all ipv4 and ipv6 interfaces on a specific port. If no ipv6 interface is found it will not listen on ipv6. This is useful where our current CI infrastructure does not support ipv6.

We use port_listener/main.go instead of netcat since this guarantees to bind to all interfaces, and keeps the held gRPC connection open during gRPC retries.

### DIFF
--- a/hub/initialize.go
+++ b/hub/initialize.go
@@ -1,6 +1,7 @@
 package hub
 
 import (
+	"context"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -36,8 +37,9 @@ func (h *Hub) Initialize(in *idl.InitializeRequest, stream idl.CliToHub_Initiali
 		return h.fillClusterConfigsSubStep(stream, in)
 	})
 
-	s.Run(idl.Substep_START_AGENTS, func(stream step.OutStreams) error {
-		return h.startAgentsSubStep(stream)
+	s.Run(idl.Substep_START_AGENTS, func(_ step.OutStreams) error {
+		_, err := RestartAgents(context.Background(), nil, h.Source.GetHostnames(), h.AgentPort, h.StateDir)
+		return err
 	})
 
 	return s.Err()
@@ -125,55 +127,4 @@ func getAgentPath() (string, error) {
 	}
 
 	return filepath.Join(filepath.Dir(hubPath), "gpupgrade"), nil
-}
-
-// TODO: use the implementation in RestartAgents() for this function and combine them
-func (h *Hub) startAgentsSubStep(stream step.OutStreams) error {
-	source := h.Source
-	stateDir := h.StateDir
-
-	// XXX If there are failures, does it matter what agents have successfully
-	// started, or do we just want to stop all of them and kick back to the
-	// user?
-	logStr := "start agents on master and hosts"
-
-	agentPath, err := getAgentPath()
-	if err != nil {
-		return errors.Errorf("failed to get the hub executable path %v", err)
-	}
-
-	// XXX State directory handling on agents needs to be improved. See issue
-	// #127: all agents will silently recreate that directory if it doesn't
-	// already exist. Plus, ExecuteOnAllHosts() doesn't let us control whether
-	// we execute locally or via SSH for the master, so we don't know whether
-	// GPUPGRADE_HOME is going to be inherited.
-	runAgentCmd := func(contentID int) string {
-		return agentPath + " agent --daemonize --state-directory " + stateDir
-	}
-
-	errStr := "Failed to start all gpupgrade agents"
-
-	remoteOutput, err := source.ExecuteOnAllHosts(logStr, runAgentCmd)
-	if err != nil {
-		return errors.Wrap(err, errStr)
-	}
-
-	errMessage := func(contentID int) string {
-		return fmt.Sprintf("Could not start gpupgrade agent on segment with contentID %d", contentID)
-	}
-	source.CheckClusterError(remoteOutput, errStr, errMessage, true)
-
-	// Agents print their port and PID to stdout; log them for posterity.
-	for content, output := range remoteOutput.Stdouts {
-		if remoteOutput.Errors[content] == nil {
-			gplog.Info("[%s] %s", source.Segments[content].Hostname, output)
-		}
-	}
-
-	if remoteOutput.NumErrors > 0 {
-		// CheckClusterError() will have already logged each error.
-		return errors.New("could not start agents on segment hosts; see log for details")
-	}
-
-	return nil
 }
--- a/test/initialize.bats
+++ b/test/initialize.bats
@@ -12,6 +12,10 @@ setup() {
     # gpdeletesystem on this cluster.
     TARGET_CLUSTER=
 
+    # The process that is holding onto the port
+    HELD_PORT_PID=
+    AGENT_PORT=
+
     gpupgrade kill-services
     gpupgrade initialize \
         --old-bindir="${GPHOME}/bin" \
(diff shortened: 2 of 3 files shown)


test/initialize.bats
```diff
@@ -164,6 +176,77 @@ outputContains() {
...
+@test "start agents fails if a process is connected on the same TCP port" {
+    # Ensure the agent is down, so that we can test port in use.
+    gpupgrade kill-services
+    rm -r "$GPUPGRADE_HOME"
+
+    # squat gpupgrade agent port
+    AGENT_PORT=6416
+    go run ./testutils/port_listener/main.go $AGENT_PORT 3>&- &
+
+    # Store the pid of the process group leader since the port are held by its child
```
dkrieger-pvtl: `s/are/is/`

testutils/port_listener/main.go
```diff
@@ -0,0 +1,86 @@
+// The port_listener utility listens on the specified port on all ipv4 and ipv6
+// interfaces. If no ipv6 interface is detect it will not listen on ipv6.
```
dkrieger-pvtl: `s/detect/detected/`

hub/initialize.go
```diff
@@ -36,8 +37,9 @@ func (h *Hub) Initialize(in *idl.InitializeRequest, stream idl.CliToHub_Initiali
 		return h.fillClusterConfigsSubStep(stream, in)
 	})
 
-	s.Run(idl.Substep_START_AGENTS, func(stream step.OutStreams) error {
-		return h.startAgentsSubStep(stream)
+	s.Run(idl.Substep_START_AGENTS, func(_ step.OutStreams) error {
```
berlin-ab: nit: it seems strange to not have a domain concept to represent the substep, even if it is a very simple function passthrough.

test/initialize.bats
```diff
@@ -164,6 +176,77 @@ outputContains() {
...
+}
+
+@test "start agents fails if a process is connected on the same TCP port" {
+    # Ensure the agent is down, so that we can test port in use.
+    gpupgrade kill-services
+    rm -r "$GPUPGRADE_HOME"
+
+    # squat gpupgrade agent port
+    AGENT_PORT=6416
+    go run ./testutils/port_listener/main.go $AGENT_PORT 3>&- &
```
berlin-ab: nit: is there any benefit to adding this to the build in the `Makefile` (ex: to get faster feedback on compilation errors)

test/initialize.bats
```diff
@@ -164,6 +176,77 @@ outputContains() {
...
+@test "start agents idempotent" {
+    run gpupgrade initialize \
+        --old-bindir="$GPHOME/bin" \
+        --new-bindir="$GPHOME/bin" \
+        --old-port="${PGPORT}" \
+        --disk-free-ratio 0 \
+        --stop-before-cluster-creation 3>&-
+    [ "$status" -eq 0 ] || fail "failed to initialize: $output"
+
+    sed  -i -e 's/"START_AGENTS": "COMPLETE"/"START_AGENTS": "FAILED"/g' "$GPUPGRADE_HOME/status.json"
```
berlin-ab: I'm very confused at the need for an acceptance level test to have such low level detail. How does this situation arise in the real world?

hub/initialize.go
dkrieger-pvtl: Yeah, we went back and forth on this but settled on eliminating the simple wrapper function.  It also helped us to realize that we're not actually using the stream.

hub/initialize.go
```diff
@@ -114,54 +116,3 @@ func getAgentPath() (string, error) {
 
 	return filepath.Join(filepath.Dir(hubPath), "gpupgrade"), nil
 }
-
-// TODO: use the implementation in RestartAgents() for this function and combine them
-func (h *Hub) startAgentsSubStep(stream step.OutStreams) error {
```
berlin-ab: suggestion: This looks to be the last usage of `ExecuteOnAllHosts`. This commit should probably remove it as well.

berlin-ab: same for `CheckClusterError`

test/initialize.bats
dkrieger-pvtl: I don't want to add test code to the main compile; that's what running tests is for.  This code is test-only.

test/initialize.bats
dkrieger-pvtl: This is a simple way to test idempotence...by setting the step status to `FAILED` we emulated an actual error during a START_AGENTS substep.  This is independent of needed to come up with a mechanism for the failure.  The next test, which grabs the agent port, is another such test but uses the explicit failure of another process grabbing its port.

test/initialize.bats
dkrieger-pvtl: done

testutils/port_listener/main.go
dkrieger-pvtl: done

hub/initialize.go
dkrieger-pvtl: I think @jchampio is working on something similar; I'll make a story for this and we can clean it up later.

testutils/port_listener/main.go
```diff
@@ -0,0 +1,86 @@
...
+	args := os.Args[1:]
+	if len(args) != 1 {
+		log.Fatalf("usage: %s <port_number>", os.Args[0])
+	}
+
+	port := args[0]
+
+	signals := make(chan os.Signal, 1)
+	done := make(chan bool, 1)
+	signal.Notify(signals, syscall.SIGINT, syscall.SIGTERM)
```
jchampio: It looks like these signal handlers are installed to exit the program, which is already the [default behavior of INT and TERM signals in Go](https://golang.org/pkg/os/signal/#hdr-Default_behavior_of_signals_in_Go_programs). Is this code really needed?

testutils/port_listener/main.go
```diff
@@ -0,0 +1,86 @@
...
+import (
+	"fmt"
+	"log"
+	"net"
+	"os"
+	"os/signal"
+	"syscall"
+)
+
+func main() {
```
jchampio: This implementation is a lot more complex than I expected. What's wrong with just

```go
func main() {
    // ... get port ...
    if _, err := net.Listen("tcp", ":"+port); err != nil {
        log.Fatalf("listen: %+v", err)
    }

    for {
    }
}

Comment (kalensk) [reply testutils/port_listener/main.go]: One consideration is that we need some way to not listen on ipv6 since the current CI infrastructure does not support that.

Comment (jchampio) [reply testutils/port_listener/main.go]: That's what `net.Listen("tcp", ...)` is documented to do. Did it not work?

Comment (kalensk) [reply testutils/port_listener/main.go]: When listening on `tcp` lsof showed the port was in use by ipv6
```
 → lsof -Pi :12345
COMMAND   PID      USER   FD   TYPE             DEVICE SIZE/OFF NODE NAME
main    33354 gpadmin    3u  IPv6 0xdba4de2a3bf587a3      0t0  TCP *:12345 (LISTEN)

When listening on `tcp4` and `tcp6` lsof showed the port was in use by ipv4 and ipv6
```
 → lsof -Pi :12345
COMMAND   PID      USER   FD   TYPE             DEVICE SIZE/OFF NODE NAME
main    35922 gpadmin    3u  IPv4 0xdba4de2a3c158f2b      0t0  TCP *:12345 (LISTEN)
main    35922 gpadmin    4u  IPv6 0xdba4de2a3bf587a3      0t0  TCP *:12345 (LISTEN)

```

Updating the code to what you suggested the test still passes locally and on the CI. I'll submit a PR shortly since it is cleaner and less code. Thanks!

jchampio: > I'll submit a PR shortly since it is cleaner and less code. Thanks!

👍 

> lsof showed the port was in use by ipv6

I suspect `lsof` is lying to you (or at least not showing the full picture) -- it shows the same on my machine, but `netstat` shows that it's actually a dual-stack socket:
```
$ lsof -i :8080
COMMAND   PID      USER   FD   TYPE             DEVICE SIZE/OFF NODE NAME
main    46464 pchampion    3u  IPv6 0x692f6533055cf093      0t0  TCP *:http-alt (LISTEN)

$ netstat -na | grep 8080
tcp46      0      0  *.8080                 *.*                    LISTEN
```

kalensk: See PR https://github.com/greenplum-db/gpupgrade/pull/198

PR #198, opened 2020-02-03
Simplify port listener

kalensk:
Simplify port listener code.

The "gpupgrade:port_listener" test pipeline is passing.

### DIFF
--- a/testutils/port_listener/main.go
+++ b/testutils/port_listener/main.go
@@ -1,5 +1,6 @@
-// The port_listener utility listens on the specified port on all ipv4 and ipv6
-// interfaces. If no ipv6 interface is detected it will not listen on ipv6.
+// The port_listener utility listens on the specified port even during gRPC
+// retries. One can inspect the ports in use on OS X with `netstat -an` rather
+// than `lsof -Pi`.
 // Usage:
 //
 //   go run port_listener <port>
@@ -11,8 +12,6 @@ import (
 	"log"
 	"net"
 	"os"
-	"os/signal"
-	"syscall"
 )
 
 func main() {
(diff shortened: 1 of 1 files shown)


berlin-ab:
Nice simplification.

kalensk:
> However, at least on my Mac(OSX High Sierra), it fails to bind to ipv4.

Right, see [the comment here](https://github.com/greenplum-db/gpupgrade/pull/196#discussion_r374397466) in PR #196

dkrieger-pvtl:
> Right, see [the comment here](https://github.com/greenplum-db/gpupgrade/pull/196#discussion_r374397466) in PR #196

Ok, then this change does the same thing in much less code.  Please add a few lines of comments from the [motivating comment](https://github.com/greenplum-db/gpupgrade/pull/196#discussion_r374397466) to this file, so we don't confuse ourselves again using `lsof` instead of `netstat`.

kalensk:
> Please add a few lines of comments from the [motivating comment](https://github.com/greenplum-db/gpupgrade/pull/196#discussion_r374397466) to this file, so we don't confuse ourselves again using `lsof` instead of `netstat`.

Do you have any specific suggestions on how to do this without simply rephrasing the man pages or knowing to use the the appropriate tool for the job? I am a little hesitant to add the requested comment since to me it seems orthogonal to the actual port_listener utility. I am not used to seeing comments like that.

However, I am curious what some other people's thoughts are, and more than happy to add something if other people feel it would be useful.

dkrieger-pvtl:
> Do you have any specific suggestions on how to do this without simply rephrasing the man pages or knowing to use the the appropriate tool for the job?

I cannot find anything relevant in the `lsof` manpage on this, hence my request for a comment like:

```
Use "netstat -na | grep PORT" to see the ports listened to by this program.  
"lsof -Pi PORT" only shows tcp6 on Mac OSX High Sierra, for instance, 
but netstat shows "tcp46".
```  

 > or knowing to use the the appropriate tool for the job? 

Unless there's something documented in the manpage on why `lsof` is not appropriate to use(or is considered general knowledge), I am not sure what "appropriate" means here.  There are often multiple utilities one can use for a given task.

My overall point is that another developer(like me in the future) might still use lsof, so to me the comment in a test utility is worth it.

dkrieger-pvtl:
> lgtm. If you want to be specific, you could specify the IP version to use, i.e.
> 
> ```
> _, err := net.Listen("tcp4", ":"+port)
> _, err := net.Listen("tcp6", ":"+port)
> ```
> 
> But what you have currently is fine. I confirmed that specifying just "tcp" listens on both ipv4 and ipv6 on mac, centos and ubuntu.

@marksliva Did you use `netstat` to determine this or lsof?

kalensk:
Added the comment ` One can inspect the ports in use on OS X with `netstat -an` rather than `lsof -Pi`.`

marksliva:
> > lgtm. If you want to be specific, you could specify the IP version to use, i.e.
> > ```
> > _, err := net.Listen("tcp4", ":"+port)
> > _, err := net.Listen("tcp6", ":"+port)
> > ```
> > 
> > 
> > But what you have currently is fine. I confirmed that specifying just "tcp" listens on both ipv4 and ipv6 on mac, centos and ubuntu.
> 
> @marksliva Did you use `netstat` to determine this or lsof?

I copied this code into 3 separate scripts. I started the first one that listened on "tcp", and attempting to start the 2 other scripts that liston on either "tcp4" and "tcp6" failed to start. I think I also ran `netstat -na` but don't remember what its output looked like.

PR #199, opened 2020-02-04
Add --delete flag to rsync for master data directory command

bhuvnesh2703:
Ideally in the PR: https://github.com/greenplum-db/gpupgrade/pull/195, the --delete flag should have been included. Adding this via this separate PR.

If there are extraneous files in the target directory they should be deleted with rsync. For instance, when upgrade process to copy is in progress and it fails, re-run of the upgrade should delete all the extra files which might exist or were left behind with the last failed step.

### DIFF
--- a/test/execute.bats
+++ b/test/execute.bats
@@ -130,7 +130,16 @@ reset_master_and_primary_pg_control_files() {
     # ensure that initialize created a backup and upgrade master refreshed the
     # target master data directory with the backup.
     rm -rf "${datadir}"/qddir_upgrade/demoDataDir-1/*
+    
+    # create an extra file to ensure that its deleted during rsync as we pass
+    # --delete flag
+    mkdir "${datadir}"/qddir_upgrade/demoDataDir-1/base_extra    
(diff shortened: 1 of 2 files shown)


test/execute.bats
```diff
@@ -130,7 +130,17 @@ reset_master_and_primary_pg_control_files() {
...
+    
+    # create an extra file to ensure that its deleted during rsync as we pass
+    # --delete flag
+    mkdir "${datadir}"/qddir_upgrade/demoDataDir-1/base_extra    
+    touch "${datadir}"/qddir_upgrade/demoDataDir-1/base_extra/1101
     gpupgrade execute --verbose
+    
+    # check that the extraneous files are deleted
+    [ ! -f "${datadir}"/qddir_upgrade/demoDataDir-1/base_extra/1101 ]
+    [ ! -d "${datadir}"/qddir_upgrade/demoDataDir-1/base_extra ]
```
jchampio: nit: technically this is all that's needed; the `-f` check is superseded by it.

PR #200, opened 2020-02-04
move SHUTDOWN_SOURCE_CLUSTER from initialize to execute

dkrieger-pvtl:
The source cluster should be up after initialize completes, as
the user may want to run initialize well in advance of execute.
Hence, the source cluster would still be used in production after
initialize completes.

Tests have been updated to reflect this change.

### DIFF
--- a/test/helpers.bash
+++ b/test/helpers.bash
@@ -26,6 +26,11 @@ skip_if_no_gpdb() {
     [ -n "${PGPORT}" ] || skip "this test requires an active GPDB cluster (set PGPORT)"
 }
 
+# start_source_cluster() ensures that database is up before returning
+start_source_cluster() {
+    "${GPHOME}"/bin/pg_isready -q || "${GPHOME}"/bin/gpstart -a
+}
+
 # Calls gpdeletesystem on the cluster pointed to by the given master data
 # directory.
 delete_cluster() {
--- a/test/initialize.bats
+++ b/test/initialize.bats
@@ -50,7 +50,6 @@ teardown_target_cluster() {
 }
 
 setup_check_upgrade_to_fail() {
-    gpstart -a
     $PSQL -d postgres -p $PGPORT -c "CREATE TABLE test_pg_upgrade(a int) DISTRIBUTED BY (a) PARTITION BY RANGE (a)(start (1) end(4) every(1));"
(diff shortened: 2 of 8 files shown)


test/helpers.bash
```diff
@@ -26,6 +26,11 @@ skip_if_no_gpdb() {
     [ -n "${PGPORT}" ] || skip "this test requires an active GPDB cluster (set PGPORT)"
 }
 
+# start_source_database() ensures that database is up before returning
+start_source_database() {
```
kalensk: Does it make more sense to rename this to `start_source_cluster` instead of database? 

Just a note that since gpstart is being used from whatever greenplum_path was sourced, are we confident this is the source cluster and people in the future won't trip over themselves. I think it would make more sense to call to be explicit when calling gpstart with `$GPHOME_OLD/bin/gpstart` or similar?

test/initialize.bats
```diff
@@ -270,3 +269,10 @@ wait_for_port_change() {
     # for check_upgrade, we always run it.
     [ "$status" -eq 1 ] || fail "$output"
 }
+
+@test "the source cluster is running at the end of initialize" {
+    set_target_cluster_var_for_teardown
+    TEARDOWN_FUNCTIONS+=( teardown_target_cluster )
```
kalensk: Probably not part of this PR, but why the need to separate out the `set` and the teardown. Seems odd to me.

dkrieger-pvtl: Not sure, but we just copied-and-pasted and deferred thinking about this.

test/helpers.bash
dkrieger-pvtl: For most bats tests, we require the source cluster to be running and available, so I thought that GPHOME, MASTER_DATA_DIRECTORY, etc, are all set to that for the source cluster.   We have other code in bats tests that depend on this, for instance `skip_if_no_gpdb()`

kalensk: Any thoughts on renaming it to `start_source_cluster` since the substep is called `Substep_SHUTDOWN_SOURCE_CLUSTER` and greenplum is a cluster?

dkrieger-pvtl: `start_source_cluster` is consistent with other functions in our bats tests.  Done.

dkrieger-pvtl: I now use "$GPHOME/bin" as the prefix in this bash function.

PR #201, opened 2020-02-05
Move CLI commands back into the cli package

berlin-ab:
Most of the code in Command()s (now HubCommand() and AgentCommand()) are cli related. Only the construction and calling of the Agent and Hub objects are specific to their subdomains.

Some of the logic around config files can probably be pushed back down into hub.

### DIFF
--- a/cli/commands/agent.go
+++ b/cli/commands/agent.go
@@ -1,16 +1,18 @@
-package agent
+package commands
 
 import (
 	"github.com/greenplum-db/gp-common-go-libs/cluster"
 	"github.com/greenplum-db/gp-common-go-libs/gplog"
 	"github.com/spf13/cobra"
 
+	"github.com/greenplum-db/gpupgrade/agent"
+
 	"github.com/greenplum-db/gpupgrade/utils"
 	"github.com/greenplum-db/gpupgrade/utils/daemon"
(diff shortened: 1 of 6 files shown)


berlin-ab:
https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:move-cli-commands-into-cli-package

cli/commands/agent.go
```diff
@@ -1,16 +1,18 @@
...
 
+	"github.com/greenplum-db/gpupgrade/agent"
+
 	"github.com/greenplum-db/gpupgrade/utils"
 	"github.com/greenplum-db/gpupgrade/utils/daemon"
 	"github.com/greenplum-db/gpupgrade/utils/log"
 )
 
-func Command() *cobra.Command {
+func agentCommand() *cobra.Command {
```
kalensk: What about just calling this `agent` since we are already in the Command package?

jchampio: > what about just calling this `Agent`

+1 to `agent` or similar

> or maybe `Command`

Is this what you meant to type?

kalensk: > Is this what you meant to type?

I think I prefer `Agent`, but was slightly waffling since it's in the command package, but in the agent file.

PR #202, opened 2020-02-05
execute: move master-to-segment copy to segment upgrade

jchampio:
...to make the overall execute implementation restartable.

Review per-commit for the best context.


agent/upgrade_primary.go
```diff
@@ -0,0 +1,72 @@
...
+
+//
+// Functions to perform an upgrade of a single segment:
+//
+
+func upgradeSegment(segment Segment, request *idl.UpgradePrimariesRequest, host string) error {
+	err := restoreBackup(request, segment)
+
+	if err != nil {
+		return err
```
berlin-ab: nit: we should consider adding some more context to this error

hub/execute.go
```diff
@@ -11,6 +14,8 @@ import (
 )
 
 func (h *Hub) Execute(request *idl.ExecuteRequest, stream idl.CliToHub_ExecuteServer) (err error) {
+	masterBackupDir := filepath.Join(h.StateDir, "master.bak")
```
kalensk: Can we use the `hub.masterBackup` variable?

berlin-ab: It's currently not a variable:
https://github.com/greenplum-db/gpupgrade/blob/c11a07b8c3c94d1825b793458f47f20bfa9a0868/hub/hub.go#L39

Also, it was only useful in the context of the copy to segment host & upgrade segment steps, not the entire hub, so we chose to limit its scope.

berlin-ab: Ah, I see what you mean.

agent/rsync_test.go
```diff
@@ -0,0 +1,176 @@
...
+
+		if err := agent.Rsync(sourceDir, targetDir, []string{}); err != nil {
+			t.Errorf("Rsync() returned error %+v", err)
+		}
+
+		targetContents, _ := ioutil.ReadFile(targetDir + "/file-that-should-get-removed")
+
+		// XXX this checks that the file is either empty or does not exist; we
+		// should just check for existence
+		if bytes.Compare(targetContents, []byte("")) != 0 {
```
kalensk: This test _and_ the one below uses bytes.Compare, when we really want to test if the file does not exist. Seems like something we want to address before merging?

kalensk: A few tests later use os.stat, I am curious why don't these test use it as well?

agent/rsync_test.go
```diff
@@ -0,0 +1,176 @@
...
+	err := ioutil.WriteFile(filepath, contents, 0644)
+
+	if err != nil {
+		t.Fatalf("error writing file '%v'", filepath)
+	}
+}
+
+func TestRsync(t *testing.T) {
+	if _, err := exec.LookPath("rsync"); err != nil {
+		t.Skipf("tests require rsync (%v)", err)
```
marksliva: How do we plan to get this running in CI? One option would be to move it into the `integrations` test package. Or add rsync to the container. It's not a blocker for this PR, but we should make a story for it.

agent/rsync_test.go
```diff
@@ -0,0 +1,176 @@
...
+		if os.IsNotExist(statError) {
+			t.Error("target directory file 'file-that-should-get-ignored' should still exist, but it does not")
+		}
+
+		_, statError = os.Stat(filepath.Join(targetDir, "another-file-that-should-get-ignored"))
+
+		if os.IsNotExist(statError) {
+			t.Error("target directory file 'another-file-that-should-get-ignored' should still exist, but it does not")
+		}
+
```
kalensk: Am I understanding this test correctly....I feel like these two assertions are backwards. (Shouldn't there be a `!` in front of them?) That is, if `"file-that-should-get-ignored"` exists, error out. Can you help me understand what I am missing?

agent/upgrade_primaries_test.go
```diff
@@ -127,4 +127,176 @@ func TestUpgradePrimary(t *testing.T) {
...
+		defer ResetCommands()
+
+		request := buildRequest(pairs)
+		request.CheckOnly = true
+
+		agent.SetRsyncCommand(
+			exectest.NewCommandWithVerifier(agent.Success, func(commandName string, _ ...string) {
+				if commandName == "rsync" {
+					t.Error("unexpected rsync call")
+				}
```
kalensk: Curious why this assertion is not a bit more generic with something like:

```go
if commandName != "pg_upgrade" {
   t.Errorf("expected pg_upgrade to be called, got %q instead", commandName)
}
```

agent/upgrade_primaries_test.go
```diff
@@ -127,4 +127,176 @@ func TestUpgradePrimary(t *testing.T) {
...
+			}
+
+			if !strings.Contains(string(exitError.Stderr), "rsync failed cause I said so") {
+				t.Errorf("wanted error message 'rsync failed cause I said so' from rsync, got %q",
+					string(exitError.Stderr))
+			}
+		}
+	})
+
+	t.Run("it grabs a copy of the master backup directory before running upgrade", func(t *testing.T) {
```
kalensk: Nice test!

agent/upgrade_primaries.go
```diff
@@ -7,106 +7,92 @@ import (
...
 	gplog.Info("agent starting %s", idl.Substep_UPGRADE_PRIMARIES)
 
-	err := UpgradePrimary(s.conf.StateDir, request)
+	err := UpgradePrimaries(s.conf.StateDir, request)
+
 	return &idl.UpgradePrimariesReply{}, err
 }
 
+// Allow exec.Command to be mocked out by exectest.NewCommand.
+var execCommand = exec.Command
```
marksliva: Is there a reason this was named generically? There were a few layers to go through when looking for it. It could be renamed to something like `execPgUpgradeCommand`.

agent/upgrade_primary.go
```diff
@@ -0,0 +1,72 @@
...
+	if request.CheckOnly {
+		options = append(options, upgrade.WithCheckOnly())
+	}
+
+	if request.UseLinkMode {
+		options = append(options, upgrade.WithLinkMode())
+	}
+
+	return upgrade.Run(segmentPair, options...)
+}
```
kalensk: I am loving the way this file and function turned out. Super easy to follow and read. Some great effort leading up to this.

agent/upgrade_primaries.go
berlin-ab: It's not a new addition, just a move. `execCommand` is the usual pattern, and we've deviated from it with `rsyncCommand` to disambiguate.

agent/rsync_test.go
kalensk: Makes sense. I think we should update the variable names to be a bit more self-documenting. Something like `source-file-that-should-be-ignored`, or `target-file-should-not-be-deleted`

agent/upgrade_primaries_test.go
```diff
@@ -127,4 +127,176 @@ func TestUpgradePrimary(t *testing.T) {
...
+	deleteFlag  string
+	exclusions  []string
+}
+
+func rsyncCall(utility string, arguments []string) rsyncRequest {
+	r := rsyncRequest{}
+	r.commandName = utility
+
+	if len(arguments) > 0 {
+		r.archiveFlag = arguments[0]
```
marksliva: nit: we store this variable but don't check for it in any of the tests at this level.

agent/upgrade_primaries_test.go
```diff
@@ -127,4 +127,176 @@ func TestUpgradePrimary(t *testing.T) {
...
+	}
+
+	if len(arguments) > 1 {
+		r.deleteFlag = arguments[1]
```
marksliva: same as above.

hub/copy_master_test.go
```diff
@@ -49,75 +66,142 @@ func TestCopyMaster(t *testing.T) {
...
 		}
 
-		hub.agentConns = agentConns
-
-		err := hub.CopyMasterDataDir(DevNull)
-		g.Expect(err).ToNot(HaveOccurred())
+		close(hosts)
 
-		g.Eventually(func() int { return testExecutor.NumExecutions }).Should(Equal(1))
+		// Collect the hostnames for validation.
```
kalensk: Again personal preference, but this comment isn't saying much to me. I'd consider removing it.

hub/copy_master_test.go
```diff
@@ -49,75 +66,142 @@ func TestCopyMaster(t *testing.T) {
...
-			gomock.Any(),
-		).Return(&idl.CopyMasterReply{}, nil)
-
-		agentConns := []*Connection{
-			{nil, client, "host1", nil},
+		// The verifier function can be called in parallel, so use a channel to
+		// communicate which hosts were actually used.
+		hosts := make(chan string, len(targetCluster.PrimaryHostnames()))
+
+		// Validate the rsync call and arguments.
```
kalensk: Again personal preference, but this comment isn't saying much to me given the code. I'd consider removing it.

agent/upgrade_primaries.go
kalensk: Just for clarity, when we can we usually try and disambiguate the various execComamnd calls in the way mark is asking for. It makes things clearer.

kalensk:
Looks good overall. Some various comments and refinements.

Great work and effort on this and the various PRs helping us get here!

agent/upgrade_primaries.go
jchampio: > Just for clarity, when we can we usually try and disambiguate the various execComamnd calls in the way mark is asking for. It makes things clearer.

Eh? We're trying to _remove_ the proliferation. `execCommand` is the first and ideally only insertion point; otherwise the cost of the system explodes. As Adam points out, this wasn't a new addition.

agent/upgrade_primaries_test.go
berlin-ab: I wanted to leave them, so that the following argument extractions made sense.

agent/upgrade_primaries.go
marksliva: Good point. Taking a step back, if we didn't need to stub them for testing, all of the `execCommand` declarations would go away. `execCommand` will be the only name left once we have a way to stub multiple calls to the same `execCommand` within a single execution flow. With that end goal in mind, we can keep the current naming scheme.

agent/rsync_test.go
berlin-ab: Good question. I think we should probably add rsync to the container.

jchampio: It is running in CI, just not in the `noinstall` case. Unfortunately `ginkgo` hides it from us entirely. We should probably move to `go test` in the CI, and I'll add a chore.

bhuvnesh2703:
question:
Once the master is upgraded -> It copies the master data dir to the segment hosts (i.e the upgraded master data directory) and it creates a backup of that under the state directory / master.bak. Could you please confirm this? If the master and the segment are on the same hosts, it will modify the old master.bak (as now it will have the upgraded master data dir) which was taken from the newly created target cluster in initialize? Is this correct?

agent/rsync.go
```diff
@@ -0,0 +1,26 @@
...
+
+func Rsync(sourceDir, targetDir string, excludedFiles []string) error {
+	arguments := append([]string{
+		"--archive", "--delete",
+		sourceDir + "/", targetDir,
+	}, makeExclusionList(excludedFiles)...)
+
+	_, err := rsyncCommand("rsync", arguments...).Output()
+
+	return err
```
marksliva: The error message when rsync fails could be improved. When I force an rsync exception, I get an error that looks like:
```
[ERROR]:-execute: substep "execute": 1 error occurred:
	* gpupgrade agent failed to convert primary segment on host MacBook-Pro-2.local: rpc error: code = Unknown desc = upgrading primaries: 3 errors occurred:
	* exit status 1
	* exit status 1
	* exit status 1
```

berlin-ab:
> question:
> Once the master is upgraded -> It copies the master data dir to the segment hosts (i.e the upgraded master data directory) and it creates a backup of that under the state directory / master.bak. Could you please confirm this? If the master and the segment are on the same hosts, it will modify the old master.bak (as now it will have the upgraded master data dir) which was taken from the newly created target cluster in initialize? Is this correct?

Great question @bhuvnesh2703. I pushed a fixup commit that creates a separate backup for the upgraded master.

In practice you shouldn't run initialize after starting to run execute, but we don't have anything preventing that from happening now. 

The cost of having a second backup is small, as the master's footprint is small.

agent/rsync.go
berlin-ab: We've hopefully improved this with our latest push. Would you mind giving it for a spin and seeing if things are improved?

jchampio:
Rebased and squashed on latest master.

agent/rsync.go
marksliva: It looks good now.
```
Execute:
    github.com/greenplum-db/gpupgrade/cli/commanders.Execute
    /Users/pivotal/go/src/github.com/greenplum-db/gpupgrade/cli/commanders/steps.go:88
  - rpc error: code = Unknown desc = substep "execute": 1 error occurred:
	* gpupgrade agent failed to convert primary segment on host MacBook-Pro-2.local: rpc error: code = Unknown desc = upgrading primaries: 3 errors occurred:
	* failed to restore master data directory backup on host MacBook-Pro-2.local for content id 2: rsync: --delete-foo: unknown option
rsync error: syntax or usage error (code 1) at /BuildRoot/Library/Caches/com.apple.xbs/Sources/rsync/rsync-52.200.1/rsync/main.c(1337) [client=2.6.9]
: rsync: --delete-foo: unknown option
rsync error: syntax or usage error (code 1) at /BuildRoot/Library/Caches/com.apple.xbs/Sources/rsync/rsync-52.200.1/rsync/main.c(1337) [client=2.6.9]

	* failed to restore master data directory backup on host MacBook-Pro-2.local for content id 0: rsync: --delete-foo: unknown option
rsync error: syntax or usage error (code 1) at /BuildRoot/Library/Caches/com.apple.xbs/Sources/rsync/rsync-52.200.1/rsync/main.c(1337) [client=2.6.9]
: rsync: --delete-foo: unknown option
rsync error: syntax or usage error (code 1) at /BuildRoot/Library/Caches/com.apple.xbs/Sources/rsync/rsync-52.200.1/rsync/main.c(1337) [client=2.6.9]

	* failed to restore master data directory backup on host MacBook-Pro-2.local for content id 1: rsync: --delete-foo: unknown option
rsync error: syntax or usage error (code 1) at /BuildRoot/Library/Caches/com.apple.xbs/Sources/rsync/rsync-52.200.1/rsync/main.c(1337) [client=2.6.9]
: rsync: --delete-foo: unknown option
rsync error: syntax or usage error (code 1) at /BuildRoot/Library/Caches/com.apple.xbs/Sources/rsync/rsync-52.200.1/rsync/main.c(1337) [client=2.6.9]
```

berlin-ab:
Woot!

PR #203, opened 2020-02-06
Move VersionString out of utils to where it is used

berlin-ab:
- clean up junk in utils
- removes some dead code along the way


cli/commands/version.go
```diff
@@ -0,0 +1,12 @@
+package commands
+
+// This global var UpgradeVersion should have a value set at build time.
+// see Makefile for -ldflags "-X etc"
+var UpgradeVersion = ""
+
+func VersionString(executableName string) string {
+	if UpgradeVersion == "" {
+		return executableName + " unknown version"
+	}
```
kalensk: I know this was here before, but I feel like we should just return empty string if we don't know the version. Keep things simple.

berlin-ab: As this effects the user-facing product, let's separate these two issues. I know it's not a big user-facing detail. It is easy to introduce a new change later.

PR #205, opened 2020-02-06
Rename Hub -> Server to clarify responsibility.

berlin-ab:
This helps match how we've defined agent.Server


kalensk:
There are gRPC client connections (to the agent) within the Hub, and this rename makes that less obvious and more confusing to me. I think some more clarity and discussion is needed.

berlin-ab:
My goal is to bring consistency to what we're naming the server for Agent and Hub. Either we name them `agent.Server` and `hub.server` - or we name them `agent.Agent` and hub.Hub`. Right now we have them as `agent.Server` and `hub.Hub`.

berlin-ab:
> There are gRPC client connections (to the agent) within the Hub, and this rename makes that less obvious and more confusing to me. I think some more clarity and team discussion is needed.

Would you clarify this concern please? Where is a spot in the codebase that this change makes more confusing? I think it should be possible to make it clearer if there are confusions.

PR #206, opened 2020-02-06
Rename agent server receivers

berlin-ab:
Agent server receiver variables are not consistent. Sometimes they are `(a *Server)` and sometimes they are`(s *Server)`. My goal is to make them consistent throughout the package.


berlin-ab:
"later"... is now :)

Issue #109, opened 2019-08-27
gometalinter is deprecated


jchampio:
[What it says on the tin.](https://github.com/alecthomas/gometalinter/issues/590)

PR #207, opened 2020-02-06
Remove linter

berlin-ab:
We're not using it, and it fails badly. Let's not continue to carry this
code with us.


marksliva:
Have we considered using https://github.com/golangci/golangci-lint ?
`gometalinter` is now deprecated, and suggests `golangci-lint` as a replacement.
I would be in favor of keeping linting around, since it does catch some useful things like `Error return value of `os.Mkdir` is not checked`.

berlin-ab:
I’m ok with using a linter if it provides us value. Our current setup does
not.

jchampio:
> I would be in favor of keeping linting around

Definitely, but [we've talked about that before](https://github.com/greenplum-db/gpupgrade/issues/109) and a move hasn't happened yet. Let's get rid of the cruft, and talk about what we want to replace it with too.

PR #208, opened 2020-02-07
execute: add a test for substep idempotence after success

jchampio:
...in the same vein as 3ddb2ea211.

This test will fail until https://github.com/greenplum-db/gpupgrade/pull/202 is merged.

### DIFF
--- a/test/execute.bats
+++ b/test/execute.bats
@@ -141,3 +141,29 @@ reset_master_and_primary_pg_control_files() {
 
     TEARDOWN_FUNCTIONS+=( reset_master_and_primary_pg_control_files )
 }
+
+# TODO: this test is a replica of one in initialize.bats. If/when we start to
+# make a third copy for finalize, decide whether the implementations should be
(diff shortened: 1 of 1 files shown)


test/execute.bats
```diff
@@ -141,3 +141,29 @@ reset_master_and_primary_pg_control_files() {
...
+    NEW_CLUSTER="$(gpupgrade config show --new-datadir)"
+
+    gpupgrade execute --verbose 3>&-
+
+    # Put the source and target clusters back the way they were.
+    gpstop -a -d "$NEW_CLUSTER"
+    gpstart -a 3>&-
+
+    # Mark every substep in the status file as failed. Then re-execute.
+    sed -i.bak -e 's/"COMPLETE"/"FAILED"/g' "$GPUPGRADE_HOME/status.json"
```
kalensk: Just an out loud note to myself, which I don't think it is an issue. This will mark the initialize steps as failed as well as the execute steps.

PR #209, opened 2020-02-07
Organize hub package by top level steps and dependent components

berlin-ab:
The hub package is starting to get large and difficult to understand how things relate. This is an initial attempt to start breaking up the hub package into sub components.

It creates sub-packages in the hub directory:

hub/cluster
hub/finalizestep
hub/initializestep
hub/steps 
hub/agent 

with these dependencies:

hub -> initializestep -> (step & agent)
hub -> finalizestep -> (step & cluster)
hub -> steps
hub -> cluster
hub -> agent

--

The key to doing this work is to decouple the Hub (Server) from the Steps and the sub-components. This keeps the gRPC handlers in the hub package, but moves the Steps and sub-components to their own packages.


berlin-ab:
I think this is a step in the right direction and is worth doing now. We can continue to move more of the steps in this direction.

berlin-ab:
I agree with Jacob in a conversation we had off-github. This decoupling could be done within the same package as `hub`. One advantage we get by subpackages is the compiler helps to drive the decoupling by not allowing circular dependencies.

berlin-ab:
Step one for me has been to achieve the decoupling.

Step two will be to take a second pass to make sure things are:
1. Named well (https://golang.org/doc/effective_go.html#package-names)
2. Actually in the right place.

berlin-ab:
I've found it to be easier to start at the leaf nodes when decoupling, rather than starting at the entry-point.

For example, rather than starting at `func (s *Server) Initialize(...)`, I started with `func RestartAgents(...)` (the last of the substeps for initialize) and started working my way backward.

PR #210, opened 2020-02-08
Experiment with golangci-lint

jchampio:
Because `golangci/golangci-lint` provides its own Docker image, this is pretty easy to slot into Concourse. [Here's some example output in a failed job](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:dev:lint/jobs/lint/builds/3), for absolutely no effort put into linter configuration.

Obviously we wouldn't merge this without first fixing the issues, but what do we think?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -91,6 +91,21 @@ installcheck:
 		@echo "# FIXME: Make, if run in parallel, hangs after test completes."
 		./installcheck.bats
 
+# To lint, you must install golangci-lint via one of the supported methods
+# listed at
+#
+#     https://github.com/golangci/golangci-lint#install
+#
+# DO NOT add the linter to the project dependencies in Gopkg.toml, as much as
+# you may want to streamline this installation process, because
+# 1. `go get` is an explicitly unsupported installation method for this utility,
+#    much like it is for gpupgrade itself, and
+# 2. adding it as a project dependency opens up the possibility of accidentally
+#    vendoring GPL'd code.
+.PHONY: lint
+lint:
+	golangci-lint run
+
 clean:
 		# Build artifacts
 		rm -f gpupgrade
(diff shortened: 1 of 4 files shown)


berlin-ab:
Some useful things in there. I'm curious how you think we could work it into our workflow so we can get this feedback quickly. I'm worried that if we only have it on CI, I won't learn about a linting mistake until long after I've made the mistake.

jchampio:
> I'm curious how you think we could work it into our workflow so we can get this feedback quickly.

Should we adopt this, I'd recommend that you add it to your editor/IDE, similar to how we suggest that you add `goimports` to your workflow.

dkrieger-pvtl:
I really like the idea and implementation of this PR.  The `golangci-lint` allows exclusions of certain errors and incremental runs(lint issues found only in your code).  

We'd have to decide on a process here for the deployment...I'm guessing we would commit the non-draft version of this PR but with the actual pipeline job disabled, fix the issues/exclude ones we disagree with and commit that as a separate PR, and then enable the pipeline job.

jchampio:
> We'd have to decide on a process here for the deployment...I'm guessing we would commit the non-draft version of this PR but with the actual pipeline job disabled, fix the issues/exclude ones we disagree with and commit that as a separate PR, and then enable the pipeline job.

Yeah, this gets my vote.

> Curiously to me, it doesn't by default flag neglecting to check the error code of Close defer file.Close() but finds many others instances of not checking error codes. Perhaps there's a flag we can throw.

Apparentely, `Close()` is one of the default exclusions because code bases are so bad at checking these things across the board, and `golangci-lint` is explicitly designed to give you as little false positives as possible out of the gate. That knob can be turned back.

jchampio:
Closing for now; I expect to reopen this very soon.

PR #211, opened 2020-02-10
Remove dependency on cluster.Cluster

jchampio:
Since we'll need to perform major surgery on `Cluster` to get mirrors in, copy its current implementation into the codebase. This implementation should be merged with `utils.Cluster` at a later date.

The first commit is in two parts for ease of review -- the first part is the code motion, and the second part is the removal of the now-unused `Executor` concept so that the tests compile again. (I chose not to bother getting the first part compiling, since I knew we wanted to get rid of `Executor` anyway; if this PR is approved, I will squash the two pieces before pushing.)

The final commit is the removal of previously acquired debt: without `Executor`, we no longer need any custom serialization code for `hub.Config`.


jchampio:
Thanks for the review!

PR #212, opened 2020-02-11
Upgrade the master standby from the hub during Finalize

berlin-ab:
Adds substep to Finalize called UPGRADE_STANDBY to the Hub, which calls off to the
gpdb utility gpinitstandby to set up a master standby.

To enable idempotency, we first remove any existing standbys from the cluster, and then perform `gpinitstandby`

- extract body of Finalize server handler to improve testability
- decouple finalize logic from server
- create testing seams for ReconfigurePorts logic
- pull up the creation of the SubstepStateStore to be able to observe
  its state from tests
- move clone_ports_from_cluster tests out of finalize_test.go into
  their own file. Same for logic that lived in finalize.go


utils/cluster.go
```diff
@@ -166,3 +167,15 @@ func (c Cluster) SegmentsOn(hostname string) ([]cluster.SegConfig, error) {
 
 	return segments, nil
 }
+
```
berlin-ab: These are hardcoded values waiting for https://github.com/greenplum-db/gpupgrade/pull/213 to be completed.

test/finalize-standby.bats
```diff
@@ -0,0 +1,60 @@
...
+#
+@test "finalize brings up the standby for the new cluster" {
+    gpupgrade initialize \
+        --old-bindir="$GPHOME/bin" \
+        --new-bindir="$GPHOME/bin" \
+        --old-port="${PGPORT}" \
+        --disk-free-ratio 0 \
+        --verbose
+
+    gpupgrade execute --verbose
```
jchampio: Does the fake cluster creation logic in `finalize.bats` help you at all here?

test/finalize-standby.bats
```diff
@@ -0,0 +1,60 @@
...
+
+    local new_datadir=$(gpupgrade config show --new-datadir)
+    local actual_standby_status=$(gpstate -d "${new_datadir}")
+    local standby_status_line=$(get_standby_status $actual_standby_status)
+    [[ $standby_status_line == *"Standby host passive"* ]] || fail "expected standby to be up and in passive mode, got **** ${actual_standby_status} ****"
+}
+
+get_standby_status() {
+    local standby_status=$1
+    echo $(echo $standby_status | grep 'Standby master state')
```
jchampio: The outer echo seems unnecessary here.

hub/upgrade_standby_test.go
```diff
@@ -0,0 +1,72 @@
...
+			utilityName = commandName
+
+			// gpinitstandby -P [some port] -s [some hostname] -S [some data directory]
+			portArgument = arguments[1]
+			hostnameArgument = arguments[3]
+			dataDirectoryArgument = arguments[5]
+
+			if len(arguments) > 6 {
+				automaticArgument = arguments[6]
+			}
```
jchampio: This test will be a good place to use `flag.FlagSet` directly. (The last example we went through together had an options use case that couldn't be handled by `flag`.)

hub/upgrade_standby.go
```diff
@@ -0,0 +1,44 @@
...
+	MasterDataDirectory string
+}
+
+func UpgradeStandby(standbyConfig StandbyConfig) error {
+	return upgradeStandbyFunc(standbyConfig)
+}
+
+func upgradeStandby(standbyConfig StandbyConfig) error {
+	os.Setenv("MASTER_DATA_DIRECTORY", standbyConfig.MasterDataDirectory)
+	os.Setenv("PGPORT", strconv.Itoa(standbyConfig.MasterPort))
```
jchampio: Set `cmd.Env` instead.

test/finalize-standby.bats
berlin-ab: I took a look at that, but wanted to know that I was using a real setup before I started faking things out.

hub/upgrade_standby.go
berlin-ab: Sweet, I was wondering if there were a way to avoid mucking with the global state.

hub/upgrade_standby_test.go
berlin-ab: Thanks for the suggestion. It took me a while to realize that I was accidentally using `flag` from golang and not `pflag`.

One downside I see to using `pflag` is that long-form options are not really valid for the underlying utility, but they're required as a part of `pflag` (from what I can see). This means that the unit test could pass, but a call-through to the utility would fail during argument parsing.

hub/upgrade_standby.go
berlin-ab: Interestingly, when I switched to the `cmd.Env` implementation, the end-to-end test started failing. I switched it back to `os.Setenv` for now.

I'm guessing (read, no idea) that gpinitstandby does not pass along its environment to the transitive utilities that it shells out to - expecting the parent environment to have set those variables.

hub/upgrade_standby_test.go
jchampio: > It took me a while to realize that I was accidentally using flag from golang and not pflag.

`flag` should be fine; is there a problem you ran into? `pflag` (as you discovered) is more flexible but also more than we need here.

berlin-ab: iirc, i had trouble setting up short-hand flags using `flag`.

hub/upgrade_standby.go
berlin-ab: Ha, false alarm. I realized that I was misusing the `fmt` library (not using `Sprintf`)

hub/greenplum_environment.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+func (e *greenplumEnv) MasterPort() int {
+	return e.masterPort
+}
+
+func (e *greenplumEnv) Run(utilityName string, arguments ...string) error {
+	commandAsString := exec.Command(
+		filepath.Join(e.binDir, utilityName), arguments...,
+	).String()
```
jchampio: [Beware!](https://golang.org/pkg/os/exec/#Cmd.String)
> String returns a human-readable description of c. It is intended only for debugging. **In particular, it is not suitable for use as input to a shell.**

berlin-ab:
Darn, it was so handy.

jchampio:
Draft subsumed by https://github.com/greenplum-db/gpupgrade/pull/218.

PR #213, opened 2020-02-11
Cluster: add mirrors concept

jchampio:
Support mirrors in our `Cluster` structs:
- `Cluster.Mirrors` maps content IDs to mirror configurations.
- `Cluster.Segments` is now `Cluster.Primaries`.
- `SegConfig` has a required `Role` member (set to either `"p"` or `"m"`).
- The master and standby are treated as a special-case primary and mirror, respectively.

Per-commit review is encouraged. Roadmap:
- migrate `cluster` tests from Ginkgo to table-driven Go style
- make the change to the `Cluster` API
- make the downstream changes necessary to consume the new API

### DIFF
--- a/hub/common_test.go
+++ b/hub/common_test.go
@@ -3,8 +3,11 @@ package hub
 import (
 	"io"
 	"io/ioutil"
+	"testing"
 
 	"github.com/greenplum-db/gpupgrade/testutils/exectest"
+	"github.com/greenplum-db/gpupgrade/utils"
+	"github.com/greenplum-db/gpupgrade/utils/cluster"
 )
 
 // Set it to nil so we don't accidentally execute a command for real during tests
@@ -49,4 +52,17 @@ type failingWriter struct {
 
 func (f *failingWriter) Write(_ []byte) (int, error) {
 	return 0, f.err
+}
+
+// MustCreateCluster creates a utils.Cluster and calls t.Fatalf() if there is
+// any error.
+func MustCreateCluster(t *testing.T, segs []cluster.SegConfig) *utils.Cluster {
+	t.Helper()
+
+	cluster, err := cluster.NewCluster(segs)
+	if err != nil {
+		t.Fatalf("%+v", err)
+	}
+
+	return &utils.Cluster{Cluster: cluster}
 }
--- a/utils/cluster/cluster.go
+++ b/utils/cluster/cluster.go
@@ -9,13 +9,27 @@ package cluster
  */
 
 import (
+	"errors"
+	"fmt"
+
 	"github.com/greenplum-db/gp-common-go-libs/dbconn"
 	"github.com/greenplum-db/gp-common-go-libs/gplog"
 )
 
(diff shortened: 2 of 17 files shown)


hub/common_test.go
```diff
@@ -50,3 +53,16 @@ type failingWriter struct {
 func (f *failingWriter) Write(_ []byte) (int, error) {
 	return 0, f.err
 }
+
+// MustCreateCluster creates a utils.Cluster and calls t.Fatalf() if there is
+// any error.
+func MustCreateCluster(t *testing.T, segs []cluster.SegConfig) *utils.Cluster {
```
bhuvnesh2703: why not CreateCluster

jchampio: `MustXXX()` is a marker for developers that the function will handle errors out-of-band (i.e. `panic`) instead of returning them. You typically only want to use a `Must` function in cases where the inputs are hardcoded and known to be correct. Tests are a good use; global initialization is another.

Some examples in the standard library:
- [`regexp.MustCompile`](https://golang.org/pkg/regexp/#MustCompile) vs [`regexp.Compile`](https://golang.org/pkg/regexp/#Compile)
- [`template.Must`](https://golang.org/pkg/text/template/#Must) vs [`template.New`](https://golang.org/pkg/text/template/#Parse)

berlin-ab: Thanks for the reference from the standard library. I always find those useful.

jchampio:
> Would now be a good time to move Cluster out of utils?

It's not a _bad_ time, but I worry that it might distract from the mirror upgrade implementation. I'll add a chore and hopefully we can get to it in the next week.

utils/cluster/cluster.go
```diff
@@ -24,40 +38,84 @@ type SegConfig struct {
...
+// ErrInvalidSegments is returned by NewCluster if the segment configuration
+// array does not map to a valid cluster.
+var ErrInvalidSegments = errors.New("invalid segment configuration")
+
 /*
  * Base cluster functions
  */
 
-func NewCluster(segConfigs []SegConfig) *Cluster {
+func NewCluster(segConfigs []SegConfig) (*Cluster, error) {
```
berlin-ab: I'm really happy about the error checking you do here. Initialization of the object ensures that it is valid.

berlin-ab:
As we introduce the concept of role, we might also want to introduce the idea of preferred role. 

Have we considered how the variability of role impacts gpupgrade?

jchampio:
> Have we considered how the variability of role impacts gpupgrade?

Yes, but the feature to check for problems hasn't been scheduled yet. @bharath1028 Is there a story yet to check for a correctly balanced cluster?

jchampio:
> As we introduce the concept of role, we might also want to introduce the idea of preferred role.

My strong preference is to not introduce this distinction unless we change gpupgrade to handle situations in which the two are different (which seems unlikely). The role-vs-preferred-role question has led to many subtle bugs in the Python code.

PR #214, opened 2020-02-12
gpupgrade link-mode: pg_upgrade --check on master needs --link

dkrieger-pvtl:
The original PR adding link mode #177  did not throw the `--link` mode flag for link-mode of gpupgrade.  However, according to the [pg_upgrade documentation](https://www.postgresql.org/docs/9.4/pgupgrade.html), this
is necessary to enable link-mode specific tests.

This is intended as a quick-fix, hence we do not add additional testing relative to the original PR.

### DIFF
--- a/hub/check_upgrade.go
+++ b/hub/check_upgrade.go
@@ -1,6 +1,8 @@
 package hub
 
 import (
+	"github.com/greenplum-db/gpupgrade/idl"
+	"github.com/greenplum-db/gpupgrade/utils"
 	"sync"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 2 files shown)


jchampio:
> This is intended as a quick-fix, hence we do not add additional testing relative to the original PR.

This doesn't pass the smell test to me. 😄

dkrieger-pvtl:
> > This is intended as a quick-fix, hence we do not add additional testing relative to the original PR.
> 
> This doesn't pass the smell test to me. 😄

Hmmmm....how about I file a follow up story to add testing.  Otherwise, I'm essentially backfilling testing for an already approved/merged PR.

jchampio:
> Otherwise, I'm essentially backfilling testing for an already approved/merged PR.

Why is that bad?

berlin-ab:
If we don't need the test, why do we need the change?

berlin-ab:
@jchampio @dkrieger-pvtl I propose we merge this commit. We do actually want this change.

In David's defense, none of `Server#CheckUpgrade` is tested directly. Neither is its caller `Server#InitializeCreateCluster`.

That said, we **should** add those tests. If not as a part of this change, as a part of the next change to either of these functions.

PR #216, opened 2020-02-12
Break Reconfigure into steps to ease idempotence behavior

bhuvnesh2703:
Earlier, Reconfigure was single step which did couple of things, for
instance:
a) stopping target cluster
b) starting target master
c) updating segment configuration in catalog with port information
d) stopping target master
e) updating postgresql.conf file
f) starting target cluster

Since, all these steps were coupled in one main step, achieiving
idempotence is complicated. In order to simply it, all these steps
are now driven from finalize directly so that we achieve idempotence
for most of the steps due to the existing behavior.


cli/commanders/steps.go
```diff
@@ -18,19 +18,25 @@ type receiver interface {
...
+	idl.Substep_COPY_MASTER:                       "Copying master to segments...",
+	idl.Substep_UPGRADE_PRIMARIES:                 "Upgrading segments...",
+	idl.Substep_START_TARGET_CLUSTER:              "Starting new cluster...",
+	idl.Substep_Finalize_SHUTDOWN_TARGET_CLUSTER:  "Stopping new cluster",
+	idl.Substep_RECONFIGURE_PORTS:                 "Changing cluster ports...",
+	idl.Substep_Finalize_START_TARGET_MASTER:      "Starting new master...",
+	idl.Substep_Finalize_UPDATE_CATALOG_WITH_PORT: "Updating new master with port information...",
+	idl.Substep_Finalize_SHUTDOWN_TARGET_MASTER:   "Stopping new master...",
+	idl.Substep_Finalize_UPDATE_POSTGRESQL_CONF:   "Updating master postgreql.conf...",
+	idl.Substep_Finalize_START_TARGET_CLUSTER:     "Starting new cluster...",
```
jchampio: Not a blocker, but I have a counterproposal: allow different steps to use the same substep IDs, and scope `status.json` by step to disambiguate; e.g.
```js
{
  "initialize": {
    "BACKUP_TARGET_MASTER": "COMPLETE",
    "CHECK_UPGRADE": "COMPLETE",
    "CONFIG": "COMPLETE",
    "CREATE_TARGET_CONFIG": "COMPLETE",
    "INIT_TARGET_CLUSTER": "COMPLETE",
    "SHUTDOWN_TARGET_CLUSTER": "COMPLETE",
    "START_AGENTS": "COMPLETE"
  },
  ...
  "finalize": {
    "SHUTDOWN_TARGET_CLUSTER": "COMPLETE",
    "RECONFIGURE_PORTS": "COMPLETE",
    "START_TARGET_MASTER": "COMPLETE",
    "UPDATE_CATALOG_WITH_PORT": "COMPLETE",
    "SHUTDOWN_TARGET_MASTER": "COMPLETE",
    "UPDATE_POSTGRESQL_CONF": "COMPLETE",
    "START_TARGET_CLUSTER": "COMPLETE",
  }
}
```

in the hope that this additional complexity in the internals would cut down on the copy-paste here.

hub/start_or_stop_cluster.go
```diff
@@ -2,76 +2,93 @@ package hub
...
 }
-func StartCluster(stream step.OutStreams, cluster *utils.Cluster) error {
-	return startStopCluster(stream, cluster, false)
+
+func StopMasterOnly(streams step.OutStreams, cluster *utils.Cluster, isSource bool) error {
+	return stopStartGpdb(streams, cluster, isSource, true, true)
 }
 
-func startStopCluster(stream step.OutStreams, cluster *utils.Cluster, stop bool) error {
+func stopStartGpdb(stream step.OutStreams, cluster *utils.Cluster, isSource, isStop, isMaster bool) error {
```
jchampio: I'm itching to refactor this now, given the lessons from the stop-cluster idempotence branch, but maybe now is not the right time. @berlin-ab 's PR is potentially introducing a good abstraction for running utilities that could simplify this in the future.

cli/commanders/steps.go
bhuvnesh2703: sure, but will keep it for a separate story.

berlin-ab:
What were the complications you ran into while attempting to implement idempotence?

I'm wondering if the user needs/wants to know these steps are happening.

hub/start_or_stop_cluster.go
berlin-ab: This uber stop-start cluster function needs to get broken up, at a minimum. We shouldn't continue to pile configuration options onto it.

We've already started down that path with: https://github.com/greenplum-db/gpupgrade/pull/212/files#diff-f3381fa58300b3af8f578952568878a2L39

kalensk:
FYI, this is needed for the reconfigure datadirs story as well.

bhuvnesh2703:
What were the complications you ran into while attempting to implement idempotence?

The current framework which tracks the status of a step does not allow tracking of steps within a step. Earlier Reconfigure ports was a bunch of steps and if you have to re-run, you have to check for each and every step which has been successful and that is unnecessary.
For instance, if stop and start was there, we should check if its stopped / started or not before proceeding to the step (we have already punted on start / stop idempotence). Whereas if these are broken into smaller steps we get the tracking  for free.
Also from a user.. What they need IMO is what's going on, and if they could see the main steps which are going on, that is always better. Simply saying Reconfigure ports, hides start / stop / update / copy operations which makes the customer blind why its taking long. Seeing these steps in a way also give minor indication of how much time it takes.

hub/start_or_stop_cluster.go
berlin-ab: We'll fix this with a follow up PR.

berlin-ab:
> What were the complications you ran into while attempting to implement idempotence?
> 
> The current framework which tracks the status of a step does not allow tracking of steps within a step. Earlier Reconfigure ports was a bunch of steps and if you have to re-run, you have to check for each and every step which has been successful and that is unnecessary.
> For instance, if stop and start was there, we should check if its stopped / started or not before proceeding to the step (we have already punted on start / stop idempotence). Whereas if these are broken into smaller steps we get the tracking for free.
> Also from a user.. What they need IMO is what's going on, and if they could see the main steps which are going on, that is always better. Simply saying Reconfigure ports, hides start / stop / update / copy operations which makes the customer blind why its taking long. Seeing these steps in a way also give minor indication of how much time it takes.
> 
> Also, for any other catalog change, that can now fit it between start of the master and stop of the master. We need not do these steps again (start / stop etc) for reconfigure data directories or other things which might require update of catalog.

Make sense, thanks.

PR #217, opened 2020-02-13
Merge cluster/cluster into utils/cluster

marksliva:
Previously, we had 2 structs and files that both defined Cluster and its
functions.  This combines them into a single struct and file.


hub/upgrade_primaries_test.go
```diff
@@ -1,25 +1,22 @@
...
+	"errors"
 	"github.com/greenplum-db/gpupgrade/hub"
 	"github.com/greenplum-db/gpupgrade/idl"
-	"github.com/greenplum-db/gpupgrade/utils/cluster"
-
+	"github.com/greenplum-db/gpupgrade/utils"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+	"path/filepath"
 )
```
jchampio: Something has gone seriously wrong with these imports, both here and in other files; can you check your `goimports` settings and reroll?

utils/cluster.go
```diff
@@ -115,3 +139,145 @@ func (c Cluster) SegmentsOn(hostname string) ([]cluster.SegConfig, error) {
...
+
+	// Make sure each mirror has a primary.
+	for _, seg := range cluster.Mirrors {
+		content := seg.ContentID
+
+		if _, ok := cluster.Primaries[content]; !ok {
+			return nil, newInvalidSegmentsError(seg, "mirror with content ID %d has no primary", content)
+		}
+	}
+
```
dkrieger-pvtl: In GPDB, I thought that in a valid cluster, either no primary has a mirror or they all do.  If so, it would be good to check for this case here too.

utils/cluster.go
```diff
@@ -115,3 +139,145 @@ func (c Cluster) SegmentsOn(hostname string) ([]cluster.SegConfig, error) {
...
+// ErrInvalidSegments is returned by NewCluster if the segment configuration
+// array does not map to a valid cluster.
+var ErrInvalidSegments = errors.New("invalid segment configuration")
+
+/*
+ * Base cluster functions
+ */
+
+func NewCluster(segConfigs []SegConfig) (*Cluster, error) {
```
dkrieger-pvtl: I may be reading the code wrong, but it seems to me that the master is counted as a "primary" and the standby as a "mirror" in `utils.Cluster`.  If so, I find this usage confusing.  It looks like this PR did not change this aspect of the code, so it could be placed as a follow up task.

hub/upgrade_primaries_test.go
marksliva: Yeah I checked. For some reason imports were being sorted by gofmt instead of goimports. I pushed a fix.

utils/cluster.go
marksliva: Created a story to address this.

utils/cluster.go
marksliva: Created a chore to address this.

PR #218, opened 2020-02-19
Upgrade master standby

jchampio:
Adds substep `FINALIZE_UPGRADE_STANDBY` to the Hub, which calls off to the gpdb utility `gpinitstandby` to set up a master standby. Idempotently upgrade master standby by removing any existing standby first.

The second entry in the `--ports` range is now reserved for the standby master, and this reservation is done during the initial upgrade configuration. This necessitates changes to the port assignment logic, which has been significantly rewritten. The end-to-end tests related to port assignment have been updated accordingly.

@berlin-ab You're listed as author currently; if too much has changed since you touched it, let me know.


hub/greenplum_runner.go
```diff
@@ -0,0 +1,53 @@
...
+	BinDir() string
+	MasterDataDirectory() string
+	MasterPort() int
+}
+
+func (e *greenplumRunner) Run(utilityName string, arguments ...string) error {
+	path := filepath.Join(e.binDir, utilityName)
+
+	arguments = append([]string{path}, arguments...)
+	script := shellquote.Join(arguments...)
```
berlin-ab: Thanks for this.

cli/commanders/steps_test.go
```diff
@@ -128,6 +129,20 @@ func TestUILoop(t *testing.T) {
 		}
 	})
 
+	t.Run("it formats the upgrade standby substep", func(t *testing.T) {
```
jchampio: This seems like a business-logic test that made its way into a unit test for UILoop; it should be moved.

hub/greenplum_runner.go
```diff
@@ -0,0 +1,53 @@
...
+
+	"github.com/kballard/go-shellquote"
+)
+
+type GreenplumRunner interface {
+	ShellRunner
+	BinDir() string
+	MasterDataDirectory() string
+	MasterPort() int
+}
```
jchampio: This interface can now be simplified -- nothing but the `Run()` provided in `ShellRunner` is actually in use.

hub/greenplum_runner.go
```diff
@@ -0,0 +1,53 @@
...
+
+	withGreenplumPath := fmt.Sprintf("source %s/../greenplum_path.sh && %s", e.binDir, script)
+
+	command := exec.Command("bash", "-c", withGreenplumPath)
+	command.Env = append(command.Env, fmt.Sprintf("%v=%v", "MASTER_DATA_DIRECTORY", e.masterDataDirectory))
+	command.Env = append(command.Env, fmt.Sprintf("%v=%v", "PGPORT", e.masterPort))
+	output, err := command.CombinedOutput()
+
+	fmt.Printf("Master data directory, %v\n", e.masterDataDirectory)
+	fmt.Printf("%s: %s \n", script, string(output))
```
jchampio: These need to be removed.

hub/greenplum_runner.go
```diff
@@ -0,0 +1,53 @@
...
+
+	arguments = append([]string{path}, arguments...)
+	script := shellquote.Join(arguments...)
+
+	withGreenplumPath := fmt.Sprintf("source %s/../greenplum_path.sh && %s", e.binDir, script)
+
+	command := exec.Command("bash", "-c", withGreenplumPath)
+	command.Env = append(command.Env, fmt.Sprintf("%v=%v", "MASTER_DATA_DIRECTORY", e.masterDataDirectory))
+	command.Env = append(command.Env, fmt.Sprintf("%v=%v", "PGPORT", e.masterPort))
+	output, err := command.CombinedOutput()
```
jchampio: ...and this can be plugged into the stream.

hub/shell_runner_test.go
```diff
@@ -0,0 +1,67 @@
+package hub_test
+
+type spyRunner struct {
+	calls map[string][]*SpyCall
+}
```
jchampio: This can be moved to the test that needs it.

hub/upgrade_standby.go
```diff
@@ -0,0 +1,49 @@
...
+		"-P", strconv.Itoa(standbyConfig.Port),
+		"-s", standbyConfig.Hostname,
+		"-S", standbyConfig.DataDirectory,
+		"-a")
+
+	if err != nil {
+		gplog.Error(fmt.Sprintf(
+			"error message from creating new standby master: %v",
+			err))
+	}
```
jchampio: Error logging should be handled at layers above this; we can remove this.

hub/initialize.go
```diff
@@ -114,13 +100,152 @@ func (s *Server) fillClusterConfigsSubStep(_ step.OutStreams, request *idl.Initi
...
+	var standbyPort int
+	if _, ok := source.Mirrors[-1]; ok {
+		// Reserve another port for the standby.
+		standbyPort = nextPort
+		nextPort++
+	}
+
+	// Reserve enough ports to handle the host with the most segments.
+	var maxSegs int
+	for _, segments := range segmentsByHost {
```
bhuvnesh2703: are we going to ever need this?

hub/initialize.go
```diff
@@ -114,13 +100,152 @@ func (s *Server) fillClusterConfigsSubStep(_ step.OutStreams, request *idl.Initi
...
+	if err := checkTargetPorts(source, ports); err != nil {
+		return PortAssignments{}, err
+	}
+
+	// Pop the first port off for master.
+	masterPort := ports[0]
+	ports = ports[1:]
+
+	var standbyPort int
+	if _, ok := source.Mirrors[-1]; ok {
```
bhuvnesh2703: Should this logic go in to StandbyPort, why the caller should worry about it?

hub/finalize.go
```diff
@@ -26,6 +26,22 @@ func (s *Server) Finalize(_ *idl.FinalizeRequest, stream idl.CliToHub_FinalizeSe
 		}
 	}()
 
+	if s.Source.HasStandby() {
```
berlin-ab: nit: This stands out as a deviation from the norm. I wonder if there's a way to work this into the step infrastructure. It would be nice to not introduce complexity to `Server#Finalize`.

hub/initialize.go
dkrieger-pvtl: @bhuvnesh2703 Can't a `gpexpand` result in a cluster with different number of segments on different segment hosts?

jchampio: Or a number of segments that isn't divisible by the number of hosts, or...

hub/finalize.go
jchampio: I'm in the camp that complexity belongs exactly where it's needed. If an abstraction doesn't reduce the complexity, but just moves it elsewhere, then all I've done is made the system harder to understand. So IMO it's a good thing that it stands out, because it is an interesting deviation from the norm and it deserves attention.

hub/initialize.go
jchampio: Not sure what you mean here?

cli/commanders/steps_test.go
jchampio: Removed after a second look; it's not actually adding coverage.

PR #219, opened 2020-02-19
Add check in gpupgrade initialize to check for preferred roles

bhuvnesh2703:
If the cluster is not in a balanced state, gpupgrade should fail
to continue during initialize.

Failure case:
```
13:59 $ gpupgrade initialize --old-bindir=/usr/local/gpdb6/bin/ --new-bindir=/usr/local/gpdb6/bin/ --old-port=6000 --disk-free-ratio 0 --stop-before-cluster-creation

Initialization in progress.

Creating state directory...                                        [COMPLETE]
Creating initial cluster config files...                           [COMPLETE]
Starting hub...                                                    [COMPLETE]
Retrieving configs...                                              [COMPLETE]
Starting agents...                                                 [COMPLETE]
Checking old cluster is balanced...                                [FAILED]

Initialize:
    github.com/greenplum-db/gpupgrade/cli/commanders.Initialize
        /Users/bhuvneshchaudhary/go/src/github.com/greenplum-db/gpupgrade/cli/commanders/steps.go:57
  - rpc error: code = Unknown desc = substep "initialize": checking cluster is balanced:
		Segment dbid [2,5] are not in balanced state.
		The cluster must be balanced for gpupgrade to continue.
		Use gprecoverseg to rebalance the cluster.

```

Success case:
```
Initialization in progress.

Creating state directory...                                        [COMPLETE]
Creating initial cluster config files...                           [COMPLETE]
Starting hub...                                                    [COMPLETE]
Retrieving configs...                                              [COMPLETE]
Starting agents...                                                 [COMPLETE]
Checking old cluster is balanced...                            [COMPLETE]
```


hub/initialize.go
```diff
@@ -41,6 +41,10 @@ func (s *Server) Initialize(in *idl.InitializeRequest, stream idl.CliToHub_Initi
 		return err
 	})
 
+	st.AlwaysRun(idl.Substep_CHECK_PREFERRED_ROLE, func(_ step.OutStreams) error {
```
bhuvnesh2703: kept it Always run..

hub/check_preferred_role_test.go
```diff
@@ -0,0 +1,74 @@
+package hub
```
dkrieger-pvtl: I had an import cycle with this, so I had to rename this to `hub_test`.    That also seems to be becoming our standard.

hub/check_preferred_role_test.go
```diff
@@ -0,0 +1,74 @@
...
+		}
+		defer finishMock(mock, t)
+
+		dbids := []int{2, 5}
+		dbidRows := sqlmock.NewRows([]string{"dbid"})
+		for _, dbid := range dbids {
+			dbidRows.AddRow(dbid)
+		}
+
+		mock.ExpectQuery("SELECT dbid FROM pg_catalog.gp_segment_configuration WHERE role != preferred_role").
```
dkrieger-pvtl: see above; replace this with the const variable the represents our query.

hub/check_preferred_role.go
```diff
@@ -0,0 +1,62 @@
...
+			xerrors.Errorf("%s failed to scan rows: %w", idl.Substep_CHECK_PREFERRED_ROLE, err)
+		}
+		dbidsSwitchedRole = append(dbidsSwitchedRole, dbid)
+	}
+
+	if len(dbidsSwitchedRole) > 0 {
+		message := fmt.Sprintf(`
+		Segment dbid %s are not in preferred role.
+		All the segments must be in their preferred role for gpupgrade to continue. 
+		Use gprecoverseg for bringing the segments in their preferred role.`,
```
dkrieger-pvtl: Change to:
```
The cluster must be balanced for gpupgrade to continue.
Use gprecoverseg to rebalance the cluster.
```

hub/initialize.go
dkrieger-pvtl: I think this should be the first substep run in `Initialize`...it relies only on the source cluster, is fast, and has no side effects.

idl/cli_to_hub.proto
```diff
@@ -60,6 +60,7 @@ enum Substep {
     FINALIZE_SHUTDOWN_TARGET_MASTER = 16;
     FINALIZE_UPDATE_POSTGRESQL_CONF = 17;
     FINALIZE_START_TARGET_CLUSTER = 18;
+    CHECK_PREFERRED_ROLE = 19;
```
dkrieger-pvtl: renamed to `CHECK_CLUSTER_BALANCED`

bhuvnesh2703:
@dkrieger-pvtl applied the feedback which you requested, please review.

hub/check_cluster_balanced.go
```diff
@@ -0,0 +1,69 @@
...
+
+	err = FindUnbalancedSegments(sourceDB)
+	if err != nil {
+		return errors.Wrap(err, "checking cluster is balanced")
+	}
+
+	return nil
+}
+
+func FindUnbalancedSegments(db *sql.DB) (err error) {
```
dkrieger-pvtl: As a design point, this function could return a slice of dbids that are not in balance.  Then the caller could return `nil` if that slice is empty or produce the error message if not.  It would also simplify the testing, as you'd be comparing against a logical result instead of an error string.

dkrieger-pvtl:
@bhuvnesh2703  Thanks for incorporating the changes.  There are two that I requested that weren't incorporated; I unresolved those.

hub/check_preferred_role.go
```diff
@@ -0,0 +1,62 @@
...
+	"fmt"
+	"strings"
+
+	"github.com/hashicorp/go-multierror"
+	"golang.org/x/xerrors"
+
+	"github.com/greenplum-db/gpupgrade/idl"
+)
+
+func CheckSegmentNotInPreferredRole(sourcePort int) error {
```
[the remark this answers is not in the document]
dkrieger-pvtl: Done.

hub/check_cluster_balanced.go
dkrieger-pvtl: Done; thx.

hub/check_cluster_balanced.go
```diff
@@ -0,0 +1,68 @@
...
+	"github.com/hashicorp/go-multierror"
+	"golang.org/x/xerrors"
+
+	"github.com/greenplum-db/gpupgrade/idl"
+)
+
+const DbidNotInBalancedStateQuery = "SELECT dbid FROM pg_catalog.gp_segment_configuration WHERE role != preferred_role"
+
+func CheckClusterIsBalanced(sourcePort int) (err error) {
+	connURI := fmt.Sprintf("postgresql://localhost:%d/template1?gp_session_role=utility&search_path=", sourcePort)
```
marksliva: Why are we making another query, when the previous substep already fills in the source cluster config? All we would have to do would be include `preferred_role` in that query. Then we could check the mismatch using the source cluster config. Am I missing something? Doing it this way only makes sense to me if we didn't already fill in the cluster config, but that isn't the case since we did that in the previous substep.

dkrieger-pvtl: > All we would have to do would be include `preferred_role` in the existing query in `utils.GetSegmentConfiguration(connection *dbconn.DBConn)`.

That's a good point.  We could check it there and return a custom error `ClusterUnbalancedError` that contains the unbalanced dbids.  We should not be filling in the cluster config if the cluster is unbalanced.  This idea makes a lot of sense to me for that reason; otherwise, we'd need to couple the CLUSTER_CONFIG and BALANCED_CLUSTER steps implicitly by placing the BALANCED_CLUSTER check first.

@bhuvnesh2703 what do you think?

bhuvnesh2703: actually a good point. let me make the change..

bhuvnesh2703:
Closing this in lieu of https://github.com/greenplum-db/gpupgrade/pull/224.

This might be reopened when we introduce this check in execute.

PR #220, opened 2020-02-20
Target cluster uses source cluster's data directory paths

berlin-ab:
After an upgrade, we want the target cluster to appear to the user in the same "location"(ports and data directories) as the source cluster.  This PR does this for the data directories.

There are two steps:

1). rename the source data directories to an archive directory AND moves the target data directories into the source's original location:

```
/source/data/directory -> /source/data/directory_old   (archive)
/target/data/directory -> /source/data/directory            (promote)
```
2). modify the target cluster's system tables to account for these new directory locations.

We do not modify the source cluster's system table, though the source cluster will of course not start up after this new substep is run.

Caveat: we are not handling filespaces(GPDB-5) or tablespaces(GPDB-6) at this time, so gpupgrade for a source cluster with those might fail.

Co-authored-by: David Krieger <dkrieger@pivotal.io>
Co-authored-by: Adam Berlin <aberlin@pivotal.io>


hub/init_target_cluster.go
```diff
@@ -131,13 +131,20 @@ func WriteInitsystemFile(gpinitsystemConfig []string, gpinitsystemFilepath strin
 }
 
 func upgradeDataDir(path string) string {
+	return dataDirWithSuffix(path, "_upgrade")
+}
+
+func oldDataDir(path string) string {
```
berlin-ab: I don't think we're using this, so let's remove it.

hub/swap_data_directories.go
```diff
@@ -0,0 +1,82 @@
...
+import (
+	"fmt"
+
+	"github.com/greenplum-db/gp-common-go-libs/gplog"
+	"github.com/hashicorp/go-multierror"
+
+	"github.com/greenplum-db/gpupgrade/idl"
+	"github.com/greenplum-db/gpupgrade/utils"
+)
+
```
berlin-ab: Any suggestions for better names for `#archive` and `#promote`

jchampio: I like `archive`. Not sure we need a verb for the corresponding new->old move, but `promote` should not be it, since promotion already has a meaning.

hub/update_master_configuration.go
```diff
@@ -0,0 +1,59 @@
...
+
+	multierror.Append(multiErr,
+		updateGpperfmonConf(source.MasterDataDir()))
+
+	multierror.Append(multiErr,
+		updatePostgresConfig(target, source))
+
+	return multiErr.ErrorOrNil()
+}
+
```
berlin-ab: I want to refactor these two `sed` calls.

berlin-ab:
We'll need to add the standby to this work after the rebase.

cli/commanders/steps.go
```diff
@@ -18,24 +18,25 @@ type receiver interface {
...
+	idl.Substep_SHUTDOWN_TARGET_CLUSTER:          "Stopping new cluster...",
+	idl.Substep_BACKUP_TARGET_MASTER:             "Backing up new master...",
+	idl.Substep_CHECK_UPGRADE:                    "Running pg_upgrade checks...",
+	idl.Substep_UPGRADE_MASTER:                   "Upgrading master...",
+	idl.Substep_COPY_MASTER:                      "Copying master to segments...",
+	idl.Substep_UPGRADE_PRIMARIES:                "Upgrading segments...",
+	idl.Substep_START_TARGET_CLUSTER:             "Starting new cluster...",
+	idl.Substep_FINALIZE_SHUTDOWN_TARGET_CLUSTER: "Stopping new cluster",
+	idl.Substep_FINALIZE_START_TARGET_MASTER:     "Starting new master...",
+	idl.Substep_FINALIZE_UPDATE_CATALOG:          "Updating new master with port information...",
```
kalensk: The string description should probably be updated to include something about data dirs.

cli/commanders/steps.go
```diff
@@ -18,24 +18,25 @@ type receiver interface {
...
+	idl.Substep_FINALIZE_SHUTDOWN_TARGET_MASTER:  "Stopping new master...",
+	idl.Substep_FINALIZE_UPDATE_POSTGRESQL_CONF:  "Updating master postgreql.conf...",
```
kalensk: This updates both postres.conf and gpperfom.conf. Should these be split at all (idempotence)? If not, the enum and description should encapsulate both ideas.

cli/commanders/steps.go
```diff
@@ -18,24 +18,25 @@ type receiver interface {
...
+	idl.Substep_FINALIZE_START_TARGET_CLUSTER:    "Starting new cluster...",
+	idl.Substep_FINALIZE_SWAP_DATA_DIRECTORIES:   "Installing upgraded data directories...",
```
kalensk: I don't think I understand the word `installing` here? Are there some other choices? Maybe "Swapping data directories"...?

utils/sys_utils.go
```diff
@@ -40,6 +41,7 @@ type SystemFunctions struct {
 	Rename       func(oldpath, newpath string) error
 	ReadFile     func(filename string) ([]byte, error)
 	WriteFile    func(filename string, data []byte, perm os.FileMode) error
+	WriteString  func(w io.Writer, s string) (n int, err error)
```
kalensk: This is no longer used.

utils/sys_utils.go
```diff
@@ -64,6 +66,7 @@ func InitializeSystemFunctions() *SystemFunctions {
 		FilePathGlob: filepath.Glob,
 		ReadFile:     ioutil.ReadFile,
 		WriteFile:    ioutil.WriteFile,
+		WriteString:  io.WriteString,
```
kalensk: This is no longer used.

utils/sys_utils.go
berlin-ab: removed, thanks.

berlin-ab:
> I don't see why we need an AgentBroker interface for this change? What is it giving us over the existing way of connecting to the agents and tests, etc.?

It simplifies SwapDataDirectory's view of the world, making it not need to deal with the more complicated hub.Connection object construction, setup, and agent pairing logic.

agent/reconfigure_data_dirs_test.go
```diff
@@ -0,0 +1,88 @@
...
+	"errors"
+	"testing"
+
+	"github.com/greenplum-db/gp-common-go-libs/testhelper"
+
+	"github.com/greenplum-db/gpupgrade/utils"
+
+	"github.com/greenplum-db/gpupgrade/idl"
+
+	"github.com/greenplum-db/gpupgrade/agent"
```
jchampio: Your import formatting is doing something very strange. Make sure you're using `goimports` and not, say, GoLand's reimplementation of almost-`goimports`-style functionality, which Mark is having a lot of trouble with on his machine.

kalensk:
> It simplifies SwapDataDirectory's view of the world, making it not need to deal with the more complicated hub.Connection object construction, setup, and agent pairing logic.

I don't think I am quite following, especially for this particular PR. For now, what is wrong with `s.agentConns` like the other gRPC calls that the hub makes to the agent? Maybe a team sync up?

hub/swap_data_directories.go
berlin-ab: I went with `publish` instead of `promote`.  I want to have some terminology there because it helps us codify the rule somewhere rather than spread around the assumption that a target segment inherits the source segment's data directory. That decision is now in one place and it has a meaningful name.

hub/swap_data_directories.go
```diff
@@ -0,0 +1,81 @@
...
+func SwapDataDirectories(hub Hub, agentBroker AgentBroker) error {
+	swapper := finalizer{agentBroker: agentBroker}
+	swapper.archive(hub.masterPair.source)
+	swapper.publish(hub.masterPair.target, hub.masterPair.source)
+	swapper.swapOnAgents(hub.agents)
+	return swapper.Errors()
+}
```
berlin-ab: These steps are not dependent, so for now we're letting them each run without short-circuiting on error.

jchampio: Why is `publish` not dependent on `archive`? If the first fails, won't the second potentially blow away the backup?

dkrieger-pvtl: While the steps are not dependent, an error in one of them likely indicates a serious problem(most likely a networking issue, I suppose).  I think we should fail quickly, to minimize the damage to the cluster and hopefully make recovery easier.  This also means the `finalizer` should go away.

hub/finalize.go
```diff
@@ -47,25 +52,35 @@ func (s *Server) Finalize(_ *idl.FinalizeRequest, stream idl.CliToHub_FinalizeSe
 		return StopCluster(streams, s.Target, false)
 	})
 
+	st.Run(idl.Substep_FINALIZE_SWAP_DATA_DIRECTORIES, func(streams step.OutStreams) error {
```
dkrieger-pvtl: We already have an idiom for calling agents; see `execute.go` for the Substep_UPGRADE_PRIMARIES substep.  It's not great, but might be best to repeat that here in favor of the agentBroker.  The agentBroker is reasonable, but we will likely want to redo all these hub-agent substeps at once, and keeping them the same will make that easier.  I see the agentBroker easing testing here, but the upgrade_primaries unit test likewise has a technique to create connections on a test Hub.

hub/hub.go
```diff
@@ -0,0 +1,97 @@
+package hub
+
+import "github.com/greenplum-db/gpupgrade/utils"
+
+//
+// Build a hub-centric model of the world:
+//
+// A hub has agents, agents have segment pairs
+//
+func MakeHub(config *Config) Hub {
```
kalensk: I appreciate the intention in wanting to simplify and clarify the code. Although, I am having some trouble seeing how this file/change is directly needed for swapping data directories? Any reason not to go with our current idiom _for now_, until we as a team make a _conscious_ and direct thought into changing that idiom?

hub/start_target_master_for_finalize.go
```diff
@@ -0,0 +1,14 @@
+package hub
+
+import "github.com/greenplum-db/gpupgrade/step"
+
+func StartTargetMasterForFinalize(streams step.OutStreams, config *Config) error {
```
kalensk: Having a single file with one function with an involved name like this seems like a code smell. What's wrong with the current code of `StartMasterOnly(streams, s.Target, false)` _for now_, until we consciously refactor the `start_or_stop_cluster.go` to be more generic. Or even adding start and stop methods to the cluster object?

hub/swap_data_directories_test.go
```diff
@@ -0,0 +1,383 @@
...
+
+func (mock *renameMock) TimesCalled() int {
+	return len(mock.calls)
+}
+
+func (mock *renameMock) Call(i int) *renameCall {
+	return mock.calls[i-1]
+}
+
+func (mock *renameMock) assertDirectoriesMoved(t *testing.T, originalName string, newName string) {
```
kalensk: blocker:  This test file does not seem to match the flow and style of all our other go-style testing. It seems like hybrid approach of go-style with ginkgo/bomega. This is making it very difficult for me to follow and understand amongst the the established convention and pattern. Can we please convert this to full-on go-style tests that match the same format, syntax, and idiom as our other tests?

test/finalize-reconfigure-datadirs.bats
```diff
@@ -0,0 +1,81 @@
...
+
+setup() {
+    skip_if_no_gpdb
+
+    [ ! -z $GPHOME ]
+    GPHOME_NEW=${GPHOME_NEW:-$GPHOME}
+    GPHOME_OLD=$GPHOME
+
+    PSQL="$GPHOME_NEW/bin/psql --no-align --tuples-only postgres"
+
```
kalensk: Is there a reason we can't combine this test with the other finalize.bats test?

test/helpers.bash
```diff
@@ -56,6 +56,37 @@ delete_cluster() {
     delete_target_datadirs "${masterdir}"
 }
 
+delete_finalized_cluster() {
```
kalensk: Is there any possibility in reusing the existing deletion of a cluster? This seems like a lot of involved/complex code.

hub/swap_data_directories.go
```diff
@@ -0,0 +1,81 @@
...
+
+type finalizer struct {
+	err         *multierror.Error
+	agentBroker AgentBroker
+}
```
kalensk: I don't understand this. Why can't we just keep it simple with an inline multierror? We should fail fast and bubble up errors like our other idioms, and if not we should have a story that addresses that across our code base in a conscious effort.

hub/finalize.go
```diff
@@ -11,6 +11,11 @@ import (
 )
 
 func (s *Server) Finalize(_ *idl.FinalizeRequest, stream idl.CliToHub_FinalizeServer) (err error) {
```
dkrieger-pvtl: We have a design issue here.  We have three sources of truth for a cluster(target and source):
- system catalog in the databases themselves
- the statedir config.json Clusters
- the in memory data structures in gpugprade

Before finalize, these should be consistent: the layout of the clusters does not change in initialize or execute.  Now, we have sources of change:
- adding the standby
   - currently only updates the system catalog
- renaming the directories
   - currently only updates the system catalog

How do we want to address this?  This should not hold up the PR entirely, but we should at least come up something reasonable for this PR.  It should be clear in the code that we're modifying the standby's datadirs, and that the datadirs in the config.json are stale.

hub/agent_broker.go
```diff
@@ -0,0 +1,48 @@
...
+//
+func (broker *AgentBrokerGRPC) ReconfigureDataDirectories(hostname string, renamePairs []*idl.RenamePair) error {
+	var connection *Connection
+
+	for _, c := range broker.agentConnections {
+		if c.Hostname == hostname {
+			connection = c
+			break
+		}
+	}
```
jchampio: I think this is just a very expensive map. If you switch `hub.Agent` to contain an `idl.AgentClient` instead of a hostname string, then you can get the same coverage without the broker concept just by spinning up `gomock` instances in the test.

See https://github.com/jchampio/gpupgrade/commit/0580af32b for an example of such an alternative design.

hub/hub.go
berlin-ab: It helped me think about and solve the problem. More specifically, I originally forgot that the rename operations needed to be pushed to the agents, and completed the story with a passing end-to-end test. After I realized that mistake, I wondered if a different model of the world would push me in the right direction - and it did.

hub/start_target_master_for_finalize.go
berlin-ab: I extracted this function from Finalize because of the extra work it needed to do to set up the call to StartMasterOnly, and finalize.go is a single-function file. Also, this file is not the only example of a single-function file.

hub/swap_data_directories.go
berlin-ab: Yep, definite oversight on my part. Will resolve.

hub/swap_data_directories_test.go
berlin-ab: This function is a direct result of reducing duplication, so if you can think of another way to do that, I'm open. Also, this looks like a nit, more than a blocker.

test/finalize-reconfigure-datadirs.bats
berlin-ab: The finalize.bats test no longer exists.

test/helpers.bash
berlin-ab: I took the approach of "duplication is preferred to the wrong abstraction". There's some further refactoring we could do to reduce the duplication between these two, but they did need to be different because they have different assumptions about the state of the world at the end of a step.

jchampio: > I took the approach of "duplication is preferred to the wrong abstraction". 

+1. Rule of Three, especially if there are already significant differences.

hub/finalize.go
kalensk: Based off David's comment, one rough example might be something like:

```go
func renameSegmentDataDirs(agentConns []*Connection,
	cluster *utils.Cluster,
	srcFunc func(path string) string,
	dstFunc func(path string) string) error {

	wg := sync.WaitGroup{}
	errs := make(chan error, len(agentConns))

		wg.Add(1)

		go func(c *Connection) {
			defer wg.Done()

			segments, err := cluster.SegmentsOn(c.Hostname)
			if err != nil {
				errs <- err
				return
			}

			req := new(idl.ReconfigureDataDirRequest)
			for _, seg := range segments {
				pair := idl.RenamePair{
					Src: filepath.Dir(srcFunc(seg.DataDir)),
					Dst: filepath.Dir(dstFunc(seg.DataDir)),
				}

				req.Pair = append(req.Pair, &pair)
			}

			_, err = c.AgentClient.ReconfigureDataDirectories(context.Background(), req)
			if err != nil {
				gplog.Error("creating segment data directories on host %s: %s", c.Hostname, err.Error())
				errs <- err
			}
		}(conn)
	}

	wg.Wait()
	close(errs)

	var mErr *multierror.Error
	for err := range errs {
	}

	return mErr.ErrorOrNil()
}
```

With callers something like:
```go
return renameSegmentDataDirsOnDisk(s.agentConns, s.Target, upgradeDataDir, noop)
// ...
return renameSegmentDataDirsOnDisk(s.agentConns, s.Source, noop, oldDataDir)
```

hub/swap_data_directories_test.go
berlin-ab: I unrolled this test helper.

hub/swap_data_directories.go
berlin-ab: This struct has been removed.

hub/finalize.go
berlin-ab: I modified Finalize Standby to modify the target cluster after upgrading a standby is successful. This modification gets stored to the JSON file, and also gets propagated to the following steps.

berlin-ab: So far, I have not made changes to the source or target cluster configurations during Reconfigure Data Directories, but we probably should.

PR #222, opened 2020-02-21
CI: use gp-upgrade-packaging for RPATH fixes

jchampio:
The `transform_rpm` task provided by `gp-upgrade-packaging` takes in a GPDB binary tarball and produces an RPM that tracks the current RelEng proposal for fixing our installation `RPATH`/`LD_LIBRARY_PATH` issues. That RPM can be passed to CCP using its gpdb_rpm input.

As a result, we can get rid of all our `RPATH` hacks, including the CI trampoline.

Dev pipeline [here](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:dev:pipeline).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -142,6 +142,7 @@ set-pipeline:
 		-l ~/workspace/gp-continuous-integration/secrets/gpdb_common-ci-secrets.yml \
 		-l ~/workspace/gp-continuous-integration/secrets/gpdb_master-ci-secrets.prod.yml \
 		-l ~/workspace/gp-continuous-integration/secrets/ccp_ci_secrets_$(FLY_TARGET).yml \
+		-l ~/workspace/gp-continuous-integration/secrets/gp-upgrade-packaging.dev.yml \
 		-v gpupgrade-git-remote=$(GIT_URI) \
 		-v gpupgrade-git-branch=$(BRANCH)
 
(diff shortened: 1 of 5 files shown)


upgrade/run.go
```diff
@@ -75,13 +73,6 @@ func Run(p SegmentPair, options ...Option) error {
 	// like PATH, and PGPORT et al are explicitly forbidden to be set.
 	cmd.Env = []string{}
 
-	// XXX ...but we make a single exception for now, for LD_LIBRARY_PATH, to
-	// work around pervasive problems with RPATH settings in our Postgres
-	// extension modules.
-	if path, ok := os.LookupEnv("LD_LIBRARY_PATH"); ok {
-		cmd.Env = append(cmd.Env, fmt.Sprintf("LD_LIBRARY_PATH=%s", path))
-	}
-
```
jchampio: I think this removal is the only thing I'm nervous about. It works fine on the CI, but anyone trying a manual upgrade on their dev machine will not be using the RPM. So this may need to stay for now.

kalensk: Agreed. I think there are enough people testing internally that do not have access to the test rpm, that leaving this makes sense. Is there a recommended way to circle back to this once the fixed RPM becomes more "mainstream"?

jchampio: I've filed a tracking bug for this code. I'll revert this section and push.

kalensk:
Approved. Looks good with the latest fixup.

PR #223, opened 2020-02-21
Move start and stop semantics to the cluster object

kalensk:
This moves start and stop semantics to the cluster object.

However, there are some potential downsides or gotchas due to import cycles and package exporting, etc.
1) The start and stop tests are in their own file (`utils/cluster_start_stop_test.go`) rather than in `cluster_test.go`.

2) `MustCreateCluster` and `DevNull` from `hub/common_test.go` are duplicated in `cluster_start_stop_test.go`.

Maybe someone can help me with the above two issue if they are doable with the current code?

[test pipeline](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:cluster_start_stop)


berlin-ab:
Ideally start/stop cluster can be done with the minimal amount of info necessary to do so. Coupling it with the cluster seems limiting long-term.

kalensk:
> Ideally start/stop cluster can be done with the minimal amount of info necessary to do so.

Can you explain what "minimal amount of info" means specifically for start & stop cluster?

I found adding it to the cluster object nice since it eliminates the need for the `isSource` parameter. That is, `StopCluster(streams, s.Source, true)` became `s.Source.Stop(streams)`.

berlin-ab:
I’m probably not exactly correct, but the greenplum_env.sh and the MASTER_DATA_DIRECTORY should be enough data to start the cluster.

berlin-ab:
The cluster should be able to provide that info. we need something similar to set up the greenplum runner while upgrading standbys.

kalensk:
> ’m probably not exactly correct, but the greenplum_env.sh and the MASTER_DATA_DIRECTORY should be enough data to start the cluster.

Ah! Now I understand where you are coming from. To start or stop a cluster one needs the master port, data directory, and bindir. To me it makes sense to use the cluster object which already contains that information rather than peeling it off and adding it to yet another object? (If I am understanding your implicit proposal.)

berlin-ab:
Should work for now. Consider the cost of extracting that object should we need to.

kalensk:
> Should work for now. Consider the cost of extracting that object should we need to.

I can't think of any specifics for why we would need to extract them. Do you have some specific concerns or scenarios in mind that we should consider? I'd prefer to write clean code for the _current_ state of the world, rather than a possible distant future.

kalensk:
> High-level thoughts for this draft:

1. I am in favor of printing datadir (or similar) to avoid threading `cluster.Destination`. Good idea assuming that is enough information for people.
2. I am not sure I see the need for both ways. Can you provide some concrete specifics on how having both would be useful? My gut says to stick to one or the other to keep things simple, and I am totally happy to close this draft PR if we stick the current implementation. Also, happy to explore a bit more about having both before deciding.
3. Sounds good.

Thanks for the comments.

jchampio:
> I am not sure I see the need for both ways. 

I'd say we don't, yet. Just pointing out that if we determine that both are useful in the future, it won't be a high-cost operation to get there.

kalensk:
Pushed a fixup that removes cluster.Destination in favor or simply printing MASTER_DATA_DIRECTORY when failing to stop or start. (I am happy to tweak the error message  if people have ideas.)

Thanks for taking a look!

kalensk:
Rebased with the latest master. And will wait until we merge upgrading mirrors and standby to the end of finalize since to make merging/rebasing easier.

jchampio:
> nit: I think we could improve our design so we don't have to thread streams everywhere

+1

PR #224, opened 2020-02-21
preferred-role

bhuvnesh2703:
If the role of the segment is not the preferred role, error while retrieving the configuration. This ensures that if the configuration of the cluster is not written into config.json unless the cluster is in balanced state.


utils/cluster.go
```diff
@@ -179,6 +180,10 @@ func NewCluster(segConfigs []SegConfig) (*Cluster, error) {
 	for _, seg := range segConfigs {
 		content := seg.ContentID
 
+		if seg.Role != seg.PreferredRole {
+			return nil, newInvalidSegmentsError(seg, "segment with dbid %d not in preferred role, run gprecoverseg to rebalance the cluster", seg.DbID)
```
marksliva: Let's tell users to include the `-r` flag when running gprecoverseg, which will rebalance each segment to its preferred role, since running gprecoverseg without that flag will not resolve this problem.
`... run 'gprecoverseg -r'`

kalensk:
Looks good overall to me.

utils/cluster.go
bhuvnesh2703: added

utils/cluster_test.go
```diff
@@ -139,18 +141,21 @@ func TestCluster(t *testing.T) {
 		{"bad role", []utils.SegConfig{
 			{Role: "x"},
 		}},
+		{"switched role", []utils.SegConfig{
+			{Role: "p", PreferredRole: "m"},
```
marksliva: nit: For completeness, you can also add the reverse of this, since it is another realistic scenario, i.e.
`{Role: "m", PreferredRole: "p"}`

marksliva:
One last thing, please update the commit message to be more descriptive of the change before merging.

dkrieger-pvtl:
The check for unbalanced contents should return all dbids that are unbalanced to the user on the first time; this will make their workflow easier.

PR #225, opened 2020-02-21
Remove db.NewDBConn in favor of dbconn.NewDBConn

kalensk:
We were not getting much from our implementation of db.NewDBConn, especially when it's useful to be explicit in connecting to the database for upgrade.

Eventually, it might be nice to have our own dbconn implementation for gpupgrade independent of common-go-libs.

[test pipeline](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:remove_db)


db/db.go
```diff
@@ -1,33 +0,0 @@
...
-func NewDBConn(masterHost string, masterPort int, dbname string) *dbconn.DBConn {
-	currentUser, _, _ := utils.GetUser()
-	username := utils.TryEnv("PGUSER", currentUser)
-	if dbname == "" {
-		dbname = utils.TryEnv("PGDATABASE", "")
-	}
-	hostname, _ := utils.GetHost()
-	if masterHost == "" {
-		masterHost = utils.TryEnv("PGHOST", hostname)
-	}
```
jchampio: Seems like this introduces functional changes? Are we ready to get rid of the reliance on `PGUSER` et al, or is that something that is already taken care of by common-libs?

kalensk: IMO, yes as was hinted towards in the commit message of it being "useful to be explicit in connecting to the database for upgrade".

If we want to keep the same semantics, I am happy to close this PR and circle back when it makes sense to, or add a similar constructor to common-go-libs, or ...?

jchampio: @bharath1028 @nirali13 This PR would remove the ability to connect to a database with a non-default `PGUSER` and `PGDATABASE` -- though we don't have any acceptance tests for that behavior, so I'm not sure we have proof that it worked before.

Thoughts? My personal preference would be to first add tests for whatever desired behavior the product teams wants, and then make changes. (In another story or chore.)

berlin-ab: Agree. Let’s put this PR down.

kalensk: If possible before closing I'd like to get some input from PM's (@bharath1028 @Nirali13 ) on the desired behavior so I can add tests if needed, merge this PR, close this PR, or something else.

kalensk: One idea might be to be explicit in the user and host by adding them as variables to the initialize config file (coming soon).

kalensk: Had an offline discussion with @Nirali13 and decided to leave the code as is for alpha to avoid any risk or unforeseen consequences. 

And before GA we will re-visit the desired behavior with the front-runner looking like adding explicit options to the initialize config file (with any further discussions happening then). Nirali has added a story has been added to track this work.

Thanks everyone!

PR #227, opened 2020-02-22
gpupgrade link-mode: pg_upgrade --check on master needs --link

dkrieger-pvtl:
The original PR adding link mode
(https://github.com/greenplum-db/gpupgrade/pull/177) did not throw the
--link mode flag for link-mode of gpupgrade.  However, according to the
documentation(https://www.postgresql.org/docs/9.4/pgupgrade.html), this
is necessary to enable link-mode specific tests.

Testing has been added.  The tests make sure that CheckUpgrade() passes
the correct arguments to the upgrade functions.

This commit also fixes a bug in CheckUpgrade() where it would
potentially deadlock filling up the error chan.  The previous code did
not fail fast, so could end up adding more than the 1 error into the
channel allocated to it.

I also collected the UpgradeMaster and UpgradePrimaries args(many) into an argument struct to make the code more readable.

This repeats the closed PR #214 but with tests added.

### DIFF
--- a/hub/check_upgrade.go
+++ b/hub/check_upgrade.go
@@ -3,12 +3,39 @@ package hub
 import (
 	"sync"
 
-	"github.com/pkg/errors"
-
 	"github.com/hashicorp/go-multierror"
+	"github.com/pkg/errors"
 
 	"github.com/greenplum-db/gpupgrade/step"
 )
+
+type UpgradeChecker interface {
+	UpgradeMaster(args UpgradeMasterArgs) error
+	UpgradePrimaries(args UpgradePrimaryArgs) error
+}
+
+type upgradeChecker struct{}
+
+func (upgradeChecker) UpgradeMaster(args UpgradeMasterArgs) error {
+	return UpgradeMaster(args)
+}
+
+func (upgradeChecker) UpgradePrimaries(args UpgradePrimaryArgs) error {
+	return UpgradePrimaries(args)
+}
+
+type AgentConnProvider interface {
+	GetAgents(s *Server) ([]*Connection, error)
+}
+
+type agentConnProvider struct{}
+
+func (agentConnProvider) GetAgents(s *Server) ([]*Connection, error) {
+	return s.AgentConns()
+}
+
+var upgrader UpgradeChecker = upgradeChecker{}
+var agentProvider AgentConnProvider = agentConnProvider{}
 
 func (s *Server) CheckUpgrade(stream step.OutStreams) error {
 	var wg sync.WaitGroup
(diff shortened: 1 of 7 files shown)


hub/check_upgrade.go
```diff
@@ -3,57 +3,91 @@ package hub
...
+
+func (upgradeChecker) UpgradeMaster(args UpgradeMasterArgs) error {
+	return UpgradeMaster(args)
+}
+
+func (upgradeChecker) UpgradePrimaries(args UpgradePrimaryArgs) error {
+	return UpgradePrimaries(args)
+}
+
+type AgentConnProvider interface {
```
berlin-ab: These setter injection seams point out an opportunity to improve the design. They're trying to hide a function on Server from other functions on Server, suggesting that CheckUpgrade shouldn't be on server in the first place.

berlin-ab: The setter injection here is interesting. The pattern has proved valuable elsewhere, because the default implementation can be set on startup, but a test version can be overridden - and the setters can be moved to an `_test.go` file, so they're not settable by the delivered executable.

berlin-ab: As more tests need to take advantage of the abstractions you've created, the setter mechanism is already in place, which is nice.

I prefer constructor injection, personally, because it calls out to the consumer the necessary dependencies to complete a function's responsibilities - but if the defaults are already set, then the consumer shouldn't need to know or worry.

dkrieger-pvtl: @berlin-ab Can you be more explicit about the actual change you are requesting?  The setters need to be in the main implementation because the controlled variables(`upgrader` and `agentProvider`) need to be private to our implementation.  

I didn't want to rework the implementation here, but wanted to add the `--link` option for the pg_upgrade check on the master and to test that.  Follow up refactoring of the main implementation can be done in a separate story.

berlin-ab: The only change that I'm suggesting here is moving the `CheckUpgrade` so it no longer has (s *Server) as its receiver. It is tightly bound to that object in a way that is making it more difficult to vary the implementations of `CheckUpgrade`'s collaborators.

berlin-ab: The second point is, if you've moved CheckUpgrade off of Server, you can pass the agents and upgraders as arguments instead of needing to create getter/setter functions.

hub/execute.go
```diff
@@ -39,7 +39,7 @@ func (s *Server) Execute(request *idl.ExecuteRequest, stream idl.CliToHub_Execut
 
 	st.Run(idl.Substep_UPGRADE_MASTER, func(streams step.OutStreams) error {
 		stateDir := s.StateDir
-		return UpgradeMaster(s.Source, s.Target, stateDir, streams, false, s.UseLinkMode)
+		return UpgradeMaster(UpgradeMasterArgs{s.Source, s.Target, stateDir, streams, false, s.UseLinkMode})
```
jchampio: nit: if we're going to be using an argument struct, why not make full use of it and name the bare values here?

hub/check_upgrade_test.go
```diff
@@ -0,0 +1,118 @@
...
+	if result.CheckOnly != true {
+		return errors.New(fmt.Sprintf("got %#v expected %#v", result.CheckOnly, true))
+	}
+	if result.UseLinkMode != expected.UseLinkMode {
+		return errors.New(fmt.Sprintf("got %#v expected %#v", result.UseLinkMode, expected.UseLinkMode))
+	}
+	return nil
+}
+
+func UpgradePrimariesMock(result hub.UpgradePrimaryArgs, expected *hub.Server) error {
```
jchampio: [putting the in-person discussion here]

By using in-band error reporting, we get the same problems Ginkgo gives us (one failure per test blocks subsequent failures). Instead, the `testing.T` can be pushed into your test doubles so you can run a `t.Errorf()` style inline. (This also lets you return in-band errors, or not, independent of whether there are failing tests.)

hub/check_upgrade.go
jchampio: > The pattern has proved valuable elsewhere, because the default implementation can be set on startup, but a test version can be overridden - and the setters can be moved to an _test.go file, so they're not settable by the delivered executable.

The nice thing about this pattern is that it doesn't have to be done with globals -- once we have a `Hub` object again, we can consider attaching seams to its internals, and either test within the `hub` package (so we can directly touch those internals in test) or provide additional methods in an `*_test.go` file to touch those internals from the `hub_test` package.

dkrieger-pvtl: I agree but will defer this to a future cleanup...pulling off CheckUpgrade from the server appears straightforward, but `(s *Server) AgentConns` will divert this PR too far from its intended path.

hub/execute.go
dkrieger-pvtl: My IDE shows them in its Editor pane, so I didn't even notice.  I've added them.

hub/check_upgrade_test.go
dkrieger-pvtl: I left this as is...in this case, we are testing help text differences and those are quick enough to iterate on that I find the current code more appropriate.

jchampio: Help text differences...? Was this for a different PR?

dkrieger-pvtl: sorry...yes, that was for a different PR.  But I meant to say that the test itself is pretty quick and checks the passed in args, so the current style is succinct.

PR #229, opened 2020-02-25
Update help text for gpupgrade/initalize/execute/finalize

dkrieger-pvtl:
We hard-coded the help output for flags:
  "-?"
  "-h"
  "--help"
  "help"
based on the provided specification.

We added tests to make sure the flags output the required values.

For gpupgrade initialize, we changed persistent flags to local flags
to be able to implement the "-?" flag.  Since we do not have any
child commands of gpupgrade initialize that require these, this should
be fine.

### DIFF
--- a/cli/bash/bash-completion.sh
+++ b/cli/bash/bash-completion.sh
@@ -289,9 +289,9 @@ _gpupgrade_config()
     noun_aliases=()
 }
 
-_gpupgrade_execute()
+_gpupgrade_execute_help()
 {
-    last_command="gpupgrade_execute"
+    last_command="gpupgrade_execute_help"
     commands=()
 
     flags=()
@@ -300,9 +300,47 @@ _gpupgrade_execute()
     flags_with_completion=()
     flags_completion=()
 
+
+    must_have_one_flag=()
+    must_have_one_noun=()
+    noun_aliases=()
+}
+
+_gpupgrade_execute()
+{
+    last_command="gpupgrade_execute"
+    commands=()
+    commands+=("help")
+
+    flags=()
+    two_word_flags=()
+    local_nonpersistent_flags=()
+    flags_with_completion=()
+    flags_completion=()
+
+    flags+=("--?")
+    flags+=("-?")
+    local_nonpersistent_flags+=("--?")
     flags+=("--verbose")
     flags+=("-v")
     local_nonpersistent_flags+=("--verbose")
+
+    must_have_one_flag=()
+    must_have_one_noun=()
+    noun_aliases=()
+}
+
+_gpupgrade_finalize_help()
+{
+    last_command="gpupgrade_finalize_help"
+    commands=()
+
+    flags=()
+    two_word_flags=()
+    local_nonpersistent_flags=()
+    flags_with_completion=()
+    flags_completion=()
+
 
     must_have_one_flag=()
     must_have_one_noun=()
@@ -313,6 +351,7 @@ _gpupgrade_finalize()
 {
     last_command="gpupgrade_finalize"
     commands=()
+    commands+=("help")
 
     flags=()
     two_word_flags=()
@@ -320,9 +359,46 @@ _gpupgrade_finalize()
     flags_with_completion=()
     flags_completion=()
 
+    flags+=("--?")
+    flags+=("-?")
+    local_nonpersistent_flags+=("--?")
     flags+=("--verbose")
     flags+=("-v")
     local_nonpersistent_flags+=("--verbose")
+
+    must_have_one_flag=()
+    must_have_one_noun=()
+    noun_aliases=()
+}
+
+_gpupgrade_help()
+{
+    last_command="gpupgrade_help"
+    commands=()
+
+    flags=()
+    two_word_flags=()
+    local_nonpersistent_flags=()
+    flags_with_completion=()
+    flags_completion=()
+
+
+    must_have_one_flag=()
+    must_have_one_noun=()
+    noun_aliases=()
+}
+
+_gpupgrade_initialize_help()
+{
+    last_command="gpupgrade_initialize_help"
+    commands=()
+
+    flags=()
+    two_word_flags=()
+    local_nonpersistent_flags=()
+    flags_with_completion=()
+    flags_completion=()
+
 
     must_have_one_flag=()
     must_have_one_noun=()
@@ -333,6 +409,7 @@ _gpupgrade_initialize()
 {
     last_command="gpupgrade_initialize"
     commands=()
+    commands+=("help")
 
     flags=()
     two_word_flags=()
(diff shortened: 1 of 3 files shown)


cli/commands/commands.go
```diff
@@ -325,22 +383,48 @@ If you would like to return the cluster to its original state, run
...
+
+	subInit.SetHelpFunc(func(cmd *cobra.Command, strs []string ) {
+		fmt.Printf(initializeHelp)
+	})
+	subInit.SetUsageTemplate(initializeHelp)
 
 	return subInit
 }
 
+//`Usage:{{if .Runnable}}
```
berlin-ab: This might be some Golang thing I don't recognize, but this does not look like it belongs here.

cli/bash/bash-completion.sh
```diff
@@ -329,10 +329,34 @@ _gpupgrade_finalize()
     noun_aliases=()
 }
 
+_gpupgrade_initialize_help()
```
berlin-ab: This doesn't look like it is called. Am I missing something?

cli/commands/commands.go
dkrieger-pvtl: We neglected to push the final code; this has been removed.

cli/bash/bash-completion.sh
dkrieger-pvtl: This is an auto-generated file from cli/bash/generate.go and is not hand-edited.

cli/commands/commands.go
```diff
@@ -54,9 +54,9 @@ func BuildRootCommand() *cobra.Command {
 	root := &cobra.Command{Use: "gpupgrade"}
 
 	root.AddCommand(config, version)
-	root.AddCommand(initialize())
-	root.AddCommand(execute())
-	root.AddCommand(finalize())
+	root.AddCommand(addHelpToCommand(initialize(), InitializeHelp))
+	root.AddCommand(addHelpToCommand(execute(), ExecuteHelp))
+	root.AddCommand(addHelpToCommand(finalize(), FinalizeHelp))
```
jchampio: nit: can we keep `initialize()` et al self-contained, and simply call `addHelpToCommand()` on the `Command` that's returned from these three functions before returning it? Similar to how you've done it for `BuildRootCommand()` below.

cli/commands/commands.go
```diff
@@ -499,3 +490,150 @@ var killServices = &cobra.Command{
...
+
+Use "gpupgrade [command] --help" for more information about a command.
+`
+)
+
+// Cobra has multiple ways to handle help text, so we want to force all of them to use the same help text
+func addHelpToCommand(cmd *cobra.Command, help string) *cobra.Command {
+	// Add a "-?" flag, which Cobra does not provide by default
+	var questionHelp bool
+	cmd.PreRunE = func(cmd *cobra.Command, args []string) error {
```
jchampio: Warning: since this technique does not save any previous `PreRunE` value, it will conflict with other tools that use `PreRunE` (say, `daemon.MakeDaemonizable()`). I don't think this necessarily has to be fixed now, but it should be noted at least.

(`daemon.MakeDaemonizable()` has a sample technique for "chaining" these callbacks.)

cli/commands/commands.go
```diff
@@ -499,3 +490,150 @@ var killServices = &cobra.Command{
...
+			fmt.Printf(help)
+			return nil
+		},
+	})
+	cmd.SetUsageTemplate(help)
+
+	// Override the built-in "-h" and "--help" flags
+	cmd.SetHelpFunc(func(cmd *cobra.Command, strs []string) {
+		fmt.Printf(help)
+	})
```
jchampio: Not a blocker, but: we're overriding `help` _and_ the `HelpFunc` _and_ the `-h` _and_ the usage template... shouldn't some of those be redundant?

cli/commands/commands_test.go
```diff
@@ -53,3 +56,28 @@ func TestParsePorts(t *testing.T) {
...
+		"initialize": InitializeHelp,
+		"execute":    ExecuteHelp,
+		"finalize":   FinalizeHelp,
+	}
+	flagList := []string{"-?", "-h", "--help", "help"}
+
+	for command, help := range helpMap {
+		for _, flag := range flagList {
+			t.Run(fmt.Sprintf("testing command %s with flag %s", command, flag), func(t *testing.T) {
+				cmd := exec.Command("gpupgrade", command, flag)
```
jchampio: Probably blocker: this belongs in the `integration` suite, which will check to make sure `gpupgrade` is on the `PATH` for you. Otherwise `go test` will fail for people who haven't built or installed yet.

dkrieger-pvtl: Done.

cli/commands/commands.go
dkrieger-pvtl: You'd think so, but unfortunately not.  The `gpupgrade initialize help` is a command that Cobra handles internally so we have to define our own to override that.  The `-h` and `--help` use the Cobra built-in "help function".  And `?` is not supported by Cobra at all so we needed to both add that as a command and change the flags that were persistent and required to ones that are local and required.

cli/commands/commands.go
dkrieger-pvtl: Good idea...I added this.

cli/commands/commands.go
dkrieger-pvtl: The current style is succinct.  Furthermore, adding all [four help syntax forms](https://github.com/greenplum-db/gpupgrade/pull/229#discussion_r384616582) requires modifying the top-level command itself.  We could separately adding the `help` command, but there's also the PreRunE that needs to be added to the top-level command.

So I do not see a straightforward to do satisfy your request, other than splitting up the current implementation into a per-syntax step: one to add PreRunE for `?`, another to set the help function, and another to add the subcommand.  That seems inelegant.

The help printout is inherent to our commands.

jchampio: > So I do not see a straightforward to do satisfy your request

All I'm asking for is e.g. at the end of `initialize()`:
```go
func initialize() *cobra.Command {
    ...

    return addHelpToCommand(subInit, InitializeHelp)
}
```

I don't think `BuildRootCommand()` needs to know details about how help is constructed for the subcommands.

integrations/commands_test.go
```diff
@@ -0,0 +1,33 @@
...
+	"fmt"
+	"os/exec"
+
+	"github.com/greenplum-db/gpupgrade/cli/commands"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("test gpupgrade help messages", func() {
```
jchampio: You don't have to use `Ginkgo` in the integrations suite, FYI. Your previous test implementation will work just fine. Use `go test -v` rather than `ginkgo` to prove to yourself that it is, in fact, being run; `ginkgo` hides standard Go-style tests.

cli/commands/commands_test.go
```diff
@@ -47,9 +47,9 @@ func TestParsePorts(t *testing.T) {
 	}
 
 	for _, c := range errorCases {
-		actual, err := parsePorts(c)
+		actual, err := ParsePorts(c)
```
jchampio: The `ParsePorts` change appears to be unrelated to this PR?

cli/commands/commands.go
dkrieger-pvtl: I agree.  I modified the code.

cli/commands/commands_test.go
dkrieger-pvtl: Thanks...when the new command help tests were in the commands_test.go file, I changed its package to `command_test`.  So I had to export `ParsePorts`.  But now that those tests are in integrations, that's no longer needed.

integrations/commands_test.go
dkrieger-pvtl: Ok, but rather than having one integration test in Ginkgo style and another in Go style, I took that path of least resistance; we can migrate both of these tests at a later time.

jchampio: I forgot that the `PATH` check was built into the Ginkgo suite. This works for now, thanks.

PR #230, opened 2020-02-25
Error during initialize for down segments mirrors

berlin-ab:
Check that all segments are up during initialize.

- Create Source Database abstraction to provide Segment Status,
  which provides SegmentStatus because we don't want to
  hold onto the state of "is a segment up or down" for any long period
  of time. (like in a Cluster's SegConfig).

- Also adds a unit test for formatting of substeps that will fail if
  the substep has not been formatted yet.

- Move logic for unbalanced segments into check

- Remove PreferredRole from SegConfig: it is no longer used for its original purpose (performing unbalanced segment check)
  and it adds unnecessary complexity to SegConfig throughout the application.

- Extract FillClusterConfigsSubStep off of Server

note: review commit by commit

There are a few commits that are applying the campsite rule.


hub/check_source_cluster_configuration.go
```diff
@@ -0,0 +1,82 @@
...
+type SegmentStatus struct {
+	IsUp bool
+	DbID int
+}
+
+type SegmentStatusError struct {
+	DownDbids []int
+}
+
+func (e SegmentStatusError) Error() string {
```
berlin-ab: Ah, I forgot to improve this error message with the full list of down dbids.

hub/initialize.go
```diff
@@ -33,8 +32,28 @@ func (s *Server) Initialize(in *idl.InitializeRequest, stream idl.CliToHub_Initi
 		}
 	}()
 
+	st.Run(idl.Substep_CHECK_SOURCE_CLUSTER_CONFIGURATION, func(streams step.OutStreams) error {
+		connection, err := getSourceDbConnection(sourcePort)
```
berlin-ab: I'd like to be able to reduce the number of times we connect to the database, but I'm not as familiar with this library.

berlin-ab:
> A new step is not needed here. I think it's cleaner to add this as part of the CONFIG substep, as was done for the [check for unbalanced contents](https://github.com/greenplum-db/gpupgrade/commit/b04e6c84c95e8c80670b2c4f73fafa2838325a0d). This is because it's naturally part of trying to create a cluster configuration from the new cluster. And checking for down segments is semantically similar to checking for unbalanced segments.

Nice catch. I moved my work into the fill cluster configs substep.

jchampio:
>  Remove PreferredRole from SegConfig.
>
> - it is no longer used for its original purpose (performing unbalanced segment check)
>  and it adds unnecessary complexity to SegConfig throughout the application.

+1, thanks.

hub/fill_cluster_configs.go
```diff
@@ -15,7 +15,7 @@ import (
 )
 
 // create old/new clusters, write to disk and re-read from disk to make sure it is "durable"
-func (s *Server) fillClusterConfigsSubStep(_ step.OutStreams, request *idl.InitializeRequest) error {
+func FillClusterConfigsSubStep(config *Config, _ step.OutStreams, request *idl.InitializeRequest, saveConfig func() error) error {
```
jchampio: This commit does not seem to belong with this PR.

hub/check_source_cluster_configuration.go
```diff
@@ -0,0 +1,169 @@
...
+	type result struct {
+		Dbid          DbID
+		Status        string
+		Role          string
+		PreferredRole string
+	}
+
+	results := make([]result, 0)
+
+	rows, err := connection.Query("select dbid as Dbid, status as Status, role as Role, preferred_role as PreferredRole from gp_segment_configuration")
```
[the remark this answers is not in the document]
berlin-ab: Thanks, i'll update.

hub/check_source_cluster_configuration.go
```diff
@@ -0,0 +1,169 @@
...
+		return err
+	}
+
+	if err := checkForDownSegments(statuses); err != nil {
+		return err
+	}
+
+	if err := checkForUnbalancedSegments(statuses); err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
berlin-ab: can do.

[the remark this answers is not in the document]
berlin-ab: This is up.

hub/fill_cluster_configs.go
berlin-ab: campsite rule cleanup. I was making a change to this function and decided to include the cleanup.

hub/check_source_cluster_configuration.go
```diff
@@ -0,0 +1,169 @@
...
+		err = rows.Scan(&r.Dbid, &r.Status, &r.Role, &r.PreferredRole)
+		results = append(results, r)
+	}
+
+	if err != nil {
+		return nil, err
+	}
+
+	for _, result := range results {
+		statuses = append(statuses, SegmentStatus{
```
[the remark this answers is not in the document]
berlin-ab: I was struggling with sqlmock, and this worked.

hub/fill_cluster_configs.go
```diff
@@ -3,21 +3,28 @@ package hub
...
 
 // create old/new clusters, write to disk and re-read from disk to make sure it is "durable"
 func (s *Server) fillClusterConfigsSubStep(_ step.OutStreams, request *idl.InitializeRequest) error {
-	conn := db.NewDBConn("localhost", int(request.SourcePort), "template1")
-	defer conn.Close()
+	connection := db.NewDBConn("localhost", int(request.SourcePort), "template1")
+
+	err := CheckSourceClusterConfiguration(func() ([]SegmentStatus, error) {
+		return GetSegmentStatuses(connection)
+	})
```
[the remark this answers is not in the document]
berlin-ab: Please open up the test suite for `CheckSourceClusterConfiguration` and see how this enables the straight-forward testing of it.

Also, the connection to the database ideally would be shared by the following Cluster load, but we are using two different mechanisms, so they're not currently shared. By keeping the connection at this level, we'll be able to make that change cheaply when we need to.

hub/check_source_cluster_configuration_test.go
```diff
@@ -0,0 +1,230 @@
...
+	"strings"
+	"testing"
+
+	"github.com/DATA-DOG/go-sqlmock"
+
+	"github.com/greenplum-db/gpupgrade/utils"
+
+	"golang.org/x/xerrors"
+
+	"github.com/greenplum-db/gpupgrade/hub"
```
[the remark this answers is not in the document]
berlin-ab: Not sure, I'll take a look tomorrow.

hub/fill_cluster_configs.go
```diff
@@ -3,21 +3,28 @@ package hub
...
 	"github.com/greenplum-db/gpupgrade/idl"
 	"github.com/greenplum-db/gpupgrade/step"
 	"github.com/greenplum-db/gpupgrade/utils"
 )
 
 // create old/new clusters, write to disk and re-read from disk to make sure it is "durable"
 func (s *Server) fillClusterConfigsSubStep(_ step.OutStreams, request *idl.InitializeRequest) error {
-	conn := db.NewDBConn("localhost", int(request.SourcePort), "template1")
-	defer conn.Close()
+	connection := db.NewDBConn("localhost", int(request.SourcePort), "template1")
```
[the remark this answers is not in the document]
berlin-ab: fixed.

hub/fill_cluster_configs.go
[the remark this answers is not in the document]
berlin-ab: I replaced this function injection. Please take a look.

cli/commanders/formatting_test.go
```diff
@@ -0,0 +1,19 @@
+package commanders
+
+import (
+	"testing"
+
+	"github.com/greenplum-db/gpupgrade/idl"
+)
+
+func TestFormatStatus(t *testing.T) {
+	t.Run("it formats all possible types", func(t *testing.T) {
```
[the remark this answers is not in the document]
berlin-ab: - I renamed lines. 
- the function called is FormatStatus, so I chose to continue using that terminology
- the test could be improved, but I'll let the next person that comes across this come up with a better test. Before , we didn't have anything that would fail, and then our end-to-end tests would fail.

hub/fill_cluster_configs.go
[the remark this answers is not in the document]
berlin-ab: I tried a few things. I like what I ended up with.

hub/reconfigure_ports.go
```diff
@@ -128,6 +128,10 @@ func ClonePortsFromCluster(db *sql.DB, src *utils.Cluster) (err error) {
 		return err
 	}
 
+	_, err = tx.Exec("SET allow_system_table_mods = true")
```
jchampio: 😞 We've moved away from this pattern in the Python because of the performance characteristics in GPDB. If you want a connection to allow system table modifications, you should make that a connection-level setting. This makes me feel like we've chosen the wrong abstraction, trying to deduplicate things that shouldn't be the same to begin with.

dkrieger-pvtl:
The approach of this PR seems overly heavyweight and introduces new concepts/interfaces(like `sourcedb.Database`) that aren't strictly needed to solve the requested issue.  Those introduced concepts, unless widely used throughout the codebase, make the overall codebase harder for me to understand.

I am thinking it is much cleaner to have `utils.GetSegmentConfiguration` return a slice containing all fields needed for a `SegConfig` plus checking for down/unbalanced segments and then have `utils.NewCluster` return an error if there are any.  For testing, we already have a SQL Mock and a way to define a cluster(see `TestGetSegmentConfiguration`).

In summary, the checking of down/unbalanced segments seems best associated with a Cluster, and to have our code reject the creation of a new Cluster if it does not meet our criteria.  This keeps the business rules inherent in our code.

Thoughts?

berlin-ab:
Cluster up/down is not valuable to store as a part of the Cluster object,
as the cluster has a long lifespan. Primary role also increases complexity
(making all other constructors of the objects ensure the roles match) of
Cluster throughout the codebase without providing value.

The SegmentStatus is a pre-cluster creation purpose-built construct. Any
changes/additions to the pre-cluster checks can go here.

kalensk:
> The approach of this PR seems overly heavyweight and introduces new concepts/interfaces(like `sourcedb.Database`) that aren't strictly needed to solve the requested issue. Those introduced concepts, unless widely used throughout the codebase, make the overall codebase harder for me to understand.

I agree with @dkrieger-pvtl in that there are _a lot_ of major changes in this PR making it quite difficult for me to review. Would it be possible to do the minimal change needed, and introduce some of the other major changes in smaller more digestible PRs that can take affect across the code all at once?

dkrieger-pvtl:
> Cluster up/down is not valuable to store as a part of the Cluster object, as the cluster has a long lifespan. Primary role also increases complexity (making all other constructors of the objects ensure the roles match) of Cluster throughout the codebase without providing value. The SegmentStatus is a pre-cluster creation purpose-built construct. Any changes/additions to the pre-cluster checks can go here.

Agreed.  What I am thinking is that GetSegmentConfiguration obtains the complete information from `gp_segment_configuration`(on 6; on 5 it uses the multi-table join that is currently in the code) and returns that to higher level code.  In some sense, it is "hardware information" and one design technique(which I prefer) argues for returning the complete hardware state.  Then a caller can decide to filter that information.   We don't want it in the Cluster/SegmentConfig, but we can use it to error out when creating one that our system considers inappropriate for upgrading.

Yes, the SegmentStatuses are purpose-built, and I am suggesting that they shouldn't be...this is built into the information we obtain from the database directly and hence it is conceptually cleaner to include it there.

jchampio:
@kalensk The original PR as filed by Adam had different commits for refactoring vs feature changes. It appears they've been squashed -- @jmcatamney can we go back to the previous commit division? I still feel that this PR was doing too much unrelated work, but that at least made it easy to review commit-by-commit.

jmcatamney:
@jchampio 
>  @jmcatamney can we go back to the previous commit division?

I didn't do any squashing yesterday, just `commit --amend`ed `true` to `dml` in the connection string in the third commit, so if there's a branch with more than his 3 + my 1 commit around I don't know where that might be.

hub/reconfigure_ports.go
jmcatamney: I figured this approach would be less far-reaching than changing things at the connection level in what was already a pretty sprawling PR, but I can give that approach a shot and see if it's workable.

berlin-ab:
Sorry, I squashed those commits together.

jchampio:
We've resurrected the original state and are working to rebase it onto latest. ([This article](https://medium.com/git-tips/githubs-reflog-a9ff21ff765f) on how we're doing that is neat.)

hub/reconfigure_ports.go
jchampio: So a point of confusion that might(?) be resolved here is that this reconfigure-ports implementation uses its _own_ connection string with the correct GUC value. We might not have to add the `SET` statements at all.

jchampio: Done.

hub/sourcedb/segment_status_test.go
```diff
@@ -0,0 +1,47 @@
...
+		if err != nil {
+			t.Errorf("unexpected error: %v", err)
+		}
+
+		if len(statuses) != 2 {
+			t.Fatalf("got %d rows, expected 2 rows to be returned", len(statuses))
+		}
+
+		first := statuses[0]
+		if first.DbID != 1 || first.IsUp != true || first.Role != Mirror || first.PreferredRole != Primary {
```
jchampio: Because the `SegmentStatus` struct is so simple, I think these verbose comparisons can be simplified by creating a series of `expected` structs. In fact you could create a slice of expected structs and `reflect.DeepEqual()` for a single-pass verification.

berlin-ab:
TLDR:  We could improve our design by creating abstractions, making our system simpler and easier to test.

(this is not a blocking comment, please merge with my approval)

--------

If you look at the `hub.Execute` handler as as tree, all of the functions it calls down to the leaf nodes are considered its responsibility. This tree of responsibility is hard to test, and therefore hard to change without falling back on our end-to-end tests.

We already see that `FillClusterConfigsSubStep` is not tested. Some of the functions called within FillClusterConfigsSubStep are tested, but their interaction is not.

This pattern repeats itself throughout the codebase. One strategy to combat this complexity is to extract components that can be tested in isolation (forming a new tree), and then use their interface as a seam to stub against while testing the complex tree. 

A rule of thumb used to drive separation here is called The Single Responsibility Principle (SRP): A class/function should have one reason to change. We can observe reasons for change by writing tests. The business logic is a reason to change, the SQL used to fill domain objects is a reason for change. Following the SRP we'd form isolate these two reasons to change from each other so that a change in one does not force a change in the other. Because these two need to agree on a contract, there needs one side of the relationship will need to change in response to the other, but the forced change should not be bi-directional.

The Dependency Inversion Principle is also used to accomplish this separation. While splitting apart these two responsibilities, we see a clear separation of High Level abstraction (business logic) and Low Level detail (SQL). To avoid coupling our high level to our low level, we provide the low level to the high level as an abstraction - rather than a concrete function call which cannot be swapped out for an alternative version of the abstraction.

A recommendation I have, is that we start to actually write tests for the top of the tree, so that we force ourselves to feel this design feedback and then change our design to make change easier. Otherwise, the more features we implement will introduce more and more complexity to the tree and make it even harder to test. Someday we'll need to make a change to someplace near the root of the tree that we'd like to get test feedback for, and we'll not be able to.

dkrieger-pvtl:
> TLDR: We could improve our design by creating abstractions, making our system simpler and easier to test.

In principle, yes.  But it's a matter of degree, and making sure the abstractions we introduce correspond to the system we are upgrading.  For instance, the abstraction of a "cluster" makes sense because upgrade is creating a new cluster from the old one.   I think it was this PR that originally had the concept of a `database` object used only in one place; to me, that adds complexity because that object didn't appear to me to correspond to the high-level idea we have in gpupgrade.

However, I find that adding single-purpose abstractions for introducing test seams for a particular feature over time reduces the conceptual integrity of the code.  This makes it harder for me to understand what the code is doing, and even to make changes.  There is a tradeoff here, at least in how I think about the code, between making sure the code is well-tested and making sure the code is understandable to read(and hence to modify) and to reason about.  For this reason, sometimes end-to-end tests can provide a valuable tradeoff against the introduction of test seams that can make the code harder to understand.

For instance, in the PR #227 (adding `--link` to linkMode pg_upgrade --check), there were two apparent testing ideas: unit testing by adding a test seam to allow testing of the arguments passed into the lower layers  and integration testing to show that a link-mode check passes before our change and fails after.  We decided on unit testing because the only integration test check required hard-linking to fail(which would require the source/target clusters to be on different filesystems) would be difficult to implement relative to the benefits of this test.  Furthermore, the responsibility of check_upgrade seemed to only be to call the underlying layers with the correct arguments.  However, the code is a bit inelegant, and had the integration test been easier, we might have decided to do that instead.

In summary, deciding on how to test a given piece of code under development or maintenance is always a tradeoff that has to be considered on a case-by-case basis for its effect on the codebase overall.

jchampio:
> However, I find that adding single-purpose abstractions for introducing test seams for a particular feature over time reduces the conceptual integrity of the code.

+1.

@berlin-ab The goal of the final commit here was to hopefully provide evidence that the `sql.DB` concept is already a usable abstraction layer; one that's widely applicable to the codebase. We had the goal of maintaining the test coverage you'd already added, and I believe we succeeded (and in the process we were able to add additional coverage to the layer closest to the database). But if there's a place you see that we clearly dropped coverage compared to the previous implementation, please let us know because it's unintended.

dkrieger-pvtl:
I still think the implementation is overly complicated for what it's doing...adding the check entirely in NewCluster seems way, way cleaner to me...and makes sense conceptually.  The current implementation is "adhoc" in the sense that we make our own db connection, fill SegmentStatus objects(which have no analogue in GPDB) and check them.  Why not manifestly refuse to construct a mal-formed cluster in NewCluster?

kalensk:
> I still think the implementation is overly complicated for what it's doing...adding the check entirely in NewCluster seems way, way cleaner to me...and makes sense conceptually. The current implementation is "adhoc" in the sense that we make our own db connection, fill SegmentStatus objects(which have no analogue in GPDB) and check them. Why not manifestly refuse to construct a mal-formed cluster in NewCluster?
> 
> That said, we decided to use the current approach, so I'll review with this design as a given.

To expand on this a bit, I think the code would be cleaner if querying from gp_segment_configuration returns rows with all data (ie: Status and PreferredRole) in a new object such as utils.GpSegRow. Then `NewCluser` would return a utils.cluster in-memory object of just the fields we care about for passing around in the code base (ie: no Status and PreferredRole fields).

We would only create a NewCluster if it does not have down segments, or if the roles are not mismatched. That is, `NewCluster` only returns valid clusters. That is, `NewCluster` could call validateRow which might looks something like:

```go
// validateRow returns an error if a gp_segment_configuration segment
// is not up, or if its role does not match its preferred role.
func validateRow(row *utils.GpSegRow) error {
	if row.Status != "u" {
		return newDownSegmentError(row)
	}

	if row.Role != row.PreferredRole {
		return newRoleMismatchError(row)
	}

	return nil
}
```

The current could is still quite confusing to me. I think the suggested approach would remove the need for the extra objects such as `SegmentStatus` and greatly increase code understanding through "the locality of reference".

berlin-ab:
@dkrieger-pvtl @kalensk You're both right. Those alternative solutions would work. What is there now does also work, and is simple from my perspective.

hub/check_source_cluster_configuration.go
```diff
@@ -0,0 +1,175 @@
...
+func (e UnbalancedSegmentStatusError) Error() string {
+	var dbidStrings []string
+
+	for _, dbid := range e.UnbalancedDbids {
+		dbidStrings = append(dbidStrings, strconv.Itoa(int(dbid)))
+	}
+
+	message := fmt.Sprintf("Could not initialize gpupgrade. These"+
+		" Greenplum segment dbids are not in their preferred role: %v."+
+		" Run gprecoverseg -r to rebalance the cluster.", strings.Join(dbidStrings, ", "))
```
dkrieger-pvtl: Do we want to say `Run gprecoverseg -r to rebalance the cluster` or something more generic like `Use gprecoverseg to rebalance the cluster`?  @kodonnell26   Looking over the gprecoverseg online reference, there are other options(like `-B`) and a comment `All segments must be valid and synchronized before running gprecoverseg -r`.  So I wonder if we want to leave the exact details of the gprecoverseg command options to the user.

jchampio: This verbiage comes from the previous implementation; it's not new to this PR.

hub/initialize.go
```diff
@@ -33,8 +31,18 @@ func (s *Server) Initialize(in *idl.InitializeRequest, stream idl.CliToHub_Initi
...
+	conn, err := sql.Open("pgx", fmt.Sprintf(connectionString, in.SourcePort))
+	if err != nil {
+		return err
+	}
+	defer func() {
+		if cerr := conn.Close(); cerr != nil {
+			err = multierror.Append(err, cerr).ErrorOrNil()
+		}
+	}()
+
```
dkrieger-pvtl: This entire code block creating a conn is only used inside FillClusterConfigSubStep.  Hence, it should be placed inside the anonymous function that runs the `idl.Substep_CONFIG` step(L45).  Also, if the `idl.Substep_CONFIG` has already run successfully, it won't run again and `conn` is created but not used.

hub/check_source_cluster_configuration.go
```diff
@@ -0,0 +1,175 @@
+package hub
```
dkrieger-pvtl: I'd prefer the identifiers in this file to be re-ordered.  I couldn't quickly find a good normative layout reference, but I think the code is more readable with:

```
constants
types
non-error functions
error functions
```

It might be faster to pair on the ordering to see if you agree it looks cleaner...see `hub/server.go` for an example.

hub/check_source_cluster_configuration.go
dkrieger-pvtl: Ok, we can defer to later then.

PR #232, opened 2020-02-27
Swap datadirs reconfigure

bhuvnesh2703:
This PR has 2 commits:

    #1 Finalize: Rename data directories

    a). Renaming the source and target data directories, example:
    Source (old) cluster:
    /data/dbfast1/demoDataDir0 becomes datadirs/dbfast1_old/demoDataDir0
    Target (new) cluster:
    /data/dbfast1_upgrade/demoDataDir0 becomes datadirs/dbfast1/demoDataDir0
    
    b). Combines multiple finalize bats files into one. The test file finalize-standby.bats was only check if the standby has been added to the cluster. That can go in finalize.bats, so merged it.

    There are a few outstanding issues which will be addressed in a followup
    PR.
    - rename the standby data directories
    - create crash-only idempotent semantics

    #2 Don't update standby datadirs or port


agent/rename_data_directories.go
```diff
@@ -0,0 +1,25 @@
...
+import (
+	"context"
+	"os"
+
+	"github.com/greenplum-db/gp-common-go-libs/gplog"
+
+	"github.com/greenplum-db/gpupgrade/idl"
+)
+
+var Rename = os.Rename
```
berlin-ab: Why not use the `utils.System` abstraction for this?

hub/finalize.go
```diff
@@ -51,21 +57,74 @@ func (s *Server) Finalize(_ *idl.FinalizeRequest, stream idl.CliToHub_FinalizeSe
...
-		return UpdateCatalogWithPortInformation(s.Source, s.Target)
+	// XXX: UpdateCatalog updates the in-memory target cluster and config.json
+	// to match what has been changed in the catalog.
+	// TODO: When in copy mode we should probably update the catalog and
+	//  in-memory object of the source cluster? Or will this cause divergence
+	//  when in link mode since the renaming datadirs has expectations on what
+	//  the incoming directory looks like. And handeling different datadirs for
+	//  link and copy mode would be too cumbersome for the rename substeps?
+	st.Run(idl.Substep_FINALIZE_UPDATE_CATALOG, func(streams step.OutStreams) error {
+		return s.UpdateCatalog()
```
berlin-ab: Putting `UpdateCatalog` back with Server as a receiver seems like a slide backward. The server's clusters are sufficient to get the job done.

hub/finalize.go
```diff
@@ -51,21 +57,74 @@ func (s *Server) Finalize(_ *idl.FinalizeRequest, stream idl.CliToHub_FinalizeSe
...
 	})
 
 	st.Run(idl.Substep_FINALIZE_SHUTDOWN_TARGET_MASTER, func(streams step.OutStreams) error {
-		return StopMasterOnly(streams, s.Target, false)
+		// Create an oldTarget cluster to pass to StopMasterOnly since
```
berlin-ab: I'd like to see tests for this behavior.

hub/update_conf_files.go
```diff
@@ -0,0 +1,50 @@
...
+	"os/exec"
+
+	"github.com/greenplum-db/gp-common-go-libs/gplog"
+	"golang.org/x/xerrors"
+
+	"github.com/greenplum-db/gpupgrade/idl"
+	"github.com/greenplum-db/gpupgrade/utils"
+)
+
+func updateGpperfmonConf(masterDataDir string) error {
```
berlin-ab: It would be nice to have tests for this behavior.

hub/update_conf_files.go
```diff
@@ -0,0 +1,50 @@
...
+			idl.Substep_FINALIZE_UPDATE_GPPERFMON_CONF, err)
+	}
+	return nil
+}
+
+// oldTargetPort is the old port on which the target cluster was initialized.
+// This is used to search the postgresql.conf rather than target.MasterPort()
+// which has been changed to match the source cluster after updating the
+// catalog in a previous substep.
+func updatePostgresqlConf(oldTargetPort int, target *utils.Cluster, source *utils.Cluster) error {
```
berlin-ab: It would be nice to have tests for this behavior.

hub/rename_data_directories_test.go
```diff
@@ -0,0 +1,183 @@
...
+		}
+
+		if len(multiErr.Errors) != 1 {
+			t.Errorf("received %d errors, want %d", len(multiErr.Errors), 1)
+		}
+
+		for _, err := range multiErr.Errors {
+			if !strings.Contains(err.Error(), expected.Error()) {
+				t.Errorf("wanted error message '%q', got %q", expected.Error(), err.Error())
+			}
```
kalensk: These types of assertions in this file should probably be updated to:
```go
	t.Errorf("wanted error %#v, got %#v", expected, err)
}
```

hub/finalize.go
kalensk: Based on @dkrieger-pvtl https://github.com/greenplum-db/gpupgrade/pull/220#discussion_r382343135 about keeping the various sources of truth in sync, we update the source and target cluster objects to match the catalog _and_ persist to disk _after_ successfully committing the catalog changes. This is done by calling `s.SaveConfig()` which is why we added it to the server recover. Is there another way without making too big of a change all at once?

berlin-ab: Give the UpdateCatalog function the SaveConfig function as an argument.

kalensk: > Give the UpdateCatalog function the SaveConfig function as an argument.

Sure that is an option, but then we would also need to pass in the source and target clusters (not sure if I am missing anything else), and at that point it made sense to just add it to the server object. Everything is still testable so far, and doesn't seem to loose us anything with the _current_ code.

kalensk: > Give the UpdateCatalog function the SaveConfig function as an argument.

That doesn't really work without a major refactor since `SaveConfig()` is a member of the Server and uses several data from the server struct like the `StateDir` and the entire `Config` object. One would have to pass through all these pieces of data into `UpdateCatalog` _and_ into the injected `SaveConfig`. That doesn't seem useful to me rather than adding UpdateCatalog as a member to the server object, which still keeps things testable and clean. Am I missing something?

hub/update_conf_files.go
kalensk: Do you have any specific ideas on how to "unit test" this without simply testing the script string we use is the one we want (ie: a simple string comparison on a script)? That unit test does not seem useful to me and not worth adding. 

I'd have to think about it some more, but perhaps an integration or bats level test?

dkrieger-pvtl: The `finalize.bats` test has an end-to-end test for this.

hub/update_conf_files.go
dkrieger-pvtl: The `finalize.bats` test has an end-to-end test for this.

dkrieger-pvtl:
The main PR description says:

```
Source (old) cluster:
/data/dbfast1/demoDataDir0 becomes datadirs/dbfast1**_old**/demoDataDir0
```

However, we want
```
Source (old) cluster:
/data/dbfast1/demoDataDir0 becomes datadirs/dbfast1/demoDataDir0**_old**
```

kalensk:
> However, we want
> 
> ```
> Source (old) cluster:
> /data/dbfast1/demoDataDir0 becomes datadirs/dbfast1/demoDataDir0_old
> ```

I am confused, why would we not want to simply "undo" what pg_upgrade does? That is, if pg_upgrade adds `_upgrade` why would we not want to do the same thing, but remove that with `_old`? Why deviate from upstream in a sense?

dkrieger-pvtl:
> > However, we want
> > ```
> > Source (old) cluster:
> > /data/dbfast1/demoDataDir0 becomes datadirs/dbfast1/demoDataDir0_old
> > ```
> 
> I am confused, why would we not want to simply "undo" what pg_upgrade does? That is, if pg_upgrade adds `_upgrade` why would we not want to do the same thing, but remove that with `_old`? Why deviate from upstream in a sense?

Once finalize is run, we consider `source` cluster to be unrunnable.  But I am not sure how stringent our requirement on directory naming is, but I thought someone could create a segment in /datadir/ and then you'd rename / to /_old  or error out.  Maybe you can create a datadir in /.

kalensk:
> Once finalize is run, we consider `source` cluster to be unrunnable. 

I think this is debatable when in copy mode.

> But I am not sure how stringent our requirement on directory naming is, but I thought someone could create a segment in /datadir/ and then you'd rename / to /_old or error out. Maybe you can create a datadir in /.

I might be confusing myself, but If that is true would pg_upgrade run into similar issues when adding _upgrade? Otherwise why doesn't pg_upgrade add _upgrade to the datadir?

jchampio:
@kalensk `pg_upgrade` doesn't create data directories at all; that's a `gpupgrade` feature. `pg_upgrade` expects you to have already created the new data directory at the location you want it.

kalensk:
Its just strange to me to have _upgrade and _old not "match". But either way its an easy fix.

hub/update_catalog.go
```diff
@@ -107,49 +135,66 @@ func commitOrRollback(tx *sql.Tx, err error) error {
...
 	}
 	defer func() {
 		err = commitOrRollback(tx, err)
+		if err == nil {
+			// After successfully changing the catalog, update the source and
+			// target cluster objects to match the catalog and persist to
+			// disk.
+			tmpTarget := s.Target
+
+			s.Target = s.Source
```
dkrieger-pvtl: I think we want to do a deep copy of Source onto Target, not a shallow copy.  Otherwise, the target Cluster is aliased to the source Cluster...this is semantically ok right now, but leaves a trap for later.

```
package main

import (
	"fmt"
)

type shallow struct {
	data []int
}

func main() {
	var src shallow
	for i:=0;i<10;i++ {
		src.data = append(src.data,i)
	}
	fmt.Printf("src orig: %v\n",src.data)

	var dst *shallow
	dst = &src
	fmt.Printf("dst orig: %v\n",dst.data)
	
	src.data[0] = 22
	fmt.Printf("src shallow: %v\n",src.data)	
	fmt.Printf("dst shallow: %v\n",dst.data)
	
	deep := &shallow{}
	deep.data = make([]int,len(src.data))
	copy(deep.data, src.data)	
	src.data[1] = 33 
	fmt.Printf("src deep: %v\n",src.data)	
	fmt.Printf("dst deep: %v\n",deep.data)
}
```

hub/finalize.go
dkrieger-pvtl: I'm wondering if we want to put the "oldTarget" concept in the server.Config and not spread the logic here.

hub/update_catalog.go
```diff
@@ -107,49 +135,66 @@ func commitOrRollback(tx *sql.Tx, err error) error {
 	return nil
 }
 
-// ClonePortsFromCluster will modify the gp_segment_configuration of the passed
+// UpdateGpSegmentConfiguration will modify the gp_segment_configuration of the passed
 // sql.DB to match the cluster port settings from the source utils.Cluster.
 //
 // As a reminder to developers, we don't have any mirrors up at this point on
 // the target cluster. We copy only the primary information. Good thing too,
```
dkrieger-pvtl: remove `Good thing too, because utils.Cluster doesn't give us mirror info.`. The source cluster is populated with mirror information now. (see `utils.NewCluster`).

hub/finalize.go
```diff
@@ -51,21 +57,74 @@ func (s *Server) Finalize(_ *idl.FinalizeRequest, stream idl.CliToHub_FinalizeSe
...
+		return nil
 	})
 
 	st.Run(idl.Substep_FINALIZE_UPDATE_POSTGRESQL_CONF, func(streams step.OutStreams) error {
-		return UpdateMasterPostgresqlConf(s.Source, s.Target)
+		return updatePostgresqlConf(s.TargetPorts.Master, s.Target, s.Source)
+	})
+
+	st.Run(idl.Substep_FINALIZE_UPDATE_GPPERFMON_CONF, func(streams step.OutStreams) error {
+		return updateGpperfmonConf(s.Source.MasterDataDir())
```
dkrieger-pvtl: The function argument should be `s.Target.MasterDataDir()`...it's value should agree with `s.Source.MasterDataDir()` but we are intending to modify the target cluster.

hub/finalize.go
```diff
@@ -51,21 +57,74 @@ func (s *Server) Finalize(_ *idl.FinalizeRequest, stream idl.CliToHub_FinalizeSe
...
-	st.Run(idl.Substep_FINALIZE_UPDATE_CATALOG_WITH_PORT, func(streams step.OutStreams) error {
-		return UpdateCatalogWithPortInformation(s.Source, s.Target)
+	// XXX: UpdateCatalog updates the in-memory target cluster and config.json
+	// to match what has been changed in the catalog.
+	// TODO: When in copy mode we should probably update the catalog and
+	//  in-memory object of the source cluster? Or will this cause divergence
+	//  when in link mode since the renaming datadirs has expectations on what
+	//  the incoming directory looks like. And handeling different datadirs for
+	//  link and copy mode would be too cumbersome for the rename substeps?
+	st.Run(idl.Substep_FINALIZE_UPDATE_CATALOG, func(streams step.OutStreams) error {
```
dkrieger-pvtl: Consider renaming `Substep_FINALIZE_UPDATE_CATALOG` to `Substep_FINALIZE_UPDATE_TARGET_CATALOG`; the other step names are explicit wrt Source/Target.

hub/update_catalog.go
```diff
@@ -14,6 +12,36 @@ import (
 	"github.com/greenplum-db/gpupgrade/utils"
 )
 
+func (s *Server) UpdateCatalog() (err error) {
```
dkrieger-pvtl: This function name `UpdateCatalog` makes me think this function just updates the database's catalog, but it also updates the in-memory target cluster and the json config.  I can't think of a great name that captures this.....`UpdateCatalogClusterConfig` is gross.  `UpgradeTriad` is too pithy.  Maybe `UpdateCatalogAndClusterConfig`?

hub/finalize.go
kalensk: I think @marksliva is putting up a PR in the next few days that should address this. That is, he has a pre-factor PR on upgrading the recovery.conf on the mirrors. Where it has a concept of an `InitializeConfig` which is the desired state of the cluster from Initialize with the temp ports used. That is, the "old" target cluster state.

```go
type InitializeConfig struct {
	Standby    utils.SegConfig
	Master	   utils.SegConfig
	Primaries  []utils.SegConfig
	Mirrors    []utils.SegConfig
}
```

Thus, the above code would become something like `StopMasterOnly(streams, &initializeConfig.Master, false)`.

hub/update_catalog.go
kalensk: Maybe we could use https://github.com/mitchellh/copystructure?

dkrieger-pvtl: We could but for now the structure is simple enough that a direct implementation in utils suffices.  I'm all for code re-use but there are a few outstanding feature requests on that repo, which hasn't been modified for 1.5 years or so.  There's the interesting issue of copying locks which the author seems to account for.

dkrieger-pvtl:
> The main PR description says:
> 
> ```
> Source (old) cluster:
> /data/dbfast1/demoDataDir0 becomes datadirs/dbfast1_old/demoDataDir0
> ```
> 
> However, we want
> 
> ```
> Source (old) cluster:
> /data/dbfast1/demoDataDir0 becomes datadirs/dbfast1/demoDataDir0_old
> ```

We have decided to move the source datadirs to /my/path/`qddir_old`/datadir, and NOT /my/path/qddir/`datadir_old`.

hub/rename_data_directories_test.go
marksliva: I thought the typical ordering for error messages is `got error %#v, want %#v`.

hub/update_catalog.go
kalensk: I used https://github.com/jinzhu/copier since it made the calling code a bit simpler.

the two outstanding issues on the other repo is because they are limitations of golang, and not due to the library. The reason for it not being modified in 1.5 years is because once you get deep copying working there isn't really anything to maintain. It just works.

hub/rename_data_directories_test.go
kalensk: oops! fixed.

agent/rename_data_directories.go
kalensk: Update to system.Utils.Rename

hub/finalize.go
kalensk: fixed.

hub/finalize.go
dkrieger-pvtl: ok...as long as this code is updated to use it once that PR goes in.

hub/finalize.go
```diff
@@ -35,6 +38,8 @@ func (s *Server) Finalize(_ *idl.FinalizeRequest, stream idl.CliToHub_FinalizeSe
 				streams:             streams,
 			}
 
+			// TODO: Persist the standby config to status.json and updating the
```
dkrieger-pvtl: s/status.json/config.json

hub/finalize.go
```diff
@@ -51,21 +56,62 @@ func (s *Server) Finalize(_ *idl.FinalizeRequest, stream idl.CliToHub_FinalizeSe
...
+	st.Run(idl.Substep_FINALIZE_RENAME_SOURCE_MASTER_DATADIR, func(streams step.OutStreams) error {
+		return RenameMasterDataDir(s.Source.MasterDataDir(), true)
+	})
+
+	// TODO: Update RenameSegmentDataDirs to include renaming the standby by
+	//  changing SegmentsOn to return the standby. Careful since this is used by
+	//  other callers. AgentConns() also needs to be updated to create an
+	//  agentConn on the standby host.
+	st.Run(idl.Substep_FINALIZE_RENAME_SOURCE_SEG_DATADIRS, func(streams step.OutStreams) error {
+		// e.g. /data/dbfast1/demoDataDir0 becomes datadirs/dbfast1_old/demoDataDir0
```
dkrieger-pvtl: nit: consider moving this to a function-level description of `RenameSegmentDatadirs`.

idl/hub_to_agent.proto
```diff
@@ -8,6 +8,7 @@ service Agent {
     rpc CheckDiskSpace (CheckSegmentDiskSpaceRequest) returns (CheckDiskSpaceReply) {}
     rpc UpgradePrimaries (UpgradePrimariesRequest) returns (UpgradePrimariesReply) {}
     rpc CreateSegmentDataDirectories (CreateSegmentDataDirRequest) returns (CreateSegmentDataDirReply) {}
+    rpc RenameDataDirectories (RenameDataDirRequest) returns (RenameDataDirReply) {}
```
dkrieger-pvtl: As with `RenameSegmentDataDirs`, this agent call is not renaming data directories, but rather their parents.  But the naming here implies otherwise.  On the hub, the call in `finalize.go` is semantically doing a renaming of segment data dirs, but just happens to be renaming their parents instead.  On the agent, you are just doing a directory move.

Maybe just call this `RenameDirectory` since that's all it is doing and leave it generic?  The hub is orchestrating the agents in our architecture, anyway.

kalensk: Part of the confusion (at least for me and @bhuvnesh2703 ) is that there is no good consistent terminology to differentiate the datadir from the segment dir, since datadir is an overloaded term. For example, `/data/primary/gpseg2` what do we call `primary` and `gpseg2` in this case? I think having good terminology might unpack and simplify a lot of what you're comments are addressing.

hub/rename_data_directories.go
```diff
@@ -0,0 +1,101 @@
...
+)
+
+// e.g. for source /data/qddir/demoDataDir-1 becomes /data/qddir_old/demoDataDir-1
+// e.g. for target /data/qddir_upgrade/demoDataDir-1 becomes /data/qddir/demoDataDir-1
+func RenameMasterDataDir(masterDataDir string, isSource bool) error {
+	destination := "new"
+	src := AddUpgradeSuffix(filepath.Dir(masterDataDir))
+	dst := filepath.Dir(masterDataDir)
+	if isSource {
+		destination = "old"
```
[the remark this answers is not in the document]
kalensk: If possible, I'd like to follow up one more time with PMs (@bharath1028 and @Nirali13 and @kodonnell26 ) and designers since this would be breaking convention with the entire code base when it comes to bubbling up user facing error messages and log lines. 

In my mind, it would be nice to do the change all at once across the code base, but I am not tied to that. Happy to switch it, just want to make sure one last time.

idl/hub_to_agent.proto
dkrieger-pvtl: At least to Greenplum developers, `data directory` is well-defined...it's the directory containing the postgres-instance data for a segment.  In your example, `/data/primary/gpseg2` is a data directory(presuming there's no subdirectory before the postrgres files).  In absence of more specific terminology, I was thinking of using standard CS language(parent, path element, etc).  There may be a cleaner technique.

To an agent however in this case, it is literally just renaming a directory, so that should provide a clean naming for the agents.  It's the parent that contains the logic.

hub/rename_data_directories.go
```diff
@@ -0,0 +1,101 @@
...
+		src = filepath.Dir(masterDataDir)
+		dst = AddOldSuffix(filepath.Dir(masterDataDir))
+	}
+	if err := utils.System.Rename(src, dst); err != nil {
+		return xerrors.Errorf("renaming %s cluster master data directory from: '%s' to: '%s': %w", destination, src, dst, err)
+	}
+	return nil
+}
+
+func RenameSegmentDataDirs(agentConns []*Connection,
```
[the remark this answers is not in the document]
kalensk: I like the intention for making the code clearer and more straightforward. Let's try and find a way to help make that happen. 

I am not sure the suggestion is getting us much benefit? To me it is a tad confusing, but I am biased. Basically, it is moving transforming the source and destination directories from parameters to a conditional if statement. I generally like to avoid if statements, and in this case having the two explicit parameters is fairly direct.

Let's see if there is a way to make things more straightforward though!

idl/hub_to_agent.proto
kalensk: Ah! After thinking for a bit, I see your point more clearly. The agents are simply renaming two directories, they don't care (or should know about) what they are renaming. Yeah, I am ok with changing the name to `RenameDirectory` (or the like), and the associated variables and functions in the agent package.

hub/rename_data_directories.go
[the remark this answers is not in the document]
dkrieger-pvtl: I don't like unnecessary conditionals either but to me it's a tradeoff between making the intent of the function clearer(it renames one directory and hence only needs one `tfrm` function) and making the implementation a little less verbose.

hub/update_catalog.go
marksliva: One approach I can think of which would avoid adding a dependency would be to load the config from `config.json` using `Config.Load(r io.Reader)`, and assign the source from there to the target.

hub/rename_data_directories.go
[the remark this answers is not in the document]
kalensk: We decided to defer until after the alpha release and possible rename all at once throughout the code base.

hub/rename_data_directories.go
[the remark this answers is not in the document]
kalensk: Updated.

Gopkg.lock
```diff
@@ -122,6 +122,14 @@
   revision = "89f1e6ac7276b61d885db5e5aed6fcbedd1c7e31"
   version = "v3.2.0"
 
+[[projects]]
+  branch = "master"
+  digest = "1:ae940e9156d0d778ce1991bcc99e58fa5ca2790ddf7c5f0bc05385854416f7a6"
+  name = "github.com/jinzhu/copier"
```
jchampio: What kind of vetting did this new dependency go through? We good with the license, the code quality, etc.? It seems tangential enough to our core goal that we could replace it with our own implementation without too much trouble.

kalensk: To me this raises the bigger question of can we create a team process or better understanding when it comes to adding dependencies? For me at least its a pretty big blind spot?

hub/update_catalog.go
dkrieger-pvtl: Instead, we did a struct copy...it works because we get a new struct.
```go
package main
import  "fmt"

type foo struct {
	vals []int
}

func main() {
	source := &foo{vals:[]int{1, 2, 3}}
	target := *source //creates a new foo

	source.vals = append(source.vals, 4)

	fmt.Printf("source: %#v\n", source)
	fmt.Printf("target: %#v\n", target)
}

source: &main.foo{vals:[]int{1, 2, 3, 4}}
target: main.foo{vals:[]int{1, 2, 3}}
```

Gopkg.lock
kalensk: No new dependency needed after we used @marksliva https://github.com/greenplum-db/gpupgrade/pull/232#discussion_r387181399 idea of loading the conf.

hub/rename_data_directories.go
```diff
@@ -0,0 +1,131 @@
...
+			if err != nil {
+				errs <- err
+				return
+			}
+
+			// When there are multiple segments under a parent data directory
+			// only call rename once on the parent directory.
+			// For example, /data/primary/gpseg1 and /data/primary/gpseg2
+			// only call rename once for /data/primary.
+			dataDirs := make(map[string]bool)
```
dkrieger-pvtl: s/dataDirs/dataDirParents/g

hub/rename_data_directories.go
```diff
@@ -0,0 +1,131 @@
...
+			for _, seg := range segments {
+				dataDirs[filepath.Dir(seg.DataDir)] = true
+			}
+
+			req := new(idl.RenameDirectoriesRequest)
+			for dataDir := range dataDirs {
```
dkrieger-pvtl: s/dataDir/dataDirParent/g

hub/rename_data_directories_test.go
```diff
@@ -0,0 +1,238 @@
+package hub_test
+
+import (
+	"errors"
+	"testing"
+
+	"github.com/golang/mock/gomock"
+	"github.com/greenplum-db/gp-common-go-libs/testhelper"
```
dkrieger-pvtl: This import should be below is current block.

hub/update_catalog.go
```diff
@@ -3,17 +3,71 @@ package hub
...
+	"github.com/greenplum-db/gpupgrade/step"
 	"github.com/greenplum-db/gpupgrade/utils"
 )
 
+// XXX: UpdateCatalogAndClusterConfig updates the in-memory target cluster and config.json
+// to match what has been changed in the catalog.
+// TODO: When in copy mode we should probably update the catalog and
+//  in-memory object of the source cluster? Or will this cause divergence
+//  when in link mode since the renaming datadirs has expectations on what
+//  the incoming directory looks like. And handeling different datadirs for
```
dkrieger-pvtl: s/handeling/handling

hub/update_catalog.go
```diff
@@ -3,17 +3,71 @@ package hub
...
+func (s *Server) UpdateCatalogAndClusterConfig(streams step.OutStreams) (err error) {
+	err = StartMasterOnly(streams, s.Target, false)
+	if err != nil {
+		return xerrors.Errorf("starting new master: %w", err)
+	}
+
+	err = WithinDbConnection(s.Target.MasterPort(), func(conn *sql.DB) error {
+		return s.UpdateGpSegmentConfiguration(conn)
+	})
+
```
dkrieger-pvtl: remove this newline

hub/update_catalog.go
```diff
@@ -3,17 +3,71 @@ package hub
...
+	}
+
+	// Create an oldTarget cluster to pass to StopMasterOnly since
+	// UpdateCatalogAndClusterConfig mutates the target cluster with the new data directories
+	// which have yet to be reflected on disk in a later substep.
+	master := s.Target.Primaries[-1]
+	master.DataDir = upgradeDataDir(master.DataDir)
+	segs := map[int]utils.SegConfig{-1: master}
+	oldTarget := &utils.Cluster{Primaries: segs, BinDir: s.Target.BinDir}
+	err = StopMasterOnly(streams, oldTarget, false)
```
dkrieger-pvtl: add a newline between above this line.

PR #234, opened 2020-03-02
finalize: add upgrade mirrors substep

marksliva:
We remove the PortAssigner, and introduce an InitializeConfig.  This allows us
to more easily track the temporary ports and other SegConfig info of the target
master, standby, primaries and mirrors.

When adding mirrors, we have to loop on a FTS probe scan until we detect that
all segments are synchronized before shutting down the cluster. Failing to do
so results in a hard to recover cluster.

We also had to manually set the port in the mirror recovery.conf files to point
to the correct port for the target cluster before starting it for the final
time.

The end-to-end test verifies that the new mirrors are valid by failing to them
and rebalancing.


dkrieger-pvtl:
I presume you mainly want high-level comments.  As for the actual creation of the mirrors, your approach is good: write the mirrors config file, create the mirrors using `gpaddmirrors` and then wait for the mirrors to synchronize.  

I am not sure if there's a better way to create a mirror quickly than `gpaddmirrors`...that is likely to take a very long time since we have to do a full `pg_basebackup`.  

Note  `waitForFTS()` isn't checking for the mirrors to be synchronized, just up:
```
rows, err = db.Query(`
				SELECT every(status = 'u')
					FROM gp_segment_configuration
					WHERE role = 'm'
			`)
```

I think you want to wait for the `mode = 's'` if checking for mirror synchronization is your goal.   You could also check for `status = 'u'`; I am not sure if a segment can be synchronized if you're not up. 

That also resolves any race conditions...I am not sure what state we guarantee for mirrors to be in after `gpaddmirrors` completes successfully.

marksliva:
> Note `waitForFTS()` isn't checking for the mirrors to be synchronized, just up:
Right, that was intentional. To fix the immediate problem we were having, we just had to check if the mirrors were up before continuing on. The mirrors need to be up in gp_segment_configuration, because if they are down when we stop the cluster, the cluster gets into a bad state which is difficult to recover from.

This brings up an interesting point though. Perhaps we should wait for mirrors to finish syncing before we consider a cluster fully finalized. Let's check with product on that one.

marksliva:
Closing in favor of: https://github.com/greenplum-db/gpupgrade/pull/236

dkrieger-pvtl:
> 
> After discussing with product, we decided to add a final substep which waits until the mirrors are fully synchronized. This feature will be implemented in a new story.

@marksliva  Just a comment...it's probably a bug that `gpaddmirrors` returns before the mirrors show up as `up` in `gp_segment_configuration`.  That is, `gpaddmirrors` should probably only return once that's true...otherwise a user would justifiably be confused.

As for sync'ing before finalize completes, I wonder how long that takes relative to the time it takes to run `gpaddmirrors`...if it takes a really long time(hours), I am not sure how a user would feel.

PR #235, opened 2020-03-02
finalize: refactor port logic and init target cluster

marksliva:
Introduce a new object called InitializeConfig. It is used to store
configuration for creating segments on the target cluster. We create a master,
standby and primaries based on the contents of this struct, and it is also
saved to the disk in config.json.

This refactor is done in preparation for finalize: upgrade mirrors https://github.com/greenplum-db/gpupgrade/pull/236

### DIFF
--- a/hub/finalize.go
+++ b/hub/finalize.go
@@ -36,7 +36,7 @@ func (s *Server) Finalize(_ *idl.FinalizeRequest, stream idl.CliToHub_FinalizeSe
 			}
 
 			return UpgradeStandby(greenplumRunner, StandbyConfig{
-				Port:          s.TargetPorts.Standby,
+				Port:          s.TargetInitializeConfig.Standby.Port,
 				Hostname:      s.Source.StandbyHostname(),
 				DataDirectory: s.Source.StandbyDataDirectory() + "_upgrade",
 			})
--- a/hub/initialize.go
+++ b/hub/initialize.go
@@ -105,7 +105,7 @@ func (s *Server) fillClusterConfigsSubStep(_ step.OutStreams, request *idl.Initi
 		ports = append(ports, int(p))
 	}
 
-	s.TargetPorts, err = assignPorts(s.Source, ports)
+	s.TargetInitializeConfig, err = AssignDatadirsAndPorts(s.Source, ports)
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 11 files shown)


hub/initialize.go
```diff
@@ -117,32 +117,73 @@ func (s *Server) fillClusterConfigsSubStep(_ step.OutStreams, request *idl.Initi
...
 
 	ports = sanitize(ports)
-	if err := checkTargetPorts(source, ports); err != nil {
-		return PortAssignments{}, err
+
+	return userSpecifiedPorts(source, ports)
+}
+
+// can return an error if we run out of ports to use
+func userSpecifiedPorts(source *utils.Cluster, ports []int) (InitializeConfig, error) {
```
bhuvnesh2703: it does assign the data dir and port to the config, could the name be assignDataDirAndPorts? Similarly AssignPort implies port only..

hub/initialize.go
```diff
@@ -165,85 +206,50 @@ func sanitize(ports []int) []int {
...
-	for content, segment := range source.Primaries {
-		// Exclude the master for now. We want to give it its own reserved port,
-		// which does not overlap with the other segments, so we'll add it back
-		// later.
-		if content == -1 {
-			continue
-		}
-		segmentsByHost[segment.Hostname] = append(segmentsByHost[segment.Hostname], segment)
-	}
+func defaultTargetPorts(source *utils.Cluster) InitializeConfig {
```
bhuvnesh2703: it updates the data dir and ports.

hub/initialize.go
marksliva: I did some renames.
`userSpecifiedPorts` -> `assignDatadirsAndCustomPorts` 
`AssignPorts` -> `AssignDatadirsAndPorts`

hub/initialize.go
marksliva: Renamed `defaultTargetPorts` to `assignDatadirsAndDefaultPorts`

hub/initialize_test.go
```diff
@@ -14,22 +14,23 @@ func TestAssignPorts(t *testing.T) {
 
 		cluster  *utils.Cluster
 		ports    []int
-		expected PortAssignments
+		expected InitializeConfig
 	}{{
-		name:     "sorts and deduplicates provided port range",
```
jchampio: This test covered the sorting functionality of `sanitize()`; is there no longer a need for that?

hub/initialize.go
```diff
@@ -117,32 +117,73 @@ func (s *Server) fillClusterConfigsSubStep(_ step.OutStreams, request *idl.Initi
...
 
 	ports = sanitize(ports)
-	if err := checkTargetPorts(source, ports); err != nil {
-		return PortAssignments{}, err
+
+	return assignDatadirsAndCustomPorts(source, ports)
+}
+
+// can return an error if we run out of ports to use
+func assignDatadirsAndCustomPorts(source *utils.Cluster, ports []int) (InitializeConfig, error) {
```
jchampio: I think this implementation is calling out to be simplified, given the nested duplication and the similarities with `assignDatadirsAndDefaultPorts()`. Maybe with the mirrors addition?

hub/finalize.go
```diff
@@ -36,7 +36,7 @@ func (s *Server) Finalize(_ *idl.FinalizeRequest, stream idl.CliToHub_FinalizeSe
 			}
 
 			return UpgradeStandby(greenplumRunner, StandbyConfig{
-				Port:          s.TargetPorts.Standby,
+				Port:          s.TargetInitializeConfig.Standby.Port,
```
jchampio: nit: if I didn't know what `TargetInitializeConfig` was before reading this PR, I'm not sure I would have understood it from the name. Something to iterate on in the future, maybe.

hub/initialize_test.go
marksliva: I put it back. Good catch!

hub/initialize.go
marksliva: Sure. I will take a look in that branch and see if there's a good way to remove the duplication.

hub/finalize.go
marksliva: OK. I'll rename it in the finalize upgrade mirrors branch.

hub/initialize.go
marksliva: refactored it.

PR #236, opened 2020-03-02
finalize: add upgrade mirrors substep

marksliva:
When adding mirrors, we have to loop on a FTS probe scan until we detect that
all segments are synchronized before shutting down the cluster. Failing to do
so results in a hard to recover cluster.

We also had to manually set the port in the mirror recovery.conf files to point
to the correct port for the target cluster before starting it for the final
time.

The end-to-end test verifies that the new mirrors are valid by failing to them
and rebalancing.


hub/update_recovery_confs.go
```diff
@@ -0,0 +1,55 @@
...
+
+			mirrors := utils.FilterSegmentsOnHost(initializeConfig.Mirrors, conn.Hostname)
+			if len(mirrors) == 0 {
+				return
+			}
+
+			var confs []*idl.RecoveryConfInfo
+			for _, mirror := range mirrors {
+				confs = append(confs, &idl.RecoveryConfInfo{
+					TargetPrimaryPort:   int32(targetCluster.Primaries[mirror.ContentID].Port),
```
bhuvnesh2703: once the renaming data dirs get merged it, the target port will be same as source port (as we make them same after the catalog is updated) at this part of code which is called in finalize. So it will be an issue and we will have to fix it..

hub/finalize.go
```diff
@@ -68,6 +81,12 @@ func (s *Server) Finalize(_ *idl.FinalizeRequest, stream idl.CliToHub_FinalizeSe
 		return UpdateMasterPostgresqlConf(s.Source, s.Target)
 	})
 
+	if s.Source.HasMirrors() {
+		st.Run(idl.Substep_FINALIZE_UPDATE_RECOVERY_CONFS, func(streams step.OutStreams) error {
+			return UpdateRecoveryConfs(context.Background(), s.agentConns, s.Source, s.Target, s.TargetInitializeConfig)
```
bhuvnesh2703: at this time, source and target configs are same except bindir and version.

test/finalize-standby-and-mirrors.bats
```diff
@@ -0,0 +1,111 @@
...
+    check_can_start_transactions
+    gprecoverseg -a
+    check_segments_are_synchronized
+    gprecoverseg -ra
+    check_segments_are_synchronized
+}
+
+check_segments_are_synchronized() {
+    for i in {1..10}; do
+        run psql -t -A -d postgres -c "SELECT count(*) FROM gp_segment_configuration WHERE content <> -1 AND mode = 'n'"
```
bhuvnesh2703: synchronized is mode = 's' ?

hub/upgrade_mirrors.go
```diff
@@ -0,0 +1,137 @@
...
+
+		if time.Since(startTime) > timeout {
+			return xerrors.Errorf("%s timeout exceeded waiting for mirrors to come up", timeout)
+		}
+
+		time.Sleep(time.Second)
+	}
+}
+
+func UpgradeMirrors(stateDir string, masterPort int, conf *InitializeConfig, targetRunner GreenplumRunner) (err error) {
```
bhuvnesh2703: Why do we use Upgrade here? I know UpgradeStandby is there too. but actually its AddMirrors and AddStandby

test/finalize-standby-and-mirrors.bats
marksliva: Perhaps we should give this function a better name. But what we are trying to do is wait until all segments are synchronized. Once all segments are synchronized, filtering on mode='n' will return '0', and the function exits. If we never satisfy that condition within the timeout, then we return an error code 1.

hub/finalize.go
marksliva: I thought we only initialize the source and target clusters once. Am I missing something?

hub/finalize.go
```diff
@@ -26,23 +27,35 @@ func (s *Server) Finalize(_ *idl.FinalizeRequest, stream idl.CliToHub_FinalizeSe
...
+
+			return UpgradeStandby(targetRunner, StandbyConfig{
+				Port:          s.TargetInitializeConfig.Standby.Port,
 				Hostname:      s.Source.StandbyHostname(),
 				DataDirectory: s.Source.StandbyDataDirectory() + "_upgrade",
 			})
 		})
 	}
 
+	if s.Source.HasMirrors() {
```
bhuvnesh2703: should add mirrors be done here? 
Because, in the subsequent steps we will rename the directories of the segments and change ports, and if synchronization process is running at the background, that will be problematic.
Why should we not add mirrors just before the last step of starting the target cluster

hub/upgrade_mirrors.go
marksliva: That's a good point. Since mirrors has gotten split into a few more substeps, it might be easier to understand and follow the intent if we rename these to 'add'. Even for the standby, we're technically not done upgrading it until we do the port swap.

hub/finalize.go
marksliva: It should be fine to do it here. The next step shuts down all segments, and replication starts up at the very end of finalize in `Substep_FINALIZE_START_TARGET_CLUSTER` after all of the necessary reconfiguration has taken place.

hub/upgrade_mirrors.go
marksliva: We checked with product and design. They prefer to leave the `UpgradeMirrors` and `UpgradeStandby` terminology as is for now.

hub/finalize.go
marksliva: Keeping this open for now, in case https://github.com/greenplum-db/gpupgrade/pull/232 gets merged in before this.

hub/update_recovery_confs.go
marksliva: Keeping this open for now, in case https://github.com/greenplum-db/gpupgrade/pull/232 gets merged in before this.

agent/update_recovery_confs.go
```diff
@@ -0,0 +1,31 @@
...
+
+func (s *Server) UpdateRecoveryConfs(ctx context.Context, request *idl.UpdateRecoveryConfsRequest) (*idl.UpdateRecoveryConfsReply, error) {
+	var mErr multierror.Error
+	for _, recoveryConfInfo := range request.RecoveryConfInfos {
+		script := fmt.Sprintf("sed -i'.bak' 's/port=%d/port=%d/' %s",
+			recoveryConfInfo.GetTargetPrimaryPort(),
+			recoveryConfInfo.GetSourcePrimaryPort(),
+			filepath.Join(recoveryConfInfo.GetTargetMirrorDataDir(), "recovery.conf"))
+
+		err := sedCommand("bash", "-c", script).Run()
```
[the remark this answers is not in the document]
marksliva: I was having problems escaping the quotes around the replacement string. We run the sed command in the same way as `UpdateMasterPostgresqlConf`, so if we do figure out a way to not have to use `bash -c` here, then we should change it in both places.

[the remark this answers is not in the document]
jchampio: > We run the sed command in the same way as UpdateMasterPostgresqlConf, so if we do figure out a way to not have to use bash -c here, then we should change it in both places.

`UpdateMasterPostgresqlConf` is using Bash features like redirection and command chaining, though. This one isn't.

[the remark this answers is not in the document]
jchampio: Does something like
```go
sedCommand("sed", "-i.bak",
    fmt.Sprintf("s/port=%d/port=%d/", info.TargetPrimaryPort, info.SourcePrimaryPort),
    filepath.Join(info.TargetMirrorDataDir, "recovery.conf"),
)
```
not work well?

test/finalize-standby-and-mirrors.bats
```diff
@@ -0,0 +1,111 @@
...
+            where role='m' and status='u' and mode='s' and content != -1
+    "
+}
+
+get_standby_status() {
+    local standby_status=$1
+    echo "$standby_status" | grep 'Standby master state'
+}
+
+# Check the validity of the upgraded mirrors - failover to them and then recover, similar to cross-subnet testing
```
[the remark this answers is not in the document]
jchampio: On second thought this would make the end-to-end suite even more costly given our current strategy. We should drive that cost down, but it might make more sense in the short term to add mirror validation to our X->Y acceptance suite in the CI.

agent/update_recovery_confs.go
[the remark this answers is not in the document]
marksliva: It works. Removed the `bash -c`.

hub/rename_data_directories.go
```diff
@@ -68,28 +64,50 @@ func RenameSegmentDataDirs(agentConns []*Connection,
...
+		}
 
+		wg.Add(1)
 		go func(c *Connection) {
 			defer wg.Done()
 
-			segments, err := cluster.SegmentsOn(c.Hostname)
-			if err != nil {
-				errs <- err
+			segments := cluster.SelectSegments(notMaster)
```
kalensk: nit: renaming `notMaster` to `excludingMaster` or similar would have it read `SelectSegments(excludingMaster)` which is kinda nice.

hub/server.go
```diff
@@ -424,3 +425,21 @@ func LoadConfig(conf *Config, path string) error {
 
 	return nil
 }
+
+func AgentHosts(c *utils.Cluster) []string {
```
kalensk: nit: I want to say adding this on the cluster object would make the callers more readable to me. For example, `s.Source.AgentHosts()` is nice.

hub/server.go
```diff
@@ -424,3 +425,21 @@ func LoadConfig(conf *Config, path string) error {
...
+	uniqueHosts := make(map[string]bool)
+
+	notMaster := func(seg *utils.SegConfig) bool {
```
kalensk: nit: similar to other comment on possible rename to `excludingMaster` for readability.

hub/upgrade_mirrors.go
```diff
@@ -0,0 +1,137 @@
...
+}
+
+func runAddMirrors(r GreenplumRunner, filepath string) error {
+	return r.Run("gpaddmirrors",
+		"-a",
+		"-i", filepath,
+	)
+}
+
+func waitForFTS(db *sql.DB, timeout time.Duration) error {
```
dkrieger-pvtl: This function actually waits for either a timeout or all mirrors to be up and synchronized, which is more than its name `waitForFTS` implies.  Maybe call it `waitForMirrorSynchronization`?  Similarity, the constant `defaultFTSTimeout` should probably be called `defaultMirrorSyncTimeout`.  

Is the current value of 2 minutes anywhere close to reasonable on a larger user cluster?

agent/update_recovery_confs_test.go
```diff
@@ -0,0 +1,126 @@
...
+	server := agent.NewServer(agent.Config{
+		Port:     -1,
+		StateDir: "",
+	})
+
+	defer func() {
+		agent.SetSedCommand(exec.Command)
+	}()
+
+	t.Run("it replaces the temporary port with the source port", func(t *testing.T) {
```
dkrieger-pvtl: This test makes sure sed is called with the correct arguments, but doesn't actually call the system's sed itself.  So there is no "replacing" happening.   So a test name like `it calls sed to replace the temporary port with the source port` is more accurate.

cli/commanders/steps.go
```diff
@@ -36,10 +35,12 @@ var lines = map[idl.Substep]string{
 	idl.Substep_UPGRADE_PRIMARIES:                                 "Upgrading primary segments...",
 	idl.Substep_START_TARGET_CLUSTER:                              "Starting target cluster...",
 	idl.Substep_FINALIZE_UPGRADE_STANDBY:                          "Upgrading standby...",
-	idl.Substep_FINALIZE_SHUTDOWN_TARGET_CLUSTER:                  "Stopping new cluster",
+	idl.Substep_FINALIZE_UPGRADE_MIRRORS:                          "Upgrading mirrors...",
+	idl.Substep_FINALIZE_SHUTDOWN_TARGET_CLUSTER:                  "Stopping new cluster...",
 	idl.Substep_FINALIZE_UPDATE_TARGET_CATALOG_AND_CLUSTER_CONFIG: "Updating new master catalog...",
 	idl.Substep_FINALIZE_RENAME_DATA_DIRECTORIES:                  "Renaming data directories...",
 	idl.Substep_FINALIZE_UPDATE_TARGET_CONF_FILES:                 "Updating new master postgreql.conf and gpperfmon.conf files...",
+	idl.Substep_FINALIZE_UPDATE_RECOVERY_CONFS:                    "Updating recovery.conf files on mirrors...",
```
dkrieger-pvtl: nit: in `finalize`, this step comes after SHUTDOWN_TARGET_CLUSTERS but before UPDATE_TARGET_CATALOG_AND_CLUSTER_CONFIG so I'd prefer it be ordered this way in this list.

idl/cli_to_hub.proto
```diff
@@ -59,6 +59,8 @@ enum Substep {
     FINALIZE_RENAME_DATA_DIRECTORIES = 16;
     FINALIZE_UPDATE_TARGET_CONF_FILES = 17;
     FINALIZE_START_TARGET_CLUSTER = 18;
+    FINALIZE_UPDATE_RECOVERY_CONFS = 19;
```
dkrieger-pvtl: This [reordering nit](https://github.com/greenplum-db/gpupgrade/pull/236/files#r388994610) applies here too.

hub/check_disk_space.go
```diff
@@ -49,12 +49,18 @@ func checkDiskSpace(ctx context.Context, cluster *utils.Cluster, agents []*Conne
 		agent := agents[i]
 		wg.Add(1)
 
+		// We want to check disk space for the standby, primaries, and mirrors.
```
dkrieger-pvtl: nit: We want to check disk space on all hosts, including the master.  But we check the master above, and we're checking hosts, not segments(primary, standby, mirrors).  I think this comment is clearer:
`// Check disk space on all hosts containing a primary, standby or mirror`.

hub/check_disk_space_test.go
```diff
@@ -78,6 +78,7 @@ func TestCheckDiskSpace(t *testing.T) {
 			{ContentID: 0, Hostname: "sdw1", DataDir: "/data/primary", Role: "p", PreferredRole: "p"},
 			{ContentID: 1, Hostname: "sdw2", DataDir: "/data/primary", Role: "p", PreferredRole: "p"},
 			{ContentID: 2, Hostname: "sdw2", DataDir: "/data/primary2", Role: "p", PreferredRole: "p"},
+			{ContentID: 2, Hostname: "sdw3", DataDir: "/data/mirror2", Role: "m", PreferredRole: "m"},
```
dkrieger-pvtl: Please add a standby to this test for both the positive and negative cases.

jchampio:
Rebased onto master. Major changes include the movement of the config filling and the removal of `PreferredRole`.

hub/init_target_cluster.go
```diff
@@ -253,7 +261,7 @@ func CreateSegmentDataDirectories(agentConns []*Connection, cluster *utils.Clust
 				req.Datadirs = append(req.Datadirs, datadir)
 			}
 
-			_, err = c.AgentClient.CreateSegmentDataDirectories(context.Background(), req)
+			_, err := c.AgentClient.CreateSegmentDataDirectories(context.Background(), req)
```
dkrieger-pvtl: Now would be a good time to address `// TODO: Use a multierror to differentiate errors between hosts.`.

idl/cli_to_hub.proto
jchampio: I'm not inclined to change this. The order keeps changing as we learn more about the system; coupling this protobuf code to the business logic just makes diffs harder to read in practice.

(Less importantly, it's also training us to do the "wrong thing" when adding enum values to protobuf -- any normal project would need to add values only at the end to maintain compatibility with existing clients. But that doesn't apply to us since clients and servers are kept in lockstep.)

hub/server.go
jchampio: I don't think `Cluster` should be coupled in any way to the higher-level `Agents` concept.

hub/upgrade_mirrors.go
jchampio: > Is the current value of 2 minutes anywhere close to reasonable on a larger user cluster?

Keep in mind that, since there's no "traffic" on the primaries and we've just gone through a basebackup, synchronization should be nearly instantaneous. Obviously we still have to put that through a stress test.

hub/server.go
```diff
@@ -424,3 +425,21 @@ func LoadConfig(conf *Config, path string) error {
...
+		return !(seg.ContentID == -1 && seg.Role == "p")
+	}
+
+	for _, seg := range c.SelectSegments(notMaster) {
```
dkrieger-pvtl: nit: this seems more readable to me as:

```go
	for _, seg := range c.SelectSegments(func(seg *utils.SegConfig) bool {
		return !(seg.ContentID == -1 && seg.Role == "p")
	}) {
		uniqueHosts[seg.Hostname] = true
	}
```

cli/commanders/steps.go
dkrieger-pvtl: As per [this comment](https://github.com/greenplum-db/gpupgrade/pull/236/files#r389014888) that's fine.

hub/rename_data_directories.go
dkrieger-pvtl: nit: [inlining notMaster](https://github.com/greenplum-db/gpupgrade/pull/236/files#r3890223380) might make this more reabable.

hub/check_disk_space.go
```diff
@@ -49,12 +49,18 @@ func checkDiskSpace(ctx context.Context, cluster *utils.Cluster, agents []*Conne
...
+				!(seg.ContentID == -1 && seg.Role == "p")
+		}
+
 		go func() {
 			defer wg.Done()
 
-			segments, err := cluster.SegmentsOn(agent.Hostname)
-			if err != nil {
-				errs <- xerrors.Errorf("finding segments on host %s: %w", agent.Hostname, err)
+			segments := cluster.SelectSegments(notMaster)
```
dkrieger-pvtl: nit: [inlining notMaster](https://github.com/greenplum-db/gpupgrade/pull/236/files#r3890223380) might make this more reabable.

hub/server.go
jchampio: You prefer having the `range` split apart from the loop body? I'll be honest, I have the exact opposite response -- I can't immediately tell what this code is doing because my eyes don't tell me which pair of braces belongs to the loop.

hub/fill_cluster_configs_test.go
```diff
@@ -61,6 +61,27 @@ func TestAssignPorts(t *testing.T) {
 				{ContentID: 1, DbID: 3, Hostname: "sdw1", DataDir: "/data/dbfast2_upgrade/seg2", Role: "p", Port: 50434},
 				{ContentID: 2, DbID: 4, Hostname: "sdw1", DataDir: "/data/dbfast3_upgrade/seg3", Role: "p", Port: 50435},
 			}},
+	}, {
+		name: "when using default ports, it sets up mirrors as expected in the InitializeConfig",
```
dkrieger-pvtl: Please add a standby to this test.

jchampio: Standby's already covered; see the five other tests that changed for the datadir bugfix.

utils/cluster.go
```diff
@@ -142,23 +156,35 @@ func (u UnknownHostError) Is(err error) bool {
...
 			segments = append(segments, segment)
 		}
 	}
 
-	if len(segments) == 0 {
-		return nil, UnknownHostError{hostname}
+	return segments
+}
+
+// SelectSegments returns a list of all segments that match the given selector
```
berlin-ab: I like the introduction of this pattern a lot. 

It makes we wish we could go one step further in the future to allow for collection pipelines:

for example:
```cluster.SelectSegments(onHost).SelectSegments(onlyMirrors)```

The technique is described well here:
https://martinfowler.com/articles/collection-pipeline/

File: utils/sys_utils.go [hunk 22]
```diff
@@ -66,6 +68,7 @@ func InitializeSystemFunctions() *SystemFunctions {
 		WriteFile:    ioutil.WriteFile,
 		Create:       os.Create,
 		Mkdir:        os.Mkdir,
+		SqlOpen:      sql.Open,
```
berlin-ab: Thanks for adding this, it should make setting up Sql driver mocks nice and easy.

When I saw this change, I expected to see a corresponding test that observes its use.

test/finalize-standby-and-mirrors.bats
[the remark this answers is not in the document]
dkrieger-pvtl: As you point out, a end-to-end test(like we do for [cross-subnet mirrors for gpdb](https://github.com/greenplum-db/gpdb/blob/cdd1e9340ae3e3ff865de2b62d2f0dd475158559/gpMgmt/test/behave/mgmt_utils/steps/cross_subnet.py)) would be great, but this can be deferred for a future PR...perhaps add a // TODO here?

test/finalize-standby-and-mirrors.bats
dkrieger-pvtl: Perhaps the query you use in waitForFTS would be clearer;
```sql

				SELECT every(status = 'u' AND mode = 's')
					FROM gp_segment_configuration
					WHERE role = 'm'
			
```

dkrieger-pvtl:
You can consider combining `finalize.bats` and `finalize-standby-and-mirrors.bats`: they both run a full upggrade with standby/mirrors and then validate the state of the cluster.

hub/server.go
dkrieger-pvtl: I find it more readable, but that's a personal style...I prefer fewer lines of code.  No problem to leave it as is.

hub/fill_cluster_configs_test.go
dkrieger-pvtl: ok.

hub/check_disk_space_test.go
```diff
@@ -78,6 +78,7 @@ func TestCheckDiskSpace(t *testing.T) {
 			{ContentID: 0, Hostname: "sdw1", DataDir: "/data/primary", Role: "p"},
 			{ContentID: 1, Hostname: "sdw2", DataDir: "/data/primary", Role: "p"},
 			{ContentID: 2, Hostname: "sdw2", DataDir: "/data/primary2", Role: "p"},
+			{ContentID: 2, Hostname: "sdw3", DataDir: "/data/mirror2", Role: "m"},
```
kalensk: Might be useful to add a test for the standby. That is, `{ContentID: -1, Hostname: "standby", DataDir: "/data/standby", Role: "m"},`

hub/fill_cluster_configs.go
```diff
@@ -51,10 +51,20 @@ func (s *Server) FillClusterConfigsSubStep(config *Config, conn *sql.DB, _ step.
 
 func AssignDatadirsAndPorts(source *utils.Cluster, ports []int) (InitializeConfig, error) {
 	if len(ports) == 0 {
-		return assignDatadirsAndDefaultPorts(source), nil
-	}
+		port := 50432
+		// 2 -> one port for the master and standby
+		numberOfSegements := len(source.Mirrors) + len(source.Primaries) + 2
```
kalensk: super minor nit: inline after the `+ 2` adding `// including the master and standby`.

I stumbled a bit with the `2 -> one` verbiage.

hub/fill_cluster_configs.go
```diff
@@ -82,7 +92,7 @@ func assignDatadirsAndCustomPorts(source *utils.Cluster, ports []int) (Initializ
 			return InitializeConfig{}, errors.New("not enough ports")
 		}
 		standby.Port = ports[nextPortIndex]
-		standby.DataDir = upgradeDataDir(standby.DataDir)
+		standby.DataDir = standby.DataDir + "_upgrade"
```
kalensk: nit:  `hub/rename_data_directories.go` has the following const for use: `const UpgradeSuffix = "_upgrade"`. Although, making it obvious inline might also be more straightforward.

hub/fill_cluster_configs.go
```diff
@@ -115,6 +125,32 @@ func assignDatadirsAndCustomPorts(source *utils.Cluster, ports []int) (Initializ
...
+	for _, content := range source.ContentIDs {
+		// Skip the standby segment
+		if content == -1 {
+			continue
+		}
+
+		if segment, ok := source.Mirrors[content]; ok {
+			if portIndex, ok := portIndexByHost[segment.Hostname]; ok {
+				if portIndex > len(ports)-1 {
+					return InitializeConfig{}, errors.New("not enough ports")
```
kalensk: nit (or for another PR): `errors.New("not enough ports")` seems repeated a couple of times here. I wonder if it would be useful to refactor that out in a way?

hub/fill_cluster_configs.go
```diff
@@ -115,6 +125,32 @@ func assignDatadirsAndCustomPorts(source *utils.Cluster, ports []int) (Initializ
 		targetInitializeConfig.Primaries = append(targetInitializeConfig.Primaries, segment)
 	}
 
+	for _, content := range source.ContentIDs {
```
kalensk: Curious if breaking this into two high-level loops would help with readability/maintainability. That is, loop over source.Primaries and then source.Mirrors, and then maybe call to a helper function that skips if content = -1 since thats the standby or master. No need to change now just a potential thought.

jchampio: It sounded like the initial approach to simplification failed, so I'll keep it for this PR, but I agree that we should take another shot at minimizing this implementation.

hub/init_target_cluster.go
jchampio: Added a bug to track this.

hub/finalize.go
```diff
@@ -27,29 +27,51 @@ func (s *Server) Finalize(_ *idl.FinalizeRequest, stream idl.CliToHub_FinalizeSe
...
+		st.Run(idl.Substep_FINALIZE_UPGRADE_MIRRORS, func(streams step.OutStreams) error {
+			// XXX this probably indicates a bad abstraction
+			targetRunner.streams = streams
+
+			return UpgradeMirrors(s.StateDir, s.Target.MasterPort(), &s.TargetInitializeConfig, targetRunner)
+		})
+	}
+
```
kalensk: Not for this PR, but wondering if putting the _if_ statement for `HasMirrors`, inside `UpgradeMirrors` and have it return early if there are no mirrors would help with readability of `finalize.go`?

jchampio: That has a direct effect on the UX -- users will see an `upgrading mirrors` substep when in fact there is nothing to do.

hub/init_target_cluster.go
```diff
@@ -234,14 +234,22 @@ func CreateSegmentDataDirectories(agentConns []*Connection, cluster *utils.Clust
 	errChan := make(chan error, len(agentConns))
 
 	for _, conn := range agentConns {
-		wg.Add(1)
+		conn := conn
+
+		// Selects all primaries belonging to this agent's host.
```
kalensk: minor nit: For me, this comment isn't needed as it's not adding much value over the descriptive name and code.

hub/fill_cluster_configs.go
jchampio: See below.

hub/server.go
kalensk: Good point.

hub/check_disk_space_test.go
jchampio: I've added (once the fixup gets pushed) a positive case. The negative case doesn't seem to get the cost/benefit right -- there's no reason now or in the future for a different code path to be taken for a standby's errors, and if there were then I'd expect coverage to be added at that point.

hub/fill_cluster_configs_test.go
```diff
@@ -110,7 +131,7 @@ func TestAssignPorts(t *testing.T) {
 		ports: []int{1, 2, 3},
 		expected: InitializeConfig{
 			Master:    utils.SegConfig{ContentID: -1, DbID: 1, Hostname: "mdw", DataDir: "/data/qddir_upgrade/seg-1", Role: "p", Port: 1},
-			Standby:   utils.SegConfig{ContentID: -1, DbID: 2, Hostname: "mdw", DataDir: "/data/qddir_upgrade/seg-1", Role: "m", Port: 2},
+			Standby:   utils.SegConfig{ContentID: -1, DbID: 2, Hostname: "mdw", DataDir: "/data/qddir/seg-1_upgrade", Role: "m", Port: 2},
```
kalensk: This line and similar are super duper confusing to me. Can we update it to not be a hybrid of a standby and master. That is, ` utils.SegConfig{ContentID: -1, DbID: 2, Hostname: "standby", DataDir: "/data/standby_upgrade", Role: "m", Port: 2},`

hub/finalize.go
kalensk: Right, I kind of forgot about that. 

Hmmm, overall I guess there is not really an easy or elegant way around that with more effort than just leaving the if statement.

hub/finalize.go
```diff
@@ -27,29 +27,51 @@ func (s *Server) Finalize(_ *idl.FinalizeRequest, stream idl.CliToHub_FinalizeSe
...
 	if s.Source.HasStandby() {
 		st.Run(idl.Substep_FINALIZE_UPGRADE_STANDBY, func(streams step.OutStreams) error {
-			greenplumRunner := &greenplumRunner{
-				masterPort:          s.Target.MasterPort(),
-				masterDataDirectory: s.Target.MasterDataDir(),
-				binDir:              s.Target.BinDir,
-				streams:             streams,
-			}
+			// XXX this probably indicates a bad abstraction
+			targetRunner.streams = streams
```
berlin-ab: A constructor for the GreenplumRunner might be in order, and it could include streams as an optional argument.

jchampio: Possibly, but I'm wondering if the bad abstraction here is how we're passing `OutStreams` to each substep...

I think I need to see it get a little uglier first.

berlin-ab: It could even be a  required argument:

```
targetRunner := NewGreenplumRunner(s.Target, streams)
```

That would reduce the duplication in a different way than a shared variable.

hub/rename_data_directories.go
jchampio: Changed.

hub/fill_cluster_configs.go
jchampio: Changed.

hub/finalize.go
jchampio: Yeah, that might be the way to go.

hub/finalize.go
berlin-ab: Possibly: 
```
st.MaybeRun(idl.Substep_FINALIZE_UPGRADE_MIRRORS, someConditionalFunc, func(streams step.OutStreams) {
    // do stuff
})
```

test/finalize.bats
```diff
@@ -5,72 +5,75 @@ load helpers
...
+        fi
+
+        [ -d "${source_datadir}/" ] || fail "expected source data directory to be located at $source_datadir"
+        [ -f "${source_datadir}/${marker_file}" ] || fail "expected ${marker_file} marker file to be in source datadir: $source_datadir"
+        [ -f "${source_datadir}/postgresql.conf" ] || fail "expected postgresql.conf file to be in $source_datadir"
+
+        # ensure the target cluster is located where the source used to be
+        [ -d "${datadir}/" ] || fail "expected target data directory to be located at $datadir"
+        [ -f "${datadir}/postgresql.conf" ] || fail "expected postgresql.conf file to be in $datadir"
+        [ ! -f "${datadir}/${marker_file}" ] || fail "unexpected ${marker_file} marker file in target datadir: $datadir"
```
kalensk: nit: Might be useful to rename `datadir` to `target_datadir`. This then corresponds to the `source_datadir` and perhaps makes things explicit. Then the comment might not be needed.

hub/finalize.go
berlin-ab: The Step owns the streams, if the streams were given to the step as an argument, then there would be no need to thread it through to each function, we'd get it through closure.

hub/finalize.go
jchampio: Hiding conditionals in layers doesn't necessarily lead to elegance or maintainability in my book.

Go has a really elegant and straightforward way for conditionally choosing not to run code -- it's the `if` statement. We don't have to choose between the two extremes of pyramids-of-doom and hiding-all-`if`s-everywhere. `if` is okay.

hub/rename_data_directories.go
```diff
@@ -68,28 +64,50 @@ func RenameSegmentDataDirs(agentConns []*Connection,
...
+				errs <- utils.UnknownHostError{conn.Hostname}
 				return
 			}
 
 			// When there are multiple segments under a parent data directory
 			// only call rename once on the parent directory.
 			// For example, /data/primary/gpseg1 and /data/primary/gpseg2
 			// only call rename once for /data/primary.
-			parentDirs := make(map[string]bool)
+			// NOTE: we keep the iteration stable for testing purposes; hence
```
berlin-ab: Would it be possible to change the tests in a way that does not require multiple data structures?

utils/cluster.go
```diff
@@ -142,23 +156,35 @@ func (u UnknownHostError) Is(err error) bool {
 	return err == ErrUnknownHost
 }
 
-// SegmentsOn returns the configurations of segments that are running on a given
-// host excluding the master. An error of type ErrUnknownHost will be returned
-// for unknown hostnames.
-func (c Cluster) SegmentsOn(hostname string) ([]SegConfig, error) {
+func FilterSegmentsOnHost(segmentsToFilter []SegConfig, hostname string) []SegConfig {
```
kalensk: nit: `FilterSegmentsOnHost` seems to only be used in one place. Could this be removed and the caller follow the `SelectSegments` idiom being used. That is, maybe something like:

```go
mirrors := func(seg *utils.SegConfig) bool {
			return seg.Hostname == conn.Hostname &&
				(seg.Role == utils.MirrorRole && seg.ContentID != -1)
		}

segments := cluster.SelectSegments(mirrors)
```

Or update all the other callers to use this helper instead? I am not sure I see the need for both functions yet?

hub/rename_data_directories.go
jchampio: We tried briefly but it would require what seems to be a fairly complicated sqlmock matcher. Maybe there's a simplification that would let us get rid of this distinction, but we didn't find one quickly.

hub/check_disk_space_test.go
jchampio: Done.

hub/upgrade_mirrors.go
```diff
@@ -0,0 +1,137 @@
...
+		}
+	}()
+
+	err = writeGpAddmirrorsConfig(conf, f)
+	if err != nil {
+		return err
+	}
+
+	err = f.Close()
+	fileClosed = true
```
kalensk: Dumb question maybe, but why not just close the file _only_ in the defer only and not here which would eliminate the need for the bool?

jchampio: We can't run addmirrors until we write the file to disk.

hub/finalize.go
berlin-ab: I don't think there's anything wrong with modeling our domain, and the domain of substeps does include some steps that optionally run. Many languages provide `if`, and yet domain modeling is a popular technique for describing the rules of a problem. That said, the above was just _a way_ of solving that problem.

jchampio: > That said, the above was just _a way_ of solving that problem.

Sure, understood.

jchampio:
> You can consider combining finalize.bats and finalize-standby-and-mirrors.bats: they both run a full upggrade with standby/mirrors and then validate the state of the cluster.

@dkrieger-pvtl Added a chore.

hub/fill_cluster_configs_test.go
jchampio: Done.

hub/fill_cluster_configs.go
jchampio: > Although, making it obvious inline might also be more straightforward.

+1

jchampio:
Running through the comments, I believe I've addressed all blockers, so I'll push. Let me know if I've missed anything (GitHub's UI leaves much to be desired) and I'll be happy to fix it.

PR #237, opened 2020-03-03
Update initialize output

bhuvnesh2703:
The initialize output now contains the below
Initialization in progress.
....
...
...
Running pg_upgrade checks...                                       [COMPLETE]   

NEXT ACTIONS
------------
Run "gpupgrade execute" to proceed with the upgrade.

After executing, you will need to finalize.

### DIFF
--- a/cli/commands/commands.go
+++ b/cli/commands/commands.go
@@ -312,12 +312,14 @@ func initialize() *cobra.Command {
 			}
 
 			fmt.Println(`
-Run "gpupgrade execute" on the command line to proceed with the upgrade.
+Initialize completed successfully.
 
-After upgrading, you will need to finalize.
+NEXT ACTIONS
+------------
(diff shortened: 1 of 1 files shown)


kalensk:
Looks good.

(Not necessarily for this PR, but for the story looks like `Initialize completed successfully.` needs to be printed after initialize succeeds and above `Next Actions`)

bhuvnesh2703:
> Looks good.
> 
> (Not necessarily for this PR, but for the story it looks like `Initialize completed successfully.` needs to be printed after initialize succeeds and above `Next Actions`)

Yep right,

Added that and merging.
Now the output.

```
Initialize completed successfully.

NEXT ACTIONS
------------
Run "gpupgrade execute" to proceed with the upgrade.

After executing, you will need to finalize.
```

PR #238, opened 2020-03-03
Update execute output

bhuvnesh2703:
The expected output should show the port and master data directory for
the target cluster in the message, thus moved loadConfig to hub. This
allows reading the config.json from the state directory and displaying
the required information.

### DIFF
--- a/cli/commanders/steps.go
+++ b/cli/commanders/steps.go
@@ -5,6 +5,11 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"path/filepath"
+	"strconv"
+
+	"github.com/greenplum-db/gpupgrade/hub"
+	"github.com/greenplum-db/gpupgrade/utils"
 
 	"github.com/greenplum-db/gp-common-go-libs/gplog"
 	"github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


cli/commanders/steps.go
```diff
@@ -92,18 +95,38 @@ func Execute(client idl.CliToHubClient, verbose bool) error {
...
-validation desired prior to finalizing your upgrade.
+	stateDir := utils.GetStateDir()
+	finfo, err := os.Stat(stateDir)
+	if os.IsNotExist(err) {
+		return fmt.Errorf("gpupgrade state dir (%s) does not exist. Did you run gpupgrade initialize?", stateDir)
+	} else if err != nil {
+		return err
+	} else if !finfo.IsDir() {
+		return fmt.Errorf("gpupgrade state dir (%s) does not exist as a directory.", stateDir)
+	}
```
kalensk: This seems like it could be more "clean" and "elegant". When I have a more time I'll have to try a few things out, and see.

I like the concept outlined in [comment 1](https://github.com/golang/go/issues/8456). I wonder if we could apply it to this code in some way.
> The proper way to check if a file exists is to not check if a file exists as that
information is unavoidably racy.
os.Stat is not about file existence but about file metadata. Even though the error
possibly returned may mean the file didn't existed when os.Stat was executing, it does
not mean the file does not exists now, when os.Stat completed.
Use os.O_EXCL when creating a file which must not exists already and the err from
os.Open when a file must exist before opening.

bhuvnesh2703: @kalensk how about this, try to read the file with os.O_RDWR|os.O_EXCL
```
	path := filepath.Join(stateDir, hub.ConfigFileName)
	_, err = os.OpenFile(path, os.O_RDWR|os.O_EXCL, 600)
	if os.IsNotExist(err) {
		return xerrors.Errorf("configuration file %s does not exist, %#v", path, err)
	} else if err != nil {
		return xerrors.Errorf("unable to read configuration file %s, %#v", path, err)
	}
	conf := &hub.Config{}
	err = hub.LoadConfig(conf, path)
	if err != nil {
		return err
	}
```

PR #242, opened 2020-03-06
Add support gpupgrade --version and -v

kalensk:
gpupgrade now supports the following options for getting the version string:
- gpupgrade version
- gpupgrade --version
- gpupgrade -V

[Test pipeline](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:version-flag)


cli/commands/commands.go
```diff
@@ -220,13 +233,19 @@ func createConfigShowSubcommand() *cobra.Command {
...
+		Short: "Version of gpupgrade",
+		Long:  `Version of gpupgrade`,
+		Run: func(cmd *cobra.Command, args []string) {
+			fmt.Println(VersionString("gpupgrade"))
+		},
+	}
+	rootCmd.AddCommand(version)
+
+	// add version flag to rootCmd
+	shouldPrintVersion = rootCmd.Flags().BoolP("version", "V", false, "print version")
```
jchampio: This is line coupled across layers (it has no relevance to the code above it because it's linked to the `root` command, not the `version` command). Switching to [the style we use elsewhere](https://github.com/greenplum-db/gpupgrade/blob/336ebd8/cli/commands/commands.go#L344-L358) would help localize the related code and get rid of the global. Then `addVersion()` can just be responsible for adding the subcommand.

cli/commands/commands.go
```diff
@@ -48,12 +48,23 @@ import (
...
+var shouldPrintVersion *bool
 
+func BuildRootCommand() *cobra.Command {
 	// TODO: if called without a subcommand, the cli prints a help message with timestamp.  Remove the timestamp.
-	root := &cobra.Command{Use: "gpupgrade"}
+	root := &cobra.Command{
+		Use: "gpupgrade",
+		Run: func(cmd *cobra.Command, args []string) {
+			if *shouldPrintVersion {
+				fmt.Println(VersionString("gpupgrade"))
```
jchampio: I will go a step further than Adam and say that the two implementations need to be deduplicated now, since future implementation drift would be a really confusing bug. Maybe just a simple `printVersion()` helper?

kalensk:
Thanks for review. Pushed a fixup that should hopefully addresses them.

PR #243, opened 2020-03-06
Refactor start stop

berlin-ab:
Refactor out complicated start/stop logic.

Please review commit by commit.

### DIFF
--- a/hub/execute.go
+++ b/hub/execute.go
@@ -34,7 +34,13 @@ func (s *Server) Execute(request *idl.ExecuteRequest, stream idl.CliToHub_Execut
 	}()
 
 	st.Run(idl.Substep_SHUTDOWN_SOURCE_CLUSTER, func(streams step.OutStreams) error {
-		return StopCluster(streams, s.Source, true)
+		err := StopCluster(streams, s.Source)
+
+		if err != nil {
+			return errors.Wrap(err, fmt.Sprintf("failed to stop source cluster"))
+		}
+
+		return nil
 	})
 
 	st.Run(idl.Substep_UPGRADE_MASTER, func(streams step.OutStreams) error {
@@ -78,7 +84,14 @@ func (s *Server) Execute(request *idl.ExecuteRequest, stream idl.CliToHub_Execut
 	})
 
 	st.Run(idl.Substep_START_TARGET_CLUSTER, func(streams step.OutStreams) error {
-		return StartCluster(streams, s.Target, false)
+		err := StartCluster(streams, s.Target)
+
+		if err != nil {
+			return errors.Wrap(err, fmt.Sprintf("failed to start target cluster"))
+		}
+
+		return nil
+
 	})
 
 	return st.Err()
--- a/hub/execute.go
+++ b/hub/execute.go
@@ -5,6 +5,7 @@ import (
 	"path/filepath"
 
 	"github.com/pkg/errors"
+	"golang.org/x/xerrors"
 
 	"github.com/greenplum-db/gp-common-go-libs/gplog"
 	"github.com/hashicorp/go-multierror"
@@ -37,7 +38,7 @@ func (s *Server) Execute(request *idl.ExecuteRequest, stream idl.CliToHub_Execut
 		err := StopCluster(streams, s.Source)
 
 		if err != nil {
-			return errors.Wrap(err, fmt.Sprintf("failed to stop source cluster"))
+			return xerrors.Errorf("failed to stop source cluster: %w", err)
 		}
 
 		return nil
@@ -87,7 +88,7 @@ func (s *Server) Execute(request *idl.ExecuteRequest, stream idl.CliToHub_Execut
 		err := StartCluster(streams, s.Target)
 
 		if err != nil {
-			return errors.Wrap(err, fmt.Sprintf("failed to start target cluster"))
+			return xerrors.Errorf("failed to start target cluster: %w", err)
 		}
 
 		return nil
--- a/hub/finalize.go
+++ b/hub/finalize.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 
 	"github.com/hashicorp/go-multierror"
+	"github.com/pkg/errors"
 
 	"github.com/greenplum-db/gp-common-go-libs/gplog"
 
@@ -47,7 +48,13 @@ func (s *Server) Finalize(_ *idl.FinalizeRequest, stream idl.CliToHub_FinalizeSe
 	}
 
 	st.Run(idl.Substep_FINALIZE_SHUTDOWN_TARGET_CLUSTER, func(streams step.OutStreams) error {
-		return StopCluster(streams, s.Target, false)
+		err := StopCluster(streams, s.Target)
+
+		if err != nil {
+			return errors.Wrap(err, fmt.Sprintf("failed to stop target cluster"))
+		}
+
+		return nil
 	})
 
 	st.Run(idl.Substep_FINALIZE_UPDATE_TARGET_CATALOG_AND_CLUSTER_CONFIG, func(streams step.OutStreams) error {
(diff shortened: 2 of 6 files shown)


hub/start_or_stop_cluster_test.go
```diff
@@ -76,7 +76,7 @@ func TestStartOrStopCluster(t *testing.T) {
 			func(path string, args ...string) {
 				g.Expect(path).To(Equal("bash"))
 				g.Expect(args).To(Equal([]string{"-c", "source /source/bindir/../greenplum_path.sh " +
-					"&& /source/bindir/gpstop  -a -d basedir/seg-1"}))
+					"&& /source/bindir/gpstop -a -d basedir/seg-1"}))
```
jchampio: nit: It's weird to see a test change during a refactor commit.

hub/execute.go
```diff
@@ -34,7 +34,13 @@ func (s *Server) Execute(request *idl.ExecuteRequest, stream idl.CliToHub_Execut
 	}()
 
 	st.Run(idl.Substep_SHUTDOWN_SOURCE_CLUSTER, func(streams step.OutStreams) error {
-		return StopCluster(streams, s.Source, true)
+		err := StopCluster(streams, s.Source)
+
+		if err != nil {
+			return errors.Wrap(err, fmt.Sprintf("failed to stop source cluster"))
```
jchampio: We're moving away from `errors.Wrap()` towards `xerrors.Errorf("...: %w")` in all code, because the standard library in Go 1.13+ supports native error wrapping. `xerrors.Errorf()` will become `fmt.Errorf()` once we update.

hub/start_or_stop_cluster.go
```diff
@@ -25,71 +23,53 @@ func IsPostmasterRunning(stream step.OutStreams, cluster *utils.Cluster) error {
...
+}
 
-	cmd := startStopCmd("bash", "-c",
-		fmt.Sprintf("source %[1]s/../greenplum_path.sh && %[1]s/%[2]s %[3]s -a -d %[4]s",
-			cluster.BinDir,
-			gpdbCmd,
-			masterOnlyFlag,
-			cluster.MasterDataDir()))
+func runStartStopCmd(stream step.OutStreams, cluster *utils.Cluster, command string) error {
+	commandWithEnv := fmt.Sprintf("source %[1]s/../greenplum_path.sh && %[1]s/%[2]s",
```
marksliva: Since this is a refactor PR, could we also replace this section with the `GreenplumRunner`?

hub/start_or_stop_cluster.go
```diff
@@ -25,71 +23,53 @@ func IsPostmasterRunning(stream step.OutStreams, cluster *utils.Cluster) error {
...
+		cluster.BinDir,
+		command)
 
+	cmd := startStopCmd("bash", "-c", commandWithEnv)
 	cmd.Stdout = stream.Stdout()
 	cmd.Stderr = stream.Stderr()
-
-	err := cmd.Run()
-	if err != nil {
```
marksliva: Nice. Pushing errors up a level cleaned this up, and also reduced the number of methods and variables we had to deal with.

hub/execute.go
berlin-ab: This was a straight refactor and was not introduced here.

hub/start_or_stop_cluster_test.go
berlin-ab: We were placing an empty string in here during the original implementation. I can put the blank back in if you'd like.

hub/execute.go
jchampio: (This is my only blocker, BTW. Overall great change.)

hub/start_or_stop_cluster.go
berlin-ab: I don't think I can get away with refactoring toward that without also changing the tests significantly. Otherwise, I'm with you. Maybe that can be the next commit along these lines. cc @jchampio

hub/execute.go
jchampio: > This was a straight refactor and was not introduced here.

I see. In that case, the change from `Errorf()` to `Wrap()` over in `hub/update_catalog.go` needs to be reverted. I think that's the only regression.

hub/start_or_stop_cluster_test.go
jchampio: Nah, I'm fine. (A note in the commit message would be nice in the future though, just to explain cases where functional changes have crept in.)

hub/update_catalog.go
```diff
@@ -17,9 +17,9 @@ import (
 // TODO: When in copy mode should we update the catalog and in-memory object of
 //  the source cluster?
 func (s *Server) UpdateCatalogAndClusterConfig(streams step.OutStreams) (err error) {
-	err = StartMasterOnly(streams, s.Target, false)
+	err = StartMasterOnly(streams, s.Target)
 	if err != nil {
-		return xerrors.Errorf("starting new master: %w", err)
+		return errors.Wrap(err, fmt.Sprintf("failed to start target master"))
```
jchampio: Here...

hub/update_catalog.go
```diff
@@ -37,9 +37,9 @@ func (s *Server) UpdateCatalogAndClusterConfig(streams step.OutStreams) (err err
 	segs := map[int]utils.SegConfig{-1: master}
 	oldTarget := &utils.Cluster{Primaries: segs, BinDir: s.Target.BinDir}
 
-	err = StopMasterOnly(streams, oldTarget, false)
+	err = StopMasterOnly(streams, oldTarget)
 	if err != nil {
-		return xerrors.Errorf("stopping new master: %w", err)
+		return errors.Wrap(err, fmt.Sprintf("failed to stop target master"))
```
jchampio: ...and here are what I'm referring to.

hub/finalize.go
```diff
@@ -63,7 +70,13 @@ func (s *Server) Finalize(_ *idl.FinalizeRequest, stream idl.CliToHub_FinalizeSe
 	})
 
 	st.Run(idl.Substep_FINALIZE_START_TARGET_CLUSTER, func(streams step.OutStreams) error {
-		return StartCluster(streams, s.Target, false)
+		err := StartCluster(streams, s.Target)
+
+		if err != nil {
+			return errors.Wrap(err, fmt.Sprintf("failed to start target cluster"))
```
marksliva: We could make this error message even more specific, something like `failed to start upgraded target cluster`.

hub/execute.go
berlin-ab: That one is just a strange looking diff. The original error message was a `errors.Wrap()` and I didn't see any reason to do `xerrors.Errorf("message: %w", errors.Wrap(err, "message"))`

That said, I pushed a change to removes all of the errors.Wrap usages for the start/stop functions.

hub/finalize.go
berlin-ab: Let's make that an exercise for the next change made here. My goal with this refactoring was to not change the externally visible behavior - even if the `xerrors.Errorf` does in fact do that.

hub/update_catalog.go
```diff
@@ -37,9 +37,9 @@ func (s *Server) UpdateCatalogAndClusterConfig(streams step.OutStreams) (err err
 	segs := map[int]utils.SegConfig{-1: master}
 	oldTarget := &utils.Cluster{Primaries: segs, BinDir: s.Target.BinDir}
 
-	err = StopMasterOnly(streams, oldTarget, false)
+	err = StopMasterOnly(streams, oldTarget)
 	if err != nil {
-		return xerrors.Errorf("stopping new master: %w", err)
+		return xerrors.Errorf("failed to stop target master: %w", err)
```
jchampio: Just as a note, this is getting further away from the desired error message structure. But I understand why it's being done in this instance and we can fix it up later.

PR #244, opened 2020-03-06
Enforce xerrors.Errorf as a preference.

berlin-ab:
Starting with the hub package, let's start to enforce our preference for `xerrors.Errorf` instead of `errors.Wrap`.

This was motivated by a PR comment about the correct API to use. So, if we can automate that portion of a PR review and catch a mistake earlier, let's automate it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,8 +40,10 @@ format:
 
 # check runs all tests against the locally built gpupgrade binaries. Use -k to
 # continue after failures.
-check: check-ginkgo check-bats
+check: check-code-style check-ginkgo check-bats
 check-ginkgo check-bats: export PATH := $(CURDIR):$(PATH)
+check-code-style:
+	bats -t scripts/enforce-code-style.bats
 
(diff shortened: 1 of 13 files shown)


jchampio:
I'm lukewarm on this approach. Maybe you can find another dev who agrees?

1. Code style enforcement is the job of a linter, and as we've effectively decided not to adopt a linter yet, I'd prefer we not add ad-hoc linter scripts to the codebase right this moment.
2. We'd talked (a long time ago, before you joined up) about cleaning up the error message style while migrating to `xerrors.Errorf()`. This effectively removes the implicit TODOs without resolving that, so although we're closer to the end goal I'm not convinced the overall effect is an improvement.

berlin-ab:
> 1. Code style enforcement is the job of a linter

I agree, but in my initial search through the Go toolset I didn't see any tools that dealt with a team-dependent deprecated API. I assumed we'd go with this approach until we find a better one. In any case we need an automated way to provide obvious PR feedback to the author earlier.

berlin-ab:
I guess we could write a custom analyzer for `go vet`.

PR #245, opened 2020-03-07
Migrate agent/server_test.go to Go style test

kalensk:
To keep things simple, I kept the semantics the same for now rather than thinking if there was a better way of testing or pulling things apart.

[Test pipeline](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:agent_server_test)

### DIFF
--- a/agent/server_test.go
+++ b/agent/server_test.go
@@ -1,73 +1,100 @@
 package agent_test
 
 import (
-	"io/ioutil"
 	"os"
+	"path"
+	"testing"
+	"time"
 
 	"github.com/greenplum-db/gp-common-go-libs/testhelper"
+	"golang.org/x/xerrors"
 
 	"github.com/greenplum-db/gpupgrade/agent"
 	"github.com/greenplum-db/gpupgrade/testutils"
-	"github.com/greenplum-db/gpupgrade/utils"
-
-	. "github.com/onsi/ginkgo"
-	. "github.com/onsi/gomega"
 )
 
-var _ = Describe("Server", func() {
-	var (
-		dir       string
-		agentConf agent.Config
-		exists    func() bool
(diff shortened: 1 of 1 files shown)


agent/server_test.go
```diff
@@ -1,73 +1,97 @@
...
-	)
+func TestServerStart(t *testing.T) {
+	testhelper.SetupTestLogger()
 
-	BeforeEach(func() {
-		testhelper.SetupTestLogger()
-		dir, err := ioutil.TempDir("", "")
-		Expect(err).ToNot(HaveOccurred())
+	t.Run("successfully starts and creates state directory if it does not exist", func(t *testing.T) {
+		stateDir := path.Join(getTempDir(t), ".gpupgrade")
```
dkrieger-pvtl: Please change this to this to clean up the tmpDir....
```go
		tmpDir := getTempDir(t)
		defer os.RemoveAll(tmpDir)

		stateDir := path.Join(tmpDir, ".gpupgrade")
```

agent/server_test.go
```diff
@@ -1,73 +1,97 @@
...
+		if !exists {
+			t.Error("expected stateDir to be created")
 		}
 	})
 
-	AfterEach(func() {
-		utils.System = utils.InitializeSystemFunctions()
-	})
+	t.Run("successfully starts if state directory already exists", func(t *testing.T) {
+		stateDir := getTempDir(t)
```
dkrieger-pvtl: Please add this after this line to clean up:
```go
		defer os.RemoveAll(stateDir)
```

kalensk:
> other than needing to delete the tmpDir. Please add that back in(it was in the original test).

Oops! Good catch.

PR #246, opened 2020-03-09
Cover configuration file sed editing with tests.

berlin-ab:


hub/replace_string_within_file.go
```diff
@@ -0,0 +1,40 @@
...
+//
+// for example:
+//
+// sed s/pattern/replacement/ /path/to/some/file.txt
+//
+func ReplaceStringWithinFile(pattern string, replacement string, file string) error {
+	return replaceStringWithinFileFunc(pattern, replacement, file)
+}
+
+func replaceStringWithinFile(pattern string, replacement string, pathToFile string) error {
```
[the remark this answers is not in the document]
berlin-ab: Thanks. I'll take a look.

hub/update_conf_files_test.go
```diff
@@ -0,0 +1,125 @@
...
+		var usedFile string
+
+		hub.SetReplaceStringWithinFile(func(pattern string, replacement string, file string) error {
+			usedPattern = pattern
+			usedReplacement = replacement
+			usedFile = file
+			return nil
+		})
+
+		sourceMaster := source.Primaries[-1]
```
marksliva: nit: there is a downside to this approach in that depending on the shared test setup prevents us from running this single test in isolation. We should revisit this in the future since other tests also use setup from `hub/services_suite_test.go` and if we get rid of ginkgo, we would lose the `BeforeSuite` and `BeforeEach` concepts at a global level anyway.
```
go test ./hub -run TestUpdatePostgresqlConf -count 1
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x466aa2f]

go test ./hub -count 1
ok  	github.com/greenplum-db/gpupgrade/hub	1.490s
```

berlin-ab: Thanks for the heads-up. I removed the dependency on the global source and target variables.

hub/replace_string_within_file.go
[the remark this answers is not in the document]
berlin-ab: I'll drop in a story to track this.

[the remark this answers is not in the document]
marksliva: Sounds good.

hub/replace_string_within_file_test.go
```diff
@@ -0,0 +1,52 @@
...
+	"os/exec"
+	"testing"
+
+	"github.com/greenplum-db/gp-common-go-libs/testhelper"
+
+	"github.com/greenplum-db/gpupgrade/hub"
+)
+
+func TestReplaceStringWithinFile(t *testing.T) {
+	t.Run("it can replace a string within a file", func(t *testing.T) {
```
marksliva: This is an integration test. We could move it to `integrations/`, leave it as is and add a comment, or rewrite it to not be an integration, but I'd be curious to see what other people think.

berlin-ab: Yes, it integrates against the filesystem, similar to how `TestRsync` works.

PR #247, opened 2020-03-09
Expose streams from step

berlin-ab:
Many substep functions do not use the stream they have been given, signaling a possible design change.

This exposes the streams as a function on Step, the owner of the streams.

### DIFF
--- a/hub/execute.go
+++ b/hub/execute.go
@@ -34,8 +34,8 @@ func (s *Server) Execute(request *idl.ExecuteRequest, stream idl.CliToHub_Execut
 		}
 	}()
 
-	st.Run(idl.Substep_SHUTDOWN_SOURCE_CLUSTER, func(streams step.OutStreams) error {
-		err := StopCluster(streams, s.Source)
+	st.Run(idl.Substep_SHUTDOWN_SOURCE_CLUSTER, func() error {
+		err := StopCluster(st.Streams(), s.Source)
 
 		if err != nil {
 			return xerrors.Errorf("failed to stop source cluster: %w", err)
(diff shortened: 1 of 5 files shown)


jchampio:
This is an interesting idea. Some further brainstorming on this based on some of the "debug-mode" work I'm doing:

I think I'd like to eventually see the step runner pulled into an interface:
```go
type substepRunner interface {
    // Run executes a substep with the given code by invoking the provided
    // implementation callback.
    Run(idl.Substep, func(step.OutStreams) error)
}
```

Your work suggests that we might be able to legitimately attach `Streams()` to that as well, and I think I like that because it allows us an additional transformation:

```go
type substepRunner interface {
    // Run executes a substep with the given code by invoking the provided
    // implementation callback.
    Run(idl.Substep, func(substepRunner) error) // <- *

    // Streams provides the output/error streams for substep communication.
    Streams() step.OutStreams
}
```

By switching our callback to take `substepRunner` itself, we allow arbitrary nesting of substeps a la `testing.T.Run()`, which other people (e.g. @dkrieger-pvtl) have floated in the past and I think I'm slowly coming around to:

```go
st.Run(idl.UPGRADE_MIRRORS_AND_STANDBY, func(st substepRunner) error {
    st.Run(idl.STOP_CLUSTER, ...)
    st.Run(idl.ADD_STANDBY, ...)
    st.Run(idl.ADD_MIRRORS, ...)
    st.Run(idl.UPDATE_CATALOG, ...)
    st.Run(idl.START_CLUSTER, ...)
    return nil
})
```

That might be a more intuitive way to express which substeps are relevant to the UX (the top level) and which ones exist merely to give us an idempotence restart point (the lower levels).

PR #248, opened 2020-03-10
Rename user facing instances of old and new to source and target.

berlin-ab:
Did a search and replace for "old " and "new " within the codebase and renamed to "source " and "target ".

I didn't find any instances of "current " used anywhere.

### DIFF
--- a/ci/scripts/upgrade-cluster.bash
+++ b/ci/scripts/upgrade-cluster.bash
@@ -79,7 +79,7 @@ for host in "${hosts[@]}"; do
     ssh centos@$host "sudo mv /tmp/gpupgrade /usr/local/bin"
 done
 
-echo 'Loading SQL dump into old cluster...'
+echo 'Loading SQL dump into source cluster...'
 time ssh mdw bash <<EOF
     set -eux -o pipefail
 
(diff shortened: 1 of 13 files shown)


kalensk:
Looks good.

Maybe the message in `hub/upgrade_primaries.go` could also be updated?
`"hostnames do not match between old and new cluster with content ID %d. Found old cluster hostname: '%s', and new cluster hostname: '%s'"`

Not for this PR, but wonder if the files in the ci directory should eventually be updated as well.

kalensk:
Not really needed, just pointing out that there are some potential references in comments in .bats files.

```
test/finalize.bats
74:    # Check to make sure the new cluster matches the old one.

test/gpinitsystem.bats
78:    # Store the data directories for the new cluster.
87:    # Ensure the new cluster has the expected ports and compare the directories

ci/scripts/upgrade-cluster.bash
107:# Dump the new cluster and compare.

ci/scripts/upgrade-cluster.bash
91:# Dump the old cluster for later comparison.
```

berlin-ab:
> Maybe the message in hub/upgrade_primaries.go could also be updated?

Thanks. I missed that one the first time around.

PR #250, opened 2020-03-12
Initialize greenplum runner with streams.

berlin-ab:
Populate the greenplum runner with the step's streams when initialized.

### DIFF
--- a/hub/finalize.go
+++ b/hub/finalize.go
@@ -33,13 +33,11 @@ func (s *Server) Finalize(_ *idl.FinalizeRequest, stream idl.CliToHub_FinalizeSe
 		masterPort:          s.Target.MasterPort(),
 		masterDataDirectory: s.Target.MasterDataDir(),
 		binDir:              s.Target.BinDir,
+		streams:             st.Streams(),
 	}
 
 	if s.Source.HasStandby() {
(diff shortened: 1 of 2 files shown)


jchampio:
This couples the substep implementations to the step. It's no longer possible for a `Run()` implementation to pass different `OutStreams` to different substeps and have this succeed, so we might as well remove the callback argument. What's your end goal?

berlin-ab:
My current goal is to simplify the setup of the greenplumRunner. They are coupled as is, there's no change in behavior. `step.Run` could theoretically pass another stream through, but we don't rely on that yet.

berlin-ab:
The callback argument is still used elsewhere, so I left it unchanged in those cases.

jchampio:
> They are coupled as is, there's no change in behavior.

These two statements aren't related, and no, they're not coupled as-is. The substep implementations aren't coupled to the step currently because they take the streams that are passed to them and push all output through that. They're isolated from the step implementation.

That doesn't mean that this change is necessarily the wrong thing to do, but I'd like to know what we're getting by reducing the isolation. Why this change, now? What's the end state?

berlin-ab:
Waiting for a story to drive the change. Thanks for the review.

PR #251, opened 2020-03-12
Test mirrors in CI and finalize.bats

dkrieger-pvtl:
We add a complete test for mirrors in the target cluster after finalize has completed.  This test cycles through failing over the cluster to the mirrors and back again to its original configuration.  
(NOTE: it probably makes sense to review the entire changeset, and then to see the sequencing of commits.)

The test cycles a full cluster through the following steps:
```
# |  step  |   mdw       | smdw         | sdw-primaries | sdw-mirrors |
# |    1   |   master    |   standby    |    primary    |  mirror     |
# |    2   |   master    |   standby    |      -        |  mirror     |
# |    3   |   master    |   standby    |      -        |  primary    |
# |    4   |   master    |   standby    |   mirror      |  primary    |
# |    5   |   master    |   standby    |   primary     |  mirror     |
```

At each stage in the test, where appropriate, we test:

    * all segments are synchronized
    * a created table with data successfully replicates to its mirror
    * new replication connections can be established


test/finalize-standby-and-mirrors.bats
```diff
@@ -83,6 +83,41 @@ check_mirror_validity() {
     check_segments_are_synchronized
 }
 
+#check_standby_validity() {
```
dkrieger-pvtl: NOTE: this entire file gets deleted in the next commit.

ci/scripts/upgrade-cluster.bash
```diff
@@ -52,6 +54,9 @@ rpm_gphome() {
 # MAIN
 #
 
+# This port is selected by our CI pipeline
+MASTER_PORT=5432
```
berlin-ab: Would $PGPORT work for us here?

test/finalize_checks.bash
```diff
@@ -0,0 +1,157 @@
+# This file provides a single high-level function check_mirror_validity()
+# that takes a cluster with mirrors "through its paces" to thoroughly test
+# the cluster's mirrors.  See the documentation of check_mirror_validity()
+# for details.
+
+run_on_master() {
+    _run_on_host "${MASTER_HOST}" "${1}"
+}
+
+_run_on_host() {
```
berlin-ab: Do we run this on multiple hosts?

test/finalize_checks.bash
```diff
@@ -0,0 +1,157 @@
...
+
+    local on_upgraded_master=$(create_table_with_name on_upgraded_master 50)
+
+    # step 2
+    kill_primaries
+
+    # step 3
+    wait_can_start_transactions $MASTER_HOST $MASTER_PORT
+
+    check_data_matches on_upgraded_master "${on_upgraded_master}"
```
kalensk: Can you explain the two references to `on_upgraded_master` on this line? I'm not clearly understanding what is going on here.

kalensk: Oh, one is the table name and the other the bash variable name. Oof.

test/finalize_checks.bash
```diff
@@ -0,0 +1,157 @@
...
+    local on_promoted_mirrors=$(create_table_with_name on_promoted_mirrors 60)
+
+    # step 4
+    run_on_master "export MASTER_DATA_DIRECTORY=${master_data_dir}; export PGPORT=$MASTER_PORT; gprecoverseg -a"  #TODO..why is PGPORT not actually needed here?
```
kalensk: Why exactly does gprecoverseg need to be run on the master?

test/finalize_checks.bash
```diff
@@ -0,0 +1,157 @@
...
+}
+
+_check_replication_connection() {
+    local primary_address=$1
+    local primary_port=$2
+    local mirror_host=$3
+
+    local cmd="PGOPTIONS=\"-c gp_session_role=utility\" psql -h $primary_address -p $primary_port  \"dbname=postgres replication=database\" -c \"IDENTIFY_SYSTEM;\""
+    _run_on_host $mirror_host "$cmd"
+}
```
kalensk: Would simply querying `gp_stat_replication` be useful here?

test/finalize_checks.bash
dkrieger-pvtl: Yes...the CI tests run our script on the "CI root host" and then we use `ssh` to run commands on every machine in the GPDB cluster: mdw, smdw, and each segment host.

test/finalize_checks.bash
```diff
@@ -0,0 +1,157 @@
...
+
+kill_primaries() {
+    local primary_data_dirs=$(run_on_master "psql -p $MASTER_PORT -t -A -d postgres -c \"SELECT hostname, port, datadir FROM gp_segment_configuration WHERE content <> -1 AND role = 'p'\"")
+    for pair in ${primary_data_dirs[@]}; do
+        local host=$(echo $pair | awk '{split($0,a,"|"); print a[1]}')
+        local port=$(echo $pair | awk '{split($0,a,"|"); print a[2]}')
+        local dir=$(echo $pair | awk '{split($0,a,"|"); print a[3]}')
+        _run_on_host $host "pg_ctl stop -p $port -m fast -D $dir -w"
+    done
```
jchampio: This `awk` construction should be restructured as a `read` loop (or something that doesn't push a line through an awk-split every time you want a variable out of it). You'll need to either set `IFS` appropriately or add a whitespace separator to psql; e.g. `-F$'\t'`. See `gpinitsystem.bats`, `initialize runs gpinitsystem based on the source cluster`, for an example.

```bash
psql ... | while read -r host port dir; do
    _run_on_host $host "pg_ctl stop -p $port -m fast -D $dir -w"
done
```

test/finalize_checks.bash
dkrieger-pvtl: I'm not sure I understand your question.  Where would you prefer it run?

test/finalize_checks.bash
dkrieger-pvtl: I don't think so.  The entire point here is to test an _actual_ connection being created from the mirror to its primary, not to infer that one has already been made.  We test that in `_check_segments_are_synchronized` by looking at `gp_segment_configuration.mode` on each mirror/primary.

ci/scripts/upgrade-cluster.bash
dkrieger-pvtl: Perhaps; I am not sure how the CI sets up the environment.  The current file on master hardcodes `5432` in several places, so I presume not.

test/finalize_checks.bash
dkrieger-pvtl: Much cleaner.  Done.

dkrieger-pvtl:
> I need some help with readability before I can really review this effectively.

@jchampio   Can you please provide an example of what you'd like to see?

test/finalize.bats
```diff
@@ -18,7 +19,10 @@ teardown() {
         return
     fi
 
-    teardown_new_cluster
+    if [ -n "$NEW_CLUSTER" ]; then
+        teardown_new_cluster
+    fi
```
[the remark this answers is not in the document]
marksliva: What if we set NEW_CLUSTER to `gpupgrade config show --target-datadir` and pass that to `teardown_new_cluster`?

test/finalize.bats
```diff
@@ -75,13 +83,21 @@ teardown() {
...
-    # this is a more accurate representation if the standby is running and
-    # in sync, since gpstate might simply check if the process is running.
     local actual_standby_status=$(gpstate -d "${new_datadir}")
     local standby_status_line=$(get_standby_status "$actual_standby_status")
     [[ $standby_status_line == *"Standby host passive"* ]] || fail "expected standby to be up and in passive mode, got **** ${actual_standby_status} ****"
+
+    local target_mirrors_count=$(number_of_mirrors)
+    local gp_segment_configuration=$(psql postgres -c "select * from gp_segment_configuration")
+    [[ $source_mirrors_count -eq $target_mirrors_count ]] || exit "expected target mirrors count '${target_mirrors_count}' to equal source mirrors count '${source_mirrors_count}'. gp_segment_configuration:
+        ${gp_segment_configuration}"
```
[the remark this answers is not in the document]
marksliva: Good point. Sure, we can remove this check.

test/finalize_checks.bash
kalensk: To make things easier and more straightforward would about adding `_tbl` to the table name or something similar? Just something to differentiate the two symbols with the exact same name.

test/finalize_checks.bash
kalensk: I think I just realized that these tests are being run from the "CI runner box" which does not have access to gpdb or the cluster, hence the need to ssh into the cluster and the need for `run_on_master`. Is that a correct understanding?

test/finalize_checks.bash
dkrieger-pvtl: I changed:
```go
local on_upgraded_master=$(create_table_with_name on_upgraded_master 50)
```

to this:
```go
local on_upgraded_master_expected=$(create_table_with_name on_upgraded_master 50)
```

test/finalize.bats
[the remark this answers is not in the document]
dkrieger-pvtl: I'm leary using any gpupgrade command after finalize completes successfully.  Technically, that gpugprade is "done" and we currently have an undefined semantic for what it means to check gpugprade state after that.  Instead, I removed `teardown_new_cluster` and inlined that function...

test/finalize.bats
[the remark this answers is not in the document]
dkrieger-pvtl: Done.

test/finalize.bats
```diff
@@ -75,13 +83,21 @@ teardown() {
...
+
+    local local_hostname=$(hostname)
```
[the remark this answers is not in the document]
dkrieger-pvtl: Done. Inlined.

dkrieger-pvtl:
> > I need some help with readability before I can really review this effectively.
> 
> @jchampio Can you please provide an example of what you'd like to see?

@jchampio We inlined all ssh calls and used a coding style that seem reasonable to us.

ci/scripts/upgrade-cluster.bash
```diff
@@ -98,17 +103,24 @@ time ssh mdw bash <<EOF
...
               --source-bindir ${GPHOME_OLD}/bin \
-              --source-master-port 5432
+              --source-master-port $MASTER_PORT
 
     gpupgrade execute
     gpupgrade finalize
 EOF
 
+# TODO: how do we know the cluster upgraded?  5 to 6 is a version check; 6 to 6 ?????
+#   currently, it's sleight of hand...old is on port $MASTER_PORT then new is!!!!
```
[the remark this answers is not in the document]
dkrieger-pvtl: I added a comment to this TODO.

test/finalize_checks.bash
```diff
@@ -0,0 +1,186 @@
...
+            WHERE primaries.role = 'p' AND mirrors.role = 'm' AND primaries.content != -1;
+        \"
+    ")
+
+    echo "${rows}" | while read -r primary_address primary_port mirror_host; do
+        ssh -n "${mirror_host}" "
+            source ${GPHOME_NEW}/greenplum_path.sh
+            PGOPTIONS=\"-c gp_session_role=utility\" psql -h $primary_address -p $primary_port \"dbname=postgres replication=database\" -c \"
+                IDENTIFY_SYSTEM;
+            \"
```
jchampio: Have you checked to be sure that errors at this level are actually being bubbled to the very top of the calling script? I don't see any use of exit-code checks or `ON_ERROR_STOP` or what have you.

test/finalize_checks.bash
```diff
@@ -0,0 +1,186 @@
...
+    local port=$2
+    for i in {1..10}; do
+        ssh -n "${host}" "
+            source ${GPHOME_NEW}/greenplum_path.sh
+            psql -p $port -At -d postgres << EOF
+                SELECT gp_request_fts_probe_scan();
+                BEGIN; CREATE TEMP TABLE temp_test(a int) DISTRIBUTED RANDOMLY; COMMIT;
+EOF
+        "
+        if [[ $? -eq 0 ]]; then
```
jchampio: nit: the use of a Bash test without any Bash features is a little strange. Why not just a regular test?

test/finalize_checks.bash
```diff
@@ -0,0 +1,186 @@
...
+        psql -AtF$'\t' -p $MASTER_PORT -d postgres -c \"
+            SELECT hostname, port, datadir FROM gp_segment_configuration
+            WHERE content <> -1 AND role = 'p'
+        \"
+    ")
+
+    echo "${primaries}" | while read -r host port dir; do
+        ssh -n "${host}" "
+            source ${GPHOME_NEW}/greenplum_path.sh
+            pg_ctl stop -p $port -m fast -D $dir -w
```
jchampio: nit: immediate-mode shutdown might be a little closer to reality for a killed-primary test.

test/finalize_checks.bash
```diff
@@ -0,0 +1,186 @@
...
+    local on_upgraded_master_expected=$(create_table_with_name on_upgraded_master 50)
+
+    # step 2
+    kill_primaries
+
+    # step 3
+    wait_can_start_transactions $MASTER_HOST $MASTER_PORT
+
+    check_data_matches on_upgraded_master "${on_upgraded_master_expected}"
+    local on_promoted_mirrors_expected=$(create_table_with_name on_promoted_mirrors 60)
```
jchampio: Shouldn't the output from `create_table_with_name`'s `psql` invocation be polluting the expected content here? (If not -- why not? If so -- why is this not failing?)

test/finalize_checks.bash
marksliva: Good point. Changed it.

test/finalize_checks.bash
marksliva: We added a comment that explains what this function outputs. The `-q` parameter suppresses output for the psql command.

test/finalize_checks.bash
marksliva: fixed.

test/finalize_checks.bash
marksliva: Good point. It fails as we expect without this change, but adding it seems like a good idea.

PR #252, opened 2020-03-13
Display the target port and master data directory after finalize.

berlin-ab:


cli/commanders/steps.go
```diff
@@ -134,20 +135,36 @@ func Finalize(client idl.CliToHubClient, verbose bool) error {
 		return err
 	}
 
-	err = UILoop(stream, verbose)
+	dataMap, err := UILoop(stream, verbose)
 	if err != nil {
 		return xerrors.Errorf("Finalize: %w", err)
 	}
 
+	port, portOk := dataMap[hub.UIFinalizeTargetPort]
```
berlin-ab: I'd like to push these constant values into the `idl` if possible.

hub/config.go
```diff
@@ -40,6 +40,8 @@ func (s *Server) GetConfig(ctx context.Context, in *idl.GetConfigRequest) (*idl.
 		resp.Value = s.Target.BinDir
 	case "target-datadir":
 		resp.Value = s.Target.MasterDataDir()
+	case "target-port":
```
berlin-ab: We ended up not using this. We should probably remove it.

jchampio:
High-level -- did the execute step's strategy for displaying the port and datadir not work here?

hub/config.go
berlin-ab: Done.

berlin-ab:
> High-level -- did the execute step's strategy for displaying the port and datadir not work here?

We weren't aware that it existed. We thought that the owner of that data should be the hub, and we didn't a way of getting it from the hub already.

jchampio:
I like the re-decoupling of the client and hub introduced here. I am concerned that we've reimplemented [gRPC metadata](https://github.com/grpc/grpc-go/blob/master/Documentation/grpc-metadata.md) though.

berlin-ab:
> I like the re-decoupling of the client and hub introduced here. I am concerned that we've reimplemented [gRPC metadata](https://github.com/grpc/grpc-go/blob/master/Documentation/grpc-metadata.md) though.

gRPC metadata seems like it is a very sophisticated solution where a `map[string]string` would do. It looks like it was made for a different usecase in mind.

cli/commanders/steps.go
```diff
@@ -92,17 +89,12 @@ func Execute(client idl.CliToHubClient, verbose bool) error {
...
 		return xerrors.Errorf("Execute: %w", err)
 	}
 
-	path := filepath.Join(utils.GetStateDir(), hub.ConfigFileName)
-	conf := &hub.Config{}
-	err = hub.LoadConfig(conf, path)
-	if err != nil {
-		return err
-	}
+	port, datadir, err := extractTargetClusterInfo(dataMap)
```
dkrieger-pvtl: You should check the error code here, as is done in Finalize().  Otherwise, a user get an empty string for both port and datadir.  In other words, please add this code snippet after this function:
```go
	port, datadir, err := extractTargetClusterInfo(dataMap)

	if err != nil {
		return xerrors.Errorf("Execute: %w", err)
	}
```

berlin-ab: Nice catch.

cli/commanders/steps.go
```diff
@@ -191,6 +209,13 @@ func UILoop(stream receiver, verbose bool) error {
 				fmt.Println()
 			}
 
+		case *idl.Message_Response:
+			// NOTE: the latest message will clobber earlier keys
+			// todo: should we throw an error when a collision occurs?
```
dkrieger-pvtl: For now, the protocol is that a given step returns at most one `Response`.  So I think this `todo` is moot and can be removed...the `NOTE` is enough.

berlin-ab: Thanks. removed.

PR #254, opened 2020-03-16
Set disk-free-ratio defaults: .20 for link & .40 for copy

bhuvnesh2703:
Set the defaults and update the help description

### DIFF
--- a/cli/commands/commands.go
+++ b/cli/commands/commands.go
@@ -263,6 +263,16 @@ func initialize() *cobra.Command {
 		Short: "prepare the system for upgrade",
 		Long:  InitializeHelp,
 		RunE: func(cmd *cobra.Command, args []string) error {
+
+			// if diskFreeRatio is not explicitly set, use defaults
+			if !cmd.Flag("disk-free-ratio").Changed {
+				if linkMode {
+					diskFreeRatio = 0.2
+				} else {
+					diskFreeRatio = 0.6
+				}
+			}
+
 			if diskFreeRatio < 0.0 || diskFreeRatio > 1.0 {
 				// Match Cobra's option-error format.
 				return fmt.Errorf(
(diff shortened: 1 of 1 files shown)


cli/commands/commands.go
```diff
@@ -263,6 +263,16 @@ func initialize() *cobra.Command {
 		Short: "prepare the system for upgrade",
 		Long:  InitializeHelp,
 		RunE: func(cmd *cobra.Command, args []string) error {
+
+			// if diskFreeRatio is not explicitly set, use defaults
+			if diskFreeRatio == -1.0 {
```
jchampio: nit: seems a little strange that a user can pass `--disk-free-ratio -1` and something useful will happen. Is there any way around that?

bhuvnesh2703: i agree, but there is no way to have a non valid default, and set it for link vs copy mode..
OTOH, if the user is setting -1, we are ignoring that and for all non valid settings we report an error.

kalensk: Viper has a [changed or HasChanged field on flags](https://godoc.org/github.com/spf13/viper#FlagValue) that may be of use. Although, my hunch is that this would not be easy to utilize in this particular scenario. That is, the place where we would check if the flag is changed would be far apart from where disk-free-ratio is passed down. Maybe worth investigating?

bhuvnesh2703: i will defer to investigate the use of viper for this story.. and create a story in backlog..

kalensk: This seems to work...
```go
if !cmd.Flag("disk-free-ratio").Changed {
    if linkMode {
        diskFreeRatio = 0.2
    } else {
        diskFreeRatio = 0.6
    }
}
```

bhuvnesh2703: thanks a lot @kalensk, works well. i thought it will be bigger change :(

jchampio: Nice find!

PR #256, opened 2020-03-16
finalize: move mirror/standby addition to the end

jchampio:
When modifying the master catalog in master-only mode (`gpstart -am`), the standby catalog is not updated alongside it. This causes failures during standby failover, because it's still looking for the *_upgrade directories and ports.

While we try to brainstorm ways around this problem, simplify the code by adding the mirrors and standby at the very end. This removes the need for recovery.conf patchup entirely. (Dead code removal is in the final commit.)

We have purposefully not modified the temporary port/datadir assignment to the `InitializeConfig` that is done for mirrors and standby. We simply ignore the results and use the source cluster's settings instead. This is for three reasons:

1. it's easier
2. we are going to try to use them again in the near future, and removing that particular code just to add it back seems like too much trouble
3. we can still get feedback on the need to reserve mirror ports from alpha users (as opposed to making requirements stricter in a future beta)

We have slightly simplified the API for `RenameSegmentDataDirs()`. In the future, we'd like to simply pass a slice of desired segments to the function directly. But doing that now resulted in a rabbit hole, due to the confusion between what `Source` and `Target` actually point to at this
point in the finalize step.

Roadmap:
1. prefactor commit
2. main implementation
3. dead code removal (recovery.conf)


hub/finalize.go
```diff
@@ -102,6 +60,51 @@ func (s *Server) Finalize(_ *idl.FinalizeRequest, stream idl.CliToHub_FinalizeSe
...
+	targetRunner := &greenplumRunner{
+		masterPort:          s.Target.MasterPort(),
+		masterDataDirectory: s.Target.MasterDataDir(),
+		binDir:              s.Target.BinDir,
+	}
+
+	// todo: we don't currently have a way to output nothing to the UI when there is no standby.
+	// If we did, this check would actually be in `UpgradeStandby`
+	if s.Source.HasStandby() {
+		st.Run(idl.Substep_FINALIZE_UPGRADE_STANDBY, func(streams step.OutStreams) error {
```
bhuvnesh2703: do we write the standby config into config.json after it's added?

hub/finalize.go
```diff
@@ -102,6 +60,51 @@ func (s *Server) Finalize(_ *idl.FinalizeRequest, stream idl.CliToHub_FinalizeSe
...
+				Hostname:      standby.Hostname,
+				DataDirectory: standby.DataDir,
+			})
+		})
+	}
+
+	// todo: we don't currently have a way to output nothing to the UI when there are no mirrors.
+	// If we did, this check would actually be in `UpgradeMirrors`
+	if s.Source.HasMirrors() {
+		st.Run(idl.Substep_FINALIZE_UPGRADE_MIRRORS, func(streams step.OutStreams) error {
```
bhuvnesh2703: do we write the target mirror configs into config.json?

bhuvnesh2703:
and no bats test changes required?

marksliva:
Note: We intentionally did not include an end-to-end bats test for the standby. This will be added in a follow-up PR.

jchampio:
> and no bats test changes required?

@bhuvnesh2703 Forgot to mention up top; the end-to-end tests for standby are being handled in a separate PR.

hub/finalize.go
jchampio: We do not. Is there a reason you ask?

hub/finalize.go
jchampio: See above.

hub/rename_data_directories.go
```diff
@@ -57,26 +62,36 @@ func RenameMasterDataDir(masterDataDir string, isSource bool) error {
 // e.g. for target /data/dbfast1_upgrade/demoDataDir0 becomes datadirs/dbfast1/demoDataDir0
 func RenameSegmentDataDirs(agentConns []*Connection,
```
berlin-ab: What do you think about passing a list of segments, instead of a boolean flag that is used to make the selector for the segments?

hub/rename_data_directories.go
```diff
@@ -57,26 +62,36 @@ func RenameMasterDataDir(masterDataDir string, isSource bool) error {
...
+			// Otherwise include mirrors and standby. (Master's excluded above.)
+			return true
 		}
 
 		wg.Add(1)
 		go func(c *Connection) {
 			defer wg.Done()
 
-			segments := cluster.SelectSegments(excludingMaster)
+			segments := cluster.SelectSegments(selector)
```
berlin-ab: Is there an advantage to doing the selection within the gofunc? What do you think about moving it out and performing the guard clause before starting the async work?

utils/cluster.go
```diff
@@ -68,6 +68,10 @@ func ClusterFromDB(conn *dbconn.DBConn, binDir string) (*Cluster, error) {
 	return c, nil
 }
 
+func (s *SegConfig) IsMaster() bool {
```
berlin-ab: nit: we should consider moving SegConfig behavior to its own file. It's took me a second to realize that the other functions in this file were methods on the Cluster object.

hub/rename_data_directories.go
berlin-ab: Ah, I just re-read your PR description - nevermind.

hub/rename_data_directories.go
jchampio: We'd prefer to do that in the future. But we ran into roadblocks making that approach readable and would like to defer until we figure out what we want `Source` and `Target` to be long-term.

jchampio: Oh wait, I confused this with the point you were making above; reading comprehension failure. That's a good idea.

berlin-ab: > We'd prefer to do that in the future. But we ran into roadblocks making that approach readable and would like to defer until we figure out what we want `Source` and `Target` to be long-term.

I was wishing for a Fluent Interface here:

not this code exactly, but something like it:

`cluster.SelectSegments().withoutMaster().withHostname(conn.Hostname)`

https://martinfowler.com/bliki/FluentInterface.html

hub/rename_data_directories.go
```diff
@@ -57,26 +62,36 @@ func RenameMasterDataDir(masterDataDir string, isSource bool) error {
...
+			if primariesOnly {
+				return seg.Role == "p"
+			}
+
+			// Otherwise include mirrors and standby. (Master's excluded above.)
+			return true
 		}
 
 		wg.Add(1)
 		go func(c *Connection) {
```
jchampio: While addressing other feedback: `conn` is correctly pinned to a new scope inside the `range` loop, so we can use it directly.

jchampio:
Updated with a slight tangential fixup (`c` -> `conn`); let me know if it makes review confusing and I can keep it for a followup PR.

hub/rename_data_directories.go
jchampio: > What do you think about moving it out and performing the guard clause before starting the async work?

Done.

PR #258, opened 2020-03-17
sed: simplify call for safety and fix pattern bugs

jchampio:
So that we don't fail halfway through (after the move to a `.bak`) and leave the cluster without a file, use sed's builtin replacement.

Additionally, the current implementation won't be idempotent if, say, we replace the port number with a number that is a superstring of the previous one.

Example, replacing 5000 with 50000:

     port=5000 -> port=50000

The second run will match `port=5000` again:

     port=50000 -> port=500000

While fixing this, we discovered some additional anchoring bugs related to the GUC name and comments, so we added a suite to fix them all. The current test subsumes the previous test, which was simply checking argument equality, so we've replaced it entirely.

This work suggests several refactors, which I will create a new PR for. This PR will wait until the standby/mirror reordering hits master.

### DIFF
--- a/hub/replace_string_within_file.go
+++ b/hub/replace_string_within_file.go
@@ -3,7 +3,6 @@ package hub
 import (
 	"fmt"
 
-	"github.com/greenplum-db/gp-common-go-libs/gplog"
 	"golang.org/x/xerrors"
 )
 
@@ -23,15 +22,12 @@ func ReplaceStringWithinFile(pattern string, replacement string, file string) er
 }
 
 func replaceStringWithinFile(pattern string, replacement string, pathToFile string) error {
-	script := fmt.Sprintf(
-		"sed 's@%[1]s@%[2]s@' %[3]s > %[3]s.updated && "+
-			"mv %[3]s %[3]s.bak && "+
-			"mv %[3]s.updated %[3]s",
-		pattern, replacement, pathToFile)
+	// XXX this isn't safe if the pattern or replacement contains @-signs
+	substitutionString := fmt.Sprintf("s@%s@%s@", pattern, replacement)
 
-	gplog.Debug("executing command: %+v", script) // TODO: Move this debug log into ExecuteLocalCommand()
+	cmd := execCommand("sed", "-i.bak", substitutionString, pathToFile)
 
-	cmd := execCommand("bash", "-c", script)
+	// TODO: stream back output through the connection
 	_, err := cmd.Output()
 	if err != nil {
 		return xerrors.Errorf("updating %s: %w", pathToFile, err)
--- a/hub/replace_string_within_file.go
+++ b/hub/replace_string_within_file.go
@@ -25,7 +25,13 @@ func replaceStringWithinFile(pattern string, replacement string, pathToFile stri
 	// XXX this isn't safe if the pattern or replacement contains @-signs
 	substitutionString := fmt.Sprintf("s@%s@%s@", pattern, replacement)
 
-	cmd := execCommand("sed", "-i.bak", substitutionString, pathToFile)
+	cmd := execCommand(
+		"sed",
+		"-E",     // use POSIX extended regexes
+		"-i.bak", // in-place substitution with .bak backup extension
+		substitutionString,
+		pathToFile,
+	)
 
 	// TODO: stream back output through the connection
 	_, err := cmd.Output()
--- a/hub/update_conf_files.go
+++ b/hub/update_conf_files.go
@@ -24,9 +24,12 @@ func UpdateGpperfmonConf(masterDataDir string) error {
 	logDir := filepath.Join(masterDataDir, "gpperfmon", "logs")
 	replacement := fmt.Sprintf("log_location = %s", logDir)
 
-	return ReplaceStringWithinFile("log_location = .*$",
+	// TODO: allow arbitrary whitespace around the = sign?
+	return ReplaceStringWithinFile(
+		"^log_location = .*$",
 		replacement,
-		configFile)
+		configFile,
+	)
 }
 
 // oldTargetPort is the old port on which the target cluster was initialized.
--- a/hub/update_conf_files.go
+++ b/hub/update_conf_files.go
@@ -35,8 +35,8 @@ func UpdateGpperfmonConf(masterDataDir string) error {
 // catalog in a previous substep.
 func UpdatePostgresqlConf(oldTargetPort int, target *utils.Cluster, source *utils.Cluster) error {
 	return ReplaceStringWithinFile(
-		fmt.Sprintf("port=%d", oldTargetPort),
-		fmt.Sprintf("port=%d", source.MasterPort()),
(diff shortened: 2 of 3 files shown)


hub/replace_string_within_file.go
```diff
@@ -23,15 +22,12 @@ func ReplaceStringWithinFile(pattern string, replacement string, file string) er
...
-	script := fmt.Sprintf(
-		"sed 's@%[1]s@%[2]s@' %[3]s > %[3]s.updated && "+
-			"mv %[3]s %[3]s.bak && "+
-			"mv %[3]s.updated %[3]s",
-		pattern, replacement, pathToFile)
+	// XXX this isn't safe if the pattern or replacement contains @-signs
+	substitutionString := fmt.Sprintf("s@%s@%s@", pattern, replacement)
 
-	gplog.Debug("executing command: %+v", script) // TODO: Move this debug log into ExecuteLocalCommand()
+	cmd := execCommand("sed", "-E", "-i.bak", substitutionString, pathToFile)
```
berlin-ab: Does `sed` have long form arguments for `-E` and `-i`? It would be nice for the reader to not have to look up their meaning.

hub/update_conf_files_test.go
```diff
@@ -1,87 +1,130 @@
...
-			return nil
-		})
-
-		hub.UpdateGpperfmonConf("/some/master/data/dir")
-
-		if usedPattern != "log_location = .*$" {
-			t.Errorf("got %v, expected log_location = .*", usedPattern)
+	defer hub.ResetExecCommand()
+
+	hub.ResetReplaceStringWithinFile()
```
marksliva: The method `SetReplaceStringWithinFile` is never used, so we can remove both `ResetReplaceStringWithinFile` and `SetReplaceStringWithinFile`. We can then clean up `hub/replace_string_within_file.go` to contain a single function.

jchampio: That PR is on its way already. Just need to write the PR roadmap.

hub/replace_string_within_file.go
jchampio: Not in a way that is portable to both BSD and GNU sed, according to their manuals.

hub/update_conf_files_test.go
jchampio: See #259.

hub/replace_string_within_file.go
```diff
@@ -23,15 +22,12 @@ func ReplaceStringWithinFile(pattern string, replacement string, file string) er
...
-	script := fmt.Sprintf(
-		"sed 's@%[1]s@%[2]s@' %[3]s > %[3]s.updated && "+
-			"mv %[3]s %[3]s.bak && "+
-			"mv %[3]s.updated %[3]s",
-		pattern, replacement, pathToFile)
+	// XXX this isn't safe if the pattern or replacement contains @-signs
+	substitutionString := fmt.Sprintf("s@%s@%s@", pattern, replacement)
 
-	gplog.Debug("executing command: %+v", script) // TODO: Move this debug log into ExecuteLocalCommand()
+	cmd := execCommand("sed", "-i.bak", substitutionString, pathToFile)
```
bhuvnesh2703: is the in place replacement available with all version of sed (atleast what we have on centos6,7)?
i vaguely remember, the same command was working on mac and not working on some machines? (I may be wrong)

jchampio: GNU and BSD sed both support `-iSUFFIX` (that is, with the suffix pinned to the option). Creating portable sed lines is tricky, but we have CI coverage for CentOS 6 to catch what doesn't show up on our local Macs.

(By the way, if you `brew install coreutils`, you'll get `g`-prefixed versions of many utilities (`gsed`, `gawk`, etc.) so that you can test GNU commands for portability on Mac too.)

hub/replace_string_within_file.go
berlin-ab: Ah, what do you think about leaving a comment instead?

hub/update_conf_files.go
```diff
@@ -19,10 +20,12 @@ func (s *Server) UpdateConfFiles() error {
...
 func UpdateGpperfmonConf(masterDataDir string) error {
-	configFile := fmt.Sprintf("%[1]s/gpperfmon/conf/gpperfmon.conf", masterDataDir)
-	replacement := fmt.Sprintf("log_location = %[1]s/gpperfmon/logs", masterDataDir)
+	configFile := filepath.Join(masterDataDir, "gpperfmon", "conf", "gpperfmon.conf")
+	logDir := filepath.Join(masterDataDir, "gpperfmon", "logs")
+	replacement := fmt.Sprintf("log_location = %s", logDir)
 
-	return ReplaceStringWithinFile("log_location = .*$",
+	// TODO: allow arbitrary whitespace around the = sign?
+	return ReplaceStringWithinFile("^log_location = .*$",
```
marksliva: nit: we should be consistent with indentation between the 2 usages of `ReplaceStringWithinFile` for better readability.

hub/update_conf_files.go
```diff
@@ -33,8 +36,8 @@ func UpdateGpperfmonConf(masterDataDir string) error {
 // catalog in a previous substep.
 func UpdatePostgresqlConf(oldTargetPort int, target *utils.Cluster, source *utils.Cluster) error {
 	return ReplaceStringWithinFile(
```
marksliva: It would be good if we tested the error case, at least in `hub/replace_string_within_file_test.go` if not both places.

hub/update_conf_files_test.go
```diff
@@ -1,87 +1,130 @@
...
+port=5000 # comment
+port = 5000 # make sure we can handle spaces
+
+# should not be replaced
+gpperfmon_port=5000
+port=50000
+#port=5000
+`)
+
+		// NOTE: we set the source and target cluster ports equal to enforce that
```
marksliva: It would be better to make these ports different, since only one of them gets used. Unless the point is that at this point in our system, they are indeed equivalent. But if that's true, why pass in source at all?

jchampio: We address this in the follow-up draft by removing the `Cluster` coupling entirely; let me know what you think.

hub/replace_string_within_file.go
jchampio: Will do.

hub/update_conf_files.go
jchampio: Was that case covered with the previous implementation? We were mostly trying to maintain current coverage for the bug fix.

hub/update_conf_files.go
jchampio: Will fix.

hub/update_conf_files.go
berlin-ab: No, it wasn't.

hub/update_conf_files_test.go
marksliva: The refactor did clean it up. Looks good.

PR #259, opened 2020-03-17
sed followup: refactor away indirection, decouple from Server

jchampio:
The first commit flattens the indirection tree. The previous units of tested/pinned behavior were the string-replacement layer, which was tested with an integration test against the local system to ensure the `sed` invocation was working properly, and the integration layer above that, which was tested by checking to make sure the arguments passed to the layer below it were exact matches for what was expected. This had several problems exposed by the previous PR work:

First, the integration layer's unit tests pinned implementation rather than observable behavior, so they would have had to be completely rewritten in order to fix bugs.

Second, the string-replacement layer was a leaky abstraction: the pattern passed in is inextricably coupled to the implementation, since it's not safe to write a `sed` pattern without knowing the flavor of `sed` in use. Not only would the pattern have to be escaped according to the separator in use (i.e. `s///` vs `s|||` vs `s@@@` -- admittedly a solvable, if painful, problem), but the new implementation requires POSIX extended regexes, and standard POSIX regexes and extended POSIX regexes are mutually exclusive.

Third, this testing strategy has very little power to catch bugs in domain logic. We can test that the first layer passes arguments as written and we can test that the second layer works as expected, but what we really care about is that both layers working together are capable of performing configuration replacement correctly.

This set of tradeoffs shows up often in utilities programming. We feel that the correct test boundaries for this particular problem (i.e. the "black box" whose behavior we want to pin) are at the integration level, not as a series of smaller units. So we remove the intermediate indirection, inlining the `sed` logic and replacing generalities with the bare minimum implementation. This makes it trivial to perform stderr streaming, which we do now, and exposes further areas of potential improvement, which we'll leave for the Rule of Three to take effect.

The second commit is by contrast relatively simple; we're just getting rid of unnecessary `Server`/`Cluster` coupling.


ivannovick:
so i gather now gpinitsystem calls a custom gobinary to do various jobs?  Thats new to me since i looked at gpinitsystem years back. How many go binaries do we have?

jchampio:
@ivannovick This is gpupgrade; gpinitsystem doesn't know anything about it.

berlin-ab:
I agree that the tests ideally should not be as fine-grained as they currently are. I'd prefer to test from the module boundary, but it's extremely difficult to do so with our current design.

I agree that the abstraction is leaky, and there are tradeoffs to be considered there:

- The rule of three should be invoked by now. We have three identical uses of sed.
- Once-and-only-once / DRY should apply here. The knowledge that our system requires edits to a file to be atomic should be captured once somewhere, or we risk coming up with yet another way to do replacement - and introduce an idempotency bug.
- I don't consider it a new layer to have a function that calls out the intention of the work being done. The code should tell me what should be happening, not how.

berlin-ab:
> from the module boundary

To be more clear here, the methods on Server that we call (via gRPC during Integration). Those are the real public API of the module. If we'd like to test functions at a lower level that the methods on Server, there's a signal that those functions belong in a different package as a separate abstraction.

jchampio:
> The rule of three should be invoked by now. We have three identical uses of sed.

Two; https://github.com/greenplum-db/gpupgrade/pull/256 is removing the third. So this file will contain the only uses of sed in the codebase.

> Once-and-only-once / DRY should apply here. The knowledge that our system requires edits to a file to be atomic should be captured once somewhere, or we risk coming up with yet another way to do replacement - and introduce an idempotency bug.

It continues to be a risk, I agree. (The current approach did not mitigate that risk, for the record. It is extremely hard to stop people from writing bugs.)

For utility codebases consisting mostly of glue code, I personally prefer to mitigate that risk with complicated-domain techniques -- the dissemination of developer knowledge and good review culture -- instead of trying to push abstractions in before they have proven their long-term helpfulness. Kind of an anti-Peter-Principle for abstractions. Because those abstractions are not going to stop us from writing bugs, and in cases like this, where the wrong abstraction is used, they increase the effort required to fix them.

So in this case, I prefer just saying "atomic operations often allow us to make stronger guarantees about system behavior in general, not just in this case, and you should think what guarantees you're trying to make any time you add a new call to a system command". That helps developers whether they're adding a call to `sed` or to `gpinitsystem` or something entirely new. And then, over time, we move from complicated-domain analysis to having end-to-end tests that can catch you when you make an idempotence mistake. But that end goal is very labor-intensive, and I feel that good review culture and developer education makes the right set of tradeoffs for right now.

> I don't consider it a new layer to have a function that calls out the intention of the work being done.

I'm not sure we're likely to agree on this, because to me it seems like redefining the word "layer".

Martin Fowler talks about this case in _Refactoring_, in the `Inline Function` step:

> I commonly use Inline Function when I see code that's using too much indirection -- when it seems like every function does simple delegation to another function, and I get lost in all the delegation. Some of this indirection may be worthwhile, but not all of it. By inlining, I can flush out the useful ones and eliminate the rest.

It's a tradeoff, and right now I'm on the side of "this was too far in that direction". Now that we have more evidence of what we want the code to do, we might find that having an `inplaceSed()` helper function might be nice to have in the future. But even then, it should be an implementation detail first, and it shouldn't try to abstract away behavior until we have clear ideas of what the abstraction boundary is. At this moment, that piece doesn't help us as a public, testable unit.

hub/finalize.go
```diff
@@ -46,8 +46,12 @@ func (s *Server) Finalize(_ *idl.FinalizeRequest, stream idl.CliToHub_FinalizeSe
 		return s.RenameDataDirectories()
 	})
 
-	st.Run(idl.Substep_FINALIZE_UPDATE_TARGET_CONF_FILES, func(_ step.OutStreams) error {
-		return s.UpdateConfFiles()
+	st.Run(idl.Substep_FINALIZE_UPDATE_TARGET_CONF_FILES, func(streams step.OutStreams) error {
```
berlin-ab: I'm curious why you chose to unpack things here rather than make that the business logic of UpdateConfigFiles.

jchampio: You mean just pass the entire server `Config` and have `UpdateConfFiles()` figure it out? That's a fine option too; we had just come off a case where too much reliance on a whole `Cluster` had made things difficult, so I think we were conservative in the API here.

jchampio:
Thanks for the review!

PR #260, opened 2020-03-17
Fix output and help text

kalensk:
This adds the following missing idl.Substeps:
- idl.Substep_CREATING_DIRECTORIES,
- idl.Substep_GENERATING_CONFIG,
- idl.Substep_START_HUB,
- idl.Substep_RETRIEVE_SOURCE_CONFIG,
- idl.Substep_CHECK_DISK_SPACE,

Thus, the correct help text for initialize is being produced. In
initialize we now use these idl.Substep enum's instead of their
hardcoded descriptions.

While adding the above enums to initialize we ran into needing to add one for checkVersion which is no longer being used, so we decided to remove it.

Further work could investigate if initialize.go could follow the same pattern of `step.Begin` as found in execute.go and finalize.go.


cli/commanders/steps_test.go
```diff
@@ -264,14 +264,12 @@ func TestSubstep(t *testing.T) {
 	d := bufferStandardDescriptors(t)
 	defer d.Close()
 
-	description := "my description"
-
 	var err error
-	s := commanders.Substep(description)
+	s := commanders.Substep(idl.Substep_CREATING_DIRECTORIES)
```
jchampio: This file's changes belong in the next commit.

kalensk:
> protobuf need to know about actions that are client-only

I wanted a single source of truth for all steps as an enum. Protobuf seemed like the best choice for now.

berlin-ab:
It makes sense to have a single source of truth. Protobuf is currently the keeper of the domain knowledge of the substeps.

jchampio:
> [Jacob] This feels like a hack...

(It was pointed out to me that this is not especially helpful. I wasn't trying to cast aspersions on the code -- I apologize. After all, I think was the one who introduced those hard-coded strings to begin with...)

> Currently, we have hard coded strings as well as enum substeps, etc. To me consolidating felt like the better option/tradeoff even if its for the CLI and not protobuf specific.

I think this is fair. I'm pretty sensitive at the moment to coupling UI-specific concerns to network concerns, as I think the substep system is already doing that a little too much at the moment, and every step we take in that direction gives that more inertia. But an ideal fully decoupled system is probably a ways off, and in the meantime this particular change shouldn't push any problems on the developer.

(To give you an idea of where I'm trying to go, I'd like to eventually switch to Go's localization framework (`x/text`/`gotext`) so that we can have the product team control aspects of the UI directly using JSON. I'm hoping that will let us put the UI specifics in a single place, while also not forcing us to push every related UI concern into a single code structure in Go.)

kalensk:
> (It was pointed out to me that this is not especially helpful. I wasn't trying to cast aspersions on the code -- I apologize. After all, I think was the one who introduced those hard-coded strings to begin with...)

From my perspective, no reason to apologize from my end. Your comment made sense to me and was coming out of a place of care for the code.
 
> I think this is fair. I'm pretty sensitive at the moment to coupling UI-specific concerns to network concerns, as I think the substep system is already doing that a little too much at the moment, and every step we take in that direction gives that more inertia. But an ideal fully decoupled system is probably a ways off, and in the meantime this particular change shouldn't push any problems on the developer.

> (To give you an idea of where I'm trying to go, I'd like to eventually switch to Go's localization framework (`x/text`/`gotext`) so that we can have the product team control aspects of the UI directly using JSON. I'm hoping that will let us put the UI specifics in a single place, while also not forcing us to push every related UI concern into a single code structure in Go.)

Sure makes sense, and as noted I don't think we are quite there yet. This PR is trying to collect everything in one place to _hopefully_ make the changes you suggested a bit easier. I think adding a story, prioritizing, and making an effort towards decoupling as suggested would be a great approach.

cli/commanders/steps_test.go
jmcatamney: Fixed.

PR #261, opened 2020-03-18
Move shutdown source cluster help text into execute.

berlin-ab:


berlin-ab:
Oops, this is resolved by #260.

PR #263, opened 2020-03-18
Migrate panic_test, commands_test, and hub_test to Go style tests

kalensk:
Migrate panic_test, commands_test, and hub_test to Go style tests. To keep things simple for now we simply ported over the similar logic.

[Test pipeline](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:go_tests)

### DIFF
--- a/agent/rsync_test.go
+++ b/agent/rsync_test.go
@@ -12,16 +12,8 @@ import (
 	"golang.org/x/xerrors"
 
 	"github.com/greenplum-db/gpupgrade/agent"
+	"github.com/greenplum-db/gpupgrade/testutils"
 )
-
-func getTempDir(t *testing.T) string {
-	sourceDir, err := ioutil.TempDir("", "rsync-source")
-	if err != nil {
-		t.Fatalf("creating temporary directory: %+v", err)
-	}
-
-	return sourceDir
-}
 
 func writeToFile(filepath string, contents []byte, t *testing.T) {
 	err := ioutil.WriteFile(filepath, contents, 0644)
@@ -42,11 +34,11 @@ func TestRsync(t *testing.T) {
 	defer func() { agent.SetRsyncCommand(nil) }()
 
 	t.Run("it copies data from a source directory to a target directory", func(t *testing.T) {
-		sourceDir := getTempDir(t)
-		defer os.RemoveAll(sourceDir)
+		sourceDir := testutils.GetTempDir(t, "rsync-source")
+		defer testutils.MustRemoveAll(t, sourceDir)
 
-		targetDir := getTempDir(t)
-		defer os.RemoveAll(targetDir)
+		targetDir := testutils.GetTempDir(t, "rsync-target")
+		defer testutils.MustRemoveAll(t, targetDir)
 
 		writeToFile(filepath.Join(sourceDir, "hi"), []byte("hi"), t)
 
@@ -64,11 +56,11 @@ func TestRsync(t *testing.T) {
 	})
 
 	t.Run("it removes files that existed in the target directory before the sync", func(t *testing.T) {
-		sourceDir := getTempDir(t)
-		defer os.RemoveAll(sourceDir)
+		sourceDir := testutils.GetTempDir(t, "rsync-source")
+		defer testutils.MustRemoveAll(t, sourceDir)
 
-		targetDir := getTempDir(t)
-		defer os.RemoveAll(targetDir)
+		targetDir := testutils.GetTempDir(t, "rsync-target")
+		defer testutils.MustRemoveAll(t, targetDir)
 
 		writeToFile(filepath.Join(targetDir, "target-file-that-should-get-removed"), []byte("goodbye"), t)
 
@@ -84,11 +76,11 @@ func TestRsync(t *testing.T) {
 	})
 
 	t.Run("it does not copy files from the source directory when in the exclusion list", func(t *testing.T) {
-		sourceDir := getTempDir(t)
-		defer os.RemoveAll(sourceDir)
+		sourceDir := testutils.GetTempDir(t, "rsync-source")
+		defer testutils.MustRemoveAll(t, sourceDir)
 
-		targetDir := getTempDir(t)
-		defer os.RemoveAll(targetDir)
+		targetDir := testutils.GetTempDir(t, "rsync-target")
+		defer testutils.MustRemoveAll(t, targetDir)
 
 		writeToFile(filepath.Join(sourceDir, "source-file-that-should-get-excluded"), []byte("goodbye"), t)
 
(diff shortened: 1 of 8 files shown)


agent/rsync_test.go
```diff
@@ -9,20 +9,13 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/greenplum-db/gpupgrade/testutils"
+
 	"golang.org/x/xerrors"
 
 	"github.com/greenplum-db/gpupgrade/agent"
```
jchampio: goimports

utils/log/panic_test.go
```diff
@@ -1,39 +1,37 @@
...
-
-		panicMsg := "aaahhhhh"
-		Expect(func() {
+		expected := "ahhh"
+		func() {
 			defer log.WritePanics()
-			panic(panicMsg)
-		}).To(Panic())
+			panic(expected)
+		}()
```
jchampio: Does this `func()` serve a purpose anymore without the `Expect()` layer?

utils/log/panic_test.go
```diff
@@ -1,39 +1,37 @@
...
+		if !called {
+			t.Errorf("expected panic func to be called")
+		}
 
-		Expect(testout.Contents()).To(BeEmpty())
-		Expect(testerr.Contents()).To(BeEmpty())
-		Expect(testlog).To(Say(`encountered panic \("%s"\); stack trace follows`, panicMsg))
+		contents := string(buffer.Contents())
+		if contents != expected {
+			t.Errorf("got %q want %q", contents, expected)
```
jchampio: This looks like a very different test; why do we pass without following the `encountered panic \("%s"\); stack trace follows` format?

utils/log/panic_test.go
```diff
@@ -1,39 +1,37 @@
...
-		Expect(func() {
+		expected := "ahhh"
+		func() {
 			defer log.WritePanics()
-			panic(panicMsg)
-		}).To(Panic())
+			panic(expected)
+		}()
+
+		if !called {
```
jchampio: This code will never be run, right? We just panicked.

utils/log/panic_test.go
jchampio: Oh, I think there's a test bug; see below.

testutils/test_utils.go
```diff
@@ -49,3 +52,35 @@ func GetOpenPort() (int, error) {
...
+func GetTempDir(t *testing.T) string {
+	sourceDir, err := ioutil.TempDir("", "gpupgrade")
+	if err != nil {
+		t.Fatalf("creating temporary directory: %+v", err)
+	}
+
+	return sourceDir
+}
+
+func RemoveTempDir(t *testing.T, dir string) {
```
jchampio: nit: this is more of a `MustRemoveAll()`; it doesn't care about the temporary-ness of its argument.

integrations/commands_test.go
```diff
@@ -25,23 +23,32 @@ var _ = Describe("test gpupgrade help messages", func() {
...
+			t.Run(fmt.Sprintf("testing command %q with flag %q", command, flag), func(t *testing.T) {
 				cmd := exec.Command("gpupgrade", command, flag)
 				if command == "" {
 					cmd = exec.Command("gpupgrade", flag)
 				}
 				output, err := cmd.Output()
-				Expect(err).ToNot(HaveOccurred())
-				Expect(string(output)).To(Equal(help))
+				if err != nil {
+					t.Errorf("unexpected err: %#v", err)
```
jchampio: IMO `%+v` is more helpful when you want to debug an expected error ("what caused this error/where did it come from?"). `%#v` helps with error equality checks ("why are these errors not the same?").

agent/rsync_test.go
```diff
@@ -42,11 +35,11 @@ func TestRsync(t *testing.T) {
 	defer func() { agent.SetRsyncCommand(nil) }()
 
 	t.Run("it copies data from a source directory to a target directory", func(t *testing.T) {
-		sourceDir := getTempDir(t)
-		defer os.RemoveAll(sourceDir)
+		sourceDir := testutils.GetTempDir(t)
+		defer testutils.RemoveTempDir(t, sourceDir)
```
jchampio: The promotion of helpers that change behavior at the same time we're trying to review to make sure no test bugs are introduced makes things difficult for me.

kalensk: I split the commit into two: 1) `Extract helpers to testutils` and 2) `Migrate hub_test.go to Go style test`.

integrations/commands_test.go
kalensk: Thanks for the clarification. I tend to forget those or mix them up. Hearing it again will hopefully help sink it in more.

testutils/test_utils.go
kalensk: Yup, you're right. I renamed it.

utils/log/panic_test.go
kalensk: Yeah, I reworked the test. Take a look and hopefully we can figure it out. 

Note that I didn't quite follow the convention of `got, want` in favor or readability on failures. Thus, I have `t.Errorf("expected %q in log file: %q"` since the log file contents is quite long and hard to read. Let me know if you have other ideas.

agent/rsync_test.go
kalensk: Fixed. Not sure how to make my editor behave...

berlin-ab: Which editor? I think i’ve figured it out for GoLand.

kalensk: Yeah intelliJ. Seems like I have goimports checked on  "file watchers"

kalensk:
Addressed comments and rebased.

utils/log/panic_test.go
```diff
@@ -1,39 +1,29 @@
...
-
-		Expect(testout.Contents()).To(BeEmpty())
-		Expect(testerr.Contents()).To(BeEmpty())
-		Expect(testlog).To(Say(`encountered panic \("%s"\); stack trace follows`, panicMsg))
+func TestWritePanics(t *testing.T) {
+	t.Run("writes panics to the log file", func(t *testing.T) {
+		_, _, buffer := testhelper.SetupTestLogger()
+
+		expected := "ahhh"
+		defer func() {
```
berlin-ab: It took me a second to realize that we were implicitly keeping the assertion that `WritePanics` does not recover panics by ordering the defer statements in this way. Is there a way to call out that behavior for future readers as an explicit intended behavior?

kalensk: Do you have any specific suggestions?

I am not sure I am completely following. My understanding is that go recovers from panic's in a defer which is placed before the funciton panic'ing.

The test is an attempt at a more straightforward and simplified way of the [panicMatcher](https://github.com/onsi/gomega/blob/master/matchers/panic_matcher.go#L27-L35).

berlin-ab: Nevermind.

utils/log/panic_test.go
kalensk: This test has been reworked. See latest changes.

integrations/commands_test.go
```diff
@@ -25,23 +23,32 @@ var _ = Describe("test gpupgrade help messages", func() {
...
-		Expect(err).ToNot(HaveOccurred())
-		Expect(string(output)).To(Equal(commands.GlobalHelp))
-	})
+		if err != nil {
+			t.Errorf("unexpected err: %#v", err)
+		}
 
-})
+		if string(output) != commands.GlobalHelp {
+			t.Errorf("got help output %q want %q", string(output), commands.GlobalHelp)
```
bhuvnesh2703, line 51: nit: got global help output..

integrations/hub_test.go
```diff
@@ -1,41 +1,102 @@
...
+		}
+	}
+
+	if err != nil {
+		t.Errorf("unexpected error got %#v", err)
+	}
+}
+
+func isPortInUse(port int) bool {
+	t := time.After(2 * time.Second)
```
bhuvnesh2703, line 85: why not use net.Listen to check if port is available?

PR #264, opened 2020-03-18
ci: Add upgrade jobs for clusters with no mirrors

marksliva:
We currently test upgrade on clusters with a standby and clusters with no
standby.
Add 2 new jobs that do the same for mirrors:
- one for each 5-to-6 upgrade (currently there is just one of these)
- a 6-to-6 job that has no standby and no mirrors

See the ccp_src PR (branch `no-mirrors-cluster`) to see the required changes there.


dkrieger-pvtl:
There are now 5 X-to-Y tests in the pipeline:
```
6-to-6
6-to-6-no-standby
6-to-6-no-standby-no-mirrors
5-to-6.1.0
5-to-6.1.0-no-mirrors
```

We don't want to test all combinations in CI(too "costly"), so we should decide which subset of full combinatorics we want to test.  I am not sure `5-to-6.1.0-no-mirrors` is useful....will a user ever have such a system?  I'd rather see `5-to-6.1.0-no-standby-no-mirrors` instead.  Then we cover what I think are the common use cases(upgrade with-mirrors-with-standby and no-mirrors-no-standby for 6-to-6 and 5-to-6) and the regression case no-standby.

ci/scripts/upgrade-cluster.bash
```diff
@@ -120,8 +120,10 @@ if ! compare_dumps /tmp/old.sql /tmp/new.sql; then
     exit 1
 fi
 
-# Test that mirrors actually work
-echo 'Doing failover tests of mirrors...'
-check_mirror_validity "${GPHOME_NEW}" mdw $MASTER_PORT
+if [ "${NO_MIRRORS}" = "0" ]; then
```
dkrieger-pvtl: I'd rather eliminate the `NO_MIRRORS` variable and use instead an explicit check on the cluster to see if it has mirrors....something like this(but one that checks for only mirrors via an additional `WHERE content<>-1` check).  The code below is in #266 .

```bash
contents_without_mirror() {
    local gphome=$1
    local host=$2
    local port=$3

    ssh -n "$host" "
        source ${gphome}/greenplum_path.sh
        psql -p $port -At -d postgres -c \"
            SELECT content
            FROM gp_segment_configuration
            GROUP BY content
            HAVING COUNT(*)!=2;
        \"
    "
}
```

To check:
```bash
    local mirrorless
    mirrorless=$(contents_without_mirror "$GPHOME_NEW" "$MASTER_HOST" "$MASTER_PORT")
    if [ -n "$mirrorless" ]; then
        echo "This test only works on full clusters but these content ids do not have mirrors: ${mirrorless}"
        exit 1
    fi
```

ci/parser/pipeline-template.yml
```diff
@@ -340,6 +394,9 @@ jobs:
         FILTER_DIFF: 1
         OLD_PACKAGE: greenplum-db-5
         NEW_PACKAGE: greenplum-db-6
+        {{- if .NoMirrors}}
+        NO_MIRRORS: 1
```
dkrieger-pvtl: See [suggestion to eliminate this variable](https://github.com/greenplum-db/gpupgrade/pull/264/files#r394685095).

ci/parser/pipeline-template.yml
```diff
@@ -262,8 +262,59 @@ jobs:
 
 {{- end}}
 
-{{range .VersionPairs}}
-- name: {{.Source}}-to-{{.Target}}
+- name: 6-to-6-no-standby-no-mirrors
```
dkrieger-pvtl: nit: `6-to-6-primaries-only` is more readable to me.

ci/parser/pipeline-template.yml
marksliva: I prefer to be intentional here. I wouldn't normally query the state of the system to figure out what my test should check for next. We know exactly when the source cluster has been created with mirrors, and it is fairly trivial to pass that context down to the test and make assertions on the mirrors that we know must now exist on the target.

Here is a concrete issue with your proposal: if the 2 substeps in `finalize.go` that add a standby and mirrors were accidentally removed, then our tests in CI would incorrectly pass.

marksliva:
> There are now 5 X-to-Y tests in the pipeline:
> 
> ```
> 6-to-6
> 6-to-6-no-standby
> 6-to-6-no-standby-no-mirrors
> 5-to-6.1.0
> 5-to-6.1.0-no-mirrors
> ```
> 
> We don't want to test all combinations in CI(too "costly"), so we should decide which subset of full combinatorics we want to test. I am not sure `5-to-6.1.0-no-mirrors` is useful....will a user ever have such a system? I'd rather see `5-to-6.1.0-no-standby-no-mirrors` instead. Then we cover what I think are the common use cases(upgrade `with-mirrors-with-standby` and `no-mirrors-no-standby` for 6-to-6 and 5-to-6) and the regression case `no-standby`.
> 
> A further consideration is the dump we are testing in each. I'd argue that we should use ICW for `6-to-6`(to test regressions in 6) and `5-to-6`(user use case) only, and a small dump for the others(which are testing cluster structure mainly).

Yeah, we can reduce the number of combinations by combining no-standby and no-mirrors. So, for all `x-to-y` version pairs, I will add an `x-to-y-primaries-only` job. Once we add more version pairs, we can then figure out which of them should include running a `x-to-y-primaries-only` job.

`6-to-6` is an ICW dump, `5-to-6` is a small dump. We can create follow-up stories to address which dumps to use where, however this PR is not concerned with making any changes in this area.

ci/parser/pipeline-template.yml
marksliva: Changed it.

dkrieger-pvtl:
We discussed using the following combinations, and we want to add a 5 ICW dump(note currently 5 has very little data in it to upgrade) for upgrading.  Adding a 5 ICW dump is a separate task.

```
6-to-6
5-to-6
5-to-6-no-standby
5-to-6-primaries-only
```

ci/scripts/upgrade-cluster.bash
```diff
@@ -121,12 +121,11 @@ if ! compare_dumps /tmp/old.sql /tmp/new.sql; then
 fi
 
 # Test that mirrors actually work
-mirrorless=$(contents_without_mirror "${GPHOME_NEW}" mdw $MASTER_PORT)
-if [ -n "$mirrorless" ]; then
-    echo "skipping validate_mirrors_and_standby since these content ids do not have mirrors: ${mirrorless}"
-else
+if [ "${MIRRORS}" = "1" ]; then
```
dkrieger-pvtl: The current `validate_mirrors_and_standby` requires mirrors and a standby.  The `MIRRORS=0` variable refers to a cluster with neither mirror nor standby in the pipeline.yml file...it'd be cleaner to have an explicitly set MIRRORS variable for just mirrors and STANDBY just for the standby, and pass that in to upgrade-cluster.yml.  Then this check would read:

```bash
if [ "${MIRRORS}" = "1"  && "${STANDBY}" = "1" ]; then
```

marksliva: Instead of 2 new single-purpose variables, why not a single one called `STANDBY_AND_MIRRORS`?

ci/tasks/upgrade-cluster.yml
```diff
@@ -19,5 +19,6 @@ run:
 
 params:
   FILTER_DIFF: 0
+  STANDBY_AND_MIRRORS: 1
```
dkrieger-pvtl: Why not have separate flags for `STANDBY` and for `MIRRORS`?  That makes the pipeline yml a bit cleaner to me and allows us to easily modify the upgrade-cluster.bash file later to test without a standby(or without mirrors) if we decide to.

marksliva: Changed it.

ci/scripts/upgrade-cluster.bash
marksliva: Will go with your suggestion.

PR #265, opened 2020-03-18
During finalize, in link mode remove standby and mirror data directories

bhuvnesh2703:
In Link mode, remove the standby and mirror data directories if configured.


idl/hub_to_agent.proto
```diff
@@ -37,8 +38,13 @@ message CreateSegmentDataDirRequest {
     repeated string datadirs = 1;
 }
 
+message DeleteDataDirectoriesRequest {
+  repeated string datadirs = 1;
+}
 message CreateSegmentDataDirReply {}
 
+message DeleteDataDirectoriesReply {}
+
```
kalensk: Should these references be renamed to just `DeleteDirectories` since the agent is agnostic of what the directories are (ie: it doesn't care that its a "data" directory)?

(This is similar to `RenameDirectories`.)

agent/delete_data_directories_test.go
```diff
@@ -0,0 +1,51 @@
...
+		StateDir: "",
+	})
+
+	req := &idl.DeleteDataDirectoriesRequest{Datadirs: []string{"/data/dbfast_mirror1/seg1"}}
+
+	t.Run("successfully deletes the data directory", func(t *testing.T) {
+		utils.System.RemoveAll = func(name string) error {
+			if name != req.Datadirs[0] {
+				t.Errorf("got %q want %q", name, req.Datadirs[0])
+			}
```
kalensk: Could a `refelect.DeepEqual(name, req.Datadirs)` on the whole slice instead of req.Datadirs[0]` be used?

hub/delete_data_directories.go
```diff
@@ -0,0 +1,60 @@
...
+
+func DeleteDataDirectories(agentConns []*Connection, cluster *utils.Cluster) error {
+	wg := sync.WaitGroup{}
+	errChan := make(chan error, len(agentConns))
+
+	for _, conn := range agentConns {
+		conn := conn
+
+		// Selects all mirrors and standby
+		mirrors := func(seg *utils.SegConfig) bool {
```
kalensk: nit: renaming to `mirrorsIncludingStandby` or similar would remove the need for the comment?

hub/rename_data_directories.go
```diff
@@ -16,15 +16,22 @@ import (
...
+	if s.Config.UseLinkMode {
+		if err := DeleteDataDirectories(s.agentConns, s.Source); err != nil {
+			return xerrors.Errorf("removing source cluster standby and mirror segment data directories: %w", err)
+		}
+	}
+
+	// if its --link mode exclude mirror and standby directories in the _old archiving
+	// as they have already been removed, else include them
+	if err := RenameSegmentDataDirs(s.agentConns, s.Source, "", OldSuffix,
+		s.Config.UseLinkMode /* rename primaries only*/); err != nil {
```
kalensk: nit: Maybe remove this comment (since its implied by the one above), or move it to be part of the function level comment. In this case, I find it hard to read inline like that.

hub/rename_data_directories.go
```diff
@@ -16,15 +16,22 @@ import (
...
 		return xerrors.Errorf("renaming source cluster master data directory: %w", err)
 	}
 
-	// Include mirror and standby directories in the _old archiving.
-	// TODO: in --link mode, shouldn't we be _removing_ the mirror and standby
-	// directories?
-	if err := RenameSegmentDataDirs(s.agentConns, s.Source, "", OldSuffix, false); err != nil {
+	// in --link mode, remove the mirror and standby data directories
+	if s.Config.UseLinkMode {
+		if err := DeleteDataDirectories(s.agentConns, s.Source); err != nil {
```
kalensk: Renaming to something like `DeleteMirrorAndStandbyDirectories` would remove the need for the comment.

We can easily change it if we find that the function needs to be more generic, and pass in the segments or directories to delete. But for now that does not seem to be needed.

hub/delete_data_directories_test.go
```diff
@@ -0,0 +1,111 @@
...
+		{ContentID: -1, DbID: 1, Port: 25431, Hostname: "standby", DataDir: "/data/standby", Role: utils.MirrorRole},
+		{ContentID: 0, DbID: 2, Port: 25432, Hostname: "sdw1", DataDir: "/data/dbfast1/seg1", Role: utils.PrimaryRole},
+		{ContentID: 1, DbID: 3, Port: 25433, Hostname: "sdw2", DataDir: "/data/dbfast2/seg2", Role: utils.PrimaryRole},
+		{ContentID: 2, DbID: 4, Port: 25434, Hostname: "sdw1", DataDir: "/data/dbfast1/seg3", Role: utils.PrimaryRole},
+		{ContentID: 3, DbID: 5, Port: 25435, Hostname: "sdw2", DataDir: "/data/dbfast2/seg4", Role: utils.PrimaryRole},
+		{ContentID: 0, DbID: 6, Port: 35432, Hostname: "sdw1", DataDir: "/data/dbfast_mirror1/seg1", Role: utils.MirrorRole},
+		{ContentID: 1, DbID: 7, Port: 35433, Hostname: "sdw2", DataDir: "/data/dbfast_mirror2/seg2", Role: utils.MirrorRole},
+		{ContentID: 2, DbID: 8, Port: 35434, Hostname: "sdw1", DataDir: "/data/dbfast_mirror1/seg3", Role: utils.MirrorRole},
+		{ContentID: 3, DbID: 9, Port: 35435, Hostname: "sdw2", DataDir: "/data/dbfast_mirror2/seg4", Role: utils.MirrorRole},
+	})
```
kalensk: Not for this PR, but I could see pulling this into a helper since this type of setup cluster could be useful for other tests.

hub/delete_data_directories_test.go
```diff
@@ -0,0 +1,111 @@
...
+	"golang.org/x/xerrors"
+	"testing"
+
+	"github.com/golang/mock/gomock"
+	"github.com/greenplum-db/gp-common-go-libs/testhelper"
+	"github.com/greenplum-db/gpupgrade/hub"
+	"github.com/greenplum-db/gpupgrade/idl"
+	"github.com/greenplum-db/gpupgrade/idl/mock_idl"
+	"github.com/greenplum-db/gpupgrade/utils"
+)
```
kalensk: nit: bottom section needs to be split. That is imports go: 
- stdlib
- 3rd party including gp-common-go-libs (I believe)
- gpupgrade specific

hub/delete_data_directories.go
```diff
@@ -0,0 +1,60 @@
+package hub
+
+import (
+	"github.com/hashicorp/go-multierror"
+	"sync"
+
+	"github.com/greenplum-db/gp-common-go-libs/gplog"
+	"github.com/greenplum-db/gpupgrade/idl"
+	"github.com/greenplum-db/gpupgrade/utils"
+	"golang.org/x/net/context"
```
kalensk: nit: imports

hub/delete_data_directories_test.go
```diff
@@ -0,0 +1,111 @@
...
+		).Return(&idl.DeleteDataDirectoriesReply{}, nil)
+
+		client4 := mock_idl.NewMockAgentClient(ctrl)
+		// NOTE: we expect no call to the master
+
+		agentConns := []*hub.Connection{
+			{nil, client1, "sdw1", nil},
+			{nil, client2, "sdw2", nil},
+			{nil, client3, "standby", nil},
+			{nil, client4, "master", nil},
```
kalensk: I know we don't do this in other places, but it seems like calling these `sdw1Client`, sdw2Client`, etc. to be more readable. Thoughts?

hub/finalize.go
```diff
@@ -43,7 +43,7 @@ func (s *Server) Finalize(_ *idl.FinalizeRequest, stream idl.CliToHub_FinalizeSe
 	})
 
 	st.Run(idl.Substep_FINALIZE_RENAME_DATA_DIRECTORIES, func(_ step.OutStreams) error {
-		return s.RenameDataDirectories()
+		return s.RenameOrDeleteDataDirectories()
```
kalensk: Part of me wants a different name than `RenameOrDeleteDataDirectories`, but I can't quite think of anything better.

hub/finalize.go
```diff
@@ -43,7 +43,7 @@ func (s *Server) Finalize(_ *idl.FinalizeRequest, stream idl.CliToHub_FinalizeSe
 	})
 
 	st.Run(idl.Substep_FINALIZE_RENAME_DATA_DIRECTORIES, func(_ step.OutStreams) error {
```
kalensk: Do we want to update `Substep_FINALIZE_RENAME_DATA_DIRECTORIES` and its step output/help text to reflect this change?

hub/delete_data_directories_test.go
jchampio: `goimports -local github.com/greenplum-db/gpupgrade`, in other words

hub/delete_data_directories_test.go
bhuvnesh2703: sure more readable.

agent/delete_data_directories_test.go
bhuvnesh2703: name is a simple string where req.Datadirs is an array..so unless i bring an array, can't use it as shared.

hub/delete_data_directories.go
bhuvnesh2703: made changes in all files accordingly now.

agent/delete_data_directories_test.go
kalensk: Sure makes sense. I just have a personal aversion to hard coding an index dereference (ie: [0]), so I'd prefer name to be a slice. But its not a blocker.

agent/delete_data_directories.go
```diff
@@ -0,0 +1,24 @@
...
+	"github.com/greenplum-db/gpupgrade/idl"
+	"github.com/greenplum-db/gpupgrade/utils"
+)
+
+func (s *Server) DeleteDirectories(ctx context.Context, in *idl.DeleteDirectoriesRequest) (*idl.DeleteDirectoriesReply, error) {
+	gplog.Info("got a request to delete data directories from the hub")
+
+	datadirs := in.Datadirs
+	for _, segDataDir := range datadirs {
+		err := utils.System.RemoveAll(segDataDir)
```
dkrieger-pvtl: I think we should add some safety here to make sure we are not deleting a "random directory" that the hub told us to...since all directories we are deleting are postgres data directories, we have a lot of options to choose from.  Perhaps checking for the two files `PG_VERSION` and `postgresql.conf`?

agent/delete_data_directories_test.go
```diff
@@ -0,0 +1,52 @@
...
+	testhelper.SetupTestLogger()
+
+	server := agent.NewServer(agent.Config{
+		Port:     -1,
+		StateDir: "",
+	})
+
+	req := &idl.DeleteDirectoriesRequest{Datadirs: []string{"/data/dbfast_mirror1/seg1"}}
+
+	t.Run("successfully deletes the data directory", func(t *testing.T) {
```
dkrieger-pvtl: Please add a test that passes multiple directories in the `req` and make sure all of them are visited by the `DeleteDataDirectories` function.

agent/delete_data_directories.go
kalensk: I am hesitant about this suggestion as it seems a bit of a "code smell" or "band-aid" to the real solution of designing a secure system as noted in @jchampio comment https://github.com/greenplum-db/gpupgrade/pull/265#pullrequestreview-377306444

jchampio: @kalensk even _with_ a secured channel, it's good practice to continue to decrease the functionality that's exposed. It reduces attack surface.

agent/delete_data_directories.go
```diff
@@ -0,0 +1,24 @@
...
+		if err != nil {
+			return &idl.DeleteDirectoriesReply{}, err
+		}
```
jchampio: nit: current best practice is to combine "parallel" operations into a `multierror.Error` as opposed to stopping on first failure. In this case, there is probably not as much of a practical benefit. What will matter more is idempotence.

Speaking of idempotence, if the directory has already been deleted, we should not error out.

agent/delete_data_directories.go
jchampio: (One way to think about this: authorized users of our system, if/when we implement authorization, have been explicitly trusted by the sysadmin and/or DBA to upgrade clusters. That does not imply that they have been trusted to delete random directories. The functionality we expose has the opportunity to create privilege escalations.)

dkrieger-pvtl: Another reason is defensive programming...the results of passing and hence removing the wrong directory is severe yet the suggested sanity check for it is lightweight and reasonably strong.

agent/delete_data_directories_test.go
```diff
@@ -0,0 +1,52 @@
...
+			if name != req.Datadirs[0] {
+				t.Errorf("got %q want %q", name, req.Datadirs[0])
+			}
+
+			return nil
+		}
+
+		_, err := server.DeleteDirectories(context.Background(), req)
+		if err != nil {
+			t.Errorf("unexpected error got %#v", err)
```
jchampio: nit: `%+v` for unexpected errors (it'll give us a detail trace if available). `%#v` for expected error comparison.

agent/delete_data_directories_test.go
jchampio: This also ties into @kalensk's request below.

hub/delete_data_directories.go
```diff
@@ -0,0 +1,61 @@
...
+	wg.Wait()
+	close(errChan)
+
+	var mErr *multierror.Error
+	for err := range errChan {
+		mErr = multierror.Append(mErr, err)
+	}
+
+	return mErr.ErrorOrNil()
+}
```
dkrieger-pvtl: nit: you can use the error channel to enforce the barrier synchronization between the caller and all the gofunc's used.

The current code uses a WaitGroup for barrier synchronization and a channel for errors.
```go
func myFunction(list []int) error {
	wg := sync.WaitGroup{}
	errChan := make(chan error, len(list))

	for _, val := range list {
		val := val
		wg.Add(1)
		go func(val int) {
			defer wg.Done()

			_, err := do_something(val)
			if err != nil {
				errChan <- err
			}
		}(val)
	}

	wg.Wait()
	close(errChan)

	var mErr *multierror.Error
	for err := range errChan {
	}

	return mErr.ErrorOrNil()
}
```

You can instead just use the channel for both.  
```go
func myFunction(list []int) error {
	errChan := make(chan error, len(list))

	for _, val := range list {
		val := val
		go func(val int) {
			_, err := do_something(val)
			errChan <- err
		}(val)
	}
	
	var mErr *multierror.Error
	for i:=0; i<len(list); i++ {
	}
	close(errChan)
	
	return mErr.ErrorOrNil()
}
```

hub/delete_data_directories.go
```diff
@@ -0,0 +1,61 @@
...
+	for _, conn := range agentConns {
+		conn := conn
+
+		mirrorsIncludingStandby := func(seg *utils.SegConfig) bool {
+			return seg.Hostname == conn.Hostname &&
+				(seg.Role == utils.MirrorRole)
+		}
+
+		wg.Add(1)
+		go func(c *Connection) {
```
jchampio: Passing `c` this way isn't needed; you've already pinned `conn` in the current iteration's scope on line 20.

hub/delete_data_directories.go
```diff
@@ -0,0 +1,61 @@
...
+			defer wg.Done()
+
+			segments := cluster.SelectSegments(mirrorsIncludingStandby)
+			if len(segments) == 0 {
+				// This can happen if there are no mirrors or standby on a host
+				return
+			}
```
jchampio: As Adam noted in a different PR, this can be moved outside the goroutine, so that we `continue` the loop instead of spawning a new goroutine that then realizes it doesn't need to do anything.

agent/delete_data_directories_test.go
jchampio: This also ties into @kalensk's request below.

agent/delete_data_directories.go
jchampio: (One way to think about this: authorized users of our system, if/when we implement authorization, have been explicitly trusted by the sysadmin and/or DBA to upgrade clusters. That does not imply that they have been trusted to delete random directories. The functionality we expose has the opportunity to create privilege escalations.)

jchampio:
(GitHub mixed up my review comments with my comments on other comments. Sorry for the mess if you're receiving notifications.)

hub/delete_data_directories.go
jchampio: Careful that you ensure the goroutine cannot panic in any case if you adopt this system; otherwise you'll deadlock if something keeps the unhandled panic from crashing the system. This has already happened with the Go test framework and `t.Fatalf()`.

berlin-ab: What mechanism makes this not true for the original? Is it the `defer wg.Done()` statement?

dkrieger-pvtl: Yes...the waitgroup is just counting starts/stops; if the goroutine panics before the errChan is written to, the calling function will never complete since the loop won't complete.

Placing the chan write in a defer should fix this:

```go
func myFunction(list []int)  error {
	errChan := make(chan error, len(list))

	for _, val := range list {
		val := val
		go func(val int) err error{
            defer func() {
                errChan <- err
            }
			_, err := do_something(val)
		}(val)
	}
	
	var mErr *multierror.Error
	for i:=0; i<len(list); i++ {
	}
	close(errChan)
	
	return mErr.ErrorOrNil()
}
```

idl/hub_to_agent.proto
dkrieger-pvtl: Note if we do [datadir sanity checks](https://github.com/greenplum-db/gpupgrade/pull/265/files#r395247588), then this should be named `DeleteDataDirectories`.

hub/delete_data_directories.go
bhuvnesh2703: for reference: we tried the below version but it hangs..
```
func DeleteMirrorAndStandbyDirectories(agentConns []*Connection, cluster *utils.Cluster) error {

		conn := conn

		mirrorsIncludingStandby := func(seg *utils.SegConfig) bool {
			return seg.Hostname == conn.Hostname &&
				(seg.Role == utils.MirrorRole)
		}

		segments := cluster.SelectSegments(mirrorsIncludingStandby)
		if len(segments) == 0 {
			// This can happen if there are no mirrors or standby on a host
			continue
		}

		go func() (err error) {
			defer func() {
				errChan <- err
			}()

			req := new(idl.DeleteDirectoriesRequest)
			for _, seg := range segments {
				datadir := seg.DataDir
				req.Datadirs = append(req.Datadirs, datadir)
			}

			_, err = conn.AgentClient.DeleteDirectories(context.Background(), req)
			if err != nil {
				gplog.Error("Error deleting segment data directories on host %s: %s",
					conn.Hostname, err.Error())
			}

			return err
		}()

	}

	var mErr *multierror.Error
	for i:= 0; i < len(agentConns); i++ {
		mErr = multierror.Append(mErr, <-errChan)
	}

	close(errChan)

	return mErr.ErrorOrNil()
}
```

jchampio: You're not guaranteed to push the exact number of errors you need due to the presence of `continue` in the loop body. Use `WaitGroup`; this is what it's for.

The takeaway for me is that the more complex your goroutine implementation gets, the nicer `WaitGroup` becomes as a way to manage the synchronization safely. IMO needing to defer `errChan <- err`, and this new deadlock you've found, is a sign that you should just be using `WaitGroup`.

agent/delete_data_directories.go
bhuvnesh2703: added check

bhuvnesh2703:
@dkrieger-pvtl applied the feedback requested.

agent/delete_data_directories.go
```diff
@@ -0,0 +1,46 @@
...
+	"path/filepath"
+
+	"github.com/greenplum-db/gp-common-go-libs/gplog"
+	"github.com/hashicorp/go-multierror"
+
+	"github.com/greenplum-db/gpupgrade/idl"
+	"github.com/greenplum-db/gpupgrade/utils"
+)
+
+var FilesToCheck = [...]string {"postgresql.conf", "PG_VERSION"}
```
marksliva: nit: This variable can be package private. Especially since you do not want it to change and nothing outside of the package ever references it.

agent/delete_data_directories.go
```diff
@@ -0,0 +1,46 @@
...
+
+func (s *Server) DeleteDirectories(ctx context.Context, in *idl.DeleteDirectoriesRequest) (*idl.DeleteDirectoriesReply, error) {
+	gplog.Info("got a request to delete data directories from the hub")
+
+	datadirs := in.Datadirs
```
marksliva: nit: This could be inlined.

hub/rename_data_directories.go
```diff
@@ -17,15 +17,20 @@ import (
...
 	if err := RenameMasterDataDir(s.Source.MasterDataDir(), true); err != nil {
 		return xerrors.Errorf("renaming source cluster master data directory: %w", err)
 	}
 
-	// Include mirror and standby directories in the _old archiving.
-	// TODO: in --link mode, shouldn't we be _removing_ the mirror and standby
-	// directories?
-	if err := RenameSegmentDataDirs(s.agentConns, s.Source, "", OldSuffix, false); err != nil {
+	// in --link mode, remove the mirror and standby data directories
+	if s.Config.UseLinkMode {
```
marksliva: There's currently no test for this conditional. We could test it in an integration bats test.

However there's no existing bats test to put this into.. why do we still not have a finalize test for a cluster upgraded using `link` mode?

bhuvnesh2703: added test in finalize with --link mode.

test/finalize.bats
```diff
@@ -106,6 +106,127 @@ teardown() {
     validate_mirrors_and_standby "${GPHOME}" "$(hostname)" "${PGPORT}"
 }
 
+@test "gpupgrade finalize with --link mode should swap the primary and master directory and delete the old mirror and standby directory" {
```
marksliva: I'm wondering if it would be worth it to combine the code of these 2 finalize tests. We could create a shared function that runs the upgrade and checks, and pass it an argument from each `@test` that specifies whether or not to run the upgrade using link mode, which would also conditionally backup and later check the source mirror directories since so much of the code is shared.

If we don't dedupe this now, it seems like it will be trickier to maintain when we want to make a change that affects both tests.

test/finalize.bats
```diff
@@ -106,6 +106,127 @@ teardown() {
...
+        cp -r ${parent_datadir} ${parent_datadir}_backup
+    done
+    gpstart -a
+
+    gpupgrade initialize \
+        --source-bindir="$GPHOME/bin" \
+        --target-bindir="$GPHOME/bin" \
+        --source-master-port="${PGPORT}" \
+        --disk-free-ratio 0 \
+        --link \
```
marksliva: This test looks good overall. I have added a follow-up story to add a link mode upgrade job for 5-to-6.

test/finalize.bats
dkrieger-pvtl: I agree strongly with this...the semantic difference between link mode and copy mode is small from upgrade's standpoint, and it seems we should be able to combine these in a straightforward way.  In addition, master's `finalize.bats` only tests if both mirrors and a standby are present in the source cluster.  We might want to test a mirrors-only or standby-only source cluster in the future.  If we don't dedup these now, that'll be more complicated in the future or we'll then have 4 tests...

berlin-ab: Copy and link have significant behavior differences that are important to gpupgrade. I think we should keep an end-to-end test for both and test their significant behavior that we care about.

bhuvnesh2703: have refactored to use a common function, with some ifs.. so pls review..

PR #266, opened 2020-03-18
validate mirrors and standby in BATS and in CI

dkrieger-pvtl:
Runs a cluster with mirrors+standby through its full paces:
* promotion of mirrors+standby to primaries+master
* readding mirrors+standby
* rebalancing the cluster
* at each relevant step, data tables are added and validated

There is a single test function, that errors out if it is called
to test a cluster that lacks either a mirror or a standby.

This runs in finalize.bats and in the CI GPDB-X->GPDB-Y tests.

NOTE TO REVIEWERS: it is probably easiest to review the test/finalize_checks.bash in side-by-side mode.

### DIFF
--- a/ci/scripts/upgrade-cluster.bash
+++ b/ci/scripts/upgrade-cluster.bash
@@ -121,7 +121,12 @@ if ! compare_dumps /tmp/old.sql /tmp/new.sql; then
 fi
 
 # Test that mirrors actually work
-echo 'Doing failover tests of mirrors...'
-check_mirror_validity "${GPHOME_NEW}" mdw $MASTER_PORT
+mirrorless=$(contents_without_mirror "${GPHOME_NEW}" mdw $MASTER_PORT)
+if [ -n "$mirrorless" ]; then
+    echo "skipping validate_mirrors_and_standby since these content ids do not have mirrors: ${mirrorless}"
+else
+    echo 'Doing failover tests of mirrors and standby...'
+    validate_mirrors_and_standby "${GPHOME_NEW}" mdw $MASTER_PORT
+fi
 
 echo 'Upgrade successful.'
--- a/test/finalize.bats
+++ b/test/finalize.bats
@@ -42,6 +42,11 @@ teardown() {
     # target cluster ends up with the source cluster's original layout
     local old_config=$(get_segment_configuration)
 
+    # set this variable before we upgrade to make sure our decision to run below
+    # is based on the source cluster before upgrade perhaps changes our cluster.
+    local no_mirrors
+    no_mirrors=$(contents_without_mirror "${GPHOME}" "$(hostname)" "${PGPORT}")
+
     gpupgrade initialize \
         --source-bindir="$GPHOME/bin" \
         --target-bindir="$GPHOME/bin" \
@@ -81,6 +86,15 @@ teardown() {
     local new_config=$(get_segment_configuration)
     [ "$old_config" = "$new_config" ] || fail "actual config: $new_config, wanted $old_config"
 
+    #
+    # The tests below depend on the source cluster having a standby and a full set of mirrors
+    #
+
+    if [ -n "$no_mirrors" ]; then
+        echo "skipping rest of this test since these content ids does not have a standby: ${no_mirrors}"
+        return 0
+    fi
+
     # TODO: Query gp_stat_replication to check if the standby is in sync.
     #   That is a more accurate representation if the standby is running and
     #   in sync, since gpstate might simply check if the process is running.
@@ -89,7 +103,7 @@ teardown() {
     local standby_status_line=$(get_standby_status "$actual_standby_status")
     [[ $standby_status_line == *"Standby host passive"* ]] || fail "expected standby to be up and in passive mode, got **** ${actual_standby_status} ****"
 
-    check_mirror_validity "${GPHOME}" "$(hostname)" "${PGPORT}"
+    validate_mirrors_and_standby "${GPHOME}" "$(hostname)" "${PGPORT}"
 }
 
 setup_state_dir() {
--- a/test/finalize_checks.bash
+++ b/test/finalize_checks.bash
@@ -1,23 +1,30 @@
-# This file provides a single high-level function check_mirror_validity()
-# that takes a cluster with mirrors "through its paces" to thoroughly test
-# the cluster's mirrors.  See the documentation of check_mirror_validity()
-# for details.
+# This file provides a single high-level function validate_mirrors_and_standby()
+# that takes a cluster with mirrors and a standby "through its paces" to
+# thoroughly test those mirrors and stadnby.
 
-check_mirrors() {
-    _check_segments_are_synchronized
-    _check_mirror_replication_connections
+check_mirrors_and_standby() {
+    local master_host=$1
+    local master_port=$2
+
+    _check_synchronized_cluster "${master_host}" "${master_port}"
+    _check_replication_connections "${master_host}" "${master_port}"
 }
 
-_check_segments_are_synchronized() {
+_check_synchronized_cluster() {
+    local master_host=$1
+    local master_port=$2
+
     for i in {1..10}; do
-        local unsynced=$(ssh -n "${MASTER_HOST}" "
+        local synced
+        synced=$(ssh -n "$master_host" "
             source ${GPHOME_NEW}/greenplum_path.sh
-            psql -p $MASTER_PORT -At -d postgres << EOF
+            psql -p $master_port -At -d postgres << EOF
                 SELECT gp_request_fts_probe_scan();
-                SELECT count(*) FROM gp_segment_configuration WHERE content <> -1 AND mode = 'n';
+                SELECT EVERY(state='streaming' AND state IS NOT NULL)
+                FROM gp_stat_replication;
 EOF
         " | tail -1)
-        if [ "$unsynced" = "0" ]; then
+        if [ "$synced" = "t" ]; then
             return 0
         fi
         sleep 5
@@ -27,15 +34,19 @@ EOF
     return 1
 }
 
-_check_mirror_replication_connections() {
-    local rows=$(ssh -n "${MASTER_HOST}" "
+_check_replication_connections() {
+    local host=$1
+    local port=$2
+
+    local rows
+    rows=$(ssh -n "${host}" "
         source ${GPHOME_NEW}/greenplum_path.sh
-        psql -p $MASTER_PORT -d postgres -AtF$'\t' -c \"
+        psql -p $port -d postgres -AtF$'\t' -c \"
             SELECT primaries.address, primaries.port, mirrors.hostname
             FROM gp_segment_configuration AS primaries
             JOIN gp_segment_configuration AS mirrors
             ON primaries.content = mirrors.content
-            WHERE primaries.role = 'p' AND mirrors.role = 'm' AND primaries.content != -1;
+            WHERE primaries.role = 'p' AND mirrors.role = 'm';
         \"
     ")
 
@@ -49,27 +60,10 @@ _check_mirror_replication_connections() {
     done
 }
 
-kill_primaries() {
-    local primaries
-    primaries=$(ssh -n "${MASTER_HOST}" "
-        source ${GPHOME_NEW}/greenplum_path.sh
-        psql -AtF$'\t' -p $MASTER_PORT -d postgres -c \"
-            SELECT hostname, port, datadir FROM gp_segment_configuration
-            WHERE content <> -1 AND role = 'p'
-        \"
-    ") || return $?
-
-    echo "${primaries}" | while read -r host port dir; do
-        ssh -n "${host}" "
-            source ${GPHOME_NEW}/greenplum_path.sh
-            pg_ctl stop -p $port -m immediate -D $dir -w
-        "
-    done
-}
-
 wait_can_start_transactions() {
     local host=$1
     local port=$2
+
     for i in {1..10}; do
         ssh -n "${host}" "
             source ${GPHOME_NEW}/greenplum_path.sh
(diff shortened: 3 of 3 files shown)


test/finalize_checks.bash
```diff
@@ -1,23 +1,30 @@
...
             source ${GPHOME_NEW}/greenplum_path.sh
-            psql -p $MASTER_PORT -At -d postgres << EOF
+            psql -p $master_port -At -d postgres << EOF
                 SELECT gp_request_fts_probe_scan();
-                SELECT count(*) FROM gp_segment_configuration WHERE content <> -1 AND mode = 'n';
+                SELECT count(*) FROM gp_stat_replication WHERE state = 'streaming';
 EOF
         " | tail -1)
-        if [ "$unsynced" = "0" ]; then
+        if [ "$synced" = $num_contents ]; then
```
kalensk: Would using an `every` remove the need for `num_contents`? 

That is, updating the query to: `SELECT every(sync_state = 'sync') FROM gp_stat_replication WHERE state = 'streaming';` and then update the if statement to match `t`.

test/finalize_checks.bash
```diff
@@ -88,102 +81,216 @@ EOF
...
     MASTER_HOST=$2
     MASTER_PORT=$3
 
-    local master_data_dir=$(ssh -n "${MASTER_HOST}" "
+    local mirrorless
+    mirrorless=$(contents_without_mirror "$GPHOME_NEW" "$MASTER_HOST" "$MASTER_PORT")
+    if [ -n "$mirrorless" ]; then
+        echo "This test only works on full clusters but these content ids do not have mirrors: ${mirrorless}"
+        exit 1
+    fi
```
kalensk: nit: would `haveMirrors` be a more descriptive name?

test/finalize_checks.bash
```diff
@@ -88,102 +81,216 @@ EOF
...
+    ")
 
-    # step 1
+    # step 1: initial
     wait_can_start_transactions $MASTER_HOST $MASTER_PORT
-    check_mirrors
+    check_mirrors_and_standby $MASTER_HOST $MASTER_PORT $num_contents
 
-    local on_upgraded_master_expected=$(create_table_with_name on_upgraded_master 50)
+    local on_upgraded_cluster_expected=$(create_table_with_name "${MASTER_HOST}" $MASTER_PORT on_upgraded_cluster 50)
```
jchampio: This doesn't read well to me; I don't know what `on_upgraded_cluster_expected` is supposed to contain. Maybe `previous_data` or `original_data` or ...?

test/finalize_checks.bash
```diff
@@ -88,102 +81,216 @@ EOF
...
+# For rebalancing the standby, we followed these instructions:
+# https://gpdb.docs.pivotal.io/6-4/admin_guide/highavail/topics/g-restoring-master-mirroring-after-a-recovery.html#topic17
+#
+# NOTE: when in a given step of this test, keep in mind that the master
+#  switches back and forth between the mdw host("MASTER") and the smdw host("standby").
+
+validate_mirrors_and_standby() {
     GPHOME_NEW=$1
     MASTER_HOST=$2
     MASTER_PORT=$3
```
jchampio: Are these globals needed anymore, or can they be made local?

test/finalize_checks.bash
```diff
@@ -88,102 +81,216 @@ EOF
...
+        source ${GPHOME_NEW}/greenplum_path.sh
+        export PGPORT=$standby_port
+        gpactivatestandby -a -d $standby_data_dir
+    "
+    wait_can_start_transactions $standby_host $standby_port
 
-    check_data_matches on_upgraded_master "${on_upgraded_master_expected}"
-    local on_promoted_mirrors_expected=$(create_table_with_name on_promoted_mirrors 60)
+    check_data_matches "${standby_host}" $standby_port on_upgraded_cluster "${on_upgraded_cluster_expected}"
+    local on_promoted_cluster_expected=$(create_table_with_name "${standby_host}" $standby_port on_promoted_cluster 60)
```
jchampio: Ditto.

test/finalize_checks.bash
```diff
@@ -88,102 +81,216 @@ EOF
...
+        export PGPORT=$standby_port
         gprecoverseg -a       # TODO..why is PGPORT not actually needed here?
     "
-    check_mirrors
+    wait_can_start_transactions $standby_host $standby_port  #TODO: is this necessary?
 
-    check_data_matches on_upgraded_master "${on_upgraded_master_expected}"
-    check_data_matches on_promoted_mirrors "${on_promoted_mirrors_expected}"
-    local on_recovered_cluster_expected=$(create_table_with_name on_recovered_cluster 70)
+    ssh -n "${MASTER_HOST}" "mv $master_data_dir $(mktemp -d)"
```
jchampio: Let's just `rm` the master directory if we're not going to clean up the temporary; otherwise we'll just fill up the disks on our dev machines.

test/finalize_checks.bash
dkrieger-pvtl: The main test function validates that every content has a mirror/standby before continuing, since the test only works on such a cluster.  This check makes sure that this is the case whenever it is called.  I tried this suggestion, but when I run it on a cluster with one mirror down(I ran `pg_ctl stop -D dir_mirror_0` and made sure that mirror showed `d` in `gp_segment_configuration`) I still get `t`:
```SQL
postgres=# select gp_segment_id, state from gp_stat_replication 
order by gp_segment_id;
 gp_segment_id |   state
---------------+-----------
            -1 | streaming
             0 |
             1 | streaming
             2 | streaming
(4 rows)

postgres=#   SELECT every(state = 'streaming') FROM gp_stat_replication;
 every
-------
 t
(1 row)
```  

I am not sure why but passing in contents explicitly works.

test/finalize_checks.bash
dkrieger-pvtl: I don't think so....`contents_without_mirrors` tests the opposite of `haveMirrors` and that function seems more natural to me returned contents that lack a mirror.

test/finalize_checks.bash
```diff
@@ -1,23 +1,30 @@
...
     for i in {1..10}; do
-        local unsynced=$(ssh -n "${MASTER_HOST}" "
+        local synced
+        synced=$(ssh -n $master_host "
             source ${GPHOME_NEW}/greenplum_path.sh
-            psql -p $MASTER_PORT -At -d postgres << EOF
+            psql -p $master_port -At -d postgres << EOF
                 SELECT gp_request_fts_probe_scan();
-                SELECT count(*) FROM gp_segment_configuration WHERE content <> -1 AND mode = 'n';
+                SELECT count(*) FROM gp_stat_replication WHERE state = 'streaming';
```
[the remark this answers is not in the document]
dkrieger-pvtl: `gp_segment_configuration` [always shows `s`](https://gpdb.docs.pivotal.io/6-4/ref_guide/system_catalogs/gp_segment_configuration.html) for `mode` in GPDB-6.  Since `gp_stat_replication.state` works on both standby and mirrors, I used it instead.

test/finalize_checks.bash
kalensk: I am not fully following. Doesn't the sync_state show something other than sync if the mirror is down? If so, `every` should work just fine.

dkrieger-pvtl: According to the [documentation](https://gpdb.docs.pivotal.io/6-4/ref_guide/system_catalogs/gp_stat_replication.html), `sync_state` is always `sync`, though a down mirror has it as `""`.  Note that `sync_error` appears to always be `none` even for a down mirror.  Apparently, the official way to check is `state='streaming'`, but somehow the `every(state = 'streaming')` SQL above still returns `t` with a `state=""`.

kalensk: Looks like sync_state turns into NULL when not synced. Thus, the query should be something like:

`SELECT every(sync_state = 'sync' and sync_state IS NOT NULL)  FROM gp_stat_replication  ;`

test/finalize_checks.bash
kalensk: Then `noMirrors`? Where if statement reads:
```
    if [ -n "$noMirrors" ]; then

```

test/finalize_checks.bash
dkrieger-pvtl: A down segment has NULL as its state, and somehow `every(state='streaming')` accepts that.

kalensk: Careful with `!=NULL` as its not the same as `IS NOT NULL`! (I don't think its giving you what you think its giving you. I'll see if I can dig up some references.)

dkrieger-pvtl: `select every(state='streaming' AND state IS NOT NULL) from gp_stat_replication;`  works as expected.  Thanks.

dkrieger-pvtl: I also double-checked with the server team that NULL is in fact returned for a content with a down mirror/standby.

test/finalize_checks.bash
```diff
@@ -1,23 +1,30 @@
...
-    _check_mirror_replication_connections
+# This file provides a single high-level function validate_mirrors_and_standby()
+# that takes a cluster with mirrors and a standby "through its paces" to
+# thoroughly test those mirrors and stadnby.
+
+check_mirrors_and_standby() {
+    local master_host=$1
+    local master_port=$2
+    local contents=$3
+    _check_synchronized_cluster $master_host $master_port $contents
```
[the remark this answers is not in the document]
dkrieger-pvtl: added here and other places in the code that lacked this.

test/finalize_checks.bash
```diff
@@ -88,102 +81,216 @@ EOF
     return 1
 }
 
+kill_contents() {
```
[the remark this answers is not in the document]
dkrieger-pvtl: I rearranged te

test/finalize_checks.bash
dkrieger-pvtl: `GPHOME_NEW` is used throughout the file, but I made the other two `local`.

test/finalize_checks.bash
dkrieger-pvtl: Done.

ci/scripts/upgrade-cluster.bash
```diff
@@ -121,7 +121,12 @@ if ! compare_dumps /tmp/old.sql /tmp/new.sql; then
 fi
 
 # Test that mirrors actually work
```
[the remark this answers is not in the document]
dkrieger-pvtl: We are updating the CI pipeline shortly and that includes knowledge of whether there are mirrors in the cluster(similarly, we could add a standby flag to the CI as well that is passed to this script).  But for now, I'd prefer to leave this as is, and refactor at a later date.

test/finalize_checks.bash
```diff
@@ -88,102 +81,216 @@ EOF
...
+    local master_info=$(ssh -n "${MASTER_HOST}" "
         source ${GPHOME_NEW}/greenplum_path.sh
-        psql -p $MASTER_PORT -t -A -d postgres -c \"
-            SELECT datadir FROM gp_segment_configuration
+        psql -p $MASTER_PORT -AtF$'\t' -d postgres -c \"
+            SELECT address, datadir FROM gp_segment_configuration
             WHERE content = -1 AND role = 'p'
         \"
     ")
+    read -r master_address master_data_dir <<<$(echo "${master_info}")
```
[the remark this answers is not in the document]
dkrieger-pvtl: These might be slightly more readable but I prefer the bulk of the code to have the same idiom since I think it's easier to maintain and we know the error return works:
```
local result
result=$(command)
```

test/finalize_checks.bash
dkrieger-pvtl: I renamed all the `*_expected` to `data_*` so the table creation reads like a sentence.

test/finalize_checks.bash
dkrieger-pvtl: good idea...I added a check as well to make sure the dir we are deleting is at least "reasonable".

test/finalize_checks.bash
dkrieger-pvtl: Done.

test/finalize_checks.bash
dkrieger-pvtl: Done...I added in sanity checks to make sure the directory to be removed at least looks somehow reasonable.

test/finalize_checks.bash
```diff
@@ -88,102 +82,224 @@ EOF
...
+    ssh -n "${standby_host}" "
+        source ${GPHOME_NEW}/greenplum_path.sh
+        export PGPORT=$standby_port; gpinitstandby -a -s $MASTER_HOST -P $MASTER_PORT -S $master_data_dir
+    "
+    check_mirrors_and_standby "${standby_host}" "${standby_port}"
+
+    check_data_matches "${standby_host}" "${standby_port}" on_upgraded_cluster "${data_on_upgraded_cluster}"
+    check_data_matches "${standby_host}" "${standby_port}" on_promoted_cluster "${data_on_promoted_cluster}"
+    local data_on_unbalanced_cluster
+    data_on_unbalanced_cluster=$(create_table_with_name "${standby_host}" "${standby_port}" on_unbalanced_cluster 70)
```
jchampio: last nit: it'd be nice if `create_table_with_name()` took the name as the first parameter; it's easier for me to infer the semantics that way.

dkrieger-pvtl: Done....I checked `check_data_matches` as well.
```
   check_data_matches on_upgraded_cluster "${data_on_upgraded_cluster}" 
                  "${standby_host}" "${standby_port}"

   data_on_promoted_cluster=$(create_table_with_name on_promoted_cluster 60
                 "${standby_host}" "${standby_port}")
```

test/finalize.bats
```diff
@@ -81,6 +81,17 @@ teardown() {
     local new_config=$(get_segment_configuration)
     [ "$old_config" = "$new_config" ] || fail "actual config: $new_config, wanted $old_config"
 
+    #
+    # The tests below depend on the source cluster having a standby and a full set of mirrors
+    #
+
+    local mirrorless
```
marksliva: For reasons discussed in #264, `mirrorless` assignment should occur at the beginning of this function based on the `source` cluster, before we run the upgrade.

marksliva: Although I now see that the check above this does enforce that both clusters are equivalent at the gp_segment_configuration level. So an argument could be made either way I suppose.
```
    # Check to make sure the new cluster matches the old one.
    local new_config=$(get_segment_configuration)
    [ "$old_config" = "$new_config" ] || fail "actual config: $new_config, wanted $old_config"
```

dkrieger-pvtl: Good point.  Done.

dkrieger-pvtl: I like placing the assignment before gpupgrade is run as then the mirrorless check is not dependent on the positioning of validation that the configs are the same....

PR #268, opened 2020-03-19
Push response to the client into the step subsystem

berlin-ab:
Helps response become a pseudo-step but not report to the UI via a new idl.Substep. We gain access to the step's MessageSender, and we also get to use the step's error tracking system.

### DIFF
--- a/step/step.go
+++ b/step/step.go
@@ -93,52 +93,63 @@ func (s *Step) Err() error {
 }
 
 func (s *Step) AlwaysRun(substep idl.Substep, f func(OutStreams) error) {
-	s.run(substep, f, true)
+	err := s.run(substep, f, true)
+
+	if err != nil {
+		s.err = xerrors.Errorf(`substep "%s": %w`, s.name, err)
+	}
 }
 
 func (s *Step) Run(substep idl.Substep, f func(OutStreams) error) {
-	s.run(substep, f, false)
+	err := s.run(substep, f, false)
+
+	if err != nil {
+		s.err = xerrors.Errorf(`substep "%s": %w`, s.name, err)
+	}
 }
 
-func (s *Step) run(substep idl.Substep, f func(OutStreams) error, alwaysRun bool) {
+func (s *Step) SendResponse(action func(sender idl.MessageSender) error) {
+	err := action(s.sender)
+
+	if err != nil {
+		s.err = xerrors.Errorf(`substep "%s": %w`, s.name, err)
+	}
+}
+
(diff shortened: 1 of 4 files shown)


step/step.go
```diff
@@ -172,3 +174,15 @@ func (s *Step) sendStatus(substep idl.Substep, status idl.Status) {
 		}},
 	})
 }
+
+func (s *Step) SetError(err error) {
```
berlin-ab: This doesn't need to be public.

step/step.go
```diff
@@ -103,9 +103,11 @@ func (s *Step) Run(substep idl.Substep, f func(OutStreams) error) {
 func (s *Step) run(substep idl.Substep, f func(OutStreams) error, alwaysRun bool) {
 	var err error
 	defer func() {
-		if err != nil {
-			s.err = xerrors.Errorf(`substep "%s": %w`, s.name, err)
```
jchampio: This change isn't equivalent. `xerrors` adds stack frames at point of call for the `%+v` format to unpack. By moving the wrap to a completely different helper, you remove the utility of the stack frame and it becomes more difficult to figure out where things came from, in the event that we need to rely on the line number to find the cause of the error.

berlin-ab: Wouldn't it be just one step up the stack?

jchampio: Yes. Maybe an example helps:
```go
package main

import (
    "errors"
    "fmt"

    "golang.org/x/xerrors"
)

type Thing struct{}

func (t Thing) WithDefer() (err error) {
    defer func() {
        if err != nil {
            err = xerrors.Errorf("helpful trace: %w", err)
        }
    }()

    return t.Break()
}

func (t Thing) WithSetError() error {
    return t.setError(t.Break())
}

func (t Thing) Break() error {
    return errors.New("it broke")
}

func (t Thing) setError(err error) error {
    return xerrors.Errorf("unhelpful trace: %w", err)
}

func main() {
    var t Thing

    fmt.Printf("%+v\n\n", t.WithDefer())
    fmt.Printf("%+v\n", t.WithSetError())
}
```

```
$ go1.14 run main.go
helpful trace:
    main.Thing.WithDefer.func1  <- ah, WithDefer failed
        /tmp/main.go:15
  - it broke

unhelpful trace:
    main.Thing.setError         <- there's no context
        /tmp/main.go:31
  - it broke
```

berlin-ab: Pushed a new change. I think it still provides the context you're looking for.

PR #271, opened 2020-03-19
Move general interface for Outstreams to utils package.

berlin-ab:
I realized while working on another PR that the utils package depended on
the step package for OutStreams, which seems like a strange dependency.

### DIFF
--- a/greenplum/cluster.go
+++ b/greenplum/cluster.go
@@ -8,7 +8,7 @@ import (
 	"github.com/pkg/errors"
 	"golang.org/x/xerrors"
 
-	"github.com/greenplum-db/gpupgrade/step"
+	"github.com/greenplum-db/gpupgrade/utils"
 )
 
 var isPostmasterRunningCmd = exec.Command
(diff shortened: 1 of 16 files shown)


jchampio:
I've been trying to move us away from the `utils` megadependency; I think this pushes things in the wrong direction.

berlin-ab:
> I've been trying to move us away from the `utils` megadependency; I think this pushes things in the wrong direction.

I understand and agree with you, but `utils` is a better spot than `step`.

jchampio:
> utils is a better spot than step

Why?

berlin-ab:
Currently:

hub depends on step (specfic to specific)
utils depends on step (general to specific)

We want to depend in from specific to general.

New:

hub -> utils
step -> utils

berlin-ab:
Can you think of a better spot to put it?

berlin-ab:
Also, currently the Cluster knows about `step` for start/stop, which doesn't seem like something it should know about.

berlin-ab:
Alternatively, we could duplicate the Outstreams into the `greenplum` package so there's not a dependency structure like this:

```
greenplum -> step
hub -> step
```

it would be more like:

```
hub -> greenplum (OutStreams)
hub -> step (OutStreams)
```

They can evolve independently, as long as hub ensures that it passes the necessary stream type to the `greenplumRunner`.

jchampio:
> Currently:
> 
> hub depends on step (specfic to specific)
> utils depends on step (general to specific)

`utils.Cluster` depends on `step`, but we already know `Cluster` doesn't belong in `utils` anyway. Having a `greenplum` -> `step` dependency is less problematic to me. Not clean, but not incredibly dirty as long as the dependency doesn't grow.

> Alternatively, we could duplicate the Outstreams into the greenplum package so there's not a dependency structure like this:

I don't think the `step` dependency is a problem in practice, but it's definitely a smell. If we want to remove the `greenplum` -> `step` dependency, this move gets my vote for now.

jchampio:
(Long-term -- and I do mean long-term; I wouldn't want to do it now -- it seems like `OutStreams` is an `io` concept. Do we have other such input/output concerns in other places?)

berlin-ab:
@jchampio https://github.com/greenplum-db/gpupgrade/pull/272

berlin-ab:
Closing in favor of alternate design.

PR #272, opened 2020-03-19
Break dependency between greenplum and step packages

berlin-ab:
Duplicates the OutputStreams type. This should let them grow independenly if the need to, and reduces
the chance that changes to the step package have downstream effects on the Greenplum package. Now,
only the hub package will need to respond to changes in the step design.

This PR has been opened in favor over https://github.com/greenplum-db/gpupgrade/pull/271

### DIFF
--- a/hub/greenplum_runner.go
+++ b/hub/greenplum_runner.go
@@ -2,14 +2,18 @@ package hub
 
 import (
 	"fmt"
+	"io"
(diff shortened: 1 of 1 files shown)


berlin-ab:
err, I'll wait on this one until the GreenplumRunner PR is in.

PR #274, opened 2020-03-20
Reduce threading of streams and cluster data through Start/Stop.

berlin-ab:
- Introduces a facade on Cluster to pass through to GpStart and GpStop utilities
  that is responsible for start/stop operations. 

- Replaces startStopCommand with Runner system.

```
// for example:

target.GpStart(streams).Start()
source.GpStop(streams).Stop()

```

### DIFF
--- a/greenplum/cluster.go
+++ b/greenplum/cluster.go
@@ -271,23 +271,30 @@ func (c *Cluster) MasterPidFile() string {
 // Cluster Manager facade
 //
 func (c *Cluster) Start(stream OutStreams) error {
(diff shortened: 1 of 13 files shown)


berlin-ab:
To @kalensk 's question offline: "What's the value of reducing threading of streams"

Streams are really only needed by the lowest layer, the system utilities that we call out to. If we can reduce the need to pass streams through every single layer to get down to the bottom,  it can simplify all of the layers in the middle.

Cluster doesn't necessarily need to know about Input/output streams, but the thing that starts and stops the cluster does need to know about them.

PR #276, opened 2020-03-20
Move agent connection and server start/stop into hub/agent package.

berlin-ab:
Move agent connection plus agent start stop logic into an `agent` package within hub.

The interesting work here is in `server.go` where we've moved all of the Agent work into an `agentClient` struct on the Server object. `AgentConns()` still remains, but the rest of the server methods for managing agent connections and agent start/stop now belong to the `agent.Client`.

This is the public interface to an agent client after the move.
```
package agent

type Client struct {
...
}

type Connection struct {
...
}

type Dialer func (...) 

func NewClient(dialer Dialer) *Client

func (c *Client) Connect(hostnames []string, port int) error

func (c *Client) Connections() []*Connection

func (c *Client) StopAllAgents() error

func (c *Client) CloseConnections() 

func (c *Client) RestartAllAgents(hostnames []string, port int, stateDir string)
```

### DIFF
--- a/hub/agent/agentclient.go
+++ b/hub/agent/agentclient.go
@@ -3,7 +3,6 @@ package agent
 import (
 	"context"
 	"fmt"
-	"net"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -35,9 +34,9 @@ func NewClient(dialer Dialer) *Client {
 	}
 }
 
-func (m *Client) Connect(hostnames []string, port int) error {
-	if m.connections != nil {
-		if err := m.ensureConnectionsAreReady(); err != nil {
+func (c *Client) Connect(hostnames []string, port int) error {
+	if c.connections != nil {
+		if err := c.ensureConnectionsAreReady(); err != nil {
 			gplog.Error("ensureConnsAreReady failed: %s", err)
 			return err
 		}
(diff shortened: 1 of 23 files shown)


kalensk:
Seeing `/hub/agent` is super duper confusing to me since it seems to be conflating the gpupgrade hub and agent processes with the gRPC client and server model. The hub has both a gRPC server and client.

```
[ gpupgrade cli ]          [gpupgrade hub]          [gpupgrade agent] 
    gRPC client     <-->      gRPC server
                              gRPC client     <-->      gRPC server
```

berlin-ab:
Good point. One thing I'm struggling with is understanding the current system. Hopefully we can improve the design such that it is not "super duper confusing" to either one of us.

berlin-ab:
The `agent` package created here currently represents the work that the `hub` needs to do to communicate with `agents`.   So the `hub gRPC client <-> agent gRPC server` portion of your diagram.

berlin-ab:
Can you think of a better name for that concept?

kalensk:
On a very quick glance perhaps teasing apart or making it more clear the gRPC client and gRPC server partitions inside the hub? Perhaps the following might help...

That is, the server portion is mostly `hub.Start()` where it creates a new server and listens and then accepts class in like initialize.go and execute.go.

The client portions are mostly the `hub.Connection` struct and the fields `agentConns` and `grpcDialer` on the hub struct. These are used in the following main functions `AgentConns()`, `closeAgentConns()`, `stopAgents()`. They are also used in initialize.go and execute.go when calling out to the gpupgrade agent processes.

Thus, initialize.go and execute.go have qualities of _both_ the client and server.

berlin-ab:
I think that's the teasing apart that I've done. Perhaps we can take it further.

kalensk:
Maybe change `hub/agent` package to `hub/client`?

berlin-ab:
> Maybe change `hub/agent` package to `hub/client`?

I'll give it a shot.

berlin-ab:
@kalensk I stuck with `agent` as the package name because `agent.Client` and `agent.Connection` ended up being the exposed public types, and I think that reads well.

PR #277, opened 2020-03-20
fixup: validate mirrors and standby in BATS and CI

dkrieger-pvtl:
add a stronger directory check before deleting a local demo
cluster to prevent accidental deletion of the wrong directory.

This passed in CI, and finalize.bats ran successfully on my machine on a demo cluster.


jchampio:
This'll break my workflow, unfortunately. I put my gpdemo under `/tmp` by using the `MASTER_DATADIR` feature of the demo-cluster scripts.

jchampio:
You could keep `datadirs` I think without loss of generality.

dkrieger-pvtl:
> You could keep `datadirs` I think without loss of generality. But `gpAux/gpdemo` is not guaranteed.

@jchampio how about now?  I just added `datadirs`....

PR #279, opened 2020-03-21
Migrate server_test.go to Go style test

kalensk:
Tried to keep test semantics the same and simply port over the tests.

### DIFF
--- a/hub/server.go
+++ b/hub/server.go
@@ -308,7 +308,7 @@ func (s *Server) AgentConns() ([]*Connection, error) {
 			host+":"+strconv.Itoa(s.AgentPort),
 			grpc.WithInsecure(), grpc.WithBlock())
 		if err != nil {
-			err = errors.Errorf("grpcDialer failed: %s", err.Error())
+			err = xerrors.Errorf("grpcDialer failed: %w", err)
 			gplog.Error(err.Error())
 			cancelFunc()
 			return nil, err
--- a/hub/server_test.go
+++ b/hub/server_test.go
@@ -6,10 +6,12 @@ import (
 	"os"
 	"reflect"
 	"sort"
-	"strconv"
+	"strings"
 	"testing"
+	"time"
 
-	multierror "github.com/hashicorp/go-multierror"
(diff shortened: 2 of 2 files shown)


hub/server.go
```diff
@@ -37,6 +37,9 @@ var ErrHubStopped = errors.New("hub is stopped")
 
 type Dialer func(ctx context.Context, target string, opts ...grpc.DialOption) (*grpc.ClientConn, error)
 
+const ServerPort = 7527
```
berlin-ab: Is it necessary for the tests to have the exact correct ports? Would stand-in values be sufficient?

kalensk: If at all possible, I'd really like to keep the semantics the same (or close), otherwise this PR could quickly spiral out of scope. Followup PRs to find a freely available port could definitely be done.

kalensk: I ended up removing this change and made the tests more robust by using `getOpenPort(t)` since the tests would fail if a spurious hub or agent process was running. This was because the test calls `h.Star()` which was also trying to listen on the same port as the hub or agent process.

hub/server_test.go
```diff
@@ -36,169 +35,315 @@ func (m *msgStream) Send(msg *idl.Message) error {
...
+		}
+		if !isReached {
+			t.Error("expected connectivity state to be reached")
+		}
+	}
+}
+
+func doesStateEventuallyReach(conn *grpc.ClientConn, state connectivity.State) (bool, error) {
+	startTime := time.Now()
+	timeout := 3 * time.Second
```
bhuvnesh2703, line 321: is this good enough timeout?

bhuvnesh2703:
requires conflict resolution..

hub/server_test.go
kalensk: I think so. Let me know if you have any other suggestions. And of course we can always change it if ends up causing any issues. For reference, gomega's defaults for `Eventually` polling are [here](https://github.com/onsi/gomega/blob/6024f5bbe35dee8d348ec498fd7eb394236a81c2/gomega_dsl.go#L37-L40), with a timeout of 1 second and a poll of 10 ms.

PR #280, opened 2020-03-23
WIP: Move config to state package

berlin-ab:
Introduces state package and State struct to hold onto the StateDir and Config.

Again, most of the value here is to pull responsibilities out of `server.go` and the `Server` object and to create a state/config concept that can be though of independently of the rest of the system.

It should simplify external callers to something like:

```
server.State.Save()
```

instead of 

```
hub.SaveConfig(server.StateDir, server.Config)
```
Eventually, I'd like to be able to hide all of Config's public functions and only use `State.Save()`, but there's more work to do to get there.e

It also moves 
```
filepath.Join(tempDir, ConfigFileName)
```

behind an interface that looks like:

```
state.GetConfigFilepath(tempDIr)
```


berlin-ab:
I'm going to attempt to do this a little slower. It looks like I broke the build along the way.

berlin-ab:
> I'm going to attempt to do this a little slower. It looks like I broke the build along the way.

I mistakenly thought this branch was broken due to tracking GPDB 6X_STABLE.

hub/update_catalog.go
```diff
@@ -174,21 +172,8 @@ func (s *Server) UpdateGpSegmentConfiguration(db *sql.DB) (err error) {
...
-			}
-
-			// TODO: this is out of sync now, as the standby/mirrors are added later.
-			//   replace with one without standby/mirrors
-			s.Target = origConf.Source
-			s.Target.BinDir = origConf.Target.BinDir
-			s.Target.Version = origConf.Target.Version
-
-			err = s.SaveConfig()
+
```
berlin-ab: This is likely an overreach. I'm waiting to see if a test catches me, but I think that we intend to not change the existing Config *except* for the three fields below.

cli/commanders/initialize.go
```diff
@@ -28,45 +26,57 @@ func CreateStateDir() (err error) {
 	stateDir := utils.GetStateDir()
 	err = os.Mkdir(stateDir, 0700)
 	if os.IsExist(err) {
-		gplog.Debug("State directory %s already present...skipping", stateDir)
+		gplog.Debug("Config directory %s already present...skipping", stateDir)
```
berlin-ab: Editor refactoring overreach.

hub/state/state.go
```diff
@@ -0,0 +1,15 @@
+package state
+
+type State struct {
+	StateDir string
+	*Config
+}
+
+func (s *State) Save() error {
```
berlin-ab: I plan to inline these private functions.

hub/update_catalog.go
dkrieger-pvtl: We aren't sure the best way to handle the complexity of the original source cluster, the planned target cluster with temporary directories for the target cluster(ie the `_upgrade` datadirs), and the current state of the target cluster during finalize.  Right now, there's a lot the developer has to keep in their head.

PR #281, opened 2020-03-23
Ignore .psqlrc during bats test.

berlin-ab:
psql currently reads the `~/.psqlrc` file which gives me red output while running the bats tests.

### DIFF
--- a/test/finalize.bats
+++ b/test/finalize.bats
@@ -6,7 +6,7 @@ load finalize_checks
 setup() {
     skip_if_no_gpdb
 
-    PSQL="$GPHOME/bin/psql --no-align --tuples-only postgres"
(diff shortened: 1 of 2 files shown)


bhuvnesh2703:
Good change.. Do we need it at other places as well..
like for instance or not needed.
```
get_datadirs() {
    $PSQL -Atc "select datadir from gp_segment_configuration"
}
```
If not LGTM.

berlin-ab:
> Good change.. Do we need it at other places as well..
> like for instance below or not needed.
> 
> ```
> get_datadirs() {
>     $PSQL -Atc "select datadir from gp_segment_configuration"
> }
> ```
> 
> where `PSQL="$GPHOME/bin/psql --no-align --tuples-only postgres"`
> or we should use something like this?

Nice catch. I've added that.

PR #286, opened 2020-03-24
Remove unused code.

berlin-ab:
integrations/integrations_suite_test.go:34:2: `hubToAgentPort` is unused (varcheck)
	hubToAgentPort = 6416
	^

### DIFF
--- a/integrations/integrations_suite_test.go
+++ b/integrations/integrations_suite_test.go
@@ -30,8 +30,7 @@ var (
 )
 
 const (
-	cliToHubPort   = 7527
-	hubToAgentPort = 6416
+	cliToHubPort = 7527
 )
 
 var _ = BeforeSuite(func() {


integrations/integrations_suite_test.go
```diff
@@ -30,8 +30,7 @@ var (
 )
 
 const (
-	cliToHubPort   = 7527
-	hubToAgentPort = 6416
+	cliToHubPort = 7527
```
marksliva: This is a little confusing.. the integration tests fail when running `make integration` on this branch. But I see that the `fix-linter-1` PR removes the code. Why not just move this commit over to that PR and close this one?

berlin-ab: That is confusing. I'll investigate.

All as one package of fixups, it worked just fine, but not all split up.

PR #288, opened 2020-03-24
Unique upgrade directories 

dkrieger-pvtl:
This DRAFT PR is a draft PR in the true sense...the code is a bit rough, but shows the essential ideas.  Please take a look at the big ideas shown in the code...we will continue working on make it higher quality, and fixing the tests.

The main idea is that upgraded datadirs now have a unique has as a suffix, NOT _upgrade as the parent dir suffix.  So, for instance,
```
qddir/demoDataDir-1  ---> qddir/demoDataDir-1_StaMaLJlyh0-1
dbfast/demoDataDir2 ---> dbfast/demoDataDir2_StaMaLJlyh0
```
This scheme ends up significantly cleaning up the code, since now we centralize upgrade datadir naming in one place in Initialize and propagate that through the code.  It also solves the motivating bug: to allow multiple datadirs under the same parent directory.

This code also allows "ownership" of directories, as the hash(`StaMaLJlyh0` in the above) is stored in config.json and allows an upgrade to know "its directories". 

This code passes all the CI's upgrade `X->Y` jobs, but not the BATS tests, as those hard-code assumptions about the upgrade datadir names.

### DIFF
--- a/hub/fill_cluster_configs.go
+++ b/hub/fill_cluster_configs.go
@@ -6,6 +6,8 @@ import (
 	"path/filepath"
 	"sort"
 
+	"github.com/greenplum-db/gpupgrade/upgrade"
+
 	"github.com/pkg/errors"
 
 	"github.com/greenplum-db/gpupgrade/db"
@@ -15,7 +17,10 @@ import (
 )
 
 // create source/target clusters, write to disk and re-read from disk to make sure it is "durable"
-func (s *Server) FillClusterConfigsSubStep(config *Config, conn *sql.DB, _ step.OutStreams, request *idl.InitializeRequest, saveConfig func() error) error {
+func FillClusterConfigsSubStep(config *Config, conn *sql.DB, _ step.OutStreams, request *idl.InitializeRequest, saveConfig func() error) error {
+	// Assign a new universal upgrade identifier.
+	config.UpgradeID = upgrade.NewID()
+
 	if err := CheckSourceClusterConfiguration(conn); err != nil {
 		return err
 	}
@@ -37,7 +42,7 @@ func (s *Server) FillClusterConfigsSubStep(config *Config, conn *sql.DB, _ step.
 		ports = append(ports, int(p))
 	}
 
-	s.TargetInitializeConfig, err = AssignDatadirsAndPorts(s.Source, ports)
+	config.TargetInitializeConfig, err = AssignDatadirsAndPorts(config.Source, ports, config.UpgradeID)
 	if err != nil {
 		return err
 	}
@@ -49,7 +54,7 @@ func (s *Server) FillClusterConfigsSubStep(config *Config, conn *sql.DB, _ step.
 	return nil
 }
 
-func AssignDatadirsAndPorts(source *greenplum.Cluster, ports []int) (InitializeConfig, error) {
+func AssignDatadirsAndPorts(source *greenplum.Cluster, ports []int, upgradeID upgrade.ID) (InitializeConfig, error) {
 	if len(ports) == 0 {
 		port := 50432
 		numberOfSegments := len(source.Mirrors) + len(source.Primaries) + 2 // +2 for master/standby
@@ -65,11 +70,11 @@ func AssignDatadirsAndPorts(source *greenplum.Cluster, ports []int) (InitializeC
 		ports = sanitize(ports)
 	}
 
-	return assignDatadirsAndCustomPorts(source, ports)
+	return assignDatadirsAndCustomPorts(source, ports, upgradeID)
 }
 
 // can return an error if we run out of ports to use
-func assignDatadirsAndCustomPorts(source *greenplum.Cluster, ports []int) (InitializeConfig, error) {
+func assignDatadirsAndCustomPorts(source *greenplum.Cluster, ports []int, upgradeID upgrade.ID) (InitializeConfig, error) {
 	targetInitializeConfig := InitializeConfig{}
 
 	nextPortIndex := 0
@@ -80,7 +85,7 @@ func assignDatadirsAndCustomPorts(source *greenplum.Cluster, ports []int) (Initi
 			return InitializeConfig{}, errors.New("not enough ports")
 		}
 		master.Port = ports[nextPortIndex]
-		master.DataDir = upgradeDataDir(master.DataDir)
+		master.DataDir = upgradeDataDirMaster(master.DataDir, upgradeID)
 		targetInitializeConfig.Master = master
 		nextPortIndex++
 	}
@@ -91,7 +96,7 @@ func assignDatadirsAndCustomPorts(source *greenplum.Cluster, ports []int) (Initi
 			return InitializeConfig{}, errors.New("not enough ports")
 		}
 		standby.Port = ports[nextPortIndex]
-		standby.DataDir = standby.DataDir + "_upgrade"
+		standby.DataDir = upgradeDataDir(standby.DataDir, upgradeID)
 		targetInitializeConfig.Standby = standby
 		nextPortIndex++
 	}
@@ -119,7 +124,7 @@ func assignDatadirsAndCustomPorts(source *greenplum.Cluster, ports []int) (Initi
 			segment.Port = ports[nextPortIndex]
 			portIndexByHost[segment.Hostname] = nextPortIndex + 1
 		}
-		segment.DataDir = upgradeDataDir(segment.DataDir)
+		segment.DataDir = upgradeDataDir(segment.DataDir, upgradeID)
 
 		targetInitializeConfig.Primaries = append(targetInitializeConfig.Primaries, segment)
 	}
@@ -144,7 +149,7 @@ func assignDatadirsAndCustomPorts(source *greenplum.Cluster, ports []int) (Initi
 				segment.Port = ports[nextPortIndex]
 				portIndexByHost[segment.Hostname] = nextPortIndex + 1
 			}
-			segment.DataDir = upgradeDataDir(segment.DataDir)
+			segment.DataDir = upgradeDataDir(segment.DataDir, upgradeID)
 
 			targetInitializeConfig.Mirrors = append(targetInitializeConfig.Mirrors, segment)
 		}
--- a/hub/fill_cluster_configs_test.go
+++ b/hub/fill_cluster_configs_test.go
@@ -1,14 +1,24 @@
 package hub
 
 import (
+	"fmt"
 	"reflect"
 	"testing"
+
+	"github.com/greenplum-db/gpupgrade/upgrade"
 
 	cluster2 "github.com/greenplum-db/gpupgrade/greenplum"
 )
 
-func TestAssignPorts(t *testing.T) {
+func TestAssignDataDirsAndPorts(t *testing.T) {
 
+	upgradeID := upgrade.NewID()
+	expectedDataDir := func(sourceDir string) string {
+		return fmt.Sprintf("%s_%s", sourceDir, upgradeID)
+	}
+	expectedDataDirMaster := func(sourceDir string) string {
+		return fmt.Sprintf("%s_%s-1", sourceDir, upgradeID)
+	}
 	cases := []struct {
 		name string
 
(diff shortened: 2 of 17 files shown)


upgrade/id.go
```diff
@@ -0,0 +1,40 @@
...
+
+import (
+	"crypto/rand"
+	"encoding/base64"
+	"encoding/binary"
+	"fmt"
+)
+
+// ID is a unique identifier for a cluster upgrade.
+type ID uint64
```
berlin-ab: What if the ID type were a string instead, so throughout the codebase we don't need to leak the knowledge about how an ID gets transformed into a string:  for example `0 -> AAAAAAAA`

hub/fill_cluster_configs.go
```diff
@@ -15,7 +17,10 @@ import (
 )
 
 // create source/target clusters, write to disk and re-read from disk to make sure it is "durable"
-func (s *Server) FillClusterConfigsSubStep(config *Config, conn *sql.DB, _ step.OutStreams, request *idl.InitializeRequest, saveConfig func() error) error {
+func FillClusterConfigsSubStep(config *Config, conn *sql.DB, _ step.OutStreams, request *idl.InitializeRequest, saveConfig func() error) error {
+	// Assign a new universal upgrade identifier.
+	config.UpgradeID = upgrade.NewID()
```
berlin-ab: Should we consider moving this to the initial creation of the Config object?

upgrade/id.go
dkrieger-pvtl: The `ID` type implements the Stringer interface, so the general usage is to use `%s` in a string-write function.  The case `0->AAAAAAAAAAA` is for testing, and we instead use fmt.Sprintf("%s",(ID)0) for that instead.  

Users should not assume the ID type is a string, but use the Stringer interface if they need to convert ID to a string.  If they assume it's a string, we are hard-coding that(leaking that it's a string), and couldn't change the type in the future.  Keeping it as not-a-string actually enforces that.

hub/fill_cluster_configs.go
dkrieger-pvtl: Probably...I was just thinking that...

upgrade/id.go
berlin-ab: What if the ID type were a struct that contained the uint64, I think the thing that's bothering me is seeing us construct a config object throughout the system like:

```
&hub.Config{source, target, hub.InitializeConfig{}, 0, port, useLinkMode, 0}
```

Which passes a zero along as the UpgradeID w/ little context as to `0` being a special type. Having it as a struct would make this clearer:

&hub.Config{source, target, hub.InitializeConfig{}, 0, port, useLinkMode, upgrade.NewId(0)}

hub/fill_cluster_configs.go
```diff
@@ -6,6 +6,8 @@ import (
 	"path/filepath"
 	"sort"
 
+	"github.com/greenplum-db/gpupgrade/upgrade"
```
bhuvnesh2703: move it down to consolidate all gpugrade related imports.

hub/init_target_cluster.go
```diff
@@ -6,19 +6,17 @@ import (
 	"path"
 	"path/filepath"
 	"strings"
-	"sync"
+
+	"github.com/greenplum-db/gpupgrade/upgrade"
```
bhuvnesh2703: move it down..

hub/fill_cluster_configs_test.go
```diff
@@ -1,14 +1,24 @@
...
 )
 
-func TestAssignPorts(t *testing.T) {
+func TestAssignDataDirsAndPorts(t *testing.T) {
 
+	upgradeID := upgrade.NewID()
+	expectedDataDir := func(sourceDir string) string {
+		return fmt.Sprintf("%s_%s", sourceDir, upgradeID)
+	}
+	expectedDataDirMaster := func(sourceDir string) string {
```
berlin-ab: I wonder if there is a way for us to not need to know about this special case?

upgrade/id.go
jchampio: I think switching to named parameters in most cases would probably help that.

hub/init_target_cluster.go
```diff
@@ -130,14 +118,23 @@ func WriteInitsystemFile(gpinitsystemConfig []string, gpinitsystemFilepath strin
...
+func upgradeDataDirMaster(path string, upgradeID upgrade.ID) string {
 	// e.g.
 	//   /data/primary/seg1
 	// becomes
-	//   /data/primary_upgrade/seg1
+	//   /data/primary/seg1_456HJLN426
 	path = filepath.Clean(path)
-	parent := fmt.Sprintf("%s_upgrade", filepath.Dir(path))
-	return filepath.Join(parent, filepath.Base(path))
+	return fmt.Sprintf("%s_%s-1", path, upgradeID)
```
bhuvnesh2703: why append _%s`-1` instead of _%s?

jchampio: I'm wondering about this too.

hub/fill_cluster_configs.go
```diff
@@ -37,7 +42,7 @@ func (s *Server) FillClusterConfigsSubStep(config *Config, conn *sql.DB, _ step.
 		ports = append(ports, int(p))
 	}
 
-	s.TargetInitializeConfig, err = AssignDatadirsAndPorts(s.Source, ports)
+	config.TargetInitializeConfig, err = AssignDatadirsAndPorts(config.Source, ports, config.UpgradeID)
```
bhuvnesh2703: possible to pass config object, instead of config.Source and config.UpgradeID?

hub/init_target_cluster.go
```diff
@@ -37,7 +35,7 @@ func (s *Server) writeConf(sourceDBConn *dbconn.DBConn) error {
 	}
 	defer sourceDBConn.Close()
 
-	gpinitsystemConfig, err := CreateInitialInitsystemConfig(s.Source.MasterDataDir())
+	gpinitsystemConfig, err := CreateInitialInitsystemConfig(s.Source.MasterDataDir(), s.Config.UpgradeID)
```
bhuvnesh2703: pass the combined string instead? what GetMasterSegPrefix does?

bhuvnesh2703:
I like it, now we change the actual data directory (gpseg-1 etc) which is much better..

hub/rename_data_directories.go
```diff
@@ -15,10 +16,9 @@ import (
 )
 
 const OldSuffix = "_old"
-const UpgradeSuffix = "_upgrade"
 
 func (s *Server) UpdateDataDirectories() error {
-	if err := RenameMasterDataDir(s.Source.MasterDataDir(), true); err != nil {
+	if err := RenameMasterDataDir(s.Source.MasterDataDir(), "", true); err != nil {
```
bhuvnesh2703: nit: can we just pass src, destination instead of doing manipulation inside. its just rename though..

hub/rename_data_directories_test.go
```diff
@@ -4,9 +4,10 @@ import (
 	"errors"
 	"testing"
 
+	"github.com/hashicorp/go-multierror"
```
bhuvnesh2703: imports placement..

hub/rename_data_directories.go
kalensk: A huge +1 to this.

jchampio: Yep. We're also combining the two calls into one for the local-master case, so that should help out too.

jchampio: Remote segment case is proving trickier to simplify in that way due to `Source/Target` and `AgentConns` tech debt.

bhuvnesh2703:
LGTM other than the comments, tested locally with/without link mode and several combinations of different data dirs setup, looks good.

upgrade/id.go
dkrieger-pvtl: And I use IntelliJ, which shows me the parameter names, so I didn't notice that....

hub/fill_cluster_configs_test.go
dkrieger-pvtl: The problem here is that `gpinitsystem` adds the `-1` suffix to the last directory in the datadir path, but only for the master.  Fixing `gpinitsystem` is one way, but that's not going to happen in this PR.

hub/init_target_cluster.go
dkrieger-pvtl: The problem here is that gpinitsystem adds the -1 suffix to the last directory in the datadir path, but only for the master. Fixing gpinitsystem is one way, but that's not going to happen in this PR.

hub/fill_cluster_configs.go
dkrieger-pvtl: There's much more to the config than just those two fields....so I'd prefer to add in separate members for easier testing.

hub/init_target_cluster.go
dkrieger-pvtl: I was thinking instead of combining the format string in `GetMasterSegPrefix` with `UpgradeDataDir*`...but your suggestion makes sense as the SegPrefix.

hub/rename_data_directories.go
dkrieger-pvtl: All of this will be heavily refactoring in our actual PR, so I am closing this.

upgrade/id.go
dkrieger-pvtl: Done...I added a constant `DefaultID` so there is only a hard-coded constant "AAAA....." in the id_test.go file...all other uses of ID are not magic constants but symbols.

hub/fill_cluster_configs.go
dkrieger-pvtl: too much of a lift now...we really need this to be in a substep....

hub/fill_cluster_configs.go
dkrieger-pvtl: Done

hub/init_target_cluster.go
dkrieger-pvtl: Done

hub/rename_data_directories_test.go
dkrieger-pvtl: @bhuvnesh2703 our usage seems to place go-common-libs in the same place as third-party imports. What change do you want here?

hub/fill_cluster_configs_test.go
```diff
@@ -242,10 +252,33 @@ func TestAssignPorts(t *testing.T) {
...
 		t.Run(c.name, func(t *testing.T) {
-			_, err := AssignDatadirsAndPorts(c.cluster, c.ports)
+			_, err := AssignDatadirsAndPorts(c.cluster, c.ports, 0)
 			if err == nil {
 				t.Errorf("AssignDatadirsAndPorts(<cluster>, %v) returned nil, want error", c.ports)
 			}
 		})
 	}
 }
+
```
berlin-ab: Do we plan on reviving this test? If not, let's remove it.

hub/init_target_cluster.go
berlin-ab: I'm not sure what should own the behavior, but rather than passing these two values around, we should group them together somehow. Maybe a function on Cluster, or on Config that can construct the path.

upgrade/id.go
jchampio: I haven't seen `DefaultID` yet, but if we've added a constant for a zero value (which already has official ways of construction in Go) then I'd say that's a code smell.

jchampio: > And I use IntelliJ, which shows me the parameter names, so I didn't notice that....

Time to turn those off...

hub/init_target_cluster_test.go
```diff
@@ -304,35 +253,24 @@ func TestGetMasterSegPrefix(t *testing.T) {
...
 
 		for _, c := range cases {
-			actual, err := GetMasterSegPrefix(c.MasterDataDir)
+			actual, err := GetMasterSegPrefix(c.MasterDataDir, 0)
 			if err != nil {
 				t.Fatalf("got %#v, want nil", err)
 			}
 
-			expected := "gpseg"
+			expected := "gpseg-1_AAAAAAAAAAA"
```
jchampio: This is not an appropriate semantic change. The segment prefix of `/data/master/gpseg-1` is not `gpseg-1_AAAAAAAAAAA`, it is `gpseg`. This is part of why the `-1` confusion exists.

upgrade/id.go
dkrieger-pvtl: See #292.  We believe we addressed this issue there.

hub/fill_cluster_configs_test.go
dkrieger-pvtl: Fixed in #292.

hub/init_target_cluster.go
dkrieger-pvtl: Fixed in #292.

hub/init_target_cluster.go
dkrieger-pvtl: fixed in #292

hub/rename_data_directories_test.go
dkrieger-pvtl: fixed in #292

hub/fill_cluster_configs.go
dkrieger-pvtl: moved to #292

hub/init_target_cluster_test.go
dkrieger-pvtl: Fixed in #292.

hub/fill_cluster_configs_test.go
dkrieger-pvtl: No...removed in #292.

dkrieger-pvtl:
See #292

PR #289, opened 2020-03-25
Cleanup in rename data directories.

berlin-ab:
This was prefactoring work as I started to take a look at making data directory rename idempotent.


dkrieger-pvtl:
I suggest we close this in favor of https://github.com/greenplum-db/gpupgrade/pull/288.  That PR will significantly modify the functionality of renaming datadirs.

berlin-ab:
Ok, I suggest taking some inspiration from this at least.

PR #291, opened 2020-03-25
Multiple host Vagrant + Centos 7 setup for gpupgrade.

berlin-ab:
Useful for performing manual tests of `gpupgrade` in a multi-host environment.

Note: Currently supports GPDB 6.

Read [`multihost/README.md`](https://github.com/greenplum-db/gpupgrade/pull/291/files#diff-fdbf6458ce2258f5190a0f204ce6614d) for more information on usage.


provision.bash
```diff
@@ -0,0 +1,135 @@
+#!/usr/bin/env bash
+
+GPUPGRADE_SOURCE_PATH=/home/vagrant/gopath/src/github.com/greenplum-db/gpupgrade
+
+own_directories() {
+    sudo chown -R vagrant /home/vagrant
```
tylarb: owning /home/vagrant should be done as part of vagrant process, but in any case most of this file could (or should) be ported to ansible.

sync-ssh-keys.bash
```diff
@@ -0,0 +1,38 @@
...
+
+echo "Collect ssh configurations"
+for hostname in $hostnames; do
+    # populate authorized keys file
+    vagrant ssh "$hostname" --command="cat $public_key_path" >> $authorized_keys_path
+    vagrant ssh "$hostname" --command="cat $guest_authorized_keys_path" >> $authorized_keys_path
+
+    # populate known hosts file
+    for inner_hostname in $hostnames; do
+        vagrant ssh "$hostname" --command="ssh-keyscan $inner_hostname.local >> $guest_known_hosts_path"
```
tylarb: To make things exceptionally simple, we use a flat file checked into the repo called "id_rsa_insecure" to handle these tasks. Especially useful when you create gpadmin user. 

Given that vagrant boxes disallow password login for ssh, manually creating a key and putting it in authorized keys is the an easy solution for these boxes.

provision.bash
berlin-ab: Vagrant's rsync owns all of the parent directories as root, and I'm placing a file within the Vagrant home directory.

berlin-ab:
Currently many of the end-to-end tests fail because they assume a single-host cluster.

berlin-ab:
For testing, I have a script that I've not checked in:

```bash
#!/usr/bin/env bash

vagrant destroy --force

vagrant up

./sync-ssh-keys.bash hub agent

./generate-cluster.bash

./end-to-end-test.bash
```

sync-ssh-keys.bash
berlin-ab: I'll probably take a second pass at doing this if we find multi-host testing valuable.

Make it work, make it right, make it fast.

berlin-ab:
Hmm, something about using dhcp for the hosts causes the test suite to hang b/c the agent host becomes unreachable via ssh (although still running).

jchampio:
(I've been playing with this; I'll leave freeform feedback as I come up with it.)

When I `up`/`halt`/`up` the cluster, the rsync task seems to overwrite the results of the previous guest-specific `make` with my host (macOS) binaries, which causes failures.

Possibly related wishlist item: I think that, rather than clone my dev tree into the guest, it'd be nice if it were shared instead, and the test script can simply `make install` from the shared source before running appropriate tests. This would, I think, also help with cycle time in addition to potentially fixing up the provisioning issue.

Other wishlist: because the Vagrant cluster can be SSH'd to directly from the host (once keys are set up, that is), it'd be really nice to be able to have the host run the test suite instead of the guest. That may be a bit further out though, since it'd require some changes to the suite itself.

berlin-ab:
> When I `up`/`halt`/`up` the cluster, the rsync task seems to overwrite the results of the previous guest-specific `make` with my host (macOS) binaries, which causes failures.
> 

I've not been using it with a `up/halt/up` scenario, but if we added another round of install in the end-to-end-test.bash script (I had it there at one time), the correct binaries will be there.

> Possibly related wishlist item: I think that, rather than clone my dev tree into the guest, it'd be nice if it were shared instead, and the test script can simply `make install` from the shared source before running appropriate tests. This would, I think, also help with cycle time in addition to potentially fixing up the provisioning issue.

I've thought about that. In the past, I've seen pretty poor performance for a shared folder setup - but that was with the GPDB codebase.

> Other wishlist: because the Vagrant cluster can be SSH'd to directly from the host (once keys are set up, that is), it'd be really nice to be able to have the host run the test suite instead of the guest. That may be a bit further out though, since it'd require some changes to the suite itself.

I can see how that would be valuable. Another similar wishlist item for me would be to have the CLI be able to run from a different host than the hub. Then the whole suite could be run from the host and not the guest - but we should wait until the product team prioritizes that feature.

tylarb:
Vagrant supports NFS at least on some hosts, to share items more easily in both directions.

However, it can be a pain to get it working - you might need to enable NFS through firewalls, etc.

multihost-test/Vagrantfile
```diff
@@ -0,0 +1,48 @@
...
+    # https://docs.vagrantup.com
+    # https://vagrantcloud.com/search
+    # https://www.vagrantup.com/docs/multi-machine/
+
+    # Setup:
+    #
+    # The latest GPDB6 RPM can be found on Google Cloud Platform:
+    # https://console.cloud.google.com/storage/browser/pivotal-gpdb-concourse-resources-intermediates-prod/gpdb6-integration-testing/rpm_gpdb_centos7/
+    #
+    # - download it and place it in the gpupgrade directory
```
marksliva, line 14: Why isn't this in the readme instead? I followed the readme and then got the error
`No package gpdb6-*.rpm available.` and then I saw this.

berlin-ab:
Good point. This came first, and then I added the readme, but I forgot to
transfer this.

berlin-ab:
I've altered this PR to only be about setting up a multi-host environment.

jchampio:
Notes from today's attempts at getting this working on my machine:
- The switch to modules means we don't have to use an explicit rsync task; we can push `Vagrantfile` up a level and have it sync automatically to `/vagrant` on the guests. (Note that the current box we're using doesn't have guest tools installed, so it's not a two-way live sync, but it still works.) This simplifies a bunch of little things.
- This might let us get rid of the `chown` stuff too; not sure.
- The switch to modules broke the CentOS 7 compilation because it's version of git is too old. (`fatal: git fetch-pack: expected shallow list`) We could solve this in the short term by installing IUS's `git2u` before installing golang from EPEL. (Long-term, I think we should solve it by cross-compiling to Linux on the host and using that for testing.)

jchampio:
Also:
- We should remove `clean` and `depend` from the make installation line. `clean` does not remove artifacts related to `install`, and `depend` is not a thing anymore.

multihost-test/Vagrantfile
berlin-ab: Transferred to the README.

berlin-ab:
@jchampio Your suggestions have been implemented:

* `Vagrantfile` at the top level
* Git 2.x installed
* `make clean depend` removed

We install bats to `/usr/local`, so it still needs to be writable by us. I've left the directory ownership change for `/usr/local` but have removed the `/home/vagrant` recursive `chown`.

multihost/.gitignore
```diff
@@ -0,0 +1,3 @@
+/.vagrant
```
jchampio, line 1: Can this line be removed now?

multihost/provision.bash
```diff
@@ -0,0 +1,139 @@
...
+
+generate_ssh_keys() {
+    ssh-keygen -f /home/vagrant/.ssh/id_rsa -t rsa -P ""
+}
+
+install_gpupgrade() {
+    cd "$GPUPGRADE_SOURCE_PATH" || exit
+
+    echo "export GOPATH=/home/vagrant/gopath" >> $HOME/.bashrc
+    echo 'export PATH=$PATH:/home/vagrant/gopath/bin:/home/vagrant/go/bin' >> $HOME/.bashrc
```
jchampio, line 104: `GOPATH` and `/home/vagrant/gopath` can go away.

multihost/gpupgrade_cluster_config
```diff
@@ -0,0 +1,50 @@
+# Set this to anything you like
+ARRAY_NAME="gpupgrade cluster"
+
+# This file must exist in the same directory that you execute gpinitsystem in
+MACHINE_LIST_FILE=/home/vagrant/gopath/src/github.com/greenplum-db/gpupgrade/multihost/hostfile
```
jchampio, line 5: `/vagrant/multihost/hostfile`

jchampio: With this last change, I'm able to run `gpupgrade` through finalize on the cluster.

jchampio:
What does ` multihost/vagrant.rb` do?

berlin-ab:
> What does ` multihost/vagrant.rb` do?

Nothing. It shouldn't be there.

I had considered using some Ruby programming to push the `Vagrantfile` contents into the `multihost` directory. I backed that out, but forgot to remove the file.

multihost/gpupgrade_cluster_config
berlin-ab: I also just tried that out. Smooth experience.

multihost/provision.bash
berlin-ab: Good point. Removed.

berlin-ab:
@jchampio I changed the sync to be over `nfs` - so now we have two-way editing.

berlin-ab:
> @jchampio I changed the sync to be over `nfs` - so now we have two-way editing.

This can be done without using the virtualbox guest host additions.  NFS sharing does not require them, only the native VirtualBox shared folders do from what I understand.

jchampio:
> I changed the sync to be over nfs - so now we have two-way editing.

Neat! I'll give it a shot.

berlin-ab:
I changed the provisioning to expect the greenplum rpm inside the `multihost` directory. I also did a `.gitignore` of the rpm.

berlin-ab:
I pushed the bulk of the Vagrant configuration into the `multihost` directory. I think it makes sense for that directory to include as much of the multi-host related code and configuration as possible. The Vagrantfile only needs to delegate responsibility.

berlin-ab:
> rather than clone my dev tree into the guest, it'd be nice if it were shared instead,

Done.

berlin-ab:
> Vagrant supports NFS at least on some hosts, to share items more easily in both directions.
> 
> However, it can be a pain to get it working - you might need to enable NFS through firewalls, etc.
> 
> When it is working, it is in our experience quite fast.

Thanks. It seems to work well.

multihost/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+* The latest GPDB6 RPM can be found on Tanzu Network. Download the Centos 7 rpm of the Greenplum Database Server and place it in the `gpupgrade/multihost` directory
+
+    https://network.pivotal.io/products/pivotal-gpdb
+
+* Create the hub and agent hosts
+
+    ```bash
+    # requires host sudo password for configuring NFS
+    vagrant up
```
bhuvnesh2703, line 25: I think the pre-requisite to it is to setup the provider, should we document anything about it?
```
15:41 $ vagrant up
No usable default provider could be found for your system.

Vagrant relies on interactions with 3rd party systems, known as
"providers", to provide Vagrant with resources to run development
environments. Examples are VirtualBox, VMware, Hyper-V.

The easiest solution to this message is to install VirtualBox, which
is available for free on all major platforms.

If you believe you already have a provider available, make sure it
is properly installed and configured. You can see more details about
why a particular provider isn't working by forcing usage with
`vagrant up --provider=PROVIDER`, which should give you a more specific
error message for that particular provider.
```

bhuvnesh2703:
Ran it locally, it came up cleanup without any issues.
the below 3 steps:
vagrant up
./multihost/sync-ssh-keys.bash
./multihost/generate-cluster.bash

PR #292, opened 2020-03-27
Unique upgrade directories

dkrieger-pvtl:
(This is the PR for the Draft PR #288)

The main idea is that upgraded datadirs now have a unique has as a suffix, NOT _upgrade as the parent dir suffix.  So, for instance,
```
qddir/demoDataDir-1  ---> qddir/demoDataDir-1_StaMaLJlyh0-1
dbfast/demoDataDir2 ---> dbfast/demoDataDir2_StaMaLJlyh0
```

This scheme ends up significantly cleaning up the code, since now we centralize upgrade datadir naming in one place in Initialize and propagate that through the code.  It also solves the motivating bug: to allow multiple datadirs under the same parent directory.

This code also allows "ownership" of directories, as the hash(`StaMaLJlyh0` in the above) is stored in config.json and allows an upgrade to know "its directories".  This feature is not used now.

This code passes all the CI's upgrade `X->Y` jobs, but and all the BATS tests.


jchampio:
Read commit-by-commit; we've split apart related functionality into separate patches that will be squashed into one later.

jchampio:
Major changes from the draft:
- `Config` creation in tests has been pared down so that we're not continually passing don't-care values without names; this should hopefully address @berlin-ab's feedback on bare `ID`/`0` creation.
- The upgrade ID is now added to the segment prefix instead of the end of the datadir name. (i.e. `gpseg.AAAAAAAAAAA.-1` instead of `gpseg-1_AAAAAAAAAAA`). This translation of source->target datadir is now consolidated in the `upgrade` package; the new rules allow us to meet `gpinitsystem` requirements in one place.
- BATS tests have been updated with the new naming requirements. (`gpupgrade config show --id` helps with that.)
- Finalize's datadir rename has been completely revamped. The long-term goal is for `UpdateDataDirectories()` to be the testable unit -- since the goal of that function is relatively simple -- and to stop pinning as many implementation details in the tests, since those divisions are now obscuring simplifications of the logic. (For instance, `RenameMasterDataDir` had become a wrapper around `os.Rename()`, which we called twice. We've instead consolidated the double-rename into a single operation, which we call `RenameDataDirs`, so that it can eventually be shared between the hub and agent.) We're not to the end yet, but this brings us a little closer.

hub/fill_cluster_configs.go
```diff
@@ -37,7 +41,7 @@ func (s *Server) FillClusterConfigsSubStep(config *Config, conn *sql.DB, _ step.
 		ports = append(ports, int(p))
 	}
 
-	s.TargetInitializeConfig, err = AssignDatadirsAndPorts(s.Source, ports)
+	config.TargetInitializeConfig, err = AssignDatadirsAndPorts(config.Source, ports, config.UpgradeID)
```
dkrieger-pvtl: In #288, @bhuvnesh2703 comments:

>possible to pass config object, instead of config.Source and config.UpgradeID?

dkrieger-pvtl: There's much more to the config than just those two fields....so I'd prefer to add in separate members for easier testing.

hub/rename_data_directories.go
```diff
@@ -14,127 +13,116 @@ import (
...
-
-			// Otherwise include mirrors and standby. (Master's excluded above.)
-			return true
-		}
-
-		segments := cluster.SelectSegments(selector)
-		if len(segments) == 0 {
-			// we can have mirror-only and standby-only hosts, which we don't
-			// care about here (they are added later)
+		if len(renames[conn.Hostname]) == 0 {
```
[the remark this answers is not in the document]
jchampio: Never mind.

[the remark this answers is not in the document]
dkrieger-pvtl: Right...it can and often will be...if the source has a standby on a different host, the agentConn list will contain that standby host but the target cluster won't have a standby...and hence will have an empty slice for the target cluster.

We could imagine removing the check, though, since the agent call to rename the dirs can handle an empty list, but I prefer to pre-optimize by not making calls to the agent that we know have no side effects.  And doing so makes testing goofy...our mocks then need to expect an "empty call" to the standby, for instance.

hub/rename_data_directories.go
```diff
@@ -14,127 +13,116 @@ import (
 	"github.com/greenplum-db/gpupgrade/utils"
 )
 
+type RenameMap = map[string][]*idl.RenamePair
+
 const OldSuffix = "_old"
```
dkrieger-pvtl: `s/OldSuffix/oldSuffix/g` since this constant is only used in this file.

hub/fill_cluster_configs.go
```diff
@@ -65,22 +69,33 @@ func AssignDatadirsAndPorts(source *greenplum.Cluster, ports []int) (InitializeC
...
+	// want to remove the "ok" check here and force NewCluster to error out
 	if master, ok := source.Primaries[-1]; ok {
 		// Reserve a port for the master.
 		if nextPortIndex > len(ports)-1 {
 			return InitializeConfig{}, errors.New("not enough ports")
 		}
+
+		// Save the segment prefix for later.
+		var err error
+		segPrefix, err = GetMasterSegPrefix(master.DataDir)
```
marksliva: Setting `segPrefix` here means that it gets passed in when we set up the standby, primary and mirror datadirs. Is it a safe assumption that mirrors follow the segPrefix convention? I thought that movemirrors allows users to change the data directory to anywhere they want, including breaking the naming convention. Not sure about how gpexpand directories work, but that would be another place to check as well.

test/config.bats
```diff
@@ -61,16 +61,17 @@ teardown() {
 
     run gpupgrade config show
     [ "$status" -eq 0 ]
-    [ "${lines[0]}" = "source-bindir - /my/old/bin/dir" ]
-    [ "${lines[1]}" = "target-bindir - /my/new/bin/dir" ]
-    [ "${lines[2]}" = "target-datadir - " ] # This isn't populated until cluster creation, but it's still displayed here
+    [[ "${lines[0]}" = "id - "* ]] # this is randomly generated
```
marksliva: We could potentially replace the `*` with a regex matcher that includes the base64 character set.

upgrade/directories_test.go
```diff
@@ -0,0 +1,50 @@
...
+		expected := fmt.Sprintf(c.expectedFormat, id)
+
+		if actual != expected {
+			t.Errorf("TempDataDir(%q, %q, id) = %q, want %q",
+				c.datadir, c.segPrefix, actual, expected)
+		}
+	}
+}
+
+func ExampleTempDataDir() {
```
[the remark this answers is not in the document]
dkrieger-pvtl: It shows up in the documentation...go doc...

hub/fill_cluster_configs.go
jchampio: > Is it a safe assumption that mirrors follow the segPrefix convention?

No, but if we designed this correctly, that shouldn't be an assumption we're making -- if, say, a mirror doesn't follow segPrefix, the temp dir will instead be something like `/path/to/mirror0.iDeNtIfIeR` and things will still work. `gpinitsystem` is not taking mirror paths at all.

jchampio: We should totally add movemirrors testing though.

upgrade/directories_test.go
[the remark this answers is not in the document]
jchampio: I started to add an example in the doc comments and then remembered that [Go has a way to do that](https://blog.golang.org/examples). `go test` will fail if the example goes out of date, too.

hub/fill_cluster_configs.go
dkrieger-pvtl: Great point, but I think we're ok...the target's temp dataDirs will still be "next to" the original source dataDir even if the standby/mirrors are created according to a different naming convention by `gpinitstandby`or `gpaddmirrors`.  The `segPrefix` determines if the temp name is a certain format.   Say the `segPrefix` is `demoDataDir`.  Then the temp master dir will be `demoDataDir.ABC123.-1`.  If the source standby is `/data/demoDataDirstandby`, then it will be mapped to `/data/demoDataDir.ABC123.standby`.  If it is `/data/standby/`, it will be mapped to `/data/standby.ABC123`.

All that matters is that the source/target are at the same level.

That said, we should probably add a test for this.

hub/rename_data_directories.go
berlin-ab: Is the same true of `RenameMap`?

dkrieger-pvtl: No, `RenameMap` is used as a type in the tests.  We hard-code the `_old` in our tests, as we want to ensure we are pinning that behavior exactly.

hub/fill_cluster_configs.go
marksliva: I have another observation. Since `GetMasterSegPrefix` returns an error if no prefix is found, it could fail for users with a master datadir which does not follow this conventional naming pattern. It is not very likely, but could be possible for them to get into this state if they go through standby recovery and then during master re-initialization they pick a different data directory path than what it was before. i.e.
`gpinitstandby -a -s localhost -P 6000 -S $HOME/workspace/gpdb6/gpAux/gpdemo/datadirs/master`

marksliva: Although the `GetMasterSegPrefix` function already existed and would have failed for the same reason before this PR anyway, so we don't have to address it at this point.

hub/rename_data_directories.go
jchampio: > No, RenameMap is used as a type in the tests. We hard-code the _old in our tests, as we want to ensure we are pinning that behavior exactly.

I thought we removed that and switched to testing external behavior instead. Where do you see that in the tests?

dkrieger-pvtl: We still have basic go tests for RenameDataDirs(master datadirs) and RenameSegmentDataDirs(all other datadirs).

hub/fill_cluster_configs.go
dkrieger-pvtl: I filed an issue for us to track this.

hub/rename_data_directories.go
jchampio: Sorry, GitHub was hiding the diff from me... sigh.

hub/fill_cluster_configs_test.go
```diff
@@ -4,10 +4,16 @@ import (
 	"reflect"
 	"testing"
 
-	cluster2 "github.com/greenplum-db/gpupgrade/greenplum"
+	cluster2 "github.com/greenplum-db/gpupgrade/greenplum" // TODO: replace with greenplum
```
kalensk: Do we want to address this TODO before merging?

hub/fill_cluster_configs_test.go
```diff
@@ -4,10 +4,16 @@ import (
...
+	"github.com/greenplum-db/gpupgrade/upgrade"
 )
 
-func TestAssignPorts(t *testing.T) {
+func TestAssignDataDirsAndPorts(t *testing.T) {
+	var upgradeID upgrade.ID
+
+	expectedDataDir := func(sourceDir string) string {
+		return upgrade.TempDataDir(sourceDir, "seg", upgradeID)
```
kalensk: Maybe inline this function since its one-line and the callers already have the variable `expected` so the function name isn't getting us much.

test/config.bats
dkrieger-pvtl: deferring this for now.  I put a comment in the code.

hub/fill_cluster_configs_test.go
dkrieger-pvtl: Yes; done.

hub/fill_cluster_configs_test.go
dkrieger-pvtl: We had that before; it is makes the lines longer and harder to see what hard-coded dirs we are seeing.

upgrade/id.go
```diff
@@ -0,0 +1,40 @@
+package upgrade
```
berlin-ab, line 1: nit: this commit does not have value on its own, maybe it should be rolled into the usage commit.

dkrieger-pvtl: that's why we put it in its own commit.  To me, it makes the code easier to review both during PR and during code archaeology as it is self-contained without dependencies.

berlin-ab: Now, if you were to use a non-fast-forward merge commit, then the archeology would make more sense.

PR #293, opened 2020-03-30
CI: stabilize install-tests

marksliva:
Specify a temp-port-range from 6020-6040 for all bats tests which is closer to
the working 6000-6007 range of the source cluster.  We have not seen any port
contention in the 6000-6007 range that the source cluster uses, so we stay near
that range for the target cluster to fix the ephemeral port conflicts we have
seen around 10-20% of the time.

### DIFF
--- a/installcheck.bats
+++ b/installcheck.bats
@@ -27,6 +27,7 @@ teardown() {
               --source-bindir "$GPHOME"/bin \
               --target-bindir "$GPHOME_NEW"/bin \
               --source-master-port $PGPORT \
+              --temp-port-range 6020-6040 \
               --disk-free-ratio=0 \
               3>&-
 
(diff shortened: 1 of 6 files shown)


dkrieger-pvtl:
How are you planning to validate/test this?  You could start a pipeline job and re-run the install tests 20 times...

berlin-ab:
Awesome thinking. Let's move forward with this. It'll either work, which is great, or it won't - which is more information to help us fix it.

marksliva:
> How are you planning to validate/test this? You could start a pipeline job and re-run the install tests 20 times...

In our pipline, we triggered `install-tests` about 30 times with these changes, and the only failing builds were due to the unrelated github timeout issue, so we are confident that these changes fix the problem.
Once it is merged, we will continue to monitor the production pipeline to see if the issue resurfaces.

PR #294, opened 2020-03-31
CI: load upgrade data in separate task

jchampio:
The overall goal of this is to split the loading of upgradable data from the upgrade itself. This has proven to be useful for the first "real-world data" story, and it has the additional debugging advantage that upgrade failures are no longer buried in hundreds of lines of SQL restoration output.

The core implementation is bookended by two optional refactors:
1. move the template file next to the actual pipeline (it is now at `ci/template.yml`)
2. provide a template function to help us stop hardcoding major versions in the pipeline template, and hopefully drive better deduplication in the future


ci/scripts/load-dump.sh
```diff
@@ -0,0 +1,19 @@
...
+apk add --no-progress openssh-client
+cp -R cluster_env_files/.ssh /root/.ssh
+
+scp sqldump/dump.sql.xz gpadmin@mdw:/tmp/
+
+echo 'Loading SQL dump into source cluster...'
+time ssh -n gpadmin@mdw "
+    set -eux -o pipefail
+
+    version=\$(rpm -q --qf '%{version}' '$OLD_PACKAGE')
```
berlin-ab, line 14: nit: what do you think about adding another parameter that is the explicit version instead of needing to parse it out of the rpm?

kalensk: Is there a way to get that from the pipeline.yml in order to pass that down? I am unaware of one.

berlin-ab: We already hardcode `$OLD_PACKAGE` as `OLD_PACKAGE: greenplum-db-6` in the yaml.

jchampio: Yeah, Bradford and I threw a couple approaches at the wall and this one fell out. Asking `rpm` directly was by far the easiest for now; maybe that will change with the "official" packaging solution.

berlin-ab: Sounds good. Thanks for the explanation.

PR #295, opened 2020-03-31
CI: add a 5-6 link mode test

dkrieger-pvtl:
We follow the current pipeline.yml generation scheme by adding a
new test in the test matrix.


ci/parser/pipeline-template.yml
```diff
@@ -257,7 +257,7 @@ jobs:
     <<: *slack_alert
 
 {{range .UpgradeJobs}}
-- name: {{.Source}}-to-{{.Target}}{{- if .PrimariesOnly }}-primaries-only{{end}}{{- if .NoStandby }}-no-standby{{end}}
+- name: {{.Source}}-to-{{.Target}}{{- if .UseLinkMode }}-link-mode{{end}}{{- if .PrimariesOnly }}-primaries-only{{end}}{{- if .NoStandby }}-no-standby{{end}}
```
marksliva: nit: this is starting to get unreadable. We could instead set a string on the `UpgradeJobs` struct that sets what to render after the '6'.

ci/parser/parse_template.go
```diff
@@ -43,17 +44,22 @@ func init() {
 	for _, sourceVersion := range sourceVersions {
 		for _, targetVersion := range targetVersions {
 			upgradeJobs = append(upgradeJobs, &UpgradeJobs{
-				Source:        sourceVersion,
-				Target:        targetVersion,
+				Source: sourceVersion,
+				Target: targetVersion,
+			})
+			upgradeJobs = append(upgradeJobs, &UpgradeJobs{
```
marksliva: nit: Although this struct was not introduced as part of this PR, it would make sense to rename `UpgradeJobs` to `UpgradeJob`.

ci/parser/pipeline-template.yml
jchampio: Make use of the space trimmer:
```
- name: {{.Source}}-to-{{.Target}}
        {{- if .UseLinkMode }}-link-mode{{end}}
        {{- if .PrimariesOnly }}-primaries-only{{end}}
        {{- if .NoStandby }}-no-standby{{end}}
```

dkrieger-pvtl: Done.

ci/parser/parse_template.go
dkrieger-pvtl: Good point.  Done.

PR #296, opened 2020-04-01
CI: deduplicate and refine Concourse template

jchampio:
The cost of duplication in Concourse YAML is higher than usual due to the extreme verbosity, ease of accidental implementation drift, and relative lack of small deduplication tools. (Anchors can only do so much, and then the next step is task `.yml` implementations, which have serious maintenance problems due to their tight and non-obvious coupling with their parent pipelines.)

The first goal of this PR is to combine the 5-to-6 and 6-to-6 logic to get rid of that copy-paste, since it's already made us hold our nose for some of the retail-demo work.

The second goal is to continue simplifying the upgrade task itself by splitting out the mirror and standby verification into its own task.

**I highly recommend reviewing commit-by-commit with whitespace differences turned off.**


ci/template.yml
```diff
@@ -307,16 +307,28 @@ jobs:
...
-        {{- end}}
-        {{- if .PrimariesOnly}}
-        MIRRORS: 0
-        STANDBY: 0
-        {{- end}}
         {{- if .UseLinkMode}}
         USE_LINK_MODE: 1
         {{- end}}
+    {{- if not .NoStandby -}}
+    {{- if not .PrimariesOnly }}
```
kalensk, line 302: nit and probably for another PR: The double negative is getting hard to read. (ie: `not .NoStandby`)

kalensk:
[Test pipeline is here](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:dev:more-pipeline) if others wanted to take a look.

PR #297, opened 2020-04-02
ci: Publish a gpupgrade binary in the prod pipeline

marksliva:
A new publish-release-candidate job is added to the new prod-pipeline.yml file.
This job requires all previous jobs to have succeeded, and will then publish
an artifact to the bucket/path specified in the bin_gpupgrade_rc resource.
The existing pipeline.yml was moved to dev-pipeline.yml.
Also, the generated files now live in ci/generated.


ci/generate.go
```diff
@@ -1,4 +1,4 @@
 package ci
 
-// Generate pipeline.yml from template.yml using parse_template
-//go:generate go run ./parser/parse_template.go ./template.yml pipeline.yml
+// Generate dev-pipeline.yml and prod-pipeline.yml from template.yml using parse_template
+//go:generate go run ./parser/parse_template.go ./template.yml generated/${TARGET}-pipeline.yml $TARGET
```
jchampio, line 4: I'd prefer that `go generate` operate on both dev and prod pipelines at once, so that we don't see "drift" on people's machines (where, for example, someone doesn't see the template propagate to the prod pipeline file until _after_ they commit the PR and set the master pipeline).

marksliva: Fixed.

README.md
```diff
@@ -190,14 +190,17 @@ NEWBINDIR, OLDBINDIR, NEWDATADIR and OLDDATADIR.
...
 files in gpupgrade/ci/tasks, and some secrets in a private repository. To set a
 pipeline, run:
 
 ```
 make set-pipeline FLY_TARGET=<CONCOURSE_INSTANCE> GIT_URI=https://github.com/<GITHUB_USERNAME>/gpupgrade.git
 ```
 
+Note: When making pipeline changes, make sure to also check in the prod pipeline changes
+into the prod-pipeline.yml file.
+
```
jchampio, line 203: This is no longer correct; we want people to modify the template and then just run `go generate` (which `make set-pipeline` will do for you).

ci/parser/parse_template.go
```diff
@@ -103,7 +105,11 @@ func deduplicate(a, b []string) []string {
 }
 
 func main() {
-	templateFilepath, pipelineFilepath := os.Args[1], os.Args[2]
+	flag.BoolVar(&data.ProdTarget, "prod", false, "the target environment")
```
jchampio, line 108: nit: `the target environment` would not tell me what this flag meant if I didn't already know.

ci/template.yml
```diff
@@ -322,3 +331,32 @@ jobs:
...
+        - install-tests
+        - 6-to-6
+        {{- range .UpgradeJobs}}
+        - {{.Source}}-to-{{.Target}}
+          {{- if .UseLinkMode }}-link-mode{{end -}}
+          {{- if .PrimariesOnly }}-primaries-only{{end -}}
+          {{- if .NoStandby }}-no-standby{{end -}}
+        {{end}}
+    - task: build-rc
+      file: gpupgrade_src/ci/tasks/build-rc.yml
```
jchampio, line 356: Can we inline this? The `.yml` hides the coupling between `build-rc` and `bin_gpupgrade_rc` (via `gpupgrade_bin_path`).

ci/tasks/build-rc.yml
```diff
@@ -0,0 +1,31 @@
...
+  - name: gpupgrade_bin_path
+
+run:
+  path: bash
+  args:
+  - -c
+  - |
+    set -ex
+
+    cd $GOPATH/src/github.com/greenplum-db/gpupgrade
```
jchampio, line 23: nit: this is what `pushd`/`popd` are for.

ci/template.yml
marksliva: Fixed.

ci/parser/parse_template.go
marksliva: Good point. Fixed.

README.md
marksliva: Updated the comment to better explain the relation between the 3 yml files and how 2 of them get generated.

ci/scripts/build-rc.sh
```diff
@@ -0,0 +1,16 @@
...
+
+export GOPATH=$PWD/go
+export PATH=$GOPATH/bin:$PATH
+
+cd $GOPATH/src/github.com/greenplum-db/gpupgrade
+    make depend
+    make
+
+    # todo: is there a cleaner way to get this format?
+    version=$(git describe | perl -pe 's/(.*)-([0-9]*)-(g[0-9a-f]*)/\1+dev.\2.\3/')
```
jchampio, line 13: If we don't care about `dev` or dots versus dashes,
```
version=$(git describe | sed 's/-/+/')
```
should result in e.g. `0.2.0+198-gc3fd8eb`.

jchampio: So it looks like semi-semantic versioning is going to care about the dots. How about
```
version=$(git describe | sed -E 's/-([0-9]+)-/+\1./')
```
for a result of `0.2.0+198.gc3fd8eb`?

jchampio: This probably isn't important enough to bikeshed right now, resolving.

Makefile
```diff
@@ -115,18 +115,18 @@ GIT_URI := $(shell git ls-remote --get-url)
...
 
 # Concourse does not allow "/" in pipeline names
 PIPELINE_NAME ?= gpupgrade:$(shell git rev-parse --abbrev-ref HEAD | tr '/' ':')
 FLY_TARGET ?= cm
 ifeq ($(FLY_TARGET),prod)
-SECRETS_TYPE := prod
+	TARGET := prod
 else
-SECRETS_TYPE := dev
+	TARGET := dev
```
kalensk, line 129: I like this change (that is consolidating SECRETS_TYPE with just TARGET).

ci/template.yml
```diff
@@ -322,3 +331,44 @@ jobs:
   on_failure:
     <<: *slack_alert
 {{end -}}
+
+{{- if .ProdTarget }}
+# todo: In the future, we want to build gpupgrade first, and input that
+# binary to each job that needs it.
+# Then, this job can also receive the binary as an input and then will
+# upload it to the rc bucket.
```
kalensk, line 339: Is there a chore for this?

jchampio: Done.

PR #302, opened 2020-04-03
basic fault injection framework

dkrieger-pvtl:
NOTE: this is currently used to test idempotence of UPGRADE_DATA_DIRECTORIES, soon to PR.

NOTE: this is very rough, and is not yet general.  Please just review the idea at a high level.  Possible comments: use package xyz123 instead of making our own, consider using reflection instead to pull function names, add a regex parser to enable/disable faults, etc

We add a basic fault injector.  To use it:

gpupgrade initialize --fault-injection ....

and then in any subsequent hub code:

if fault_injector.Insert("callMe", 10) {
    fmt.Println("callMe executing arbitrary code")
}

NOTE: this currently only runs fault injection code on the cli once
gpugprade initialize has been called and on the hub during and after
the call to gpugrade initialize starts the hub.


berlin-ab:
We should strongly consider white-box integration testing in Go rather than black-box integration testing with fault injections.

PR #307, opened 2020-04-07
Idempotent update data dirs

dkrieger-pvtl:
You should review this one commit at a time.  All commits(other than the REMOVE BEFORE MERGE tagged commits, which will be removed before merge) will be squashed together prior to merge.  This replaces https://github.com/dkrieger-pvtl/gpupgrade/pull/1.

** YOU WILL LIKELY FIND IT EASIER TO REVIEW IN THIS ORDER(commit order) **
```
a7608ca (REMOVE BEFORE MERGE) use fault injector to test
7a225ba (REMOVE BEFORE MERGE) fault injection
551db11 test: finalize.bats fixes for new archive datadir scheme
bca18af add a few comment observations in code from this implementation
b23d1e2 agent: refactor delete_data_directories
04ae154 finalize: implement upgrade_data_dirs idempotence
a85c163 test: utilities for unit testing renaming
aff6539 utils: add helper functions for general use
984e618 proto: small refactor of hub-agent rename rpc command
```

```
    We establish idempotence for the finalize substep FINALIZE_UPDATE_DATA_DIRECTORIES.

    We use the atomicity of os.Rename() to make the actual UPDATE step 
    idempotent. This is detailed and tricky, as we must also handle idempotence
    of the deletion of the mirrors/standby of the source cluster(in link mode)
    as well as idempotence on the agents.

    In addition, we use a new archive directory naming scheme to allow
    a further cross-check on renaming.  Instead of the current scheme,
           DIR_PATH/sourceDir -->  DIR_PATH/sourceDir_old
   we embed the ugprade ID:
           DIR_PATH/sourceDir --> DIR_PATH/sourceDir.ABC123.old 
```


utils/sys_utils.go
```diff
@@ -6,17 +6,17 @@ import (
...
-	"github.com/greenplum-db/gp-common-go-libs/gplog"
 	"golang.org/x/xerrors"
 )
 
 var (
 	System = InitializeSystemFunctions()
 )
 
-const markerFile = ".gpupgrade"
+var PostgresFiles = []string{"postgresql.conf", "PG_VERSION"}
```
berlin-ab, line 19: This likely belongs elsewhere. My first thought would be with the code that uses it in `IsPostgres`.

testutils/test_utils.go
```diff
@@ -49,3 +59,46 @@ func GetOpenPort() (int, error) {
...
+func createDataDir(t *testing.T, name, tmpDir string) (source string) {
+
+	source = filepath.Join(tmpDir, name)
+
+	err := os.Mkdir(source, 0700)
+	if err != nil {
+		t.Errorf("unexpected err: %v", err)
+	}
+
+	for _, fileName := range utils.PostgresFiles {
```
berlin-ab, line 95: I would recommend using the literal list here to call out the externally expected behavior, instead of tying the behavior to implementation.

utils/sys_utils.go
dkrieger-pvtl: This list is used in two places in two different files, though for a similar purpose: either in testing to create a directory with the correct files that `IsPostgres` will accept, or `IsPostgres` itself.  I could pull `IsPostgres` into this file, but right now it's only used on the agent.  The test usage is used on both hub and agent.

dkrieger-pvtl: As you point out below, a literal list can be used in the testing code and `PostgresFiles` can be pulled into the one file that uses it.

agent/create_files.go
```diff
@@ -0,0 +1,29 @@
...
+import (
+	"context"
+
+	"github.com/greenplum-db/gpupgrade/utils"
+
+	"github.com/hashicorp/go-multierror"
+
+	"github.com/greenplum-db/gp-common-go-libs/gplog"
+
+	"github.com/greenplum-db/gpupgrade/idl"
```
jchampio, line 12: Please, please fix your IDE. It's not fair to us to have to point these out every time a PR is filed.

hub/rename_data_directories.go
```diff
@@ -34,79 +49,132 @@ func UpdateDataDirectories(conf *Config, agentConns []*Connection) error {
...
+		}
+		if target == "" {
+			if !OnlySourceRenamed(source, archive, upgradeID) {
+				return xerrors.Errorf("inconsistent state for source only: %v to archive:%v err:%w", source, archive, err)
+			}
+		} else {
+			if !(OnlySourceRenamed(source, archive, upgradeID) || BothRenamed(source, target, archive, upgradeID)) {
+				return xerrors.Errorf("inconsistent state for source/target: %v to target:%v err:%w", source, target, err)
+			}
+		}
```
jchampio, line 154: Doesn't the error returned by `Rename` tell us which problem exists? Why do we need the two-step marker file solution?

- If you try to archive source and you get `ENOENT`, we've already renamed it and need to move on to target.
- If you try to archive source and you get `EEXIST`, we've already renamed both the source and the target.

hub/marker_files_test.go
```diff
@@ -0,0 +1,210 @@
...
+	"github.com/greenplum-db/gpupgrade/idl"
+	"github.com/greenplum-db/gpupgrade/idl/mock_idl"
+	"github.com/greenplum-db/gpupgrade/utils"
+)
+
+func TestAddMarkerFiles(t *testing.T) {
+
+	upgradeID := upgrade.ID(10) // "CgAAAAAAAAA" in base64
+
+	t.Run("passes correct filenames to OpenFile", func(t *testing.T) {
```
berlin-ab, line 25: nit: I think the number of segments could be reduced and we'd still get the intended coverage.

jchampio:
> The need for a fault injection library signals to me that we need to start integration testing in Go with the ability to stand up the system and use dependency injection to inject faulty components.

For most simple cases, that would probably work. But fault injection gives us a complicated-domain tool for very specific types of failures that wouldn't otherwise be considered a "dependency" (e.g. the hub crashing in between two critical changes to the system).

That said, I'm not sure adding fault injection as part of this work is warranted yet.

hub/rename_data_directories.go
dkrieger-pvtl: If the archive source has already been archived and target has not, the [rename POSIX spec](https://pubs.opengroup.org/onlinepubs/009695399/functions/rename.html) appears to allow either `ENOENT` or `EEXIST/ENOTEMPTY` to be returned.  That is, there appears to be no situation where a non-empty directory can be the target of a rename, so an implementation could check that first.  

This is more of a pedantic issue, as we can still achieve what is required:

* if source has been renamed and target has not: a reattempt to rename the source will return `ENOENT` or `EEXIST/ENOTEMPTY` _and_ the source archive directory will exist.
* if the source and target have both been renamed, a reattempt to rename the source will return `EEXIST/ENOTEMPTY` for the attempted source _and_ the source directory will exist.  A reattempt to rename the target will return either `ENOENT` or `EEXIST/ENOTEMPTY` _and_ the target directory will not exist.

dkrieger-pvtl:
> When you and I paired on that, we noted that it'd be good to move the archive handling (`_old`) to the same scheme, to make it easy to prove that we were not overwriting an `seg-1_old` that someone else had dropped on the system. That seems to have been forgotten with this PR, and I'm not convinced that the marker file solution is giving us anything we didn't already have; see my comment inline.

Using a sourceDir archive directory that includes the upgradeID(say, `SOURCE_DIR/demoDataDir.ABC123_old`) will prevent us from over-writing the source archive directory from a previous/stale/pre-existing archive directory.

The marker file solution seems more robust to me...we have a positive way to identify a renamed target cluster datadir(it is in the source location with a target-like marker file).  Without the marker files, we are left to rely on the atomicity of rename _and_ the exact reasoning we use.  The marker files use this as well as the existence of the marker files to cross-check this reasoning.

jchampio:
> we have a positive way to identify a renamed target cluster datadir(it is in the source location with a target-like marker file).

We already have a positive way: there's an archived source and a renamed target.

I understand that a marker file might give us an additional sanity check, but let's talk about what abuses we're trying to protect against in that situation, and not fold it into what's ostensibly about adding idempotence.

dkrieger-pvtl:
> I understand that a marker file might give us an additional sanity check, but let's talk about what abuses we're trying to protect against in that situation, and not fold it into what's ostensibly about adding idempotence.

Yes, the marker files are for an additional sanity check.  Without them, the one check we have relies on reasoning involving the atomicity/return codes of the `rename` system call.  The marker files are a straightforward way to provide an explicit check that the correct directories are in the correct location.  For the source cluster, we can use the directory name of the archived cluster(i.e. move the source cluster to `_<upgradeID>_old`) and not add a marker file.  (This also simplifies revert).  For the target cluster, its datadirs are moved to the source's dirs.

So, there are three options:
1). no marker files, archive source to `_old`: works, with no "safety check"
2). no marker files, archive source to `_<upgradeID>_old`: works, with the safety check that the sourceDir we are currently working on in this upgrade was in fact moved.  There's no safety check on the target, but the safety check on the source should be sufficient.
3). marker files, archive source to `_<upgradeID>_old`: works, with a safety check on the source cluster and on the target cluster.

2). doesn't involve the use of marker files _and_ provides safety that we are not failing due to some stale archived source from a previous run of gpugprade.  It also provides protection against "unknown unknown" flaws in the reasoning based on atomicity the existence of the archive dir is positive evidence that that source dataDir has been moved.

3). also works but I can see the argument that, in absence of an explicit failure mode, the added complexity isn' justified.

agent/rename_directories_test.go
```diff
@@ -2,62 +2,153 @@ package agent_test
...
+				os.RemoveAll(tmpDir)
 			}
+		}()
 
-			if dst != pair.Dst {
-				t.Errorf("got %q want %q", dst, pair.Dst)
+		// set up function spies
+		numCalls := 0
+		var fileCheckSourceDirs, fileCheckTargetDirs []string
+		agent.IsPostgresFunc = func(dataDir string) bool {
```
berlin-ab, line 61: This seems like some coupling to the implementation.  Can we think of a way to avoid doing that?

agent/delete_data_directories_test.go
```diff
@@ -16,91 +18,75 @@ import (
...
+			}
+			if utils.DoesPathExist(dataDir2) {
+				t.Errorf("expected false")
+			}
 
-		var multiErr *multierror.Error
-		if !xerrors.As(err, &multiErr) {
-			t.Fatalf("got error %#v, want type %T", err, multiErr)
-		}
+			if c.checkIdempotence {
```
berlin-ab, line 60: Why not just always test for this rather than making it conditional? Its always a behavior we want.

agent/create_files.go
dkrieger-pvtl: @jchampio understood.  I have `go fmt` and `goimports` run on file save, but neither reformats across blank lines in the import block.  I don't see a way to have either do so in their options.  Suggestions?

agent/delete_data_directories_test.go
dkrieger-pvtl: I have two tests: one to test basic functionality and the other to test that the call is idempotent.  Separating out these calls allows us to determine which is failing.

agent/rename_directories_test.go
dkrieger-pvtl: This function calls two underlying functions, each of which is unit tested elsewhere.  So the approach of this test is to make sure those underlying functions are called with the correct arguments.  Otherwise, we end up repeating that test code here.  So, by design, this unit test couples to its implementation.

hub/marker_files_test.go
dkrieger-pvtl: This file/feature has been removed from this PR.

agent/create_files.go
jchampio: > I have go fmt and goimports run on file save

`goimports` subsumes `go fmt`; run only `goimports`.

> but neither reformats across blank lines in the import block.

Correct. The key is to get your IDE not to add those lines in the first place; `goimports` will not fight you once you've decided to introduce a new import "section".

agent/utils.go
```diff
@@ -0,0 +1,25 @@
...
+	if !utils.DoesPathExist(dir) {
+		return false
+	}
+
+	for _, fileName := range utils.PostgresFiles {
+		filePath := filepath.Join(dir, fileName)
+		_, err := utils.System.Stat(filePath)
+		if err != nil {
+			return false
+		}
```
jchampio, line 21: Why doesn't this use `DoesPathExist()`?

utils/sys_utils.go
```diff
@@ -102,31 +101,9 @@ func GetStateDir() string {
...
-	mFile := filepath.Join(dataDir, markerFile)
-	gplog.Info("creating marker file %s", mFile)
-	err = System.WriteFile(mFile, []byte{}, 0644)
-	if err != nil {
-		return xerrors.Errorf("create gpupgrade marker file %s: %w", mFile, err)
-	}
-	return nil
+// DoesPathExist returns true if the path argument can be successfully accessed
+//  by the caller and false otherwise.
+func DoesPathExist(path string) bool {
```
jchampio, line 106: nit: "if does path exist" doesn't read particularly well in English. `PathExists()` maybe?

agent/utils_test.go
```diff
@@ -0,0 +1,49 @@
...
+	testhelper.SetupTestLogger()
+
+	t.Run("returns false if dir does not exist", func(t *testing.T) {
+		if agent.IsPostgres("/does/not/exist/ABC1234") {
+			t.Errorf("expected false")
+		}
+	})
+
+	t.Run("returns true if required files exist", func(t *testing.T) {
+		source, _, tmpDir := testutils.SetupDataDirs(t)
```
jchampio, line 24: I can't find `SetupDataDirs` in this commit (`utils: add helper functions for general use`); has it been added to the wrong patch?

utils/sys_utils.go
```diff
@@ -69,6 +67,7 @@ func InitializeSystemFunctions() *SystemFunctions {
 		Create:       os.Create,
 		Mkdir:        os.Mkdir,
 		SqlOpen:      sql.Open,
+		Close:        func(f *os.File) error { return f.Close() },
```
jchampio, line 70: This seems like a code smell to me. We have well-known interfaces that include `Close()` already.

utils/sys_utils_test.go
```diff
@@ -95,188 +89,3 @@ var _ = Describe("user utils", func() {
 
 	})
 })
-
-func TestCreateAllDataDirectories(t *testing.T) {
```
jchampio, line 99: The commit message should probably reference the removal of dead code here.

upgrade/directories.go
```diff
@@ -56,3 +56,14 @@ func TempDataDir(datadir, segPrefix string, id ID) string {
 
 	return filepath.Join(dir, newBase)
 }
+
+// ArchiveDirectoryForSource returns the archive directory for a given source cluster
+//   data directory.
+func ArchiveDirectoryForSource(dataDir string, upgradeID ID) string {
```
jchampio, line 62: nit: this function doesn't need to care against the higher concepts of source and target. It just does its thing.

agent/delete_data_directories_test.go
```diff
@@ -13,94 +13,79 @@ import (
...
+	}
 
-		if len(multiErr.Errors) != 4 {
-			t.Errorf("received %d errors, want %d", len(multiErr.Errors), 1)
-		}
+	t.Run("correctly reports failure of deleting a data directory", func(t *testing.T) {
+		dataDir1, dataDir2, tmpDir := testutils.SetupDataDirs(t)
+		defer func() {
+			os.RemoveAll(tmpDir)
+		}()
```
kalensk, line 72: nit: I don't believe this is needed.

dkrieger-pvtl: Why not?  The creation of a tempdir is [documented](https://golang.org/pkg/io/ioutil/#TempDir) to require the creator to manage the removal of the dir.  Otherwise, we leak tmpdirs onto the users filesystem.

agent/rename_directories.go
```diff
@@ -2,22 +2,50 @@ package agent
...
 	"github.com/greenplum-db/gpupgrade/utils"
 )
 
-func (s *Server) RenameDirectories(ctx context.Context, in *idl.RenameDirectoriesRequest) (*idl.RenameDirectoriesReply, error) {
+// The right-hand side functions are tested elsewhere, so to test RenameDataDirectories
+//   we only need to spy on the calls made in this function.  We enable this via global
+//   function injection.
+var (
+	IsPostgresFunc     = IsPostgres         //func IsPostgres(dataDir string) error
+	RenameDataDirsFunc = hub.RenameDataDirs //func RenameDataDirs(source, target string, upgradeID upgrade.ID) error
```
kalensk, line 21: nit: Seeing `hub.RenameDataDirs` makes me think this might be a better fit for the upgrade package as a place to share hub and agent code.

agent/rename_directories.go
```diff
@@ -2,22 +2,50 @@ package agent
...
+		//   any args, we make sure that we are called with proper arguments.  A non-existent dir path might
+		//   be ok for idempotence, hence we accept that here and defer to RenameDataDirsFunc().
+		if utils.DoesPathExist(dirs.Source) && !IsPostgresFunc(dirs.Source) {
+			mErr = multierror.Append(mErr, fmt.Errorf("could not delete non-postgres data dir: %s", dirs.Source))
+			continue
+		}
+		if utils.DoesPathExist(dirs.Target) && !IsPostgresFunc(dirs.Target) {
+			mErr = multierror.Append(mErr, fmt.Errorf("could not delete non-postgres data dir: %s", dirs.Target))
+			continue
+		}
```
jchampio, line 43: Is there a reason why these checks aren't performed as part of `hub.RenameDataDirs()`? That would make `agent.Server.RenameDataDirectories()` light enough that end-to-end verification would be fine.

agent/rename_directories.go
```diff
@@ -2,22 +2,50 @@ package agent
...
+	"github.com/greenplum-db/gpupgrade/upgrade"
 	"github.com/greenplum-db/gpupgrade/utils"
 )
 
-func (s *Server) RenameDirectories(ctx context.Context, in *idl.RenameDirectoriesRequest) (*idl.RenameDirectoriesReply, error) {
+// The right-hand side functions are tested elsewhere, so to test RenameDataDirectories
+//   we only need to spy on the calls made in this function.  We enable this via global
+//   function injection.
+var (
+	IsPostgresFunc     = IsPostgres         //func IsPostgres(dataDir string) error
```
kalensk, line 20: nit: Im not sure these comments on the function signatures is helping us with IDEs and modern code navigation utilities. I could see them getting out of date if the functions were to change.

agent/rename_directories_test.go
```diff
@@ -2,62 +2,159 @@ package agent_test
...
-	}
+		// set up expectations
+		var expectedSourceDirs, expectedTargetDirs, tmpDirs []string
+		var expectedUpgradeIDs []upgrade.ID
+		var renamePairs []*idl.RenameDataDirs
+		for i := 0; i < numPairs; i++ {
+			source, target, tmpDir := testutils.SetupDataDirs(t)
+			expectedSourceDirs = append(expectedSourceDirs, source)
+			if i == numPairs-1 {
+				// test the use case of a target of ""(target mirror/standby doesn't exist)
```
jchampio, line 36: Is this special case documented somewhere and I missed it? I don't know what the significance of an empty target is.

agent/rename_directories_test.go
```diff
@@ -2,62 +2,159 @@ package agent_test
...
+			}
+			expectedUpgradeIDs = append(expectedUpgradeIDs, testutils.UpgradeID)
+			tmpDirs = append(tmpDirs, tmpDir)
+			renamePairs = append(renamePairs,
+				&idl.RenameDataDirs{
+					Source: expectedSourceDirs[i],
+					Target: expectedTargetDirs[i],
+				})
+		}
+		// rename gets call with all args, but IsPostgres not with a target of ""
```
jchampio, line 49: related: this comment doesn't parse for me.

hub/rename_data_directories_test.go
```diff
@@ -2,75 +2,192 @@ package hub_test
...
+	"github.com/greenplum-db/gpupgrade/upgrade"
 	"github.com/greenplum-db/gpupgrade/utils"
 )
 
-func TestRenameDataDirs(t *testing.T) {
-	t.Run("renames both source and target", func(t *testing.T) {
-		numCalls := 0
-		utils.System.Rename = func(src, dst string) error {
-			if numCalls == 0 {
+// Since this simple function is heavily relied upon, test it thoroughly.
```
jchampio, line 22: This is an odd comment; it implies that less-used code should not be thoroughly tested maybe?

agent/delete_data_directories_test.go
kalensk: [utils.System.RemoveAll](https://github.com/greenplum-db/gpupgrade/blob/82ef8f0caf207e9340e5b4f6a9f522642f7d2c29/utils/sys_utils.go#L63) is just a function reference, where the production code will use the real `os.RemoveAll` and the test is reassigning.

hub/rename_data_directories_test.go
```diff
@@ -79,57 +196,57 @@ func TestRenameSegmentDataDirs(t *testing.T) {
...
-					Dst: "/data/dbfast2/seg2",
+			&idl.RenameDataDirectoriesRequest{
+				DataDirs: []*idl.RenameDataDirs{{
+					Source: "/data/dbfast2/seg2_123ABC",
+					Target: "/data/dbfast2/seg2",
 				}, {
-					Src: "/data/dbfast2/seg4_123ABC",
-					Dst: "/data/dbfast2/seg4",
+					Source: "/data/dbfast2/seg4_123ABC",
+					Target: "/data/dbfast2/seg4",
```
jchampio, line 246: (The code motion throughout this patch makes it much more difficult to review.)

agent/delete_data_directories_test.go
kalensk: Ah! Never mind, I guess reassigning it back is needed in case other tests rely on the actual implementation. In some ways I wish there was a global test cleanup.

hub/rename_data_directories.go
```diff
@@ -34,79 +45,116 @@ func UpdateDataDirectories(conf *Config, agentConns []*Connection) error {
...
-	m := make(RenameMap)
+// IsRenameErrorIdempotent interprets an error returned from os.Rename().  If that error is acceptable, it returns true.
+// The error code options are taken from the POSIX rename(2) manpage:
+//   https://pubs.opengroup.org/onlinepubs/009695399/functions/rename.html
+// These are consistent with the Mac OSX manpage and Linux manpage for rename(2.
+// Note that if rename(2) could return both ENOENT and (EEXIST,ENOTEMPTY),  the standard does
+// not specify which is returned.
+func IsRenameErrorIdempotent(err error) bool {
+	switch x := err.(type) {
+	case *os.LinkError:
```
jchampio, line 107: Use the `xerrors.As` workflow instead.

hub/rename_data_directories_test.go
```diff
@@ -2,75 +2,192 @@ package hub_test
...
-					t.Errorf("got %q want %q", src, expectedSrc)
+	cases := []struct {
+		name             string
+		clusterHasMirror bool
+		deleteSource     bool
+		deleteTarget     bool
+		deleteArchiveDir bool
+		expected         bool
+	}{
+		{name: "returns true when called properly setup with mirror",
```
jchampio, line 35: It's very difficult for me to understand what this test is supposed to be doing and how to maintain it. Maybe a table-driven test isn't what's needed here?

testutils/test_utils.go
```diff
@@ -1,13 +1,22 @@
...
+
+	"github.com/greenplum-db/gpupgrade/upgrade"
+	"github.com/greenplum-db/gpupgrade/utils"
 
 	"github.com/greenplum-db/gp-common-go-libs/dbconn"
 
 	"github.com/greenplum-db/gpupgrade/greenplum"
 )
 
+var UpgradeID = upgrade.ID(10) // "CgAAAAAAAAA" in base64
```
jchampio, line 18: What's special about 10?

agent/delete_data_directories_test.go
```diff
@@ -13,94 +13,79 @@ import (
...
-
-		_, err := server.DeleteDirectories(context.Background(), req)
+	cases := []struct {
+		name             string
+		checkIdempotence bool
+	}{
+		{"does delete directories",
+			false},
+		{"is idempotent",
+			true},
```
jchampio, line 35: Probably not a use case for table-driven tests. Just run the code without subtests.

agent/rename_directories.go
dkrieger-pvtl: There's no harm in applying these checks to the master segment as well, though we consider them unnecessary currently on the master.   They are there so that the agent doesn't just rename a "random" directory.  But it does clean up the code.

agent/rename_directories_test.go
```diff
@@ -2,62 +2,159 @@ package agent_test
...
+				expectedTargetDirs = append(expectedTargetDirs, target)
+			}
+			expectedUpgradeIDs = append(expectedUpgradeIDs, testutils.UpgradeID)
+			tmpDirs = append(tmpDirs, tmpDir)
+			renamePairs = append(renamePairs,
+				&idl.RenameDataDirs{
+					Source: expectedSourceDirs[i],
+					Target: expectedTargetDirs[i],
+				})
+		}
```
kalensk, line 48: I am having a difficult time following this....What is numPairs and why is it hardcoded? Why is there a special condition when `i == numPairs-1`? 

This block or test could be clearer.

agent/rename_directories_test.go
dkrieger-pvtl: An empty target means the source path represents a mirror/standby, which don't exist on the target cluster yet.  I played around with various function signatures, and though it's ugly in this test, in the implementation code it seems pretty clean to me to do it this way.

agent/rename_directories_test.go
dkrieger-pvtl: yeah...a target dir of "" means the target lacks the corresponding segment that's on the source(since mirror/standby aren't on the target yet), but in agent code we check for an empty dir before calling `IsPostgresFunc()`.

agent/utils.go
dkrieger-pvtl: Oversight; good idea.

agent/utils_test.go
dkrieger-pvtl: It's in `f6425a5d1e125b11bd1f67dea3fa65a9f67963e5`(`f6425a5 test: utilities for unit testing renaming`).   It's a test function, so I added it to a different commit.

testutils/test_utils.go
dkrieger-pvtl: It's not `0`...I wanted to use a value where an accidental use of a default initialized ID would fail in the tests.  I can add a comment.

upgrade/directories.go
dkrieger-pvtl: What would you recommend for naming?  In our code, it should only be used for the purpose for which it is named, so that's why I didn't given it a generic name.

agent/delete_data_directories_test.go
kalensk: +1

upgrade/directories.go
jchampio: > In our code, it should only be used for the purpose for which it is named,

In the end, that's not really up to you; it's up to the person writing the clients of this code. Naming functions this way not only increases verbosity, it leads to strange copy-paste situations where we have `HammerForDrivingNails` and `HammerForPullingNails` and `HammerForUseAsAPaperweight` when a single `Hammer` does all those things.

Make the implementation work consistently and document it appropriately, then trust future developers to use the tools in ways that are helpful.

testutils/test_utils.go
jchampio: If a particular test needs to verify that an implementation doesn't do something accidentally, that seems like something that should be baked into the test. It requires black-box knowledge of the implementation.

agent/utils_test.go
jchampio: That commit comes later, though.

agent/rename_directories.go
dkrieger-pvtl: right...I'll remove them...it was an overside.

hub/rename_data_directories_test.go
dkrieger-pvtl: This comment has been made in prior PRs, and I'm not sure of the best way to handle it.  In this case, I think the code motion was from a significant amount of refactoring of the existing code.  If I stuck to trying to make the diff easy to review, it would constrain this.  For files with a lot of diffs, how do I make the tradeoff between easy to review diffs and what I think is cleaner code?  For files with a lot of diffs, a review of the new file itself might be better than looking at a diff...

hub/rename_data_directories.go
```diff
@@ -34,79 +46,128 @@ func UpdateDataDirectories(conf *Config, agentConns []*Connection) error {
...
-		sourceSeg := source.Primaries[content]
+	return false
+}
 
-		host := targetSeg.Hostname
-		m[host] = append(m[host], &idl.RenamePair{
-			Src: targetSeg.DataDir,
-			Dst: sourceSeg.DataDir,
-		})
+func AlreadyRenamed(source, target, archive string, noTargetMirror bool) bool {
```
marksliva, line 120: Is `noTargetMirror` supposed to be a bool or a string? It looks like a string gets passed in by the caller.

dkrieger-pvtl:
> I'm finding this pretty hard to review. It maybe needs to be pared down into multiple PR passes, depending on what parts are critical for idempotence and what's just additional protections that are being added.

This PR is being closed to make multiple PR passes...the code here has been archived at https://github.com/dkrieger-pvtl/gpupgrade/tree/idempotent_update_data_dirs_PR307.

agent/delete_data_directories_test.go
```diff
@@ -13,94 +13,79 @@ import (
...
 
-	t.Run("fails to remove one segment data directory", func(t *testing.T) {
-		expected := errors.New("permission denied")
-		expectedDataDirectories := []string{"/data/dbfast_mirror1/seg1", "/data/dbfast_mirror2/seg2"}
-		actualDataDirectories :=  []string{}
+		actualDataDirectories := []string{}
 		utils.System.RemoveAll = func(name string) error {
 			actualDataDirectories = append(actualDataDirectories, name)
-			if name == "/data/dbfast_mirror1/seg1" {
+			if name == dataDir2 {
```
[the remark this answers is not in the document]
dkrieger-pvtl: That test is to introduce the error `EPERM` directly, though I could explicitly create a file with `0000` permissions, I guess.

agent/rename_directories.go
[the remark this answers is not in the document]
dkrieger-pvtl: right, I plan to remove them.  Kalen made the same comment, but it somehow is showing up in a different place in the UI....

hub/rename_data_directories.go
```diff
@@ -34,79 +46,128 @@ func UpdateDataDirectories(conf *Config, agentConns []*Connection) error {
...
 }
 
-func getSourceRenameMap(source *greenplum.Cluster, primariesOnly bool) RenameMap {
+// getRenameMap() splices together the source and target clusters by combining the corresponding segment from
+//   each cluster.  It does so per-host.
+// TODO: Do we want to sanity-check that the source and target clusters "match"?  At this point in finalize,
+//   this should be a reasonable assumption.
+func getRenameMap(source *greenplum.Cluster, target InitializeConfig, sourcePrimariesOnly bool) RenameMap {
 	m := make(RenameMap)
+	tMap := make(map[int]string)
```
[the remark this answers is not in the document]
dkrieger-pvtl: to my eyes, having short names makes the code easier to read.  I guess `targetMap` makes it easier to read, but `m` seems simpler since its type is clear.

hub/rename_data_directories.go
```diff
@@ -34,79 +46,128 @@ func UpdateDataDirectories(conf *Config, agentConns []*Connection) error {
...
 		}
 
 		seg, ok := source.Mirrors[content]
-		if !primariesOnly && ok {
-			m[seg.Hostname] = append(m[seg.Hostname], &idl.RenamePair{
-				Src: seg.DataDir,
-				Dst: seg.DataDir + oldSuffix,
+		if !sourcePrimariesOnly && ok {
+			m[seg.Hostname] = append(m[seg.Hostname], &idl.RenameDataDirs{
+				Source: seg.DataDir,
```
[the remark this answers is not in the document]
dkrieger-pvtl: @marksliva I'm not sure what you mean here...the comment refers to code on the agent but is tagged to the hub.

hub/rename_data_directories.go
dkrieger-pvtl: `bool`...it gets used as such.

hub/rename_data_directories_test.go
kalensk: I think "pre-factor" PRs can help a lot with this. That is, a PR that just focuses on the minimal refactors needed for an upcoming PR. Being sure to keep things small and self-contained.

jchampio: > If I stuck to trying to make the diff easy to review, it would constrain this.

It doesn't constrain the refactor, it just constrains _when_ you perform that refactor. Seeing `Src` being renamed to `Source` in the same commit as significant implementation changes adds noise to both sides of the review -- I am more likely to miss problems in the implementation, and to miss incorrect refactor changes, because my attention is split between both.

hub/rename_data_directories.go
marksliva: oh sorry, I misread this line. For some reason, I was reading this as a double initialization.

hub/rename_data_directories.go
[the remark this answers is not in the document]
marksliva: I see how it works now. We could add a boolean to make it easier to follow.

PR #310, opened 2020-04-09
Add revert command for the initialize step

marksliva:
The revert command is only being tested after running a full initialize,
however it should be idempotent enough to handle reverting from any substep or
failure encountered during initialize. The revert process includes:
- delete segment and standby datadirs
- stop agents
- delete master datadir
- stop hub
- delete gpupgrade state directory


test/revert.bats
```diff
@@ -0,0 +1,83 @@
...
+        --source-bindir="$GPHOME/bin" \
+        --target-bindir="$GPHOME/bin" \
+        --source-master-port="${PGPORT}" \
+        --temp-port-range 6020-6040 \
+        --disk-free-ratio 0 \
+        --verbose 3>&-
+
+    # parse config.json for the datadirs
+    local target_datadirs=$(jq ".Target.Primaries[].DataDir" "${GPUPGRADE_HOME}/config.json")
+
```
kalensk, line 56: Should this be exposed via `gpupgrade config show --target-datadirs` instead? That way it follows the idiom we have been using throughout the BATs tests.

marksliva: Potentially. However, we would have to add a new flag to the config that returns all target datadirs, since the only datadir flag `target-datadir` returns only the masterDataDir
```
	case "target-datadir":
		resp.Value = s.Target.MasterDataDir()
```
Other tests that have needed the target datadirs have queried the database for the datadirs. However, the reason we did not do that was because the target cluster is shut down at the end of initialize.

kalensk: Wouldn't using a different variable be easier than a flag. That is `target-datadir` and `target-directories` or something?

marksliva: I'm not sure what you mean by easier. Either way, we'd have to call config with a specific flag or arg regardless.

hub/delete_master_data_directory.go
```diff
@@ -0,0 +1,44 @@
...
+	_, err := utils.System.Stat(masterDataDir)
+	if err != nil {
+		if xerrors.Is(err, os.ErrNotExist) {
+			return nil
+		}
+		mErr = multierror.Append(mErr, err)
+		return mErr
+	}
+
+	for _, fileName := range agent.PostgresFiles {
```
berlin-ab, line 25: This seems to be duplication between the agent code. Also, the `hub` package shouldn't have a dependency on the `agent` for this work.

test/revert.bats
marksliva: We checked with product, and they don't think any customers need this, so there's not much reason to add it to the binary. In our test, it is a one-liner, so we'll stick with that for now.

kalensk: Sure, but we are deviating from previous patterns and I guess I have a strong preference for consistency.

For example, a previous [PR](https://github.com/greenplum-db/gpupgrade/pull/179) introduced a new [gpupgrade config](https://github.com/greenplum-db/gpupgrade/commit/1fca144a08b470c9bcfb83f5122e5a619ea19cdd#diff-ab4ba84babe11ee4bb488285f23e46f8
) option simply for testing such as in [pg_ugprade.bats]( https://github.com/greenplum-db/gpupgrade/blob/140885913865233367ce55cf0b36b1288ad4ccd2/test/pg_upgrade.bats#L59) and other places. 

I guess my comment is pointing out that now we have two different ways of getting information used in our tests: 1) gpupgrade config show, and 2) using jq to parse the config file.

agent/delete_data_directories.go
```diff
@@ -7,6 +7,10 @@ import (
 	"github.com/greenplum-db/gpupgrade/idl"
 	"github.com/greenplum-db/gpupgrade/upgrade"
 )
+// todo: needs to go in its own file
+func (s *Server) DeleteDirectory(ctx context.Context, in *idl.DeleteDirectoryRequest) (*idl.DeleteDirectoryReply, error) {
+	return nil, nil
+}
```
kalensk, line 13: I know this is still in progress so pardon this comment if its already being worked on. Similar to delete_data_directories.go we will probably want to add an additional safety check to ensure we are deleting the appropriate directory.

idl/hub_to_agent.proto
```diff
@@ -36,8 +37,12 @@ message UpgradePrimariesReply {}
 message DeleteDirectoriesRequest {
   repeated string datadirs = 1;
 }
+message DeleteDirectoryRequest {
+  string directory = 1;
+}
```
kalensk, line 42: I think we should consider calling this `DeleteStateDirectory` to emphasize that it won't/shouldn't delete any arbitrary directory for safety/security reasons. Rather it should check and make sure to delete only sate directories.

dkrieger-pvtl:
> The revert command is only being tested after running a full initialize,
> however it should be idempotent enough to handle reverting from any substep or
> failure encountered during initialize.

I think you mean to say that this story is only handling revert after a full initialize has completed, right?  Recovering from a failure mid-initialize is more work(might have to stop postgres processes, etc), and will be handled later.  As will saving off the state directory.

Note the gpAdminLogs are not touched by this PR, but I suppose we consider those system logs and hence we do not need to delete them.

agent/delete_directories_test.go
```diff
@@ -0,0 +1,102 @@
...
+	"reflect"
+	"testing"
+
+	"github.com/greenplum-db/gp-common-go-libs/testhelper"
+	"github.com/greenplum-db/gpupgrade/agent"
+	"github.com/greenplum-db/gpupgrade/idl"
+	"github.com/greenplum-db/gpupgrade/utils"
+)
+
+func TestDeleteDirectories(t *testing.T) {
```
dkrieger-pvtl, line 15: Consider creating actual directories here to test...you can use a temp dir and create the files you need underneath.  A strict mock-style test pattern would mock out the system calls, but in this case, you can test the actual functionality directly in a straightforward manner.

test/revert.bats
jchampio: > We checked with product, and they don't think any customers need this, so there's not much reason to add it to the binary.

`config show` isn't for users, it's for us (and potentially support). It exists to help us partially decouple hub internals from the tests that rely on configuration values like this.

upgrade/directories.go
```diff
@@ -56,3 +59,30 @@ func TempDataDir(datadir, segPrefix string, id ID) string {
 
 	return filepath.Join(dir, newBase)
 }
+
+func DeleteDirectories(directoriesToDelete []string, filesThatMustExist []string) *multierror.Error {
+	var mErr *multierror.Error
+	for _, segDataDir := range directoriesToDelete {
+
+		var postgresFilesError *multierror.Error
```
kalensk, line 67: I might be misreading, but why the need for this variable rather than keep things simple and just use `mErr`? Why separate the errors to just combine them in `mErr` anyway?

upgrade/directories_test.go
```diff
@@ -48,3 +55,112 @@ func ExampleTempDataDir() {
...
+				return expected
+			}
+			return nil
+		}
+
+		statCalls := 0
+		utils.System.Stat = func(name string) (os.FileInfo, error) {
+			statCalls++
+			return nil, nil
+		}
```
kalensk, line 139: I think this test should be stronger. Something like verifying that stat is being called for on the inputs which in this case is foo and bar for each of the number of directories.

hub/revert.go
```diff
@@ -0,0 +1,48 @@
...
+			gplog.Error(fmt.Sprintf("revert: %s", err))
+		}
+	}()
+
+	err = DeleteSegmentAndStandbyDirectories(s.agentConns, s.Config.Target)
+	if err != nil {
+		return err
+	}
+
+	// todo: DeleteStateDirectories()
```
dkrieger-pvtl, line 35: Did you mean to fix this as part of this PR?  

One potential issue if it is enabled: how do we know no call after this one will, in fact, attempt to read/write the state directory?

agent/delete_directories.go
```diff
@@ -0,0 +1,27 @@
...
+
+	mErr := upgrade.DeleteDirectories([]string{in.Directory}, stateDirectoryFiles)
+	return &idl.DeleteStateDirectoryReply{}, mErr.ErrorOrNil()
+}
+
+func (s *Server) DeleteDirectories(ctx context.Context, in *idl.DeleteDirectoriesRequest) (*idl.DeleteDirectoriesReply, error) {
+	gplog.Info("got a request to delete data directories from the hub")
+
+	mErr := upgrade.DeleteDirectories(in.Datadirs, postgresFiles)
+	return &idl.DeleteDirectoriesReply{}, mErr.ErrorOrNil()
```
kalensk, line 26: Can these be renamed to `DeleteDataDirectories` for clarity? I find `DeleteDirectories` to be ambiguous when trying to follow the code.

hub/revert.go
kalensk: > One potential issue if it is enabled: how do we know no call after this one will, in fact, attempt to read/write the state directory?

Aren't we in control of that? That is, wouldn't that be a developer bug?

idl/hub_to_agent.proto
```diff
@@ -36,8 +37,12 @@ message UpgradePrimariesReply {}
 message DeleteDirectoriesRequest {
```
dkrieger-pvtl, line 37: nit: This message(and the corresponding rpc call and req/reply fields) should probably be called `DeleteDataDirectories`.  It's not part of this PR, though.

upgrade/directories.go
bhuvnesh2703, line 67: now its not only postgresFiles, it could either be the state files? so we should use a generic name.

dkrieger-pvtl: Also, this function is generic, so a name like `missingFilesError` would likely be better...the state dir isn't a postgres dir...

agent/delete_directories.go
```diff
@@ -0,0 +1,27 @@
...
+var stateDirectoryFiles = []string {"config.json", "status.json"}
+
+func (s *Server) DeleteStateDirectory(ctx context.Context, in *idl.DeleteStateDirectoryRequest) (*idl.DeleteStateDirectoryReply, error) {
+	gplog.Info("got a request to delete the state directory from the hub")
+
+	mErr := upgrade.DeleteDirectories([]string{in.Directory}, stateDirectoryFiles)
+	return &idl.DeleteStateDirectoryReply{}, mErr.ErrorOrNil()
+}
+
+func (s *Server) DeleteDirectories(ctx context.Context, in *idl.DeleteDirectoriesRequest) (*idl.DeleteDirectoriesReply, error) {
```
bhuvnesh2703, line 22: do you want the name to be `DeleteDataDirectories` similar to DeleteStateDirectory so that its obvious from the name what it does..?

agent/delete_directories_test.go
kalensk: I think I would prefer this as well. Seems quite simple and "low cost" to have a temp dir, and more "realistic".

upgrade/directories_test.go
```diff
@@ -48,3 +55,112 @@ func ExampleTempDataDir() {
 	// /data/standby.AAAAAAAAAAA
 	// /data/primary/seg.AAAAAAAAAAA.3
 }
+
+func TestDeleteDataDirectories(t *testing.T) {
```
dkrieger-pvtl, line 59: Consider creating actual directories here to test...you can use a temp dir and create the files you need underneath.  A strict mock-style test pattern would mock out the system calls, but in this case, you can test the actual functionality directly in a straightforward manner.

cli/commands/commands.go
```diff
@@ -109,6 +111,7 @@ func BuildRootCommand() *cobra.Command {
 	root.AddCommand(initialize())
 	root.AddCommand(execute())
 	root.AddCommand(finalize())
+	root.AddCommand(revert())
```
bhuvnesh2703, line 114: how do we know what we plan to revert, the work done by initialize or the work done by execute?

cli/commanders/steps.go
```diff
@@ -157,6 +156,28 @@ func Finalize(client idl.CliToHubClient, verbose bool) error {
...
+	fmt.Println("Revert in progress.")
+	fmt.Println()
+
+	stream, err := client.Revert(context.Background(), &idl.RevertRequest{})
+	if err != nil {
+		gplog.Error(err.Error())
+		return err
+	}
+
+	_, err = UILoop(stream, false)
```
dkrieger-pvtl, line 170: maybe add a // TODO: add verbose option

cli/commanders/steps.go
```diff
@@ -157,6 +156,28 @@ func Finalize(client idl.CliToHubClient, verbose bool) error {
...
+	if err != nil {
+		return xerrors.Errorf("Revert: %w", err)
+	}
+
+	fmt.Println("")
+	fmt.Printf("The target cluster is now restored to its original state. [more information TBD]\n")
```
dkrieger-pvtl, line 176: There is no longer a concept of a target cluster after a successful revert.  Do you mean to say _The source cluster_...?

test/revert.bats
```diff
@@ -0,0 +1,83 @@
...
+    while read -r datadir; do
+        if [ -d $(echo "${datadir}" | tr -d '"') ]; then
+            echo "expected datadir ${datadir} to have been deleted"
+            exit 1
+        fi
+    done <<< "${target_datadirs}"
+}
+
+@test "the state directory gets deleted" {
```
bhuvnesh2703, line 67: instead of doing the initialize 3 times (with stop-before-cluster-creation/or not) why not do it once and have all the validations...

hub/delete_data_directories.go
```diff
@@ -12,20 +12,36 @@ import (
 )
 
 func DeleteMirrorAndStandbyDirectories(agentConns []*Connection, cluster *greenplum.Cluster) error {
+	return DeleteDataDirectories(agentConns, cluster, false)
+}
+
+func DeleteSegmentAndStandbyDirectories(agentConns []*Connection, cluster *greenplum.Cluster) error {
```
dkrieger-pvtl, line 18: We don't have a great vocabulary for describing the role of segments, but the intent here is to delete ALL data directories other than master.  So I think something like `DeleteAllDataDirectoriesButNotMaster` is clearer.  The current name leaves me wondering what is left after it finishes.

ci/tasks/install-tests.yml
```diff
@@ -4,7 +4,7 @@ image_resource:
   type: docker-image
   source:
     repository: pivotaldata/centos-gpdb-dev
-    tag: "6-gcc6.2-llvm3.7"
+    tag: "7-gcc6.2-llvm3.7"
```
dkrieger-pvtl, line 7: Is this meant to be a diff as part of this PR?

cli/commanders/steps.go
jchampio, line 176: There's a sample output linked to in the story, or is that for a follow-up PR?

cli/commands/commands.go
dkrieger-pvtl: `gpupgrade revert` should undo all of the work involved in the current upgrade, whether part of initialize or execute.  The result should be the original source cluster.  For finalize, I think we require manual user intervention to recover the source, as it's "gone".

cli/commands/commands.go
```diff
@@ -447,6 +450,40 @@ func finalize() *cobra.Command {
...
+		Long:  "reverts the upgrade and returns the cluster to its original state",
+		Run: func(cmd *cobra.Command, args []string) {
+			client := connectToHub()
+			err := commanders.Revert(client)
+			if err != nil {
+				gplog.Error(err.Error())
+				os.Exit(1)
+			}
+
+			// todo: is there a way to instruct the hub to interrupt its own process gracefully?
```
jchampio, line 466: Isn't that what we added `hub.Server.StopServices()` to do?

hub/delete_state_directories_test.go
```diff
@@ -0,0 +1,95 @@
...
+	"testing"
+
+	"github.com/golang/mock/gomock"
+	"github.com/greenplum-db/gp-common-go-libs/testhelper"
+	"github.com/greenplum-db/gpupgrade/hub"
+	"github.com/greenplum-db/gpupgrade/idl"
+	"github.com/greenplum-db/gpupgrade/idl/mock_idl"
+	"github.com/greenplum-db/gpupgrade/utils"
+	"github.com/hashicorp/go-multierror"
+	"golang.org/x/xerrors"
```
jchampio, line 14: `goimports`, here and elsewhere.

hub/delete_data_directories.go
```diff
@@ -12,20 +12,36 @@ import (
...
-				(seg.Role == greenplum.MirrorRole)
+		filterFunc := func(seg *greenplum.SegConfig) bool {
+			if seg.Hostname != conn.Hostname {
+				return false
+			}
+
+			// we don't have mirrors at this point after initialize, but we may have them later
+			// if the finalize step crashes after adding mirrors or a standby
+			if includePrimaries {
+				return !(seg.ContentID == -1 && seg.Role == greenplum.PrimaryRole)
```
dkrieger-pvtl, line 37: nit: `!seg.IsMaster()` might be clearer here.

hub/delete_state_directories.go
```diff
@@ -0,0 +1,51 @@
...
+	for _, conn := range agentConns {
+		conn := conn
+
+
+		if conn.Hostname == masterHostName {
+			continue
+		}
+
+		wg.Add(1)
+		go func(c *Connection) {
```
jchampio, line 29: `conn` is already pinned in scope.

hub/delete_state_directories.go
```diff
@@ -0,0 +1,51 @@
...
+	"context"
+	"sync"
+
+	"github.com/greenplum-db/gp-common-go-libs/gplog"
+	"github.com/greenplum-db/gpupgrade/idl"
+	"github.com/greenplum-db/gpupgrade/utils"
+	"github.com/hashicorp/go-multierror"
+)
+
+//func (s *Server) DeleteDirectories(ctx context.Context, in *idl.DeleteDirectoriesRequest) (*idl.DeleteDirectoriesReply, error) {
```
dkrieger-pvtl, line 13: nit: remove this line and the next one.

hub/revert.go
```diff
@@ -0,0 +1,48 @@
...
+
+	err = s.StopAgents()
```
jchampio, line 37: Speaking of `StopServices()`, this seems like a duplication of that logic.

hub/delete_state_directories.go
```diff
@@ -0,0 +1,51 @@
...
+//func (s *Server) DeleteDirectory(ctx context.Context, in *idl.DeleteDirectoryRequest) (*idl.DeleteDirectoryReply, error) {
+func DeleteStateDirectories(agentConns []*Connection, masterHostName string) error {
+	wg := sync.WaitGroup{}
+	errChan := make(chan error, len(agentConns))
+	stateDir := utils.GetStateDir()
+
+	for _, conn := range agentConns {
+		conn := conn
+
```
dkrieger-pvtl, line 22: nit: remove this blank line.

hub/delete_state_directories.go
```diff
@@ -0,0 +1,51 @@
...
+
+		if conn.Hostname == masterHostName {
+			continue
+		}
```
jchampio, line 26: Some explanation of why would be good.

hub/delete_state_directories_test.go
```diff
@@ -0,0 +1,95 @@
...
+			}
+
+			err := hub.DeleteStateDirectories(agentConns, "master")
+			if err != nil {
+				t.Errorf("unexpected err %#v", err)
+			}
+		})
+	})
+
+	t.Run("DeleteDataDirectories", func(t *testing.T) {
```
dkrieger-pvtl, line 54: This should be "DeleteStateDirectories"

hub/revert.go
dkrieger-pvtl: Yes, we are in control of that, but how do we design the code to make it easy to do?  It's a dynamic thing...any developer needs to know which function can be called "after revert".  For instance, if I were working on the internals of the next call(`s.StopAgents`) as part of the upcoming pidfile work, how would I know this?  Maybe an annotation-like comment?  There shouldn't be too many things we do after revert.

cli/commands/commands.go
kalensk: This PR is just a first step towards a fully functional customer facing revert command.

dkrieger-pvtl:
Overall, I think this PR is well-designed and tested.  But I have a lot of code comments/suggestions(see comments I've made).  I also haven't looked at the tests yet.

test/revert.bats
kalensk: I was thinking the same thing...

cli/commands/commands.go
```diff
@@ -447,6 +450,40 @@ func finalize() *cobra.Command {
...
+			err = exec.Command("pkill", "-f", "gpupgrade hub").Run()
```
dkrieger-pvtl, line 467: I wonder if the technique in agent/server.go::StopAgent() used to gracefully stop the agent will work here.  Although I think we might want to switch to pid files and a process-level lifecycle for the hub and agents anywa.

hub/delete_data_directories.go
kalensk: FWIW, I tend prefer to phrase things _explicitly_ and in the positive such as `DeleteSegmentAndStandbyDataDirectories`. This way it makes it explicit and clear what is being operated on, rather than "notMaster" which is more implicit and opaque to me.

ci/tasks/install-tests.yml
kalensk: I believe it was to include `jq` for testing, but that is under discussion in comment https://github.com/greenplum-db/gpupgrade/pull/310#discussion_r406885951. I am not sure if jq is needed in other places.

hub/delete_data_directories.go
kalensk: +1

hub/revert.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/greenplum-db/gp-common-go-libs/gplog"
+	"github.com/greenplum-db/gpupgrade/upgrade"
+	"github.com/hashicorp/go-multierror"
+
+	"github.com/greenplum-db/gpupgrade/idl"
+	"github.com/greenplum-db/gpupgrade/step"
+)
+
+func (s *Server) Revert(_ *idl.RevertRequest, stream idl.CliToHub_RevertServer) (err error) {
+	st, err := step.Begin(s.StateDir, "revert", stream)
```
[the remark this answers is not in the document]
kalensk: Agreed. I am not sure trying to make this fit the` step.Steap` pattern is helping us here, and in fact may not actually be helping.

hub/delete_state_directories.go
jchampio: (run the autoformatter please)

marksliva:
Closing this PR, and making some structural changes before opening a new PR.

cli/commands/commands.go
marksliva: I ended up re-using the `StopServices` that `gpupgrade kill-services` uses, which was Jacob's suggestion.

hub/revert.go
marksliva: Fixed.

hub/revert.go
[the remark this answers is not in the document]
marksliva: Done.

hub/revert.go
```diff
@@ -0,0 +1,48 @@
...
+	}
+
+	// todo: DeleteStateDirectories()
+
+	err = s.StopAgents()
+	if err != nil {
+		return err
+	}
+
+	err = upgrade.DeleteDirectories([]string{s.Config.Target.MasterDataDir()}, nil)
```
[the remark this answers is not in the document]
marksliva: Not exactly. The latest implementation passes in `PostgresFiles` for added safety.

test/revert.bats
```diff
@@ -0,0 +1,83 @@
...
+        --target-bindir="$GPHOME/bin" \
+        --source-master-port="${PGPORT}" \
+        --temp-port-range 6020-6040 \
+        --disk-free-ratio 0 \
+        --stop-before-cluster-creation \
+        --verbose 3>&-
+
+    gpupgrade revert
+
+    if [ -d $(echo "${GPUPGRADE_HOME}") ]; then
```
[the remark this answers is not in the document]
marksliva: Fixed.

test/revert.bats
```diff
@@ -0,0 +1,83 @@
...
+        --disk-free-ratio 0 \
+        --verbose 3>&-
+
+    # parse config.json for the datadirs
+    local target_datadirs=$(jq ".Target.Primaries[].DataDir" "${GPUPGRADE_HOME}/config.json")
+
+    gpupgrade revert
+
+    while read -r datadir; do
+        if [ -d $(echo "${datadir}" | tr -d '"') ]; then
```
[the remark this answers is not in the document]
marksliva: Done.

test/revert.bats
```diff
@@ -0,0 +1,83 @@
...
+        --temp-port-range 6020-6040 \
+        --disk-free-ratio 0 \
+        --stop-before-cluster-creation \
+        --verbose 3>&-
+
+    if [[ $(process_is_running "[g]pupgrade hub") -eq 0 ]]; then
+        # todo: this makes a single host assumption
+        echo 'expected hub to be running'
+        exit 1
+    fi
```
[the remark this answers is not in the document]
marksliva: Fixed.

agent/delete_directories.go
```diff
@@ -0,0 +1,27 @@
...
+
+import (
+	"context"
+
+	"github.com/greenplum-db/gp-common-go-libs/gplog"
+	"github.com/greenplum-db/gpupgrade/idl"
+	"github.com/greenplum-db/gpupgrade/upgrade"
+)
+
+// todo: add back the [...]
```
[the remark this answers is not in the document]
marksliva: Done. Now I am passing these slices in like this: `stateDirectoryFiles[:]`.

hub/delete_state_directories.go
marksliva: Done.

hub/delete_state_directories.go
marksliva: Done.

hub/delete_data_directories.go
marksliva: Good catch. Done.

ci/tasks/install-tests.yml
marksliva: Yes, this was done so we could use `jq`, which is isolated to just `revert.bats` for now.

hub/delete_data_directories.go
marksliva: It seems fine to list the 2 things that do want to delete. If there were 3 things, saying what we are not deleting would start being a better option.

test/revert.bats
marksliva: Done. This was the result of TDDing the feature, but since these are long running integration tests, it does make sense to consolidate them when possible.

cli/commanders/steps.go
marksliva: Good catch.

cli/commanders/steps.go
marksliva: Done.

agent/delete_directories.go
marksliva: Fixed.

upgrade/directories.go
marksliva: Fixed.

idl/hub_to_agent.proto
marksliva: This is a good idea though. It'll make things easier to follow. I changed it.

agent/delete_directories.go
marksliva: Renamed the server function. But leaving the `upgrade.DeleteDirectories` as is since it can delete either state directories or data directories.

hub/revert.go
marksliva: I had commented this mainly because I was still thinking about how it would be tested in a multi-host setting. Which it turns out we can use the multi-host vagrant code to do that. I added it back in.

upgrade/directories.go
marksliva: The reason it is a separate multi error from `mErr` is so that it can catch all missing files, and if it finds at least one missing file, all the errors get appended to `mErr`. Either way, I renamed this to `missingFilesError`.

idl/hub_to_agent.proto
marksliva: Done.

agent/delete_data_directories.go
marksliva: Done.

test/revert.bats
marksliva: The jq command can be written like this to extract hostname info as well:
`jq -r '.Target.Primaries[] | .Hostname + " " + .DataDir' "${GPUPGRADE_HOME}/config.json"`
So that can still work.

hub/delete_master_data_directory.go
marksliva: This has been deduplicatd.

cli/commands/commands.go
marksliva: Good point. Initially, we thought we needed to control the stopping of the services before removing the state dirs. But it turns out we can reuse the `StopServices` logic.

hub/delete_state_directories.go
marksliva: Done.

hub/delete_state_directories_test.go
marksliva: Fixed.

marksliva: Fixed.

agent/delete_directories_test.go
marksliva: Done.

> > The revert command is only being tested after running a full initialize,
> > however it should be idempotent enough to handle reverting from any substep or
> > failure encountered during initialize.
> 
> I think you mean to say that this story is only handling revert after a full initialize has completed, right? Recovering from a failure mid-initialize is more work(might have to stop postgres processes, etc), and will be handled later. As will saving off the state directory.
> 
> Note the gpAdminLogs are not touched by this PR, but I suppose we consider those system logs and hence we do not need to delete them.

Sure. It will be useful to do a more thorough pass at idempotence at some later point. I updated the commit message to reflect this.

upgrade/directories_test.go
marksliva: Since this test is now integrated to the file system, this comment is no longer relevant.

cli/commands/commands.go
marksliva: Yes. In the long term, `revert` will be able to revert from any point up until finalize. This PR is just handling revert after initialize.

PR #314, opened 2020-04-16
Add revert command for the initialize step

marksliva:
The revert command can be run after completing the initialize step. More
granular idempotence will be added later on.

The revert process includes:

- delete all target datadirs
- delete non-master host state directories
- delete master host state directory
- stop hub and agents

This PR addresses all of the comments in this original closed one: https://github.com/greenplum-db/gpupgrade/pull/310


agent/delete_directories.go
```diff
@@ -0,0 +1,27 @@
...
+	gplog.Info("got a request to delete the state directory from the hub")
+
+	err := upgrade.DeleteDirectories([]string{in.Directory}, stateDirectoryFiles[:])
+	return &idl.DeleteStateDirectoryReply{}, err
+}
+
+func (s *Server) DeleteDataDirectories(ctx context.Context, in *idl.DeleteDataDirectoriesRequest) (*idl.DeleteDataDirectoriesReply, error) {
+	gplog.Info("got a request to delete data directories from the hub")
+
+	err := upgrade.DeleteDirectories(in.Datadirs, PostgresFiles[:])
```
kalensk, line 25: nit: Any reason not to simply pass `PostgresFiles` directly (ie: not `PostgresFiles[:]`) by not using the ellipse notation when creating the slice literal. Similarly for `stateDirectoryFiles`.

agent/delete_directories_test.go
```diff
@@ -0,0 +1,103 @@
...
+	"github.com/greenplum-db/gpupgrade/agent"
+	"github.com/greenplum-db/gpupgrade/idl"
+	"github.com/greenplum-db/gpupgrade/utils"
+)
+
+func TestDeleteDirectories(t *testing.T) {
+	server := agent.NewServer(agent.Config{
+		Port:     -1,
+		StateDir: "",
+	})
```
kalensk, line 20: nit: Can just use an empty config for these tests. `server := agent.NewServer(agent.Config{})`

agent/delete_directories_test.go
```diff
@@ -0,0 +1,103 @@
...
+	"testing"
+
+	"github.com/greenplum-db/gp-common-go-libs/testhelper"
+
+	"github.com/greenplum-db/gpupgrade/agent"
+	"github.com/greenplum-db/gpupgrade/idl"
+	"github.com/greenplum-db/gpupgrade/utils"
+)
+
+func TestDeleteDirectories(t *testing.T) {
```
kalensk, line 16: Rename to `TestDeleteDataDirectories` for clarity and consistency.

agent/delete_directories.go
marksliva: If you get rid of the ellipses when passing `PostgresFiles` into the function, there's a compilation error:
`err := upgrade.DeleteDirectories(in.Datadirs, PostgresFiles)`
...
`cannot use PostgresFiles (type [2]string) as type []string in argument to upgrade.DeleteDirectories`

kalensk: Works for me with the following when changing both the variable `var PostgresFiles = []string{"postgresql.conf", "PG_VERSION"}` and the caller `upgrade.DeleteDirectories(in.Datadirs, PostgresFiles)`.

```go

var PostgresFiles = []string{"postgresql.conf", "PG_VERSION"}
var stateDirectoryFiles = []string{"config.json", "status.json"}

func (s *Server) DeleteStateDirectory(ctx context.Context, in *idl.DeleteStateDirectoryRequest) (*idl.DeleteStateDirectoryReply, error) {
	gplog.Info("got a request to delete the state directory from the hub")

	err := upgrade.DeleteDirectories([]string{in.Directory}, stateDirectoryFiles)
	return &idl.DeleteStateDirectoryReply{}, err
}

func (s *Server) DeleteDataDirectories(ctx context.Context, in *idl.DeleteDataDirectoriesRequest) (*idl.DeleteDataDirectoriesReply, error) {
	gplog.Info("got a request to delete data directories from the hub")

	err := upgrade.DeleteDirectories(in.Datadirs, PostgresFiles)
	return &idl.DeleteDataDirectoriesReply{}, err
}

```

cli/commands/commands.go
```diff
@@ -542,22 +578,30 @@ var killServices = &cobra.Command{
...
-			errCode := grpcStatus.Code(err)
-			errMsg := grpcStatus.Convert(err).Message()
-			// XXX: "transport is closing" is not documented but is needed to uniquely interpret codes.Unavailable
-			// https://github.com/grpc/grpc/blob/v1.24.0/doc/statuscodes.md
-			if errCode != codes.Unavailable || errMsg != "transport is closing" {
-				return err
-			}
-			return nil
+		err, done := stopAllServices()
+		if done {
```
bhuvnesh2703, line 582: ideally it should be if err return err.. if done we are good..

cli/commands/commands.go
```diff
@@ -542,22 +578,30 @@ var killServices = &cobra.Command{
...
+	_, err := connectToHub().StopServices(context.Background(), &idl.StopServicesRequest{})
+	if err != nil {
+		errCode := grpcStatus.Code(err)
+		errMsg := grpcStatus.Convert(err).Message()
+		// XXX: "transport is closing" is not documented but is needed to uniquely interpret codes.Unavailable
+		// https://github.com/grpc/grpc/blob/v1.24.0/doc/statuscodes.md
+		if errCode != codes.Unavailable || errMsg != "transport is closing" {
+			return err, true
+		}
+		return nil, true
```
bhuvnesh2703, line 600: could you explain why err is marked as nil here?

agent/delete_directories.go
marksliva: Sorry, when I said ellipses, I meant `:`. The reason to keep the ellipses is because that's the closest we can get to defining the slice as a constant, since ellipses make the contents of the slice immutable. It's still not perfect, since the variable could be reassigned to a different immutable slice, but it's better than using a mutable slice.

kalensk:
In some sense the `agent/delete_directories_test.go` tests seem like a repeat of the `upgrade/directories_test.go` tests, rather than testing the agent specific logic. I could see two potential types of agent specific tests: 1) Are the agents passing/using the correct required files (less useful), and 2) An integration test where the actual directories are getting deleted. (much more useful).

agent/delete_directories_test.go
```diff
@@ -0,0 +1,103 @@
...
+	testhelper.SetupTestLogger()
+
+	t.Run("successfully deletes the data directories if all required files exist for that directory", func(t *testing.T) {
+		directory := "/gpupgrade/my/state/dir"
+		var actualDirectory string
+		utils.System.RemoveAll = func(name string) error {
+			actualDirectory = name
+			return nil
+		}
+
```
kalensk, line 76: It seems like it would be clearer and less lines to move the test inside the func as shown below. Similarly for the other tests. No need to do reflect.DeepEqual on strings.

```go
expected := "/gpupgrade/my/state/dir"
utils.System.RemoveAll = func(dir string) error {
    if dir != expected {
        t.Errorf("got %s, want %s", dir, expected)
    }
    return nil
}
```

agent/delete_directories_test.go
```diff
@@ -0,0 +1,103 @@
...
+	t.Run("successfully deletes the data directories if all required files exist for that directory", func(t *testing.T) {
+		expectedDataDirectories := []string{"/data/dbfast_mirror1/seg1", "/data/dbfast_mirror2/seg2"}
+		actualDataDirectories := []string{}
+		utils.System.RemoveAll = func(name string) error {
+			actualDataDirectories = append(actualDataDirectories, name)
+			return nil
+		}
+
+		expectedFilesStatCalls := []string{"/data/dbfast_mirror1/seg1/postgresql.conf",
```
kalensk, line 32: nit: put postgresql.conf on a new line.

testutils/test_utils.go
```diff
@@ -49,3 +53,36 @@ func GetOpenPort() (int, error) {
 
 	return l.Addr().(*net.TCPAddr).Port, nil
 }
+
+func SetupDirs(t *testing.T, subDirectories []string, requiredFiles []string) (tmpDir string, createdDirectories []string) {
```
jchampio, line 57: This new implementation is used only in one place; IMO it doesn't belong in `testutils` yet. It also seems pretty delete-specific.

testutils/test_utils.go
```diff
@@ -49,3 +53,36 @@ func GetOpenPort() (int, error) {
...
+	var err error
+	tmpDir, err = ioutil.TempDir("", "")
+	if err != nil {
+		t.Fatalf("unexpected err: %v", err)
```
jchampio, line 61: This needs to be fleshed out with context now that it's been pulled into a helper. `creating temporary directory: %v` or similar.

This applies to all the other "unexpected err" strings here. If the top-level test name itself isn't enough context to figure out what's going on, we need more detail; we don't want to have to jump into the source just to figure out what broke.

testutils/test_utils.go
```diff
@@ -49,3 +53,36 @@ func GetOpenPort() (int, error) {
...
+	dirPath = filepath.Join(tmpDir, name)
+
+	err := os.MkdirAll(dirPath, 0700)
+	if err != nil {
+		t.Errorf("unexpected err: %v", err)
+	}
+
+	for _, fileName := range requiredFiles {
+		filePath := filepath.Join(dirPath, fileName)
+		_, err = os.Create(filePath)
```
jchampio, line 81: This file hangs open until the GC collects it. Consider `ioutil.WriteFile` if you're just wanting to touch a zero-length file on disk.

agent/delete_directories.go
kalensk: If that's the goal you might be better off with a function initialization, which returns a new slice every time as shown below. Although, I find this all unnecessary and a bit excessive for simply passing to `upgrade.DeleteDirectories`.

```go
func getPostgresFiles() []string {
	return []string{"postgresql.conf", "PG_VERSION"}
}
```

upgrade/directories.go
```diff
@@ -56,3 +60,31 @@ func TempDataDir(datadir, segPrefix string, id ID) string {
 
 	return filepath.Join(dir, newBase)
 }
+
+func DeleteDirectories(directoriesToDelete []string, filesThatMustExist []string) error {
```
jchampio, line 64: nit: in a function named `DeleteDirectories`, having the first argument named `directoriesToDelete` seems needlessly verbose. We know that from context. And `filesThatMustExist` seems like we're trying to make up for the lack of documentation with a four-word identifier. I think this function needs documentation, to describe why it exists and how to use it, and then the parameter names can shrink to match.

[Some good suggestions for identifier choice and length.](https://dave.cheney.net/practical-go/presentations/gophercon-singapore-2019.html#_identifiers)

marksliva:
> In some sense the `agent/delete_directories_test.go` tests seem like a repeat of the `upgrade/directories_test.go` tests, rather than testing the agent specific logic. I could see two potential types of agent specific tests: 1) Are the agents passing/using the correct required files (less useful), and 2) An integration test where the actual directories are getting deleted. (much more useful).

So in that case are you suggesting that `delete_directories_test.go` replace `directories_test.go#TestDeleteDataDirectories`? As long as we test it only there, then sure, but this is not ideal since that shared delete logic would not be tested in its associated `_test.go` file.

The other approach (which seems to be what you're hinting at in `1`) is to create a mockable interface, which we override in our test, and then check that `upgrade.DeleteDirectories` gets called with the expected `directoriesToDelete` and `filesThatMustExist`.

So, can you explain more about why option `1` would not be useful while still keeping the existing integration tests in `directories_test.go`? I believe the mockable interface would work well in this scenario, so I'm curious if you have other reasons that you prefer the other option.

upgrade/directories.go
```diff
@@ -56,3 +60,31 @@ func TempDataDir(datadir, segPrefix string, id ID) string {
...
+	var mErr *multierror.Error
+	for _, segDataDir := range directoriesToDelete {
+
```
jchampio, line 67: unnecessary empty line

hub/revert.go
```diff
@@ -0,0 +1,48 @@
...
+
+import (
+	"fmt"
+
+	"github.com/greenplum-db/gp-common-go-libs/gplog"
+	"github.com/hashicorp/go-multierror"
+
+	"github.com/greenplum-db/gpupgrade/agent"
+	"github.com/greenplum-db/gpupgrade/upgrade"
+
```
bhuvnesh2703, line 11: new line not needed

cli/commands/commands.go
```diff
@@ -447,6 +449,40 @@ func finalize() *cobra.Command {
...
+	cmd := &cobra.Command{
+		Use:   "revert",
+		Short: "reverts the upgrade and returns the cluster to its original state",
+		Long:  "reverts the upgrade and returns the cluster to its original state",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			client := connectToHub()
+			err := commanders.Revert(client)
+			if err != nil {
+				gplog.Error(err.Error())
+				os.Exit(1)
```
kalensk, line 462: Since this is using `RunE ` why the need for `os.Exit(1)` everywhere rather than returning the error and letting cobra handle the exit code?

upgrade/directories.go
```diff
@@ -56,3 +60,31 @@ func TempDataDir(datadir, segPrefix string, id ID) string {
...
+		var missingFilesError *multierror.Error
```
jchampio, line 68: Though I understand the reason for using it, I'm finding the nested multierror difficult to follow in my head. Can we switch to just a `bool` flag that gets set if there are any missing files, and append to the top-level multierror directly?

upgrade/directories_test.go
```diff
@@ -48,3 +57,114 @@ func ExampleTempDataDir() {
...
+
+var dataDirs []string
+var filesThatMustExist []string
+var dataDirectories = []string{"/data/dbfast_mirror1/seg1", "/data/dbfast_mirror2/seg2"}
+
+func setup() {
+	filesThatMustExist = []string{"pg_file1", "pg_file2"}
+	_, dataDirs = testutils.SetupDirs(nil, dataDirectories, filesThatMustExist)
+
+	println(fmt.Sprintf("the datadirs: %#v", dataDirs))
```
jchampio, line 69: This is a case for `t.Log()`.

upgrade/directories_test.go
```diff
@@ -48,3 +57,114 @@ func ExampleTempDataDir() {
...
+}
+
+func teardown() {
+	defer func() {
```
jchampio, line 73: The `defer` here seems to be unnecessary.

hub/revert.go
```diff
@@ -0,0 +1,48 @@
...
+			gplog.Error(fmt.Sprintf("revert: %s", err))
+		}
+	}()
+
+	err = DeleteSegmentAndStandbyDataDirectories(s.agentConns, s.Config.Target)
+	if err != nil {
+		return err
+	}
+
+	err = upgrade.DeleteDirectories([]string{s.Config.Target.MasterDataDir()}, agent.PostgresFiles[:])
```
kalensk, line 37: Seeing `agent.PostgresFiles[:])` is telling me that it should be moved to the upgrade package in probably directories.go.

agent/delete_directories_test.go
marksliva: It might seem clearer, but then for correctness we would have to add a boolean variable called `removeAllCalled` and assert that it was called in our test after calling `DeleteDirectories`. Otherwise, how would we know that `RemoveAll` was actually called?

upgrade/directories_test.go
```diff
@@ -48,3 +57,114 @@ func ExampleTempDataDir() {
...
+		actualFilesStatCalls := []string{}
+		utils.System.Stat = func(name string) (os.FileInfo, error) {
+			actualFilesStatCalls = append(actualFilesStatCalls, name)
+
+			return nil, nil
+		}
+
+		defer func() {
+			utils.System.Stat = os.Stat
+		}()
```
jchampio, line 96: Seems like this should be deleted? Why do we need to mock out `Stat` for a live test?

upgrade/directories_test.go
```diff
@@ -48,3 +57,114 @@ func ExampleTempDataDir() {
 	// /data/standby.AAAAAAAAAAA
 	// /data/primary/seg.AAAAAAAAAAA.3
 }
+
+var dataDirs []string
+var filesThatMustExist []string
+var dataDirectories = []string{"/data/dbfast_mirror1/seg1", "/data/dbfast_mirror2/seg2"}
```
jchampio, line 63: It seems like this should be done without global variables. The `setup` pattern gives us a way to set up and handle per-test state.

cli/commands/commands.go
```diff
@@ -447,6 +449,40 @@ func finalize() *cobra.Command {
...
+			}
+
+			err, done := stopAllServices()
+			if done {
+				return err
+			}
+
+			fmt.Println("Revert completed successfully.")
```
kalensk, line 476: Probably for another story and a design decision , but Im wondering if we want to say something more such as "Your cluster has now been restored to before gpupgrade was run."

cli/commands/commands.go
marksliva: Good point. Changed it.

ci/tasks/install-tests.yml
```diff
@@ -4,7 +4,7 @@ image_resource:
   type: docker-image
   source:
     repository: pivotaldata/centos-gpdb-dev
-    tag: "6-gcc6.2-llvm3.7"
+    tag: "7-gcc6.2-llvm3.7"
```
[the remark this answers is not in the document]
marksliva: The `7` image has `jq`, which is used in `revert.bats`.

agent/delete_directories_test.go
```diff
@@ -0,0 +1,103 @@
...
+
+func TestDeleteStateDirectory(t *testing.T) {
+	server := agent.NewServer(agent.Config{
+		Port:     -1,
+		StateDir: "",
+	})
+
+	testhelper.SetupTestLogger()
+
+	t.Run("successfully deletes the data directories if all required files exist for that directory", func(t *testing.T) {
```
bhuvnesh2703, line 69: test description to have state directories not data directories.

hub/delete_data_directories.go
```diff
@@ -11,37 +11,53 @@ import (
...
-		mirrorsIncludingStandby := func(seg *greenplum.SegConfig) bool {
-			return seg.Hostname == conn.Hostname &&
-				(seg.Role == greenplum.MirrorRole)
+		filterFunc := func(seg *greenplum.SegConfig) bool {
+			if seg.Hostname != conn.Hostname {
+				return false
+			}
+
+			// we don't have mirrors at this point after initialize, but we may have them later
+			// if the finalize step crashes after adding mirrors or a standby
```
bhuvnesh2703, line 35: could you elaborate this a bit more..

cli/commands/commands.go
```diff
@@ -542,22 +578,30 @@ var killServices = &cobra.Command{
...
-		if err != nil {
-			errCode := grpcStatus.Code(err)
-			errMsg := grpcStatus.Convert(err).Message()
-			// XXX: "transport is closing" is not documented but is needed to uniquely interpret codes.Unavailable
-			// https://github.com/grpc/grpc/blob/v1.24.0/doc/statuscodes.md
-			if errCode != codes.Unavailable || errMsg != "transport is closing" {
-				return err
-			}
-			return nil
+		err, done := stopAllServices()
```
bhuvnesh2703, line 581: stopAllServices (could it be stopHubAndAgents) ?

cli/commands/commands.go
kalensk: > could you explain why err is marked as nil here?

Its nil since we want ignore errors that are caused by the gRPC sever shutting down (which is Unavailable - "transport is closing"). That is, we only want to return errors that are not of that type. See https://github.com/grpc/grpc/blob/v1.24.0/doc/statuscodes.md


What I don't yet fully understand is the need for the boolean.

kalensk:
I am advocating for the following agent tests, and delete the current agent tests since they are already covered in upgrade/data_directories.

The following agent test does not know anything about the internals of `upgrade.DeleteDirectories` (which is tested in upgrade/data_directories_test.go). It's straightforward while using real directories containing the required files without "mocking" anything.

```go
	t.Run("deletes data directory", func(t *testing.T) {
		dir1 := getDataDir(t)
		defer os.RemoveAll(dir1)

		dir2 := getDataDir(t)
		defer os.RemoveAll(dir2)

		req := &idl.DeleteDataDirectoriesRequest{Datadirs: []string{dir1, dir2}}
		_, err := server.DeleteDataDirectories(context.Background(), req)
		if err != nil {
			t.Errorf("unexpected error got %+v", err)
		}

		if upgrade.PathExists(dir1) {
			t.Errorf("expected directory %q to not exist", dir1)
		}

		if upgrade.PathExists(dir2) {
			t.Errorf("expected directory %q to not exist", dir2)
		}
	})


	t.Run("does not delete data directory", func(t *testing.T) {
		dir := getTempDir(t)
		defer os.RemoveAll(dir)

		req := &idl.DeleteDataDirectoriesRequest{Datadirs: []string{dir}}
		
		expected := errors.New("no such file or directory") // FIXME to be expected error
		_, err := server.DeleteDataDirectories(context.Background(), req)
		if !errors.Is(err, expected) {
			t.Errorf("got error %#v, want %#v", err, expected)
		}

		if !upgrade.PathExists(dir) {
			t.Errorf("expected directory %q to exist", dir)
		}
	})
	
func getDataDir(t *testing.T) string {
	dir := getTempDir(t)

	for _, f := range agent.PostgresFiles {
		err := ioutil.WriteFile(path.Join(dir, f), []byte(""), 0644)
		if err != nil {
			t.Fatalf("writing file %q: %+v", f, err)
		}
	}

	return dir
}

func PathExists(path string) bool {
	return err == nil
}

```

marksliva:
> I am advocating for the following agent tests, and delete the current agent tests since they are already covered in upgrade/data_directories_test.go.

@kalensk 
If we did that then we'll have nearly identical tests in 2 different files (`upgrade/directories_test.go` and `agent/delete_directories_test.go`), and we'd have to maintain this duplicated code. To me, this does not seem ideal.

agent/delete_directories.go
marksliva: Since golang ultimately fails at having a way to define a slice const, I will simplify this to use a mutable slice, i.e. `[]string{"postgresql.conf", "PG_VERSION"}`.

kalensk:
I am more concerned that the agent tests have knowledge of the internals of `upgrade.DeleteDirectories`

marksliva:
> I am more concerned that the agent tests have knowledge of the internals of `upgrade.DeleteDirectories`
@kalensk 
OK. I agree with you on that.
I'll refactor these tests by mocking the `upgrade.DeleteDirectories` function.

agent/delete_directories_test.go
marksliva: Done.

hub/revert.go
marksliva: Moved it.

agent/delete_directories.go
```diff
@@ -0,0 +1,27 @@
...
+	"context"
+
+	"github.com/greenplum-db/gp-common-go-libs/gplog"
+
+	"github.com/greenplum-db/gpupgrade/idl"
+	"github.com/greenplum-db/gpupgrade/upgrade"
+)
+
+var PostgresFiles = [...]string{"postgresql.conf", "PG_VERSION"}
+var stateDirectoryFiles = [...]string{"config.json", "status.json"}
```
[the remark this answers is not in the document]
marksliva: Done.

agent/delete_directories_test.go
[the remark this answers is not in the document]
marksliva: Done.

agent/delete_directories_test.go
```diff
@@ -0,0 +1,103 @@
...
+		}
+
+		req := &idl.DeleteDataDirectoriesRequest{Datadirs: expectedDataDirectories}
+		_, err := server.DeleteDataDirectories(context.Background(), req)
+
+		if !reflect.DeepEqual(actualDataDirectories, expectedDataDirectories) {
+			t.Errorf("got %s, want %s", actualDataDirectories, expectedDataDirectories)
+		}
+
+		if !reflect.DeepEqual(actualFilesStatCalls, expectedFilesStatCalls) {
```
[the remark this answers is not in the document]
marksliva: I'm going to mock `DeleteDirectories` in order to simplify these tests.

upgrade/directories.go
marksliva: Fixed.

testutils/test_utils.go
marksliva: Fixed.

testutils/test_utils.go
marksliva: Moved.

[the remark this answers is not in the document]
marksliva: Done.

testutils/test_utils.go
marksliva: Done.

upgrade/directories.go
marksliva: ok. I renamed the arguments to `directories` and `requiredPaths`.

upgrade/directories_test.go
marksliva: Good catch. I removed all stubbing of `utils.System.Stat` from this test file.

agent/delete_directories_test.go
marksliva: Fixed.

upgrade/directories_test.go
marksliva: Removed.

upgrade/directories_test.go
marksliva: Fixed.

marksliva: On second thought, this can be removed since these directories get deleted anyways.
I'll move directory logging to the place that removes the directories at the end if there's an error.

agent/delete_directories.go
```diff
@@ -0,0 +1,27 @@
...
+
+func (s *Server) DeleteStateDirectory(ctx context.Context, in *idl.DeleteStateDirectoryRequest) (*idl.DeleteStateDirectoryReply, error) {
+	gplog.Info("got a request to delete the state directory from the hub")
+
+	err := upgrade.DeleteDirectories([]string{in.Directory}, stateDirectoryFiles[:])
```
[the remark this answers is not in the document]
marksliva: fixed.

upgrade/directories.go
marksliva: Done.

upgrade/directories_test.go
marksliva: Done.

test/revert.bats
```diff
@@ -0,0 +1,51 @@
...
+        --disk-free-ratio 0 \
+        --verbose 3>&-
+
+    # parse config.json for the datadirs
+    local target_hosts_dirs=$(jq -r '.Target.Primaries[] | .Hostname + " " + .DataDir' "${GPUPGRADE_HOME}/config.json")
+
+    # check that the target datadirs exist
+    while read -r hostname datadir; do
+        run ssh "${hostname}" stat "${datadir}"
+        [ $status -eq 0 ] || fail "expected datadir ${datadir} to exist"
```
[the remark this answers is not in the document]
marksliva: Fixed.

agent/delete_directories_test.go
```diff
@@ -0,0 +1,103 @@
...
+			actualDataDirectories = append(actualDataDirectories, name)
+			return nil
+		}
+
+		expectedFilesStatCalls := []string{"/data/dbfast_mirror1/seg1/postgresql.conf",
+			"/data/dbfast_mirror1/seg1/PG_VERSION",
+			"/data/dbfast_mirror2/seg2/postgresql.conf",
+			"/data/dbfast_mirror2/seg2/PG_VERSION",
+		}
+		actualFilesStatCalls := []string{}
```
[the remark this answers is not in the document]
marksliva: No longer relevant since `DeleteDirectories` is now mocked.

agent/delete_directories_test.go
```diff
@@ -0,0 +1,103 @@
...
+			actualFilesStatCalls = append(actualFilesStatCalls, name)
+
+			return nil, nil
+		}
+
+		req := &idl.DeleteDataDirectoriesRequest{Datadirs: expectedDataDirectories}
+		_, err := server.DeleteDataDirectories(context.Background(), req)
+
+		if !reflect.DeepEqual(actualDataDirectories, expectedDataDirectories) {
+			t.Errorf("got %s, want %s", actualDataDirectories, expectedDataDirectories)
```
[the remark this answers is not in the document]
marksliva: Fixed.

agent/delete_directories_test.go
```diff
@@ -0,0 +1,103 @@
...
+		utils.System.Stat = func(name string) (os.FileInfo, error) {
+			actualFilesStatCalls = append(actualFilesStatCalls, name)
+
+			return nil, nil
+		}
+
+		req := &idl.DeleteStateDirectoryRequest{Directory: directory}
+		_, err := server.DeleteStateDirectory(context.Background(), req)
+
+		if !reflect.DeepEqual(actualDirectory, directory) {
```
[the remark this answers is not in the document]
marksliva: This was refactored and it is now a slice.

agent/delete_directories_test.go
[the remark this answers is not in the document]
marksliva: No longer relevant since `DeleteDirectories` is now mocked.

cli/commanders/steps.go
```diff
@@ -157,6 +156,29 @@ func Finalize(client idl.CliToHubClient, verbose bool) error {
...
+	}
+
+	// TODO: add verbose option
+	_, err = UILoop(stream, false)
+	if err != nil {
+		return xerrors.Errorf("Revert: %w", err)
+	}
+
+	fmt.Println("")
+	fmt.Printf("The source cluster is now restored to its original state. [more information TBD]\n")
```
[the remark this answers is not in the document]
marksliva: Fixed.

cli/commands/commands.go
```diff
@@ -447,6 +449,40 @@ func finalize() *cobra.Command {
...
+		Long:  "reverts the upgrade and returns the cluster to its original state",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			client := connectToHub()
+			err := commanders.Revert(client)
+			if err != nil {
+				gplog.Error(err.Error())
+				os.Exit(1)
+			}
+
+			err = utils.System.RemoveAll(utils.GetStateDir())
```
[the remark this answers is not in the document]
marksliva: Moved.

cli/commands/commands.go
```diff
@@ -447,6 +449,40 @@ func finalize() *cobra.Command {
...
+				os.Exit(1)
+			}
+
+			err, done := stopAllServices()
+			if done {
+				return err
+			}
+
+			fmt.Println("Revert completed successfully.")
+			fmt.Println("")
```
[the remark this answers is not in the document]
marksliva: Removed.

hub/delete_data_directories_test.go
```diff
@@ -30,83 +29,131 @@ func TestDeleteSegmentDataDirs(t *testing.T) {
...
+	t.Run("DeleteSegmentAndStandbyDataDirectories", func(t *testing.T) {
+		t.Run("deletes standby, primary, and mirror data directories", func(t *testing.T) {
+			ctrl := gomock.NewController(t)
+			defer ctrl.Finish()
+
+			sdw1Client := mock_idl.NewMockAgentClient(ctrl)
+			sdw1Client.EXPECT().DeleteDataDirectories(
+				gomock.Any(),
+				&idl.DeleteDataDirectoriesRequest{Datadirs: []string{
+					"/data/dbfast1/seg1", "/data/dbfast_mirror1/seg1", "/data/dbfast1/seg3", "/data/dbfast_mirror1/seg3",
```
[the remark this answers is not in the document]
marksliva: Done.

hub/delete_data_directories_test.go
```diff
@@ -30,83 +29,131 @@ func TestDeleteSegmentDataDirs(t *testing.T) {
...
-		}
-
-		if len(multiErr.Errors) != 1 {
-			t.Errorf("received %d errors, want %d", len(multiErr.Errors), 1)
-		}
-
-		for _, err := range multiErr.Errors {
-			if !xerrors.Is(err, expected) {
-				t.Errorf("got error %#v, want %#v", expected, err)
+	t.Run("DeleteDataDirectories", func(t *testing.T) {
```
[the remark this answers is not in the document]
marksliva: Good point. I moved the failure test that was using it under the `DeleteSegmentAndStandbyDataDirectories` test block and now `deleteDataDirectories` can be private.

agent/delete_directories_test.go
marksliva: This file was significantly refactored.

agent/delete_directories_test.go
marksliva: This file ended up being refactored to use a mock, so this is no longer relevant.

agent/delete_directories_test.go
marksliva: Done.

hub/delete_state_directories_test.go
```diff
@@ -0,0 +1,94 @@
...
+				&idl.DeleteStateDirectoryRequest{Directory: expected},
+			).Return(&idl.DeleteStateDirectoryReply{}, nil)
+
+			masterClient := mock_idl.NewMockAgentClient(ctrl)
+			// NOTE: we expect no call to the master
+
+			agentConns := []*hub.Connection{
+				{nil, sdw1Client, "sdw1", nil},
+				{nil, standbyClient, "standby", nil},
+				{nil, masterClient, "master", nil},
```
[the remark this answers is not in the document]
marksliva: I thought this is a case, don't we have an agent client that runs on the master host? (i.e. when running gpupgrade on a local cluster)
Should this be renamed to `masterHostClient` for clarity?

cli/commands/commands.go
marksliva: Sure. Renamed.

hub/revert.go
```diff
@@ -0,0 +1,48 @@
...
+
+	"github.com/greenplum-db/gpupgrade/agent"
+	"github.com/greenplum-db/gpupgrade/upgrade"
+
+	"github.com/greenplum-db/gpupgrade/idl"
+	"github.com/greenplum-db/gpupgrade/step"
+)
+
+func (s *Server) Revert(_ *idl.RevertRequest, stream idl.CliToHub_RevertServer) (err error) {
+	st, err := step.Begin(s.StateDir, "revert", stream)
```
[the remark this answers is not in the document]
marksliva: I think initially we were thinking about using the stream. But it does appear that we never created a step for this and also this story is not about verbose mode anyway. Removed.

hub/revert.go
marksliva: Well this dependency went away anyway. Let me know if this still looks off.

hub/delete_data_directories.go
marksliva: Hmm on second thought.. we are not planning to support revert after running finalize, even if it crashes near the end of finalize. So in this case, I have changed the logic to delete only the primaries, and deleted this comment.

cli/commands/commands.go
```diff
@@ -447,6 +449,40 @@ func finalize() *cobra.Command {
...
+
+			err = utils.System.RemoveAll(utils.GetStateDir())
+			if err != nil {
+				gplog.Error(err.Error())
+				os.Exit(1)
+			}
+
+			err, done := stopAllServices()
+			if done {
+				return err
```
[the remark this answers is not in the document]
marksliva: Refactored this to return just an error.

cli/commands/commands.go
```diff
@@ -542,22 +578,30 @@ var killServices = &cobra.Command{
...
+		err, done := stopAllServices()
+		if done {
+			return err
 		}
 
 		return nil
 	},
 }
 
+func stopAllServices() (error, bool) {
```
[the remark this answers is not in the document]
marksliva: Got rid of `done`.

cli/commands/commands.go
marksliva: I got rid of the boolean.

cli/commands/commands.go
marksliva: Done.

cli/commands/commands.go
```diff
@@ -462,12 +463,12 @@ func revert() *cobra.Command {
...
 
-			err, done := stopHubAndAgents()
-			if done {
+			err = stopHubAndAgents()
+			if err != nil {
 				return err
 			}
 
-			err = utils.System.RemoveAll(utils.GetStateDir())
+			err = upgrade.DeleteDirectories([]string{utils.GetStateDir()}, upgrade.StateDirectoryFiles)
```
jchampio, line 471: For the future: the construction here implies to me that we should maybe move the `for` loop inside `DeleteDirectories()` out to the callers, and just have `DeleteDirectory()`.

jchampio:
Looking good. Is this in a "final" state or is there still work in progress?

hub/delete_state_directories_test.go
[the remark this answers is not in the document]
jchampio: Yes, but we're not guaranteed to have one, and it's not the client for the master.
> Should this be renamed to masterHostClient for clarity?

That might be a little better, sure.

I think we need to work on a way to uniformly map our agents to the cluster. Every interface does this differently.

marksliva:
> Looking good. Is this in a "final" state or is there still work in progress?

I did expose another flaw, which occurs when running revert after an initialize when no cluster was created. In that case, the test below fails because the `DeleteDirectories` func returns an error when a stat errors (since the datadir does not exist). As a temporary workaround, I'm thinking about checking if there are primaries/master on the `target` cluster before attempting a delete. A better fix would be to make this actually idempotent by reworking `DeleteDirectories` to not error out when the directory doesn't exist, but that seems like a future story.

```
# todo: this should also work
#@test "revert cleans up the state dir when no cluster was created" {
#    gpupgrade initialize \
#        --source-bindir="$GPHOME/bin" \
#        --target-bindir="$GPHOME/bin" \
#        --source-master-port="${PGPORT}" \
#        --temp-port-range 6020-6040 \
#        --disk-free-ratio 0 \
#        --stop-before-cluster-creation \
#        --verbose 3>&-
#
#    gpupgrade revert
#
#    # the GPUPGRADE_HOME directory is deleted
#    if [ -d "${GPUPGRADE_HOME}" ]; then
#        echo "expected GPUPGRADE_HOME directory ${GPUPGRADE_HOME} to have been deleted"
#        exit 1
#    fi
#}

...
     - rpc error: code = Unknown desc = 2 errors occurred:
   	* stat postgresql.conf: no such file or directory
   	* stat PG_VERSION: no such file or directory

```

cli/commands/commands.go
marksliva: I added this suggestion to the clean up revert output story.

marksliva:
> > Looking good. Is this in a "final" state or is there still work in progress?
> 
> I did expose another flaw, which occurs when running revert after an initialize when no cluster was created...

I fixed this by only deleting the datadirs if the target cluster config has primaries.

upgrade/directories_test.go
```diff
@@ -316,3 +321,133 @@ func verifyRename(t *testing.T, source, archive, target string) {
 		t.Errorf("expected target %q to not exist", target)
 	}
 }
+
+func setup() (directories []string, requiredPaths []string) {
+	requiredPaths = []string{"pg_file1", "pg_file2"}
+	var dataDirectories = []string{"/data/dbfast_mirror1/seg1", "/data/dbfast_mirror2/seg2"}
+	_, directories = setupDirs(nil, dataDirectories, requiredPaths)
```
jchampio, line 328: Looks like the root temporary directory is being "leaked" here.

marksliva: Fixed.

marksliva: Fixed.

upgrade/directories_test.go
```diff
@@ -322,29 +322,26 @@ func verifyRename(t *testing.T, source, archive, target string) {
...
 	var dataDirectories = []string{"/data/dbfast_mirror1/seg1", "/data/dbfast_mirror2/seg2"}
-	_, directories = setupDirs(nil, dataDirectories, requiredPaths)
-
-	return directories, requiredPaths
-}
-
-func teardown(t *testing.T, directories []string) {
-	for _, directory := range directories {
-		err := os.RemoveAll(directory)
+	rootDir, directories := setupDirs(nil, dataDirectories, requiredPaths)
```
jchampio, line 328: `t` needs to be passed here, doesn't it?

marksliva: Good catch. Fixed.

upgrade/directories_test.go
```diff
@@ -316,3 +321,130 @@ func verifyRename(t *testing.T, source, archive, target string) {
 		t.Errorf("expected target %q to not exist", target)
 	}
 }
+
+func setup(t *testing.T) (teardown func(), directories []string, requiredPaths []string) {
+	requiredPaths = []string{"pg_file1", "pg_file2"}
+	var dataDirectories = []string{"/data/dbfast_mirror1/seg1", "/data/dbfast_mirror2/seg2"}
```
jchampio, line 327: Future improvement: have the test control these, rather than burying them in setup. `setup()` can take them as parameters.

hub/delete_data_directories_test.go
```diff
@@ -30,83 +29,134 @@ func TestDeleteSegmentDataDirs(t *testing.T) {
...
+			).Return(&idl.DeleteDataDirectoriesReply{}, nil)
+
+			standbyClient := mock_idl.NewMockAgentClient(ctrl)
+			standbyClient.EXPECT().DeleteDataDirectories(
+				gomock.Any(),
+				&idl.DeleteDataDirectoriesRequest{Datadirs: []string{"/data/standby"}},
+			).Return(&idl.DeleteDataDirectoriesReply{}, nil)
+
+			masterClient := mock_idl.NewMockAgentClient(ctrl)
+			// NOTE: we expect no call to the master
```
jchampio, line 64: Still waiting on `masterClient` to be removed from this and the next test. We don't run an agent for the master segment, so this doesn't add coverage. And unlike the state directory deletion, which wants to skip the master host even if there's an agent running on it, we have no such constraint here.

marksliva: Done.

PR #315, opened 2020-04-17
rename data directories idempotently

kalensk:
For ease of reviewing see each commit _and_ its message one at a time in order. This is a followup to https://github.com/greenplum-db/gpupgrade/pull/307.

This renames data directories idempotently. We add `RenameDataDirectory` in the upgrade package so it can be used by both the hub for renaming the master data directories, and by the agent for renaming the segment directories.

To enable idempotence we allow ENOENT errors when renaming the source to archive as this may occur when renaming the target fails resulting in the source directory not existing.

When renaming segments, in link mode the mirrors have been deleted to save disk space. Since the upgraded mirrors will be added later to the correct directory there is no need to rename target to source. Thus, the `RenameTarget` boolean is used to indicate whether renaming the source to target is needed.

[Test pipeline](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:data_dirs)

**_Note:_** If desired followup PRs could be included to either:
1) have the old directory include the upgrade.ID to mitigate collisions
2) for safety check that the directories look like postgres data directories before renaming


testutils/test_utils.go
```diff
@@ -49,3 +51,12 @@ func GetOpenPort() (int, error) {
 
 	return l.Addr().(*net.TCPAddr).Port, nil
 }
+
+func GetTempDir(t *testing.T, prefix string) string {
+	sourceDir, err := ioutil.TempDir("", prefix+"-")
```
kalensk, line 56: nit: rename `sourceDir` to `dir`

kalensk: This function needs `t.Helper()`

agent/server_test.go
```diff
@@ -17,7 +17,7 @@ func TestServerStart(t *testing.T) {
 	testhelper.SetupTestLogger()
 
 	t.Run("successfully starts and creates state directory if it does not exist", func(t *testing.T) {
-		tempDir := getTempDir(t)
+		tempDir := testutils.GetTempDir(t, "")
```
marksliva, line 20: nit: We could create another method called `GetTempDir` that takes a single argument which then calls the new `GetTempDir`. This could be used to simplify `agent/rsync_test.go` as well.

hub/rename_data_directories.go
```diff
@@ -10,19 +10,19 @@ import (
...
 
 func (s *Server) UpdateDataDirectories() error {
 	return UpdateDataDirectories(s.Config, s.agentConns)
 }
 
 func UpdateDataDirectories(conf *Config, agentConns []*Connection) error {
-	if err := RenameDataDirs(conf.Source.MasterDataDir(), conf.TargetInitializeConfig.Master.DataDir); err != nil {
+	source := conf.Source.MasterDataDir()
+	target := conf.TargetInitializeConfig.Master.DataDir
+	if err := upgrade.RenameDataDirectory(source, source+upgrade.OldSuffix, target, true); err != nil {
```
marksliva, line 25: nit: add space around `+`. Or also, consider making a function that returns `dir + upgrade.OldSuffix` since we do this 3 times in this file.

upgrade/directories.go
```diff
@@ -56,3 +64,41 @@ func TempDataDir(datadir, segPrefix string, id ID) string {
...
+
+// renames source to archive, and target to source
+// e.g. for source /data/dbfast1/demoDataDir0 becomes /data/dbfast1/demoDataDir0_old
+// e.g. for target /data/dbfast1/demoDataDir.123ABC.0 becomes /data/dbfast1/demoDataDir0
+func RenameDataDirectory(source, archive, target string, renameTarget bool) error {
+	if alreadyRenamed(archive, target) {
+		return nil
+	}
+
+	err := utils.System.Rename(source, archive)
```
marksliva, line 76: Do we want to first do a safety check that both source and target contain `postgresFiles`?

upgrade/directories.go
```diff
@@ -56,3 +64,41 @@ func TempDataDir(datadir, segPrefix string, id ID) string {
...
+	if err != nil {
+		if !xerrors.Is(err, syscall.ENOENT) {
+			return err
+		}
+
+		gplog.Debug("Renaming '%q' to '%q'. Source directory does not exist, was it already renamed from a previous re-run.", source, archive)
```
marksliva, line 82: I would change this log message to the following:
`Renaming '%q' to '%q' - Source directory does not exist. It was already renamed during a previous re-run.`

agent/rename_directories.go
```diff
@@ -6,14 +6,14 @@ import (
...
-	"github.com/greenplum-db/gpupgrade/utils"
+	"github.com/greenplum-db/gpupgrade/upgrade"
 )
 
 func (s *Server) RenameDirectories(ctx context.Context, in *idl.RenameDirectoriesRequest) (*idl.RenameDirectoriesReply, error) {
 	gplog.Info("agent received request to rename segment data directories")
 
-	for _, pair := range in.GetPairs() {
-		err := utils.System.Rename(pair.Src, pair.Dst)
+	for _, pair := range in.GetDirs() {
```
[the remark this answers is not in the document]
kalensk: yeah that was `dir` at one point, but got lost in the git-wrangling. Will update.

agent/rsync_test.go
```diff
@@ -42,10 +34,10 @@ func TestRsync(t *testing.T) {
 	defer func() { agent.SetRsyncCommand(nil) }()
 
 	t.Run("it copies data from a source directory to a target directory", func(t *testing.T) {
-		sourceDir := getTempDir(t)
+		sourceDir := testutils.GetTempDir(t, "rsync-source")
```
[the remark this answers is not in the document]
kalensk: I wanted to keep the same meaning as what I removed, as shown [here](https://github.com/greenplum-db/gpupgrade/pull/315/commits/7e43d595f1e90ee0a8f5ce2bbb23e01ede6eaf63#diff-eacb628c7a6f9287af34de32eab256c0L18).

agent/rename_directories_test.go
```diff
@@ -1,63 +0,0 @@
-package agent_test
```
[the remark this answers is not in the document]
kalensk: The tests no longer apply since the agent now calls `upgrade.RenameDataDirectory` which has its own tests. Let me know if there any specific tests needed.

upgrade/directories.go
kalensk: As noted in the PR description I am happy to do that in a followup PR if people would like that. I'd like avoid doing that in this PR to keep it manageable and not too dense for reviewing.

hub/rename_data_directories.go
kalensk: Formatter does not allow spaces around `+` when the variable is inlined like that. I'll ponder about using a function for adding the old suffix.

marksliva: Oh interesting, I didn't realize that `gofmt` removes spaces in this case.

agent/rename_directories_test.go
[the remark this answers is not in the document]
marksliva: True for the most complicated of the logic. But looping through the directories doesn't get tested. For example, when I delete this entire block:
```
	for _, pair := range in.GetDirs() {
		err := upgrade.RenameDataDirectory(pair.Source, pair.Archive, pair.Target, pair.RenameTarget)
		if err != nil {
			return &idl.RenameDirectoriesReply{}, err
		}
	}
```
All the unit tests still pass.

upgrade/directories_test.go
```diff
@@ -48,3 +56,266 @@ func ExampleTempDataDir() {
...
+	})
+
+	t.Run("it returns other LinkErrors when renaming the source fails for errors other than ENOENT", func(t *testing.T) {
+		source, archive, target, cleanup := mustCreateDirs(t)
+		defer cleanup(t)
+
+		expected := &os.LinkError{Err: syscall.EEXIST}
+		utils.System.Rename = func(old, new string) error {
+			return expected
+		}
```
kalensk, line 130: It might make things more clear to do `old == source` since this test is for when the source fails.


```go
utils.System.Rename = func(old, new string) error {
    if old == source {
        return expected
    }
    return os.Rename(old, new)
}

```

agent/rename_directories.go
```diff
@@ -6,14 +6,14 @@ import (
...
+		err := upgrade.RenameDataDirectory(pair.Source, pair.Archive, pair.Target, pair.RenameTarget)
```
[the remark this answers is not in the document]
kalensk: Fixed.

agent/rsync_test.go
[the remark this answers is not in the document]
kalensk: I believe the `rysnc-` prefix was used to differentiate between temp dirs used throughout the gpupgrade unit tests. That is, these are temp dirs for the rsync tests.

agent/rename_directories_test.go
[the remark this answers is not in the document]
kalensk: Added back a test that bubbles up errors now that we are returning multiple errors.

agent/server_test.go
kalensk: Since golang doesn't support function overloading I am going to punt on this for now.

hub/rename_data_directories.go
kalensk: You're right that `OldSuffix` is used 3 times (rule of three). However, I would like to punt on this now, and address it in the upcoming PR that uses upgrade.ID in the archive directory to mitigate collisions. That is, `OldSuffix` will go away in this upcoming PR.

upgrade/directories.go
kalensk: Will address this in a followup PR.

upgrade/directories.go
kalensk: updated.

upgrade/directories_test.go
```diff
@@ -48,3 +56,269 @@ func ExampleTempDataDir() {
...
+			t.Errorf("expected source %q to not exist", source)
+		}
+
+		if !upgrade.PathExists(archive) {
+			t.Errorf("expected archive %q to exist", archive)
+		}
+
+		if upgrade.PathExists(target) {
+			t.Errorf("expected target %q to not exist", target)
+		}
```
marksliva, line 189: Since `target := ""` above, this check can be removed, and also the `target` variable can be inlined.

PR #317, opened 2020-04-20
Add VMware copyright to all sources

jchampio:
What it says on the tin.


multihost/vagrant.rb
```diff
@@ -1,4 +1,7 @@
 #
+# Copyright (c) 2017-2020 VMware, Inc. or its affiliates
+# SPDX-License-Identifier: Apache-2.0
+#
```
kalensk, line 4: nit: I would have expected the copyright to start on line 1.

jchampio: First line is not a requirement; just needs to be near the top. I kept the existing block style where I could.

marksliva:
I rebased and updated these 7 files that had been added to master, and then force pushed.
```
git diff 3680b10989dc0cd55d9743a642e9da1f56252663..head --name-status | grep A
A	agent/common_test.go
A	agent/delete_directories.go
A	agent/delete_directories_test.go
A	hub/delete_state_directories.go
A	hub/delete_state_directories_test.go
A	hub/revert.go
A	test/revert.bats
```
Ready to merge when the pipeline goes green.

PR #318, opened 2020-04-23
finalize: archive source with upgradeID and verify before renaming

kalensk:
This is a followup to https://github.com/greenplum-db/gpupgrade/pull/315 and https://github.com/greenplum-db/gpupgrade/pull/307 which 
- use upgradeID when archiving the source directory to mitigate collisions
- for safety check that the directories look like postgres data directories before renaming

For ease of reviewing see each commit _and_ its message one at a time in order. Github is not displaying the commits in the correct order which should be:
- rename RenameDataDirectory
- refactor getRenameMap
- bats: move expected_datadir to helpers.bash
- use upgradeID when archiving the source directory
- bats: refactor delete_cluster
- verify data directories before renaming
- ArchiveAndSwapDirectories: verify log contents

[Test Pipeline](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:archieve_with_upgradeid)


agent/rename_directories.go
```diff
@@ -10,14 +10,14 @@ import (
...
-var RenameDataDirectory = upgrade.RenameDataDirectory
+var ArchiveSourceAndRenameTarget = upgrade.ArchiveSourceAndRenameTarget
 
 func (s *Server) RenameDirectories(ctx context.Context, in *idl.RenameDirectoriesRequest) (*idl.RenameDirectoriesReply, error) {
 	gplog.Info("agent received request to rename segment data directories")
 
 	var mErr *multierror.Error
 	for _, dir := range in.GetDirs() {
-		err := RenameDataDirectory(dir.Source, dir.Archive, dir.Target, dir.RenameTarget)
+		err := ArchiveSourceAndRenameTarget(dir.Source, dir.Target, dir.OnlyArchive)
```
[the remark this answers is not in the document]
kalensk: Updated. The function is now called `ArchiveSource` with parameter `RenameTarget`.

hub/rename_data_directories_test.go
```diff
@@ -175,7 +174,7 @@ func TestUpdateDataDirectories(t *testing.T) {
 		},
 	}
 
-	utils.System.Rename = func(src, dst string) error {
+	hub.ArchiveSourceAndRenameTarget = func(source, target string, onlyArchive bool) error {
```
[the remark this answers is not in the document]
kalensk: Added a test. See last two commits (one to move helpers to testutils, and another adding tests.)

test/helpers.bash
```diff
@@ -113,3 +118,39 @@ require_gnu_stat() {
...
+# Takes an original datadir and echoes the expected temporary datadir containing
+# the upgradeID.
+#
+# NOTE for devs: this is just for getting the expected data directories, which
+# is an implementation detail. If you want the actual location of the new master
+# data directory after an initialization, you can just ask the hub with
+#
+#    gpupgrade config show --target-datadir
+#
+expected_datadir() {
```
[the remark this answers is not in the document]
kalensk: Done. See commit "bats: move expected_datadir to helpers.bash"

agent/rename_directories.go
```diff
@@ -13,14 +13,14 @@ import (
...
-var RenameDataDirectory = upgrade.RenameDataDirectory
+var ArchiveSource = upgrade.ArchiveSource
 
 func (s *Server) RenameDirectories(ctx context.Context, in *idl.RenameDirectoriesRequest) (*idl.RenameDirectoriesReply, error) {
 	gplog.Info("agent received request to rename segment data directories")
 
 	var mErr *multierror.Error
 	for _, dir := range in.GetDirs() {
-		err := RenameDataDirectory(dir.Source, dir.Archive, dir.Target, dir.RenameTarget)
+		err := ArchiveSource(dir.GetSource(), dir.GetTarget(), dir.GetRenameTarget())
```
marksliva, line 23: It seems unnecessary to call the function `dir.GetRenameTarget()`. Why don't we do `dir.RenameTarget` instead since it is a boolean that defaults to false?

kalensk: It seemed like good defensive programming that the generated code gives is for free by checking for nulls.

marksliva: I agree with your general idea, however in this case if `dir` itself is nil, would we even want to call that function? I'd think that we would want to error out as early as possible before allowing `ArchiveSource` to get called with empty strings and false.

marksliva: Although I suppose this would prevent the program from crashing, which would be nice..

PR #321, opened 2020-04-28
Add centos6 jobs

marksliva:
Previously we were only testing upgrade on centos7.  Now, all upgrade jobs are
also ran on centos6, and to limit the workload, there are serial groups so that
a centos6 job must finish first before the respective centos7 job can start.

Another strategy to limit the strain on resources was setting the ccp reap time
to 3 hours instead of the default 8 hours.

The only job that currently fails is 6-to-6-centos-6 and has been removed for
now.


ci/parser/parse_template.go
```diff
@@ -29,16 +29,22 @@ import (
...
 	UseLinkMode    bool
 	RetailDemo     bool
+	CentosVersion  string
 }
 
 func (j *UpgradeJob) Name() string {
+	return fmt.Sprintf("%s-centos-%s", j.BaseName(), j.CentosVersion)
+}
+
+func (j *UpgradeJob) BaseName() string {
```
kalensk, line 47: It wasn't easily obvious to me what BaseName was used for. Perhaps a comment to the effect of the following?

> BaseName returns the pipeline job name without the operating system. This is used as a tag in Concourse's serial group to limit similar jobs between operating systems from running at once to avoid overloading Concourse.

ci/parser/parse_template.go
```diff
@@ -55,12 +61,18 @@ func (j *UpgradeJob) Name() string {
...
 }
 
+type CentosGPVersion struct {
+	CentosVersion string
+	GPVersion     string
+}
+
 type Data struct {
 	SourceVersions, TargetVersions []string
 	AllVersions                    []string // combination of Source/Target
```
kalensk, line 71: Is `AllVersions` still used? Seems like it got superseded by `CentosGPVersion`, which I think renaming to just `Version` might be nice.

kalensk:
Looks good overall! Glad we are adding centos6 as I believe that is the primary platform for upgrading from 5 to 6.

ci/template.yml
```diff
@@ -40,14 +40,14 @@ resources:
     private_key: ((retail-demo-git-key))
     branch: ((retail-demo-git-branch))
 
-{{range .AllVersions}}
-- name: rpm_gpdb{{.}}_centos7
+{{range .CentosGPVersions}}
```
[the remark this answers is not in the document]
marksliva: Since there are 4 other variables with `Version` in their name, perhaps renaming from `Centos` to `OS` would help, but I would prefer to keep the name a little longer so it's more clear what is in there.
```
	SourceVersions, TargetVersions []string
	AllVersions                    []string // combination of Source/Target
	UpgradeJobs                    []*UpgradeJob
	LastTargetVersion              string
	PrimariesOnly                  []bool
	CentosGPVersions               []*CentosGPVersion
```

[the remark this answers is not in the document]
marksliva: Actually, most of these fields are unused. I was able to clean it up significantly, and now `Data` looks like this (I ended up renaming it to `Versions` after all):
```
type Data struct {
	UpgradeJobs       []*UpgradeJob
	LastTargetVersion string
	Versions          []*Version
}
```

ci/parser/parse_template.go
marksliva: Removed, along with `SourceVersions, TargetVersions` and `PrimariesOnly`.

ci/parser/parse_template.go
marksliva: Done.

ci/template.yml
```diff
@@ -236,6 +236,9 @@ jobs:
 
 {{range .UpgradeJobs}}
 - name: {{ .Name }}
+  serial_groups: [{{ .BaseName }}]
+  # Specifying serial groups so that only one platform runs at a time. For
+  # example, 5-to-6-centos7 will only run after 5-to-6-centos6 completes.
```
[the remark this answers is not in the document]
marksliva: Added.

PR #322, opened 2020-04-29
initialize: add upgrade mode flag

kalensk:
Add upgrade mode flag to initialize that takes either copy or link.

**_Note_** the help text is not critical as this will be changing soon as we move to using a config file.

[Test pipeline](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:mode-flag)


ci/scripts/upgrade-cluster.bash
```diff
@@ -79,7 +79,7 @@ dump_sql $MASTER_PORT /tmp/old.sql
 # Now do the upgrade.
 LINK_MODE=""
 if [ "${USE_LINK_MODE}" = "1" ]; then
-    LINK_MODE="--link"
+    LINK_MODE="--mode=link"
```
marksliva, line 82: We could add a bats tests that also ensures that `--mode=copy` is also a valid option, in addition to the unit tests that were added.

cli/commands/commands.go
```diff
@@ -529,6 +533,20 @@ func parsePorts(val string) ([]uint32, error) {
...
 }
 
+// isLinkMode parses the mode flag returning an error if it is not copy or link.
+// It returns true if mode is link.
+func isLinkMode(mode string) (bool, error) {
+	choices := []string{"copy", "link"}
+
+	for _, choice := range choices {
+		if strings.ToLower(strings.TrimSpace(mode)) == choice {
+			return mode == "link", nil
```
marksliva, line 543: There's a slight issue here.. wouldn't this incorrectly return `false` when the user passes in `Link`? That would be a good test case to add.

cli/commands/commands.go
```diff
@@ -529,6 +533,20 @@ func parsePorts(val string) ([]uint32, error) {
...
+		}
+	}
+
+	return false, fmt.Errorf("Invalid mode %q. Please specify either copy or link mode.", mode)
```
marksliva, line 547: We could join `choices` with an ` or `, and render that instead of hard coding `copy` and `link`.

cli/commands/commands_test.go
```diff
@@ -56,3 +56,72 @@ func TestParsePorts(t *testing.T) {
...
+			}
+		})
+	}
+
+	errCases := []struct {
+		name string
+		mode string
+	}{
+		{
+			name: "empty string",
```
marksliva, line 106: nit: Add a few more error cases, or get rid of the cases loop.

hub/rename_data_directories.go
```diff
@@ -31,7 +31,7 @@ func UpdateDataDirectories(conf *Config, agentConns []*Connection) error {
 		return xerrors.Errorf("renaming master data directories: %w", err)
 	}
 
-	// in --link mode, remove the source mirror and standby data directories; otherwise we create a second copy
+	// in link mode, remove the source mirror and standby data directories; otherwise we create a second copy
```
marksliva, line 34: ..and could run out of space? Or is there a different reason that they are deleted instead of archived?

test/finalize.bats
```diff
@@ -132,8 +132,8 @@ upgrade_cluster() {
     upgrade_cluster
 }
 
-@test "gpupgrade finalize with --link mode should swap the primary and master directory and delete the old mirror and standby directory" {
-    upgrade_cluster "--link"
+@test "gpupgrade finalize with --mode=link mode should swap the primary and master directory and delete the old mirror and standby directory" {
```
marksliva, line 135: This is a little hard to read. It could be worded like: `gpupgrade finalize using hard links...`.

kalensk:
There are 2 files that still reference the old `--link` flag though:
> 
> ```
> upgrade/run.go
> upgrade/run_test.go
> ```

My understanding is that those are what actually call [pg_upgrade with --link](https://www.postgresql.org/docs/9.4/pgupgrade.html) and thus should use --link.

cli/commands/commands.go
kalensk: IMO that should be valid. Otherwise no need for strings.ToLower or perhaps even this function.

marksliva:
> There are 2 files that still reference the old `--link` flag though:
> 
> > ```
> > upgrade/run.go
> > upgrade/run_test.go
> > ```
> 
> My understanding is that those are what actually call [pg_upgrade with --link](https://www.postgresql.org/docs/9.4/pgupgrade.html) and thus should use --link.

Oh, oops, you are correct.

ci/scripts/upgrade-cluster.bash
kalensk: Added a story to track this called "add copy mode test to upgrade-cluster.bash".

cli/commands/commands.go
kalensk: From a brief offline discussion PM's are ok with allowing mode to be case insensitive. And we can always re-evaluate.

cli/commands/commands.go
kalensk: When writing that code, I was really debating between doing this or not. I'll update it.

cli/commands/commands.go
marksliva: Sure, however, let's say a user passes in `--mode=LINK`. It will match the 'link' choice because it gets lowercased which is fine, but then this function would still incorrectly return `false` for `linkMode` because it'll return the value of `"LINK" == "link"` when it should actually return `true`.

marksliva: And the same issue would happen for spaces as well: `--mode=' link '`.

kalensk: Thanks! fixed and updated the tests.

PR #323, opened 2020-05-01
revert: add verbose option

marksliva:
Always show revert substeps. Stream the directories being deleted to the CLI
when the verbose option is set.

Currently we are only streaming the directories that are being deleted on the
hub, but once we introduce a way to stream output to the hub from the agents,
we'll want to print the directories being deleted on each agent as well.


cli/commanders/steps.go
```diff
@@ -50,6 +50,11 @@ var SubstepDescriptions = map[idl.Substep]substepText{
 	idl.Substep_FINALIZE_START_TARGET_CLUSTER:                     substepText{"Starting target cluster...", "Start target cluster"},
 	idl.Substep_FINALIZE_UPGRADE_STANDBY:                          substepText{"Upgrading standby master...", "Upgrade standby master"},
 	idl.Substep_FINALIZE_UPGRADE_MIRRORS:                          substepText{"Upgrading mirror segments...", "Upgrade mirror segments"},
+	idl.Substep_DELETE_PRIMARY_DATADIRS:                           substepText{"Deleting primary datadirs...", "Delete primary datadirs"},
```
bhuvnesh2703, line 53: Delete primary segments datadirs (to be in sync with Upgrade primary segments)

cli/commanders/steps.go
```diff
@@ -50,6 +50,11 @@ var SubstepDescriptions = map[idl.Substep]substepText{
...
+	idl.Substep_DELETE_MASTER_DATADIR:                             substepText{"Deleting master datadir...", "Delete master datadir"},
+	idl.Substep_DELETE_NON_MASTER_STATEDIRS:                       substepText{"Deleting non-master state directories...", "Delete non-master state directories"},
```
bhuvnesh2703, line 55: Delete gpupgrade config / state directories?

kalensk, line 55: Maybe rephrase to something like "Deleting state directory on the segments"? Reading "non-master" is a little clunky.

cli/commanders/steps.go
kalensk, line 53: Since this UI facing maybe change `datadirs` to `data directories`, and similarly below.

cli/commands/commands.go
```diff
@@ -484,9 +502,37 @@ func revert() *cobra.Command {
...
+}
+
+func stopHubAndAgentsSubStep() error {
+	var err error
+
+	s := commanders.Substep(idl.Substep_STOP_HUB_AND_AGENTS)
+	defer s.Finish(&err)
+
+	err = stopHubAndAgents()
+	if err != nil {
```
bhuvnesh2703, line 529: can potentially return  stopHubAndAgents()

cli/commands/commands.go
```diff
@@ -453,25 +454,42 @@ func finalize() *cobra.Command {
 	return addHelpToCommand(cmd, FinalizeHelp)
 }
 
+type stdStream struct {
+}
```
kalensk, line 458: I know its used only here, but I could also see it being in `utils` next to `DevNull`, so they are all in one place.

cli/commands/commands.go
```diff
@@ -453,25 +454,42 @@ func finalize() *cobra.Command {
...
 			}
 
-			err = stopHubAndAgents()
+			err = stopHubAndAgentsSubStep()
 			if err != nil {
 				return err
 			}
 
-			err = upgrade.DeleteDirectories([]string{utils.GetStateDir()}, upgrade.StateDirectoryFiles)
+			err = deleteMasterStateDirectory()
```
kalensk, line 492: Im curious if there is a way of inlining these functions since they aren't used anywhere else. It seems like it would make the code a lot cleaner.

hub/revert.go
```diff
@@ -5,21 +5,32 @@ package hub
...
 import (
 	"github.com/greenplum-db/gpupgrade/idl"
+	"github.com/greenplum-db/gpupgrade/step"
 	"github.com/greenplum-db/gpupgrade/upgrade"
 )
 
 func (s *Server) Revert(_ *idl.RevertRequest, stream idl.CliToHub_RevertServer) (err error) {
+	st, err := step.Begin(s.StateDir, "revert", stream)
+	if err != nil {
+		return nil
```
bhuvnesh2703, line 15: why return nil when err is not nil?

utils/sys_utils.go
```diff
@@ -133,3 +134,16 @@ func CreateDataDirectory(dataDir string) error {
 	}
 	return nil
 }
+
+// DevNull implements OutStreams by just discarding all writes.
+var DevNull = devNull{}
```
kalensk, line 139: For future reference, I think it would make reviewing easier if this move was done in a separate commit.

hub/revert.go
```diff
@@ -5,21 +5,32 @@ package hub
...
+			hostname := s.Config.Target.MasterHostname()
+
+			return upgrade.DeleteDirectories([]string{datadir}, upgrade.PostgresFiles, hostname, streams)
+		})
 	}
 
-	return DeleteStateDirectories(s.agentConns, s.Source.MasterHostname())
+	st.Run(idl.Substep_DELETE_NON_MASTER_STATEDIRS, func(_ step.OutStreams) error {
+		return DeleteStateDirectories(s.agentConns, s.Source.MasterHostname())
+	})
```
kalensk, line 33: Its a little confusing that the substep description is `Substep_DELETE_NON_MASTER_STATEDIRS` but then it calls `DeleteStateDirectories` with the `masterHostname`? Is that all correct?

hub/revert.go
```diff
@@ -5,21 +5,32 @@ package hub
...
-		err = DeletePrimaryDataDirectories(s.agentConns, s.Config.Target)
-		if err != nil {
-			return err
-		}
-
-		err = upgrade.DeleteDirectories([]string{s.Config.Target.MasterDataDir()}, upgrade.PostgresFiles)
-		if err != nil {
-			return err
-		}
+		st.Run(idl.Substep_DELETE_PRIMARY_DATADIRS, func(_ step.OutStreams) error {
```
bhuvnesh2703, line 19: might need to review with the PM team about the description of these substeps and how many Steps we want to show. if you have already checked ignore..

step/step_test.go
```diff
@@ -348,24 +348,24 @@ func (t *TestStore) Write(substep idl.Substep, status idl.Status) (err error) {
 	return t.WriteErr
 }
 
-// DevNull implements step.OutStreamsCloser as a no-op. It also tracks calls to
+// DevNullWithClose implements step.OutStreamsCloser as a no-op. It also tracks calls to
```
kalensk, line 351: Maybe not for this PR, but I wonder if all these OutStream implementations should go together since its easy to find and see which one to use. I think that might be nice.

upgrade/directories_test.go
```diff
@@ -328,13 +348,20 @@ func TestDeleteDirectories(t *testing.T) {
 				t.Errorf("dataDir %s exists", dataDir)
 			}
 		}
+
+		expected := regexp.MustCompile("Deleting directory: .*/data/dbfast_mirror1/seg1\" on host \"localhost.local\"\nDeleting directory: \".*/data/dbfast_mirror2/seg2\" on host \"localhost.local")
+
+		actual := buf.String()
+		if !expected.MatchString(actual) {
+			t.Errorf("got stream output %s want %s", actual, expected)
+		}
```
kalensk, line 357: We should not be matching error strings.

Some of the other alternatives include:
1) Matching error types. For example, we do this in [cluster_test.go](https://github.com/greenplum-db/gpupgrade/blob/cb3561cb60bfacc9fd9b9bc23a923362c9df7666/greenplum/cluster_test.go#L160-L167) with [ErrInvalidSegments](https://github.com/greenplum-db/gpupgrade/blob/master/greenplum/cluster.go#L178-L180) and [InvalidSegmentsError](https://github.com/greenplum-db/gpupgrade/blob/master/greenplum/cluster.go#L230-L251).

2) Simply see if we get _any_ kind of error back. We do this in [commands_test.go](https://github.com/greenplum-db/gpupgrade/blob/master/cli/commands/commands_test.go#L52-L57)

3) Other ways...

step/step_test.go
kalensk: Also for future reference doing this rename to DevNullWithClose in a separate commit might make things a bit easier for reviewing.

agent/delete_directories.go
```diff
@@ -10,20 +10,25 @@ import (
...
+	"github.com/greenplum-db/gpupgrade/utils"
 )
 
 var deleteDirectories = upgrade.DeleteDirectories
 
 func (s *Server) DeleteStateDirectory(ctx context.Context, in *idl.DeleteStateDirectoryRequest) (*idl.DeleteStateDirectoryReply, error) {
 	gplog.Info("got a request to delete the state directory from the hub")
 
-	err := deleteDirectories([]string{s.conf.StateDir}, upgrade.StateDirectoryFiles)
+	hostname, _ := utils.System.Hostname()
```
kalensk, line 21: I am not sure we should be ignoring these errors everywhere.

upgrade/directories_test.go
marksliva: This is not testing an error message. This checks that the correct output was streamed to the buffer through `Stdout`.
```
		devNull := devNullSpy{
			outStream: &buf,
		}
```

step/step_test.go
marksliva: I was thinking about that, however, since this is used only in testing, this seems like a better place for it than putting it in `sys_utils`, which has code that's used in both production code and the tests.

agent/delete_directories.go
marksliva: I'll remove this from the agents, since we know it is not currently being used anyway.

cli/commands/commands.go
marksliva: The main reason to do it this way was to utilize the `defer s.Finish(&err)` pattern. But I suppose we could inline it if that makes it easier to read.

hub/revert.go
marksliva: Exactly. The `masterHostname` is used to exclude any agent that is running on the master host. The variable could be renamed to `excludeHostname` if that would make it easier to follow?

cli/commands/commands.go
marksliva: Deleting the state directory looks like this when inlined:
```
			hostname, err := os.Hostname()
			if err != nil {
				s.Finish(&err)
				return err
			}

			err = upgrade.DeleteDirectories([]string{utils.GetStateDir()}, upgrade.StateDirectoryFiles, hostname, &utils.StdStream{})
			if err != nil {
				s.Finish(&err)
				gplog.Error(err.Error())
				return err
			}
			s.Finish(&err)
```

cli/commanders/steps.go
marksliva: What do you mean by config? By deleting the root level state directory, wouldn't that end up deleting all of the configs?

hub/revert.go
marksliva: good catch.

cli/commanders/steps.go
marksliva: Done.

cli/commands/commands.go
marksliva: not possible anymore, since this was inlined.

cli/commanders/steps.go
marksliva: Done.

cli/commands/commands.go
marksliva: Moved it to utils.

hub/revert.go
marksliva: I renamed it to `excludeHostname`.

cli/commanders/steps.go
marksliva: Done.

upgrade/directories_test.go
```diff
@@ -328,13 +348,20 @@ func TestDeleteDirectories(t *testing.T) {
 				t.Errorf("dataDir %s exists", dataDir)
 			}
 		}
+
+		expected := regexp.MustCompile("Deleting directory: .*/data/dbfast_mirror1/seg1\" on host \"localhost.local\"\nDeleting directory: \".*/data/dbfast_mirror2/seg2\" on host \"localhost.local")
```
kalensk, line 352: nite: I believe if you switch the outer quotes to back tics you won't have to escape the double quotes inside. It sometimes makes things much more readable.

upgrade/directories_test.go
kalensk: Ah, my bad!

cli/commands/commands.go
```diff
@@ -454,28 +454,48 @@ func finalize() *cobra.Command {
...
 			if err != nil {
 				return err
 			}
 
-			err = upgrade.DeleteDirectories([]string{utils.GetStateDir()}, upgrade.StateDirectoryFiles)
+			s = commanders.Substep(idl.Substep_DELETE_MASTER_STATEDIR)
+
+			hostname, err := os.Hostname()
 			if err != nil {
+				s.Finish(&err)
```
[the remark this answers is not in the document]
marksliva: 2 of them can be combined, but we should only call `s.Finish(&err)` a single time in our flow. So this first `s.Finish` should stay since we return right afterwards, and the 2 after this can be combined and called first, since we'll either have an error or nil.
```
			err = upgrade.DeleteDirectories([]string{utils.GetStateDir()}, upgrade.StateDirectoryFiles, hostname, &utils.StdStream{})
			s.Finish(&err)
```

agent/delete_directories.go
```diff
@@ -10,20 +10,21 @@ import (
...
+	"github.com/greenplum-db/gpupgrade/utils"
 )
 
 var deleteDirectories = upgrade.DeleteDirectories
 
 func (s *Server) DeleteStateDirectory(ctx context.Context, in *idl.DeleteStateDirectoryRequest) (*idl.DeleteStateDirectoryReply, error) {
 	gplog.Info("got a request to delete the state directory from the hub")
 
-	err := deleteDirectories([]string{s.conf.StateDir}, upgrade.StateDirectoryFiles)
+	err := deleteDirectories([]string{s.conf.StateDir}, upgrade.StateDirectoryFiles, "", utils.DevNull)
```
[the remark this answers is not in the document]
marksliva: Done.

upgrade/directories_test.go
marksliva: Done.

cli/commands/commands.go
```diff
@@ -454,28 +454,48 @@ func finalize() *cobra.Command {
...
-			err := commanders.Revert(client)
+
+			err := commanders.Revert(client, verbose)
 			if err != nil {
 				gplog.Error(err.Error())
 				return err
 			}
 
+			s := commanders.Substep(idl.Substep_STOP_HUB_AND_AGENTS)
+
```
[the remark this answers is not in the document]
marksliva: Done.

PR #324, opened 2020-05-01
remove ginkgo

kalensk:
- migrates `daemon_test.go` and `sys_utils_test.go` to Go style tests
  - I kept the test logic and code the same with the goal of simply porting it to go style.
- Removes dangling references to ginkgo
- Removes remaining usages of gomega from tests to Go style

**_Note:_** It seems like one of the remaining uses of gomega is `testhelper.SetupTestLogger()` and `testutils.VerifyLogContains()` which use `gbytes.Buffer` from gomega.

[Test Pipeline](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:go-tests)


utils/daemon/daemon_test.go
```diff
@@ -105,76 +106,117 @@ func (m *MockDaemonizableCommand) StderrPipe() (io.ReadCloser, error) {
...
 
 		cmd := MockDaemonizableCommand{StdoutBuf: output, StderrBuf: errput}
 		err := waitForDaemon(&cmd, outbuf, errbuf, 0)
-		Expect(err).NotTo(HaveOccurred())
+		if err != nil {
+			t.Errorf("unexpected error %#v", err)
+		}
+
+		if !bytes.Equal(outbuf.Bytes(), output) {
+			t.Errorf("got %v want %v", outbuf, output)
```
marksliva, line 209: Each time the expected output is printed in the error string, it should be cast to a string using `string(output)`. Otherwise, the error message looks like this:
```
 daemon_test.go:209: got this is output
         want [116 104 105 115 32 105 115 32 111 117 116 112 117 116 10]
```
And it would also help to format both strings with `%q` to see all differences including trailing spaces:
`got "1this is output\n" want "this is output\n"`

utils/sys_utils_test.go
```diff
@@ -12,102 +12,121 @@ import (
...
+		defer resetSystemFunctions()
+
+		expected := "val"
+		System.Getenv = func(s string) string {
+			return expected
+		}
 
-	AfterEach(func() {
-		System = InitializeSystemFunctions()
+		rc := TryEnv("VAR", "default")
```
marksliva, line 30: nit: `rc` could be renamed to `actual` or `value`.

utils/sys_utils_test.go
```diff
@@ -12,102 +12,121 @@ import (
...
+			}, nil
+		}
+
+		user, dir, err := GetUser()
+		if err != nil {
+			t.Errorf("unexpected error %#v", err)
+		}
+
+		if user != expectedUser {
+			t.Errorf("got %q want %q", user, expectedUser)
```
marksliva, line 68: nit: add `got user ...` for better context in the error message.

utils/sys_utils_test.go
```diff
@@ -12,102 +12,121 @@ import (
...
+		}
+
+		if dir != expectedDir {
+			t.Errorf("got %q want %q", dir, expectedDir)
```
marksliva, line 72: nit: add `got dir ...` for added context.

kalensk:
I believe all comments have been addressed. Thanks!

PR #327, opened 2020-05-06
Initialize step logs in gpAdminLogs subdirectory

dgkimura:
Logs created in ~/gpAdminLogs/gpupgrade directory with same file pattern
as gpupgrade_cli_[timestamp]. Unfortunately for us, gplog is currently a
singleton, so we cannot easily reuse that library to create multiple log
files.

### DIFF
--- a/cli/commands/agent.go
+++ b/cli/commands/agent.go
@@ -24,7 +24,11 @@ func Agent() *cobra.Command {
 		Hidden: true,
 		Args:   cobra.MaximumNArgs(0), //no positional args allowed
 		RunE: func(cmd *cobra.Command, args []string) error {
-			gplog.InitializeLogging("gpupgrade agent", "")
+			logdir, err := utils.GetLogDir()
+			if err != nil {
+				return err
+			}
+			gplog.InitializeLogging("gpupgrade_agent", logdir)
 			defer log.WritePanics()
 
 			conf := agent.Config{
--- a/cli/commands/hub.go
+++ b/cli/commands/hub.go
@@ -32,7 +32,11 @@ func Hub() *cobra.Command {
 		Hidden: true,
 		Args:   cobra.MaximumNArgs(0), //no positional args allowed
 		RunE: func(cmd *cobra.Command, args []string) error {
-			gplog.InitializeLogging("gpupgrade hub", "")
+			logdir, err := utils.GetLogDir()
+			if err != nil {
+				return err
+			}
+			gplog.InitializeLogging("gpupgrade_hub", logdir)
 			debug.SetTraceback("all")
 			defer log.WritePanics()
 
--- a/cmd/gpupgrade/main.go
+++ b/cmd/gpupgrade/main.go
@@ -12,25 +12,26 @@ import (
 	_ "github.com/lib/pq"
 
 	"github.com/greenplum-db/gpupgrade/cli/commands"
+	"github.com/greenplum-db/gpupgrade/utils"
 	"github.com/greenplum-db/gpupgrade/utils/daemon"
 )
 
 func main() {
-	setUpLogging()
+	debug.SetTraceback("all")
+	logdir, err := utils.GetLogDir()
+	if err != nil {
+		fmt.Printf("\n%+v\n", err)
+		os.Exit(1)
+	}
+	gplog.InitializeLogging("gpupgrade_cli", logdir)
 
 	root := commands.BuildRootCommand()
 	root.SilenceErrors = true // we'll print these ourselves
 
-	err := root.Execute()
+	err = root.Execute()
 	if err != nil && err != daemon.ErrSuccessfullyDaemonized {
 		// Use v to print the stack trace of an object errors.
 		fmt.Printf("\n%+v\n", err)
 		os.Exit(1)
 	}
-}
-
-func setUpLogging() {
-	debug.SetTraceback("all")
-	//empty logdir defaults to ~/gpAdminLogs
-	gplog.InitializeLogging("gpupgrade_cli", "")
 }
--- a/step/step.go
+++ b/step/step.go
@@ -9,10 +9,12 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/greenplum-db/gp-common-go-libs/operating"
 	multierror "github.com/hashicorp/go-multierror"
 	"golang.org/x/xerrors"
 
 	"github.com/greenplum-db/gpupgrade/idl"
+	"github.com/greenplum-db/gpupgrade/utils"
 )
 
 type Step struct {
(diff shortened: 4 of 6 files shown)


cmd/gpupgrade/main.go
```diff
@@ -5,6 +5,7 @@ package main
 
 import (
 	"fmt"
+	"github.com/greenplum-db/gp-common-go-libs/operating"
```
bhuvnesh2703, line 8: this goes into the next block with "github.com/greenplum-db/gp-common-go-libs/gplog" etc..

step/step.go
```diff
@@ -33,7 +34,10 @@ func New(name string, sender idl.MessageSender, store Store, streams OutStreamsC
 }
 
 func Begin(stateDir string, name string, sender idl.MessageSender) (*Step, error) {
-	path := filepath.Join(stateDir, fmt.Sprintf("%s.log", name))
+	currentUser, _ := operating.System.CurrentUser()
+	logDir := fmt.Sprintf("%s/gpAdminLogs/gpupgrade", currentUser.HomeDir)
+
+	path := filepath.Join(logDir, fmt.Sprintf("%s_%s.log", name, operating.System.Now().Format("20060102")))
```
bhuvnesh2703, line 40: may be a PM questions, but how to we want to name the log? should we prepend gpupgrade_ to the name?
gpupgrade_intitialize_<timestamp>.log
gpupgrade_execute_<timestamp>.log
gpupgrade_finalize_<timestamp>.log

step/step.go
```diff
@@ -33,7 +34,10 @@ func New(name string, sender idl.MessageSender, store Store, streams OutStreamsC
 }
 
 func Begin(stateDir string, name string, sender idl.MessageSender) (*Step, error) {
-	path := filepath.Join(stateDir, fmt.Sprintf("%s.log", name))
+	currentUser, _ := operating.System.CurrentUser()
```
bhuvnesh2703, line 37: may be these 2 lines goes to common function called here and in setUpLogging?

cmd/gpupgrade/main.go
```diff
@@ -31,6 +32,9 @@ func main() {
 
 func setUpLogging() {
 	debug.SetTraceback("all")
-	//empty logdir defaults to ~/gpAdminLogs
-	gplog.InitializeLogging("gpupgrade_cli", "")
+
+	currentUser, _ := operating.System.CurrentUser()
```
marksliva, line 36: We should check the error that comes back, and return it to the caller of `setUpLogging` if it is not nil.

step/step.go
marksliva, line 37: Same thing here, about inspecting the error that `CurrentUser()` returns.

marksliva:
There is a test failure in `pg_upgrade.bats` in the `install-tests` job because it is hard coded to look for the `initialize.log` in the old location:
`fatal" "$GPUPGRADE_HOME"/initialize.log' failed with status 2`

utils/sys_utils.go
```diff
@@ -106,6 +107,16 @@ func GetStateDir() string {
 	return stateDir
 }
 
+func GetLogDir() (string, error) {
+	user, err := System.CurrentUser()
```
marksliva, line 111: I get this warning in my IDE `Variable 'user' collides with imported package name`. This is due to importing `"os/user"` above. It could be renamed to something like `currentUser`.

utils/sys_utils.go
```diff
@@ -106,6 +107,16 @@ func GetStateDir() string {
...
+	if err != nil {
+		return "", err
+	}
+
+	logDir := fmt.Sprintf("%s/gpAdminLogs/gpupgrade", user.HomeDir)
```
marksliva, line 116: This could be rewritten as `logDir := filepath.Join(user.HomeDir, "gpAdminLogs", "gpupgrade")`.

dgkimura:
On Thu, May 7, 2020 at 12:12 PM Mark Sliva wrote:
>
> There is a test failure in pg_upgrade.bats in the install-tests job because it is hard coded to look for the initialize.log in the old location:
> fatal" "$GPUPGRADE_HOME"/initialize.log' failed with status 2

Ah, thanks! I was wondering about that.

cli/commands/agent.go
```diff
@@ -24,7 +24,11 @@ func Agent() *cobra.Command {
 		Hidden: true,
 		Args:   cobra.MaximumNArgs(0), //no positional args allowed
 		RunE: func(cmd *cobra.Command, args []string) error {
-			gplog.InitializeLogging("gpupgrade agent", "")
```
bhuvnesh2703, line 27: InitializeLogging
```
func InitializeLogging(program string, logdir string) {
	if logger != nil {
		return
	}
	currentUser, _ := operating.System.CurrentUser()
	if logdir == "" {
		logdir = fmt.Sprintf("%s/gpAdminLogs", currentUser.HomeDir)
	}

	createLogDirectory(logdir)

	logfile := GenerateLogFileName(program, logdir)
	logFileHandle := openLogFile(logfile)

	logger = NewLogger(os.Stdout, os.Stderr, logFileHandle, logfile, LOGINFO, program)
	SetExitFunc(defaultExit)
}
```
calls GenerateLogFileName(program, logdir)
```
func GenerateLogFileName(program, logdir string) string {
	var logfile string
	if logFileNameFunc != nil {
		logfile = logFileNameFunc(program, logdir)
	} else {
		timestamp := operating.System.Now().Format("20060102")
		logfile = fmt.Sprintf("%s/%s_%s.log", logdir, program, timestamp)
	}
	return logfile
}
```

where if we are passing the program name with spaces, i think it will cause issues, but since we don't get here when we run the command, we dont see error.
But we should pass gpupgrade_hub `without spaces`

In a followup story, we can potentially add tests or fix run gpupgrade hub --daemonize etc commands to ensure that we are doing the right thing.

cli/commands/agent.go
```diff
@@ -24,7 +24,11 @@ func Agent() *cobra.Command {
...
+			logdir, err := utils.GetLogDir()
+			if err != nil {
+				return err
+			}
+			gplog.InitializeLogging("gpupgrade agent", logdir)
```
bhuvnesh2703, line 31: gpupgrade_agent

cli/commands/hub.go
```diff
@@ -32,7 +32,11 @@ func Hub() *cobra.Command {
 		Hidden: true,
 		Args:   cobra.MaximumNArgs(0), //no positional args allowed
 		RunE: func(cmd *cobra.Command, args []string) error {
-			gplog.InitializeLogging("gpupgrade hub", "")
+			logdir, err := utils.GetLogDir()
+			if err != nil {
+				return err
+			}
+			gplog.InitializeLogging("gpupgrade hub", logdir)
```
bhuvnesh2703, line 39: ditto

dgkimura:
Thanks for the reviews!

PR #328, opened 2020-05-06
Upgrade tablespace in gpdb5 clusters to gpdb6 clusters

bhuvnesh2703:
This PR introduces the functionality of upgrading gpdb5 clusters with tablespaces. I have tried to arrange the commits which makes reviewing easy. Here is the flow of the upgrade process

1. During Initialize
  a) query the database to extract the tablespace information which is used for upgrade. Populate the information to config.json so that hub is aware of it.
  b) Write the tablespace information to a mapping file which is passed as a parameter to pg_upgrade while upgrading segments. Master doesn't need it.
2. During execute
  a) Copy the master tablespaces & mapping file to ~/.gpupgrade/tablespace location on the segments
  b) Pass the information for the segment tablespaces to the segments
  c) Use the supplied information during upgrade of segments to perform rsync of tablespaces to segment tablespace location and recreate the symbolic links.
3) During finalize: No action required, gpaddmirrors and gpinitstandby takes care of addition of the mirror tablespaces

Also corresponding GPDB PR: https://github.com/greenplum-db/gpdb/pull/10053
There is no need to pass the tablespaces mapping file during --check as there are no tablespace specific check which are performed. Happy to take input here if the team wants it otherwise.

Note:
1) The bats tests currently works on gpdb6 clusters, so have not included a test for upgrade tablespaces. I will be happy to do it as a followup commit, also, we should make bats more flexible to upgrade gpdb5 or gpdb6 cluster.
2) Don't see a point of adding a test of upgrading tablespaces from gpdb6 to gpdb6 version. We can revisit it.


greenplum/tablespace.go
```diff
@@ -0,0 +1,141 @@
...
+
+	return clusterTableSpaceMap
+}
+
+// write the tuples returned from the database to a csv file
+func (t TableSpaceTuples) Write(w io.Writer) error {
+	writer := csv.NewWriter(w)
+	for _, tableSpace := range t {
+		var s []string
+		s = append(
```
kalensk, line 101: I would inline this function. I am not sure its getting us much.


Why not just create a slice literal? No need to append or anything. 
```go
line := []string{
            strconv.Itoa(tableSpace.DbId),
			strconv.Itoa(tableSpace.Oid),
			tableSpace.Name,
			tableSpace.Location,
			strconv.Itoa(tableSpace.IsUserDefined))
}
```

Not needed, but another possibility is just `csv.NewWriter(file).Write(line)`

hub/fill_cluster_configs.go
```diff
@@ -49,6 +50,18 @@ func FillClusterConfigsSubStep(config *Config, conn *sql.DB, _ step.OutStreams,
 		return err
 	}
 
+	// major version upgrade requires upgrading tablespaces
+	if dbconn.Version.Is("5") {
+		if err := utils.System.MkdirAll(utils.GetTableSpaceDir(), 0700); err != nil {
+			return errors.Wrapf(err, "failed to create tablespace directory %q", utils.GetTableSpaceDir())
+		}
+		config.TableSpacesMappingFile = filepath.Join(utils.GetTableSpaceDir(), greenplum.TableSpacesMappingFile)
```
kalensk, line 58: Does `TableSpacesMappingFile` really need to be on the `config` object? Looks like its just used here locally.

utils/sys_utils.go
```diff
@@ -22,6 +22,10 @@ var (
 
 const markerFile = ".gpupgrade"
 
+// directory in which the tablespaces will be copied from the master to the
+// segment hosts
+const TableSpacesDir = "tablespaces"
```
kalensk, line 28: I would inline this since its only used once which is locally in this file.

hub/copy_master.go
```diff
@@ -47,17 +38,18 @@ func (s *Server) CopyMasterDataDir(streams step.OutStreams, destinationDir strin
...
+			cmd := execCommand("rsync", args...)
 
 			result := Result{}
 			cmd.Stdout = &result.stdout
 			cmd.Stderr = &result.stderr
 
 			err := cmd.Run()
 			if err != nil {
-				err = xerrors.Errorf("copying master data directory to host %s: %w", hostname, err)
+				err = xerrors.Errorf("copying directory to host %s: %w", hostname, err)
```
kalensk, line 52: Do we want to include the source and destination directories in the error message? Just a thought.

testutils/test_utils.go
```diff
@@ -165,3 +165,50 @@ func verifyLog(t *testing.T, testlog *gbytes.Buffer, expected string, shouldCont
 		t.Errorf("\nexpected log: %q\n%s:   %q", contents, text, expected)
 	}
 }
+
+func CreateSampleTableSpaces() greenplum.TableSpaces {
```
kalensk, line 169: This doesn't seem needed to me. Just inline a much simpler down version in server_internal_test.go.

hub/server_internal_test.go
```diff
@@ -27,10 +27,12 @@ func TestConfig(t *testing.T) {
...
 			targetInitializeConfig,
-			12345,           // Port
-			54321,           // AgentPort
-			false,           // UseLinkMode
-			upgrade.NewID(), // UpgradeID
+			12345,                               // Port
+			54321,                               // AgentPort
+			false,                               // UseLinkMode
+			upgrade.NewID(),                     // UpgradeID
+			testutils.CreateSampleTableSpaces(), // TableSpaces
```
kalensk, line 34: We should replace with either:
```go
map[int]greenplum.SegmentTableSpaces{
		1: {
			1663: {
				Location:      "/tmp/master/my_tablespace/1663",
				IsUserDefined: 1,
			},
   },
},
```

Or 

```go
map[int]greenplum.SegmentTableSpaces{}
```

hub/server_internal_test.go
```diff
@@ -27,10 +27,12 @@ func TestConfig(t *testing.T) {
...
+			"/tmp/tablespaces.txt",              // TableSpacesMappingFile
```
kalensk, line 35: Let's use the constant `TableSpacesMappingFile = "tablespaces.txt"`

greenplum/tablespace_test.go
```diff
@@ -0,0 +1,367 @@
...
+					&TableSpaceInfo{
+						"/tmp/master/gpseg-1",
+						1,
+					},
+				},
+			},
+			expected: "1,1663,default,/tmp/master/gpseg-1,0\n2,1664,my_tablespace,/tmp/master/gpseg-1,1\n",
+		},
+	}
+	for _, tt := range tests {
```
kalensk, line 355: minor nit: rename `tt` to `test` for clarity.

greenplum/tablespace_test.go
```diff
@@ -0,0 +1,367 @@
...
+// response to a tablespace query.
+func MockTableSpaceQueryResult() *sqlmock.Rows {
+	rows := sqlmock.NewRows([]string{"dbid", "oid", "name", "location", "isuserdefined"})
+	rows.AddRow(1, 1663, "pg_default", "/tmp/master_tablespace", 0)
+	rows.AddRow(2, 1663, "pg_default", "/tmp/my_tablespace", 0)
+
+	return rows
+}
+
+func TestTableSpacesFromDB(t *testing.T) {
```
kalensk, line 232: Running code coverage shows we are missing error case tests for when utils.System.Create` fails and `tableSpaceTuples.Write` fails.

greenplum/tablespace_test.go
```diff
@@ -0,0 +1,367 @@
...
+			name:           "tablespace query execution failed",
+			rows:           nil,
+			versionStr:     "",
+			expectedTuples: nil,
+			error:          errors.New("failed to execute query"),
+		},
+	}
+
+	for _, c := range cases {
+		t.Run(fmt.Sprintf("%s", c.name), func(t *testing.T) {
```
kalensk, line 71: There are a few lint issues from `make lint`. 

The `fmt.Sprintf` is not needed for this particular line.

greenplum/tablespace_test.go
```diff
@@ -0,0 +1,367 @@
...
+			}
+
+			if c.versionStr != "" {
+				testhelper.SetDBVersion(conn, c.versionStr)
+			}
+
+			results, err := GetTableSpaceTuples(conn)
+			if c.error != nil && !strings.Contains(err.Error(), c.error.Error()) {
+				t.Errorf("got %+v, want %+v", err, c.error)
+			}
```
kalensk, line 97: I know this is a hold over from cluster.go, but we need a chore to compare error types and not strings. This is needs to be prioritized.

Instead of comparing error strings we could just do:
```go
	if err == nil {
			t.Fatal("GetTableSpaceTubles() returned no error")
		}
```

hub/copy_master.go
```diff
@@ -84,3 +84,19 @@ func (s *Server) CopyMasterDataDir(streams step.OutStreams, destination string)
...
 }
+
+func (s *Server) CopyMasterTableSpaces(streams step.OutStreams, destinationDir string) error {
+	// include tablespace mapping file which is used as a parameter to pg_upgrade
+	sourcePaths := []string{s.TableSpacesMappingFile}
+
+	// include all the master tablespace directories
+	masterTableSpaces := s.TableSpaces[1]
+	for _, tableSpace := range masterTableSpaces {
+		if tableSpace.IsUserDefined == 0 {
```
kalensk, line 95: This seems perfect for adding a helper function to tabeleSpace. That is, 

```
if !tableSpace.IsUserDefined() {
   continue
}
```

where 
```go
func (t *TableSpaceInfo) IsUserDefined() bool {
	return t.IsUserDefined == 0
}
```

**_However_**, you will have to rename the `IsUserDefined` to not collide with the function with the same name.

hub/copy_master.go
```diff
@@ -84,3 +84,19 @@ func (s *Server) CopyMasterDataDir(streams step.OutStreams, destination string)
 	source := []string{filepath.Clean(s.Target.MasterDataDir()) + string(filepath.Separator)}
 	return Copy(streams, destination, source, s.Target.PrimaryHostnames())
 }
+
+func (s *Server) CopyMasterTableSpaces(streams step.OutStreams, destinationDir string) error {
+	// include tablespace mapping file which is used as a parameter to pg_upgrade
+	sourcePaths := []string{s.TableSpacesMappingFile}
+
+	// include all the master tablespace directories
+	masterTableSpaces := s.TableSpaces[1]
```
kalensk, line 93: Adding a method for s.TableSpaces[1] makes the code self-documenting and no need for the comment.

Try adding this helper:
```go
func (t TableSpaces) MasterTableSpaces() SegmentTableSpaces {
	return t[1]
```

To allow:
```go
for _, tableSpace := range s.TableSpaces.MasterTableSpaces() {
   // ...
}
```

agent/upgrade_primary.go
```diff
@@ -46,6 +46,12 @@ func performUpgrade(segment Segment, request *idl.UpgradePrimariesRequest) error
 
 	if request.CheckOnly {
 		options = append(options, upgrade.WithCheckOnly())
+	} else {
+		// During gpupgrade execute, tablepace mapping file is copied after
+		// the master has been upgraded. So, don't pass this option during
+		// --check mode. There is no test in pg_upgrade which depends on the
+		// existence of this file.
+		options = append(options, upgrade.WithTableSpaceFile(request.TableSpacesMappingFile))
```
kalensk, line 66: Needing an else like this seems like a code smell in a sense. Let's spend some time together to see if we can simplify this code a bit or make it "cleaner" in a sense.

My initial thought is to always pass `upgrade.WithTableSpaceFile(request.TableSpacesMappingFile))` and have the upgrade.Run layer further down know whether to call that with pg_upgrade or not.

agent/upgrade_primary.go
```diff
@@ -76,3 +88,30 @@ func restoreBackup(request *idl.UpgradePrimariesRequest, segment Segment) error
...
 	})
 }
+
+func RestoreTableSpaces(request *idl.UpgradePrimariesRequest, segment Segment) error {
+	if request.CheckOnly {
+		return nil
+	}
+
+	for oid, tableSpace := range segment.TableSpaces {
+		if tableSpace.IsUserDefined == 0 {
```
kalensk, line 98: Another good place to have a helper function. `if !tableSpace.IsUserDefined()`

agent/upgrade_primary.go
```diff
@@ -76,3 +88,30 @@ func restoreBackup(request *idl.UpgradePrimariesRequest, segment Segment) error
...
+			continue
+		}
+
+		targetDir := filepath.Join(tableSpace.Location, strconv.Itoa(int(segment.DBID)))
+
+		tableSpaceMasterBackupDir := filepath.Dir(request.TableSpacesMappingFile)
+		sourceDir := filepath.Join(tableSpaceMasterBackupDir, strconv.Itoa(int(oid)), strconv.Itoa(greenplum.MasterDbid))
```
kalensk, line 105: Without fully knowing all the nitty gritty details, it seems like the logic in these lines:
`targetDir := filepath.Join(tableSpace.Location, strconv.Itoa(int(segment.DBID)))`
`sourceDir := filepath.Join(tableSpaceMasterBackupDir, strconv.Itoa(int(oid)), strconv.Itoa(greenplum.MasterDbid))`

would make a lot more sense as a helper function of the TabeleSpace struct or something similar.

utils/sys_utils.go
```diff
@@ -168,3 +173,20 @@ func (m *StdStream) Stderr() io.Writer {
 func GetTableSpaceDir() string {
 	return filepath.Join(GetStateDir(), TableSpacesDir)
 }
+
+func ReCreateSymLink(sourceDir, symLinkName string) error {
```
kalensk, line 177: I am not sure this belongs in sys_utils.go just yet since it's only used in upgrade_primary.go in a specific context. I don't really see it being used in multiple places as a utility function, at least not any time soon. Thoughts?

hub/execute.go
```diff
@@ -60,7 +60,19 @@ func (s *Server) Execute(request *idl.ExecuteRequest, stream idl.CliToHub_Execut
 	})
 
 	st.Run(idl.Substep_COPY_MASTER, func(streams step.OutStreams) error {
-		return s.CopyMasterDataDir(streams, upgradedMasterBackupDir)
+		err := s.CopyMasterDataDir(streams, upgradedMasterBackupDir)
+		if err != nil {
+			return err
+		}
+
+		if s.TableSpaces != nil {
```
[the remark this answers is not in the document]
kalensk: +1

hub/upgrade_primaries.go
```diff
@@ -99,14 +99,19 @@ func (s *Server) GetDataDirPairs() (map[string][]*idl.DataDirPair, error) {
 				contentID, sourceSeg.Hostname, targetSeg.Hostname)
 		}
 
+		tableSpaces := map[int32]*idl.TableSpaceInfo{}
+		if s.TableSpaces != nil {
```
kalensk, line 103: Huge caveat warning in that I haven't fully looked at all the details concerning this.

Seeing this reminds me of this concept of whenever a resource is initialized then it should be available to use and thus preventing the need for any sort of checks. This greatly simplifies code and makes reasoning about things much easer with less bugs. That is, don't create a tablespace if its not needed. This concept is somewhat tied to RAII in C++ as explained [here](https://en.wikipedia.org/wiki/Resource_acquisition_is_initialization) and [here](https://stackoverflow.com/questions/2321511/what-is-meant-by-resource-acquisition-is-initialization-raii). Its very powerful and worth seeing if we can get the code towards that.

agent/upgrade_primary_test.go
```diff
@@ -0,0 +1,194 @@
+package agent
```
dgkimura, line 1: What is the convention for test packages? I see existing pattern typically puts test in package suffixed with `_test` (e.g. `agent_test`). However, I assume you did this on purpose to access rsyncCommand in agent package? Though, it seems maybe you can use `SetRsyncCommand()` to avoid exposing that abstraction.

kalensk: The `_test` suffix is needed by the go testing framework.

> Test files that declare a package with the suffix "_test" will be compiled as a
separate package, and then linked and run with the main test binary.

https://golang.org/pkg/cmd/go/internal/test/

kalensk: Some tests are in `agent_test` and some in `agent` package for example. Ideally, everything that can be tested externally as if it was public would be in `agent_test` package for example. But, sometimes private things need to be tested or its difficult to test something in particular, so we have the test file be in `agent` package.

hub/execute.go
```diff
@@ -60,7 +60,19 @@ func (s *Server) Execute(request *idl.ExecuteRequest, stream idl.CliToHub_Execut
...
+			err = s.CopyMasterTableSpaces(streams, utils.GetTableSpaceDir())
```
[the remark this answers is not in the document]
bhuvnesh2703: But right now as well, you can change the 2nd param `utils.GetTableSpaceDir()` to anything you want when you call CopyMasterTableSpaces..for instance in the `TestCopyMasterTableSpaces` test.

agent/upgrade_primary_test.go
```diff
@@ -0,0 +1,194 @@
...
+				"--archive", "--delete",
+				"/tmp/tablespaces/1664/1/", "/tmp/other/1664/2",
+			},
+		}
+
+		var actualArgs [][]string
+		rsyncCommand = exectest.NewCommandWithVerifier(Success, func(name string, args ...string) {
+			expected := "rsync"
+			if name != expected {
+				t.Errorf("Copy() invoked %q, want %q", name, expected)
```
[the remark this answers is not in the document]
bhuvnesh2703: to Rsync() actually.

agent/upgrade_primary_test.go
```diff
@@ -0,0 +1,194 @@
...
+			t.Errorf("Remove() invoked with %q, want %q", actualRemovedLinks, expectedLstatLinks)
+		}
+
+		if !reflect.DeepEqual(actualNewSymLink, expectedLstatLinks) {
+			t.Errorf("SymLink() invoked with new name %q, want %q", actualNewSymLink, expectedLstatLinks)
+		}
+
+		if !reflect.DeepEqual(actualOldName, expectedOldName) {
+			t.Errorf("SymLink() invoked with old name %q, want %q", actualOldName, expectedOldName)
+		}
```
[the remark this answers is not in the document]
bhuvnesh2703: sure

greenplum/tablespace.go
```diff
@@ -0,0 +1,141 @@
...
+}
+
+// function to query the database to extract the tablespace information
+func GetTableSpaceTuples(connection *dbconn.DBConn) (TableSpaceTuples, error) {
+	query := ""
+	if connection.Version.Is("5") {
+		query = TableSpacesQuery
+	} else {
+		return nil, errors.New("version not supported to retrieve tablespace information")
+	}
```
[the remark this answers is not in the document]
bhuvnesh2703: simplified it the way you suggested.
I would prefer returning an error if its called on a version != 5, the caller should call only for version 5.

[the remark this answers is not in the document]
bhuvnesh2703: as of now `FillClusterConfigsSubStep` calls only for when version is 5.
```
	// major version upgrade requires upgrading tablespaces
	if dbconn.Version.Is("5") {
		if err := utils.System.MkdirAll(utils.GetTableSpaceDir(), 0700); err != nil {
			return errors.Wrapf(err, "failed to create tablespace directory %q", utils.GetTableSpaceDir())
		}
		config.TableSpacesMappingFilePath = filepath.Join(utils.GetTableSpaceDir(), greenplum.TableSpacesMappingFile)
		config.TableSpaces, err = greenplum.TableSpacesFromDB(dbconn, config.TableSpacesMappingFilePath)
		if err != nil {
			return errors.Wrap(err, "could not extract tablespace information")
		}
	}
```

greenplum/tablespace.go
bhuvnesh2703: > I would inline this function. I am not sure its getting us much.
> 
> Why not just create a slice literal? No need to append or anything.
> 
> ```go
> line := []string{
>             strconv.Itoa(tableSpace.DbId),
> 			strconv.Itoa(tableSpace.Oid),
> 			tableSpace.Name,
> 			tableSpace.Location,
> 			strconv.Itoa(tableSpace.IsUserDefined))
> }
sure
> 
> Another possibility is just keep it simple with `csv.NewWriter(file).Write(line)`
need to flush on the writer object and i will prefer the existing one.. so will keep it as is..

hub/fill_cluster_configs.go
bhuvnesh2703: its passed to the agents.. as an argument to pg_upgrade for segments.

utils/sys_utils.go
bhuvnesh2703: sure

hub/copy_master.go
bhuvnesh2703: sure made the change.

hub/server_internal_test.go
bhuvnesh2703: used 1 so that we can ensure that something is actually written and read..

testutils/test_utils.go
bhuvnesh2703: sure..

hub/server_internal_test.go
bhuvnesh2703: sure

greenplum/tablespace_test.go
bhuvnesh2703: done.

greenplum/tablespace_test.go
bhuvnesh2703: thansks, didnt run make lint earlier.. did now.

hub/copy_master.go
bhuvnesh2703: sure

utils/sys_utils.go
bhuvnesh2703: i agree, will move it to upgrade_primary.go

agent/upgrade_primary_test.go
bhuvnesh2703: nothing special in this case, used agent_test now..

hub/upgrade_primaries.go
bhuvnesh2703: We have to pass the tablespaces in the messages, so considering that in mind, the protobuf generated code will always have that field, i chose what you suggested in the 3 paragraph above `If the code`

agent/upgrade_primary.go
bhuvnesh2703: could do it either way.
Always set it in performUpgrade, then in Run.go if its not a check mode, and a segment then only pass it. But that seems like, i gave u something, and you still filtered that..
So basically my thought was performUpgrade passes it only when its doing an upgrade, and run.go just adds it if there.
In gpdb, before the PR greenplum-db/gpdb#10053, in check mode it requires the mapping file, however does nothing special IMO with that. If tablespaces exists, and its not passed pg_ugprade check will fail for segment. Thus, in that PR, skipped initializing tablespaces during --check mode. (which corresponds to the else part)
We can syncup on it.

greenplum/tablespace_test.go
bhuvnesh2703: added test for utils.System.Create, might need to refactor little bit to test tableSpaceTuples.Write

agent/upgrade_primary.go
bhuvnesh2703: Without GPDB PR greenplum-db/gpdb#10053. GPDB expects a tablespace mapping file passed when pg_upgrade --check is run on segments. The PR changes the behavior and doesn't require the mapping file (by skipping initializing tablespaces in check mode)

So, we could:
1) Copy the tablespaces file in gpupgrade initialize to the segments before running the upgrade --check. (We will have to add a new steps)
2) Or pg_upgrade doesn't need the file during --check mode which the GPDB PR does but technically for gpupgrade it translates to the else block as we don't have the file yet and we don't want to pass in pg_upgrade --check on segments. (As is in this current PR)

agent/upgrade_primary_test.go
```diff
@@ -0,0 +1,283 @@
+package agent_test
```
kalensk, line 1: Newly added files are missing the copyright.

greenplum/tablespace_test.go
```diff
@@ -0,0 +1,395 @@
+package greenplum
```
kalensk, line 1: missing copyright.

agent/upgrade_primary.go
```diff
@@ -18,6 +25,12 @@ func upgradeSegment(segment Segment, request *idl.UpgradePrimariesRequest, host
 			host, segment.Content, err)
 	}
 
+	err = RestoreTableSpaces(request, segment)
+	if err != nil {
+		return errors.Wrapf(err, "failed to restore tablespace on host %s for content id %d: %s",
```
kalensk, line 30: nit: The errors throughout this PR don't need to start with "failed"/etc., but rather simply state the action as explained in the following links:

> When adding context to returned errors, keep the context succinct by avoiding phrases like "failed to", which state the obvious and pile up as the error percolates up through the stack:

https://github.com/golang/go/wiki/CodeReviewComments#error-strings

testutils/test_utils.go
kalensk, line 169: This appears to be no longer used and can be deleted.

hub/copy_master_test.go
```diff
@@ -175,27 +148,174 @@ func TestCopyMaster(t *testing.T) {
...
+						"/tmp/primary2/tblspc1",
+						0,
+					},
+					1664: {
+						"/tmp/primary2/tblspc2",
+						1,
+					},
+				},
+			},
+			TableSpacesMappingFilePath: "/tmp/mapping.txt",
```
kalensk, line 236: linter is showing:

```
hub/copy_master_test.go:206:12: composites: `github.com/greenplum-db/gpupgrade/greenplum.TableSpaceInfo` composite literal uses unkeyed fields (govet)
					1663: {
					      ^
hub/copy_master_test.go:210:12: composites: `github.com/greenplum-db/gpupgrade/greenplum.TableSpaceInfo` composite literal uses unkeyed fields (govet)
					1664: {
					      ^
hub/copy_master_test.go:216:12: composites: `github.com/greenplum-db/gpupgrade/greenplum.TableSpaceInfo` composite literal uses unkeyed fields (govet)
					1663: {

```

greenplum/tablespace_test.go
kalensk: Could a chore been added for this?

agent/upgrade_primary.go
kalensk: Thanks for looking into this some more. As discussed we are going to go with option 2.

kalensk:
nit: This PR uses TableSpaces with camel casing, when I believe its referred to just tablespace. Maybe an additional commit that does a full rename to minimize rebase conflicts?

agent/upgrade_primary_test.go
bhuvnesh2703: added now in all the files which had it missing. Thanks.

agent/upgrade_primary.go
bhuvnesh2703: sure made the change..

testutils/test_utils.go
bhuvnesh2703: removed..

PR #330, opened 2020-05-11
Archive upgrade logs

dgkimura:
Previously the upgrade log files were lost after running revert command
because all files in state directory are deleted, including upgrade
logs. This commit adds a new substep to revert command that moves
upgrade log files from the state directory into a local directory under
gpAdminLog named "gupgrade-YYYY-MM-DDTHH:MM". The timestamp is time when
revert was issued.

It may be worth explicitly calling out that this does *not* move the
logs from the segments to the client. It simply moves the logs to a
directory by the same name on each segment. The intent of this is to
save the logs from deletion and allow a log collector to gather the
upgrade logs from the segments after an upgrade revert.

Additional details in individual commit messages.


cli/commanders/steps.go
```diff
@@ -55,6 +56,7 @@ var SubstepDescriptions = map[idl.Substep]substepText{
 	idl.Substep_DELETE_SEGMENT_STATEDIRS:                          substepText{"Deleting state directories on the segments...", "Delete state directories on the segments"},
 	idl.Substep_STOP_HUB_AND_AGENTS:                               substepText{"Stopping hub and agents...", "Stop hub and agents"},
 	idl.Substep_DELETE_MASTER_STATEDIR:                            substepText{"Deleting master state directory...", "Delete master state directory"},
+	idl.Substep_ARCHIVE_SEGMENT_LOGDIR:                            substepText{"Archiving master/segment log directory...", "Archive master/segment log directory"},
```
dgkimura, line 59: Look into single substep for both segment and master. Also need to call function explicitly for master.

agent/archive_log_directories_test.go
```diff
@@ -0,0 +1,82 @@
...
+		if renameCallCount > 0 {
+			t.Errorf("Renamed archive files, expected 0, actual %d", renameCallCount)
+		}
+	})
+
+	t.Run("a log directory with one directory archives that directory", func(t *testing.T) {
+		ctrl := gomock.NewController(t)
+		defer ctrl.Finish()
+
+		mfinfo := testutils.NewMockFileInfo(ctrl)
```
dgkimura, line 49: Look into rsync. Then may not need mock.

dgkimura:
Pivoted a bit to lower the scope by taking state directory files out of the picture.  Instead this PR now only archives (i.e. renames) the gpupgrade log directories on master/segments.  A future PR can deal with moving log files from the state directory to the gpupgrade log directory.

cli/commanders/steps.go
```diff
@@ -55,6 +55,7 @@ var SubstepDescriptions = map[idl.Substep]substepText{
 	idl.Substep_DELETE_SEGMENT_STATEDIRS:                          substepText{"Deleting state directories on the segments...", "Delete state directories on the segments"},
 	idl.Substep_STOP_HUB_AND_AGENTS:                               substepText{"Stopping hub and agents...", "Stop hub and agents"},
 	idl.Substep_DELETE_MASTER_STATEDIR:                            substepText{"Deleting master state directory...", "Delete master state directory"},
+	idl.Substep_ARCHIVE_LOGDIR:                                    substepText{"Archiving master/segment log directory...", "Archive master/segment log directory"},
```
kalensk, line 58: Since this is UI facing I think having `Archiving log directories` is clearer.

testutils/mock_agent/mock_agent_server.go
```diff
@@ -98,6 +98,11 @@ func (m *MockAgentServer) DeleteStateDirectory(context.Context, *idl.DeleteState
 	return &idl.DeleteStateDirectoryReply{}, nil
 }
 
+func (m *MockAgentServer) ArchiveLogDirectory(context.Context, *idl.ArchiveLogDirectoryRequest) (*idl.ArchiveLogDirectoryReply, error) {
+	m.increaseCalls()
+	return &idl.ArchiveLogDirectoryReply{}, nil
+}
+
```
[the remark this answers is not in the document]
dgkimura: Unfortunately, I _think_ it's still needed since we've added a new gRPC interface method.  Else, compiler complains about MockAgentServer not implementing ArchiveLogDir method from idl.AgentServer.

dgkimura:
On Tue, May 12, 2020 at 11:11 PM Kalen Krempely <notifications@github.com> wrote:
>
> Consider adding a BATs test similar to this one.
>

I'm thinking about piggybacking the check on an existing test like that one.  Would that be frowned upon?

> For a bit more realistic unit tests that actually rename real directories consider using testutils.GetTempDir() and defer testutils.MustRemoveAll().
>
Sounds like a good functional test.  Do we make any separation between unit tests and functional tests?  I couldn't quite find a division.

> Either in this PR or a followup PR we should probably verify that the log directories look like actual log directories before renaming.
>
> ...

Cool!  Those are interesting ideas.

kalensk:
> I'm thinking about piggybacking the check on an existing test like that one. Would that be frowned upon?

For this case I think it makes sense to piggyback.
 
> Sounds like a good functional test. Do we make any separation between unit tests and functional tests? I couldn't quite find a division.

Currently, there is not really much of a distinction. A lot of the unit tests create temp dirs when needed for more realistic testing, which seems to be working well.

agent/archive_log_directory.go
```diff
@@ -0,0 +1,16 @@
...
+	"github.com/greenplum-db/gpupgrade/utils"
+)
+
+func (s *Server) ArchiveLogDirectory(ctx context.Context, in *idl.ArchiveLogDirectoryRequest) (*idl.ArchiveLogDirectoryReply, error) {
+	if err := utils.System.Rename(in.GetOldDir(), in.GetNewDir()); err != nil {
+		return &idl.ArchiveLogDirectoryReply{}, err
+	}
+
+	return &idl.ArchiveLogDirectoryReply{}, nil
+}
```
kalensk, line 16: nit: I believe this could be simplified to:
```go
err := utils.System.Rename(in.GetOldDir(), in.GetNewDir());
return &idl.ArchiveLogDirectoryReply{}, err
```

nit: Most (but apparently) not all of these agent functions that receive gRPC calls have a log line at the top. Something like:
`gplog.Info("agent starting %s", idl.Substep_ARCHIVE_LOG_DIRECTORIES)`

test/revert.bats
```diff
@@ -69,4 +69,10 @@ setup() {
         echo "expected GPUPGRADE_HOME directory ${GPUPGRADE_HOME} to have been deleted"
         exit 1
     fi
+
+    # check that archive directory has been created
+    if [ ! -d "$HOME/gpAdminLogs/gpupgrade-"* ]; then
+        echo "expected directory matching $HOME/gpAdminLogs/gpupgrade-* to be created"
+        exit 1
+    fi
```
kalensk, line 77: nit: I know this bats file does't fully follow it, but the fail idiom is preferred.

`[ ! -d "$HOME/gpAdminLogs/gpupgrade-"* ] || fail  "expected directory matching $HOME/gpAdminLogs/gpupgrade-* to be created"`

dgkimura:
Committed. Thanks for review!

PR #331, opened 2020-05-13
ci: add back 6-to-6-centos-6 job

kalensk:
This test was failing when centos6 jobs were added in commit 37c3764. The new beta.4 rpm `greenplum-db-6.7.1+beta.4-rhel6-x86_64.rpm` fixes this.

[Test pipeline](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:add_666_job)


kalensk:
**_Note:_** Need to re-fly the production pipeline after merging.

PR #332, opened 2020-05-14
Add revert command help text

dgkimura:


kalensk:
I would think we would want to add this revert help text to the "global help text" (ie: `gpupgrade help`) like we have for the other substeps?

kalensk:
> Before closing out the story we should consider adding revert help text to the global help text.

Looks like we are going to add this at a later time.

dgkimura:
Thanks for the review!

PR #333, opened 2020-05-15
add agent-port flag to initialize

kalensk:
- This adds the add agent-port flag to initialize with the default port still being 6416.
  - Note that the help text will be superseded by the upcoming config file implementation.
- As I was adding the agent test I noticed some possible simplification/cleanup in the hub test which the last commit includes.

[Test pipeline](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:agent_port_flag)


integrations/agent_test.go
```diff
@@ -0,0 +1,65 @@
...
+				if err := cmd.Process.Kill(); err != nil {
+					log.Fatalf("failed to kill gpupgrade agent: %#v", err)
+				}
+			}()
+			if err != nil {
+				t.Errorf("gpupgrade agent returned unexpected error %#v", err)
+			}
+
+			// connect to the specified port
+			timeout := 3 * time.Second
```
jchampio, line 50: nit: three seconds is probably overkill for a localhost TCP connection.

integrations/hub_test.go
```diff
@@ -39,10 +33,19 @@ func TestHub(t *testing.T) {
...
 
 		cmd := exec.Command("gpupgrade", "hub")
-		errChan := make(chan error, 1)
+		err = cmd.Start()
+		if err != nil {
+			t.Errorf("unexpected error %#v", err)
+		}
+		defer func() {
+			if err := cmd.Process.Kill(); err != nil {
+				log.Fatalf("failed to kill gpupgrade hub: %#v", err)
```
jchampio, line 42: ditto

integrations/hub_test.go
```diff
@@ -4,24 +4,18 @@
...
 
 	"github.com/greenplum-db/gpupgrade/cli/commanders"
 	"github.com/greenplum-db/gpupgrade/testutils"
 )
 
-const HubPort = 7527
-
 func TestHub(t *testing.T) {
 	t.Run("the hub does not daemonizes unless --deamonize is passed", func(t *testing.T) {
-		killHub(t)
```
jchampio, line 22: Now that `killHub()` has been removed, this test is going to fail if someone's running a hub on their local machine, right? That was the purpose of the helper.

The cleanup is nice, but I don't think we can actually do it until we address the `XXX`:
```go
// XXX we should really use a PID file for this, and allow side-by-side hubs,
// rather than blowing away developer state.
```

integrations/agent_test.go
kalensk: What did you have in mind?

cli/commands/agent.go
```diff
@@ -47,7 +49,7 @@ func Agent() *cobra.Command {
 			return nil
 		},
 	}
-
+	cmd.Flags().IntVar(&port, "port", upgrade.AgentPort, "the port to listen for commands on")
```
dgkimura, line 52: Curious, what error is propagated to the user if the port specified is already in use? (e.g. What if segment 42 has something using the port, but all other segments are free?)

integrations/hub_test.go
kalensk: Honestly, I am skeptical the helper was fully working (at least the part that checks if the port is in use) since lsof does not appear to be installed on the container.

Are you suggesting to bring back the killProcess code _simply_ in case a developer is running a hub? I would push back a bit on that notion.

jchampio: > Honestly, I am skeptical the helper was fully working (at least the part that checks if the port is in use) since lsof does not appear to be installed on the container.

Can you confirm?

> Are you suggesting to bring back the killProcess code simply in case a developer is running a hub?

Yes, mostly because that's how I run the tests all the time. Anyone else?

cli/commands/agent.go
kalensk: The hub would return "bind address already in use error" when attempting to start the agent.

https://github.com/greenplum-db/gpupgrade/blob/adb636f7997b3fbf7041ad8128b76de527744a59/hub/server.go#L258-L264

dgkimura: Ah, thanks. And because it's exec-ing through ssh command the message should include the host where it's already in use.  Perfect!

integrations/agent_test.go
jchampio: Normally I'd say half a second. When we're running in Concourse, though, that can get a little dicey if the instance is heavily loaded. Maybe just a second?

testutils/test_utils.go
```diff
@@ -45,19 +45,23 @@ func CreateMultinodeSampleClusterPair(baseDir string) (*greenplum.Cluster, *gree
 	return sourceCluster, targetCluster
 }
 
-func GetOpenPort() (int, error) {
-	addr, err := net.ResolveTCPAddr("tcp", "localhost:0")
-	if err != nil {
-		return 0, err
-	}
+func MustGetPort(t *testing.T) int {
+	t.Helper()
```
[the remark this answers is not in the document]
kalensk: Can you be more clear in your desired request? I am assuming you would prefer the commit message to say "reimplement getOpenPort"?

integrations/hub_test.go
kalensk: > Can you confirm?

Just logged into the noinstall-tests and there is no lsof in PATH which is using the following image:
```
  type: docker-image
  source:
    repository: golang
    tag: '1.14'
```
```
root@eb787014-5619-4f0d-40d7-1b09f6b30a9c:/go/src/github.com/greenplum-db/gpupgrade# /bin/sh -c "lsof "
/bin/sh: 1: lsof: not found
```

The script to install the tests does not seem to add it and there is [no full path when running it or anything special](https://github.com/greenplum-db/gpupgrade/blob/adb636f7997b3fbf7041ad8128b76de527744a59/integrations/hub_test.go#L92-L97). If the error returned is not an [exitError then it is not being surfaced](https://github.com/greenplum-db/gpupgrade/blob/3e7bbf822b664383e7c731ebbff1ad1937f18676/integrations/integrations_suite_test.go).


> Yes, mostly because that's how I run the tests all the time. Anyone else?

Im curious to see what others think, and will definitely want to add it back if it because an issue on the CI or for most of the team.

testutils/test_utils.go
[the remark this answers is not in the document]
jchampio: Either that, or pull the reimplementation into a different commit to help reviewers.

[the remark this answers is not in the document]
kalensk: Reworded commit message.

upgrade/run.go
```diff
@@ -12,6 +12,9 @@ import (
 	"strconv"
 )
 
+// AgentPort is the default port that the gpupgrade agent process starts on.
+var AgentPort = 6416
```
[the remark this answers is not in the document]
kalensk: updated

integrations/agent_test.go
```diff
@@ -0,0 +1,65 @@
...
+		},
+	}
+
+	for _, c := range cases {
+		t.Run(c.name, func(t *testing.T) {
+			cmd := exec.Command("gpupgrade", "agent", "--port", strconv.Itoa(c.port))
+			err := cmd.Start()
+			defer func() {
+				if err := cmd.Process.Kill(); err != nil {
+					log.Fatalf("failed to kill gpupgrade agent: %#v", err)
```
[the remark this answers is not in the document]
kalensk: Updated to use `t.Fatalf` and use `%+v` in all places except when explicitly comparing errors.

integrations/agent_test.go
kalensk: Updated.

kalensk: Tests fail with 1 second and take a little over that to run, will update to 2 seconds.

```
--- PASS: TestAgent (2.02s)
=== RUN   TestAgent/agent_starts_on_default_port
    --- PASS: TestAgent/agent_starts_on_default_port (1.02s)
=== RUN   TestAgent/agent_starts_on_specified_port
    --- PASS: TestAgent/agent_starts_on_specified_port (1.01s)
PASS
```

integrations/hub_test.go
bhuvnesh2703: I personally am fine to kill the hub manually before running the test. Because if i am writing and running test, most likely i am not working to perform upgrade using gpupgrade simultaneously. This is not a blocker for me either way.

jchampio: Sure, let's remove it and see.

agent/server_test.go
```diff
@@ -25,7 +25,7 @@ func TestServerStart(t *testing.T) {
 		stateDir := path.Join(tempDir, ".gpupgrade")
 
 		server := agent.NewServer(agent.Config{
-			Port:     getOpenPort(t),
+			Port:     testutils.MustGetPort(t),
```
jchampio, line 28: I know we've talked about it before, but I wanted to point this out again since the usage is being expanded. This implementation is inherently flaky. Opening and closing a port doesn't reserve it, so there's no guarantee that the port number we've grabbed will still be there by the time we create the listener later.

PR #334, opened 2020-05-19
Deduplicate substeps that are shared between steps

jchampio:
[We've recently discussed several issues with the substep tracking system, including UI reporting and difficulty with changing the granularity of substeps. This particular fix is mostly cosmetic, but it's low-hanging fruit.]

We shouldn't have to come up with new copies of substeps just because multiple steps need to use the same ones. Give `status.json` a hierarchical organization; for example,
```
  {
    "execute": {
      "COPY_MASTER": "COMPLETE",
      "SHUTDOWN_SOURCE_CLUSTER": "COMPLETE",
      "START_TARGET_CLUSTER": "COMPLETE",
      ...
    },
    "initialize": {
      "BACKUP_TARGET_MASTER": "COMPLETE",
      "CHECK_UPGRADE": "COMPLETE",
      "CREATE_TARGET_CONFIG": "COMPLETE",
      ...
    },
    ...
  }
```

...and then remove the `FINALIZE_` substep prefix from all substeps. See commit messages for more details.


step/file_store_test.go
```diff
@@ -90,14 +156,24 @@ func TestFileStore(t *testing.T) {
...
-		if raw[key] != status.String() {
-			t.Errorf("status[%q] = %q, want %q", key, raw[key], status.String())
+		if raw[section][key] != status.String() {
+			t.Errorf("status[%q][%q] = %q, want %q", section, key, raw[section][key], status.String())
 		}
 	})
 }
+
+// clear writes an empty JSON map to the given FileStore backing path.
+func clear(t *testing.T, path string) {
```
dgkimura, line 172: Admittedly, I don't know details of how the tests are run.  But seems like this means we have no intention of ever running tests in parallel?

step/file_store.go
```diff
@@ -28,6 +28,8 @@ func NewFileStore(path string) *FileStore {
 	return &FileStore{path}
 }
 
+type prettyMap = map[string]map[string]PrettyStatus
```
dgkimura, line 31: Seems I am still wrapping my head around Go syntax. ;)

step/file_store_test.go
jchampio: Tests have to be [explicitly marked as parallel-safe](https://golang.org/pkg/testing/#T.Parallel); in general they run fast enough that I don't think we've really been interested in taking on the additional complexity.

step/file_store.go
jchampio: Yep 😁 nested maps are kind of nonintuitive. Should I try to make that easier to read, maybe with a type alias for the value type `map[string]PrettyStatus`?

PR #335, opened 2020-05-19
Validate logs before archiving

dgkimura:
Adds sanity check that log directory contains expected log files before
archiving. Similar to archiving data directories, the intent is to have
some form of verificiation before committing changes.


agent/archive_log_directory.go
```diff
@@ -9,12 +9,14 @@ import (
...
+	"github.com/greenplum-db/gpupgrade/upgrade"
 )
 
+var archiveLogs = upgrade.ArchiveLogs
+
 func (s *Server) ArchiveLogDirectory(ctx context.Context, in *idl.ArchiveLogDirectoryRequest) (*idl.ArchiveLogDirectoryReply, error) {
 	gplog.Info("agent starting %s", idl.Substep_ARCHIVE_LOG_DIRECTORIES)
 
-	err := utils.System.Rename(in.GetOldDir(), in.GetNewDir())
+	err := archiveLogs(in.GetOldDir(), in.GetNewDir())
```
jchampio, line 20: This pattern is pretty heavyweight for the test value we receive -- the test is just checking to make sure we pass through arguments correctly, and I'd argue that's a case for an integration or end-to-end test. The real value comes from the `upgrade.ArchiveLogs` tests you've written.

(We overuse the global function pointer mock technique in this codebase; we should try to clean up and consolidate the current uses.)

upgrade/directories_test.go
```diff
@@ -449,3 +449,54 @@ func createDataDir(t *testing.T, name, tmpDir string, requiredPaths []string) (d
 
 	return dirPath
 }
+
+func TestArchiveLogs(t *testing.T) {
+	t.Run("fails to  archive a directory that is not contain expected logs", func(t *testing.T) {
```
jchampio, line 454: I think this test should also validate that the archive didn't happen, in addition to checking returned errors.

upgrade/directories.go
```diff
@@ -155,6 +156,41 @@ func alreadyRenamed(archive, target string) bool {
...
+	var mErr multierror.Error
+	for _, f := range LogDirectoryFiles {
+		fp := filepath.Join(source, f)
+		matches, _ := filepath.Glob(fp)
+		if len(matches) <= 0 {
+			mErr = *multierror.Append(&mErr, &InvalidLogDirectoryError{source, f})
+		}
+	}
+	if mErr.ErrorOrNil() != nil {
+		return mErr.ErrorOrNil()
```
jchampio, line 169: The `multierror` API isn't very intuitive. I recommend using the following pattern instead, for now:
```go
var err error

for ... {
    err = multierror.Append(err, ...).ErrorOrNil()
}

if err != nil {
    return err
}
```

We should probably try to get rid of `multierror` altogether, and replace it with something that does exactly what we want.

dgkimura:
On Wed, May 20, 2020 at 8:56 AM Jacob Champion wrote:
>
> ... we let the hub tell the agent what directory to move, but the agent should already know what the log directory is (because it's already writing data to it). We should just move that responsibility to the agent.
>

Ah, that's an interesting point.  So you're saying `utils.GetLogDir()` is the source of truth on each agent, hence no need to perform validation.  I like it.

agent/archive_log_directory_test.go
```diff
@@ -23,12 +21,12 @@ func TestArchiveLogDirectories(t *testing.T) {
 
 	t.Run("bubbles up errors", func(t *testing.T) {
 		expected := errors.New("permission denied")
-		utils.System.Rename = func(old, new string) error {
+
+		mockArchiveLogs := func(source, target string) error {
```
dkrieger-pvtl, line 25: I think you can easily run on the actual filesystem using a temp. directory and [os.Chmod](https://golang.org/pkg/os/#Chmod).  I see no use to use function substitution in the actual implementation; the code seems much cleaner this way and you're actually testing the functionality directly.

agent/archive_log_directory_test.go
```diff
@@ -41,22 +39,21 @@ func TestArchiveLogDirectories(t *testing.T) {
 		newLogDir := "/home/gpAdmin/newlogdir"
 		calls := 0
 
-		utils.System.Rename = func(old, new string) error {
+		mockArchiveLogs := func(source, target string) error {
```
dkrieger-pvtl, line 42: same as above...you can run these on the actual filesystem in a temp dir and check the results directly.

agent/archive_log_directory_test.go
dgkimura: That's a good point.  It does make the test easier to read.  Originally I didn't do that because I figured that the on disk level testing could be handled in directories_test and I wanted to minimize external inputs as much as possible.  But I think your point is valid.

If I go with Jacob's high level suggestion, I'm thinking about removing changes to `directories.go` and these tests would almost go back to what they used to be.  In that case, what are your thoughts about function substitution on `utils.System.Rename()`?  Pushing next diff to get your thoughts.

dkrieger-pvtl: It's fine; I experimented with using os.Rename() directly here.  Since the ArchiveLogDirectory() function uses utils.GetLogDir() directly, we'd have to mock _that_ out which isn't clean either.

Since redirecting os.Rename is an idiom we use, I think that's cleaner here. 

Another option is to refactor ArchiveLogDirectory() to separate out the rename from getting the log dir, but that's just another way to prevent redirecting os.Rename().  So I still think redirecting os.Rename() is cleaner here...

```
func (s *Server) ArchiveLogDirectory(ctx context.Context, in *idl.ArchiveLogDirectoryRequest) (*idl.ArchiveLogDirectoryReply, error) {
	gplog.Info("agent starting %s", idl.Substep_ARCHIVE_LOG_DIRECTORIES)

	logdir, err := utils.GetLogDir()
	if err != nil {
		return &idl.ArchiveLogDirectoryReply{}, err
	}

	err = ArchiveLogDirectoryInternal(logdir, in.GetNewDir())
	return &idl.ArchiveLogDirectoryReply{}, err
}

func ArchiveLogDirectoryInternal(src, dst string) error {
	return utils.System.Rename(src, dst)
}

```

idl/hub_to_agent.proto
```diff
@@ -47,7 +47,6 @@ message DeleteStateDirectoryRequest {}
 message DeleteStateDirectoryReply {}
 
 message ArchiveLogDirectoryRequest {
-    string OldDir = 1;
     string NewDir = 2;
```
dkrieger-pvtl, line 50: nit: change to `string NewDir = 1;`.  We don't store old rpc requests, so there's no need to maintain binary compatibility of the protobuf message.

agent/archive_log_directory_test.go
```diff
@@ -37,19 +38,19 @@ func TestArchiveLogDirectories(t *testing.T) {
 	})
 
 	t.Run("archives log directories", func(t *testing.T) {
-		oldLogDir := "/home/gpAdmin/oldlogidr"
+		oldLogDir, _ := utils.GetLogDir()
 		newLogDir := "/home/gpAdmin/newlogdir"
 		calls := 0
 
-		utils.System.Rename = func(old, new string) error {
+		utils.System.Rename = func(source, target string) error {
```
dkrieger-pvtl, line 45: nit: above you use `old` and `new` in the signature and here `source` and `target`.  These should be consistent, and the [os.Rename](https://golang.org/pkg/os/#Rename) uses `oldPath` and `newPath`, so I prefer `old` and `new` here.

hub/archive_log_directory_test.go
```diff
@@ -17,8 +17,7 @@ import (
 	"github.com/greenplum-db/gpupgrade/idl/mock_idl"
 )
 
-const oldDir = "AnOldDirectory"
-const newDir = "AnANewDirectory"
+const newDir = "ANewDirectory"
```
dkrieger-pvtl, line 20: nit: I find `ANewDirectory` hard to read; rename to `NewDirectory`.

agent/archive_log_directory_test.go
dgkimura: I changed this after realizing `new` is a built in function in golang. I like consistency with `os.Rename` so planning to go with your `oldPath` and `newPath` suggestion.

dgkimura:
Thanks for reviews!

PR #336, opened 2020-05-22
Add jobs to upgrade a 5->6 cluster

bhuvnesh2703:
In this commit, we are not comparing the dump before and after the
upgrade of 5->6 clusters except for retail demo. It will be enabled in future commits after the upgrade issues are resolved.


ci/scripts/filter/filter.go
```diff
@@ -88,13 +88,21 @@ func write(out io.Writer, lines ...string) {
 }
 
 func Filter(in io.Reader, out io.Writer) {
-	scanner := bufio.NewScanner(in)
+	reader := bufio.NewReader(in)
```
jchampio, line 91: You can instead set the scanner's buffer size with `scanner.Buffer(nil, SomeHugeTokenSize)`, which would make this patch just a one-liner.

(What line lengths are more than 64K?)

ci/scripts/upgrade-cluster.bash
```diff
@@ -12,7 +12,7 @@ dump_sql() {
     echo "Dumping cluster contents from port ${port} to ${dumpfile}..."
 
     ssh -n mdw "
-        source ${GPHOME_NEW}/greenplum_path.sh
+        source ${GPHOME_OLD}/greenplum_path.sh
```
jchampio, line 15: It's _very_ important that we use the new `pg_dump` to dump both the old and new databases. This change will introduce more diffs later on.

What bug was being fixed here?

ci/template.yml
```diff
@@ -357,7 +357,12 @@ jobs:
         {{- end}}
         {{- if .RetailDemo}}
         DIFF_FILE: retail_demo.diff
-        {{- end}}
+        COMPARE_DIFF: 1
+        {{- else }}
+          {{- if ne (majorVersion .Source) "5" }}
+          COMPARE_DIFF: 1
```
jchampio, line 363: This introduces a change in indentation (significant in YAML).

Seems like it might be simpler to move to something like
```go
COMPARE_DIFF: {{if ne (majorVersion .Source) "5"}}1{{else}}0{{end}
```
?

bhuvnesh2703: the indentation looks good in the pipeline yml, examples below.
```
    params:
      COMPARE_DIFF: 1
```

```
    params:
      COMPARE_DIFF: 1
      DIFF_FILE: retail_demo.diff
      FILTER_DIFF: 1
```

ci/scripts/filter/filter.go
bhuvnesh2703: Instead of manually setting the values, i would prefer the methods which implicitly takes care of them. Any reason why not the NewReader?

ci/scripts/upgrade-cluster.bash
bhuvnesh2703: i thought thats how it should be. I see your point. changed it back to use the new binaries to take dump.

ci/scripts/filter/filter.go
jchampio: > Any reason why not the NewReader?

`Scanner` gives us the API we want ("read the next line"), and that has to be manually recreated in this new patch. `scanner.Buffer` is the official way to tune the token limit.

ci/template.yml
jchampio: I see four-space indents in the actual pipeline though?

ci/scripts/load-dump.sh
```diff
@@ -18,3 +18,213 @@ time ssh -n gpadmin@mdw "
...
+                                            ('pg_constraint'::regclass, 'pg_class'::regclass, 0)
+                    AND refobjid = con.oid AND deptype = 'i' AND
+                                            contype IN ('u', 'p')
+                    JOIN pg_class c ON dep.objid = c.oid AND relkind = 'i'
+                    JOIN pg_class rel on (con.conrelid = rel.oid)
+                WHERE conname <> c.relname;
+SQL_EOF
+")
+
+if [ -n "${tables_keys}" ]; then
```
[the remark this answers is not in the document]
dkrieger-pvtl: @bhuvnesh2703 Here's an example of what I mean by combining the two parts(find objects then act on them) into a single function.

```sql
drop table if exists ts;
create table ts (a varchar);
insert into ts values ('aa');
insert into ts values ('bb');
insert into ts values ('cc');

CREATE OR REPLACE FUNCTION loopdloop() RETURNS void as $$
    DECLARE
    values RECORD;
BEGIN
    FOR values IN
    SELECT a from ts
    LOOP
        EXECUTE 'CREATE TABLE ' || quote_ident(values.a) || '(a int)';
    END LOOP;
    RETURN;
END;
$$ LANGUAGE plpgsql;

SELECT loopdloop();
DROP function loopdloop();
```

ci/template.yml
bhuvnesh2703: fixed.. thanks..

ci/scripts/load-dump.sh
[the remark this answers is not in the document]
bhuvnesh2703: i see what you mean, existing setup can run 3 such drop statements in 1 sec, so the performance impact i will have to evaluate.. which i am thinking to defer now.. ok?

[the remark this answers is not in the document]
dkrieger-pvtl: Yes, this can be deferred.  To me, there are two issues: performance(speculative) and code readability.  I find it easier to see one block of SQL to do everything, vs. two blocks that grab objects and then act on them.

ci/scripts/filter/filter.go
bhuvnesh2703: used Scanner.

ci/scripts/load-dump.sh
```diff
@@ -18,3 +18,213 @@ time ssh -n gpadmin@mdw "
...
+
+echo 'Dropping unique and primary keys...'
+tables_keys=$(ssh -n mdw "
+    set -x
+
+    source /usr/local/greenplum-db-old/greenplum_path.sh
+    export MASTER_DATA_DIRECTORY=/data/gpdata/master/gpseg-1
+
+    psql -d regression --tuples-only --no-align --field-separator ' ' <<SQL_EOF
+        SELECT rel.relname table_name, conname constraint_name
```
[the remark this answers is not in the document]
bhuvnesh2703: i have added a previous step to delete on partitioned tables and then on non partitioned.

ci/scripts/filter/filter.go
```diff
@@ -89,6 +89,9 @@ func write(out io.Writer, lines ...string) {
 
 func Filter(in io.Reader, out io.Writer) {
 	scanner := bufio.NewScanner(in)
+	// there are lines in icw regression suite requiring buffer
+	// to be atleast 10000000, so keeping it a little higher for now.
+	scanner.Buffer(nil, 9800 * 4024)
```
dkrieger-pvtl, line 94: Is there something special about `9800 * 4024` as opposed to using, say, `40000000` here?

bhuvnesh2703: no nothing special, i can change it.

dkrieger-pvtl: That'd be good...I'm otherwise left wondering what's so special about those numbers....

PR #337, opened 2020-05-22
revert after execute in copy mode

kalensk:
Add support to revert a cluster after a successfully running execute in copy mode. This included stopping the target cluster and starting the source cluster.

Since we currently only support reverting after a successful initialize and execute in copy mode a test was removed which was stopping before the target cluster was created. Since the target cluster was not created this was causing references to hub.Config.Target to fail which are needed to check if the target cluster is running. Revert currently does not yet support when initialize or execute fails part way through.

[Test pipeline](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:revert_after_execute)


hub/revert.go
```diff
@@ -18,6 +23,29 @@ func (s *Server) Revert(_ *idl.RevertRequest, stream idl.CliToHub_RevertServer)
...
+
+		if err != nil {
+			gplog.Error(fmt.Sprintf("revert: %s", err))
+		}
+	}()
+
+	// Since revert needs to work at any point, and stop is not yet idempotent
+	// check if the cluster is running before stopping.
+	// TODO: This will fail if the target does not exist which can occur when
+	//  initialize fails part way through and does not create the target cluster.
```
dkrieger-pvtl, line 39: Currently, we only want to have `revert` working after `initialize` or `execute` have _completed_ successfully.   But these comments are for cases other than these...if those steps fail partway.  It's probably better just to have a comment block at the top of `Revert()` that mentions the limitations of the current code.

However, a user is allowed(and supposed) to use target cluster after a copy mode `execute` to test their new cluster.  This may involve bringing the target cluster down, so checking if the target cluster is running is still a good idea.

hub/revert.go
```diff
@@ -49,5 +77,16 @@ func (s *Server) Revert(_ *idl.RevertRequest, stream idl.CliToHub_RevertServer)
 		return DeleteStateDirectories(s.agentConns, s.Source.MasterHostname())
 	})
 
+	// Since revert needs to work at any point, and start is not yet idempotent
```
dkrieger-pvtl, line 80: This comment should be removed: see https://github.com/greenplum-db/gpupgrade/pull/337/files#r430616991.  It's probably better to not have a comment here as it's part of the core of `Revert()`: at the end of `initialize`, the source cluster is running and we need not run this.   At the end of `execute`, the source cluster isn't running so we do want to run this. And a user can start their source cluster after `execute` for testing and then decide to `revert`.  There's no need to explain this detail in a comment in the code, as it's a core part of the implementation.

test/revert.bats
```diff
@@ -11,25 +11,8 @@ setup() {
...
 
-    # the GPUPGRADE_HOME directory is deleted
-    if [ -d "${GPUPGRADE_HOME}" ]; then
-        echo "expected GPUPGRADE_HOME directory ${GPUPGRADE_HOME} to have been deleted"
-        exit 1
-    fi
+    PSQL="$GPHOME"/bin/psql
 }
 
 @test "revert cleans up state dir and data dirs" {
```
dkrieger-pvtl, line 18: nit: this test should probably be called `"after initialize, revert cleans up state dir and data dirs"`.

dkrieger-pvtl, line 18: The checks on L30-36 that check that `initialize` ran properly are not core to this test...they should be tested elsewhere if we really want to check that `initialize` works as we expect it to.

hub/revert.go
kalensk: I'll let others chime in and see if there is a general censuses.

I much prefer to have comments directly next to the relevant code unless its a very broad overarching explanation such as in `daemon.go`.

hub/revert.go
kalensk: Again, I'll let others chime in and see if there is a general censuses.

To me, the comment explains _why_ we need to check the cluster before starting or stopping, and gives an indication on improving the code by making start & stop idempotent.

test/revert.bats
kalensk: Should I create a chore to address this since that change is a bit beyond the particular scope of this current PR? Or is there another idea in mind?

Makefile
```diff
@@ -103,7 +103,7 @@ install:
 installcheck:
 		@echo "--------------------------------------------------------------"
 		@echo "# FIXME: Make, if run in parallel, hangs after test completes."
-		./installcheck.bats
+		./test/installcheck.bats
```
jchampio, line 106: This will run `installcheck.bats` during `make check` which is not desirable. We should add a quick comment to this line explaining why this doesn't live in `./test`.

test/helpers.bash
```diff
@@ -27,6 +27,7 @@ abort() {
 skip_if_no_gpdb() {
     [ -n "${GPHOME}" ] || skip "this test requires an active GPDB cluster (set GPHOME)"
     [ -n "${PGPORT}" ] || skip "this test requires an active GPDB cluster (set PGPORT)"
+    [ -n "${MASTER_DATA_DIRECTORY}" ] || skip "this test requires an active GPDB cluster (set MASTER_DATA_DIRECTORY)"
```
jchampio, line 30: If only one test relies on this being set, we should keep it in that test for now. This would couple every other test to `MASTER_DATA_DIRECTORY`.

hub/revert.go
jchampio: > But these comments are for cases other than these...if those steps fail partway.

I think there's also the case where `revert` fails partway, not just `initialize` and `execute` (which, as you mention, we're not tackling yet). We need to handle `revert` idempotence too.

dkrieger-pvtl: Yes, we realized that when working on this story.  This is another comment that can go in the comment block at the top of `Revert()`...the implementation isn't designed to handle idempotence nor intermediate states of initialize/execute...

test/revert.bats
dkrieger-pvtl: I was thinking a chore that puts those checks in the proper test...

test/revert.bats
```diff
@@ -53,8 +53,10 @@ setup() {
         exit 1
     fi
 
-    # check that archive directory has been created
-    [ -d "$HOME/gpAdminLogs/gpupgrade-"* ] || fail "expected directory matching $HOME/gpAdminLogs/gpupgrade-* to be created"
+    # check that the archived log directory was created within the last 3 minutes
```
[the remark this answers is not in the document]
dkrieger-pvtl: The glob `[ -d "$HOME/gpAdminLogs/gpupgrade-"* ]` would expand to  `[ -d /home/bengie//gpAdminLogs/gpupgrade-1234  /home/bengie/gpAdminLogs/gpupgrade-5678]` and hence be malformed bash if there were two archive dirs matching the glob.  So instead we changed it to making sure there's a "recently created" archive directory.

test/helpers.bash
jchampio: Turns out there are already other tests that implicitly rely on this, so never mind; we should be explicit.

Makefile
kalensk: Updated.

greenplum/cluster.go
```diff
@@ -277,14 +277,13 @@ func (c *Cluster) Start(stream OutStreams) error {
 }
 
 func (c *Cluster) Stop(stream OutStreams) error {
-	// TODO: why can't we call isPostmasterRunning for the !stop case?  If we do, we get this on the pipeline:
+	// TODO: why can't we call IsPostmasterRunning for the !stop case?  If we do, we get this on the pipeline:
 	// Usage: pgrep [-flvx] [-d DELIM] [-n|-o] [-P PPIDLIST] [-g PGRPLIST] [-s SIDLIST]
 	// [-u EUIDLIST] [-U UIDLIST] [-G GIDLIST] [-t TERMLIST] [PATTERN]
 	//  pgrep: pidfile not valid
 	// TODO: should we actually return an error if we try to gpstop an already stopped cluster?
```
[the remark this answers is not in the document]
kalensk: Updated.

hub/revert.go
```diff
@@ -32,6 +33,19 @@ func (s *Server) Revert(_ *idl.RevertRequest, stream idl.CliToHub_RevertServer)
...
 	}()
 
+	// Since revert needs to work at any point, and stop is not yet idempotent
+	// check if the cluster is running before stopping.
+	// TODO: This will fail if the target does not exist which can occur when
+	//  initialize fails part way through and does not create the target cluster.
+	if s.Target.IsPostmasterRunning(st.GetStream()) {
+		st.Run(idl.Substep_SHUTDOWN_TARGET_CLUSTER, func(streams step.OutStreams) error {
+			if err := s.Target.Stop(streams); err != nil {
+				return errors.Wrap(err, "stopping target cluster")
```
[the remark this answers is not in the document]
kalensk: Updated.

step/step.go
```diff
@@ -90,6 +90,10 @@ func GetStatusFile(stateDir string) (path string, err error) {
 	return path, nil
 }
 
+func (s *Step) GetStream() OutStreamsCloser {
```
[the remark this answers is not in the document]
kalensk: Updated.

greenplum/cluster.go
```diff
@@ -320,17 +318,17 @@ func runStartStopCmd(stream OutStreams, binDir, command string) error {
...
 		))
 
 	cmd.Stdout = stream.Stdout()
 	cmd.Stderr = stream.Stderr()
 
-	return cmd.Run()
+	if err := cmd.Run(); err != nil {
+		return false
+	}
+	return true
```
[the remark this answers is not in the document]
kalensk: Updated.

test/revert.bats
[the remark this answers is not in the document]
jchampio: I see, thanks for the explanation.

Someone running the `revert.bats` test back-to-back could still run into this if the tests are fast enough. I think we should consider a more solid solution; is there a way to get the exact archive directory name? (That seems potentially useful to end users as well.)

hub/revert.go
jchampio: > the implementation isn't designed to handle idempotence

But the comment is explaining (by my reading) that this implementation is trying to make steps towards idempotence.

dkrieger-pvtl: Which is why I wanted that comment removed...for now, the conditional here is in case the user stops the cluster during testing.

greenplum/cluster.go
```diff
@@ -318,17 +332,30 @@ func runStartStopCmd(stream OutStreams, binDir, command string) error {
 	return cmd.Run()
 }
 
-func (c *Cluster) IsPostmasterRunning(stream OutStreams) bool {
-	cmd := isPostmasterRunningCmd("bash", "-c",
-		fmt.Sprintf("pgrep -F %s/postmaster.pid",
-			c.MasterDataDir(),
-		))
+// IsMasterRunning returns whether the cluster's master process is running.
+func (c *Cluster) IsMasterRunning(stream OutStreams) (bool, error) {
```
jchampio, line 336: The new implementation in this function needs test coverage; it looks like the existing tests have been updated to succeed but there are no new tests for the additional functionality.

test/revert.bats
```diff
@@ -65,11 +65,8 @@ setup() {
 
     gpupgrade revert --verbose
 
-    run $PSQL postgres -c "SELECT 1;"
-    [ "$status" -eq 0 ] || fail "expected source cluster to be running on port ${PGPORT}"
-
-    run $PSQL postgres -p ${target_master_port} -c "SELECT 1;"
-    [ "$status" -ne 0 ] || fail "expected target cluster to not be running on port ${target_master_port}"
+    pg_isready -q || fail "expected source cluster to be running on port ${PGPORT}"
+    pg_isready -qp ${target_master_port} || fail "expected target cluster to not be running on port ${target_master_port}"
```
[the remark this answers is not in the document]
kalensk: updated.

greenplum/cluster.go
kalensk: added

greenplum/start_or_stop_cluster_test.go
```diff
@@ -70,40 +90,51 @@ func TestStartOrStopCluster(t *testing.T) {
...
 
-	t.Run("IsPostmasterRunning fails", func(t *testing.T) {
+	t.Run("IsMasterRunning fails", func(t *testing.T) {
 		isPostmasterRunningCmd = exectest.NewCommand(IsPostmasterRunningCmd_Errors)
 
-		if source.IsPostmasterRunning(utils.DevNull) {
+		running, err := source.IsMasterRunning(utils.DevNull)
+		var expected *exec.ExitError
+		if !xerrors.As(err, &expected) {
+			t.Errorf("returned error %#v want %#v", err, expected)
```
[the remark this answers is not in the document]
kalensk: sure.

kalensk:
rebased with master

kalensk:
rebased with master and squashed feedback.

test/revert.bats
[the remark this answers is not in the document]
jchampio: In the future I'd like to see the archive directory stamped with our upgrade ID to prove ownership. That should help this case. If others agree, let's put in a chore.

greenplum/cluster.go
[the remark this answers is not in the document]
jchampio: It looks like the TODO is still there.

hub/revert.go
dkrieger-pvtl: This is fine; we'll eventually want to have this comment there.

hub/revert.go
dkrieger-pvtl: Fine as is; we'll eventually need idempotent revert.

greenplum/cluster.go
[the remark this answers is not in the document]
kalensk: Hmm I guess I need to better learn how to apply fixups and squashes when there are conflicts and the order in which things are applied.

test/revert.bats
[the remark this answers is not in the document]
kalensk: Chore added in the ice box to have a team discussion and possibly turn into a chore. I think this current commit is better than before even if not ideal. It keeps things rolling forward.

greenplum/cluster.go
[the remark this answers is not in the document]
kalensk: Comment hopefully removed.

greenplum/cluster.go
```diff
@@ -321,17 +330,30 @@ func runStartStopCmd(stream OutStreams, binDir, command string) error {
...
-			masterDataDir,
-		))
+// IsMasterRunning returns whether the cluster's master process is running.
+func (c *Cluster) IsMasterRunning(stream OutStreams) (bool, error) {
+	path := filepath.Join(c.MasterDataDir(), "postmaster.pid")
+	if !upgrade.PathExists(path) {
+		return false, nil
+	}
+
+	cmd := isPostmasterRunningCmd("pgrep", "-F", path)
```
bhuvnesh2703, line 340: could we log this command to cli logs..

kalensk: Sure, added:
`gplog.Debug("checking if master process is running with %s", cmd.String())`. 
Let me know if you prefer a different wording.

greenplum/start_or_stop_cluster_test.go
```diff
@@ -66,42 +94,78 @@ func TestStartOrStopCluster(t *testing.T) {
...
+		}
+	})
+
+	t.Run("returns false with no error when master data directory does not exist", func(t *testing.T) {
+		source := MustCreateCluster(t, []SegConfig{
+			{ContentID: -1, DbID: 1, Port: 15432, Hostname: "localhost", DataDir: "/does/not/exist", Role: "p"},
+		})
+		running, err := source.IsMasterRunning(utils.DevNull)
+		if err != nil {
+			t.Errorf("IsMasterRunning returned error: %#v", err)
```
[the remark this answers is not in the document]
kalensk: updated

PR #340, opened 2020-05-28
migrate errors.Wrap to xerrors.Errorf

kalensk:
The xerrors backport package gives stack frames in our error values compared to the old errors.Wrap style. This also consolidates all error messages to use a single consistent idiom to avoid confusion.

Error context was also minimized with "failed to" type prefixes such that the return context is easier to read. For example avoiding

`failed to x: failed to y: failed to create new store: the error`

Compared to:
`x: y: new store: the error`

This example is from:

[Test pipeline](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:xerrors)

### DIFF
--- a/agent/upgrade_primary.go
+++ b/agent/upgrade_primary.go
@@ -9,7 +9,7 @@ import (
 	"path/filepath"
 	"strconv"
 
-	"github.com/pkg/errors"
+	"golang.org/x/xerrors"
 
 	"github.com/greenplum-db/gpupgrade/greenplum"
 	"github.com/greenplum-db/gpupgrade/idl"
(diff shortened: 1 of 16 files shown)


agent/upgrade_primary.go
```diff
@@ -123,14 +123,14 @@ func reCreateSymLink(sourceDir, symLinkName string) error {
...
+			return xerrors.Errorf("unlink %q: %w", symLinkName, err)
 		}
 	} else if !os.IsNotExist(err) {
-		return errors.Wrapf(err, "stat symbolic link %q", symLinkName)
+		return xerrors.Errorf("stat symbolic link %q: %w", symLinkName, err)
 	}
 
 	if err := utils.System.Symlink(sourceDir, symLinkName); err != nil {
-		return errors.Wrapf(err, "create symbolic link %q to directory %q", symLinkName, sourceDir)
+		return xerrors.Errorf("create symbolic link %q to directory %q: %w", symLinkName, sourceDir, err)
```
jchampio, line 133: We'll probably want to further clean these in a future pass. The errors given by `os` are self-describing; e.g.
```
symlink /tmp/hello /tmp/there: file exists
```
so context only needs to be added in the layers above.

agent/upgrade_primary_test.go
```diff
@@ -200,7 +201,7 @@ func TestReCreateSymLink(t *testing.T) {
 			t.Errorf("got nil, want %+v", os.ErrPermission)
 		}
 
-		if os.ErrPermission != errors.Cause(err) {
+		if os.ErrPermission != xerrors.Unwrap(err) {
```
jchampio, line 204: `xerrors.Is()` is the new workflow for this; we should rarely need to use `Unwrap` directly.

agent/upgrade_primary.go
kalensk: I'll leave that for future work. This is just a step forward in ensuring our error usage is using a consistent API, etc.

agent/upgrade_primary_test.go
kalensk: updated.

PR #341, opened 2020-06-01
generate SQL scripts that drop non-upgradable objects

dkrieger-pvtl:
Iniitialize runs a `pg_upgrade --check`, which can find various
database objects that cannot be upgraded.  These scripts generate
SQL scripts that can be used to drop different types of these
objects.

The scripts do not drop the objects directly, as a user might want
to inspect the objects first.

This is a hack, as ideally `pg_upgrade` would generate a script itself.  But this is expedient.

### DIFF
--- a/scripts/drop/README.md
+++ b/scripts/drop/README.md
@@ -1,0 +1,44 @@
+Copyright (c) 2017-2020 VMware, Inc. or its affiliates
+
+SPDX-License-Identifier: Apache-2.0
+
+Background
+==========
+
+Iniitialize runs a "pg_upgrade --check", which can find various
+database objects that cannot be upgraded.  
+
+NOTE: right now, these scripts are to be run against a gpdb-5 database.
+
+Introduction
+============
+
+The scripts here generate SQL scripts that drop non-upgradable objects. 
+
+The scripts do not drop the objects directly, as a user might want
+to inspect the objects first.
+
+Notes
+======
+
+These generating scripts can be run in any order BUT THE RESULTING DROP SCRIPTS
+SHOULD BE RUN IN THIS ORDER:
+
+* script_to_drop_constraints_step_1.sql
+* script_to_drop_constraints_step_2.sql
+* script_to_drop_constraints_step_3.sql
+
+BEFORE DROPPING
+===============
+
+Run the script_to_recreate_partition_indexes.sql first and save off the
+results:
+
+    psql -d <database> -t -f script_to_recreate_partition_indexes.sql
+
+Running
+=======
+
+This will only output the tuple results(not the number of rows or column headers)
+
+    psql -d <database> -t -f script_to_drop_constraints_step_1.sql
--- a/scripts/drop/script_to_drop_constraints_step_1.sql
+++ b/scripts/drop/script_to_drop_constraints_step_1.sql
@@ -1,0 +1,16 @@
+-- Copyright (c) 2017-2020 VMware, Inc. or its affiliates
+-- SPDX-License-Identifier: Apache-2.0
+
+-- Step 1 of 3
+-- generates a script to Drop foreign/unique/primary key constraints from partitioned tables.
+--    Order is significant. Remove constraints on root partition table before
+--    non-partition tables so that we can cascade deleting constraints from
+--    child partition tables.
+SELECT 'ALTER TABLE ' || nspname || '.' || relname || ' DROP CONSTRAINT ' || conname || ' CASCADE;'
+FROM pg_constraint cc
+    JOIN
+    (SELECT DISTINCT c.oid, n.nspname, c.relname
+     FROM pg_catalog.pg_partition p
+        JOIN pg_catalog.pg_class c ON (p.parrelid = c.oid)
+        JOIN pg_catalog.pg_namespace n ON (n.oid = c.relnamespace)) as sub ON sub.oid=cc.conrelid
+WHERE cc.contype IN ('f', 'u', 'p');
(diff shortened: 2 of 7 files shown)


kalensk:
- Thinking out loud here, would these be better placed in the user facing documentation for gpupgrade until we automate things? I could see having them in two places (the documentation and gpupgrade repo) causing two sources of truth and one to drift.

- To simplify things have we considered combining them into a single drop script "_for now_" separated with comments. Just offering it up for discussion.

- Would it be possible to combine the three scripts to drop constraints into a single file and simply execute the SQL in order?

- I find the `script_to_` prefix in the filenames unnecessary.

- I am not sure the some of the SQL comments are adding much over the name of the file. For example, `generates a sql script to drop gphdfs roles in the cluster` and `script_to_drop_gphdfs_roles.sql`.

- Github is showing that `scripts/drop/script_to_recreate_partition_indexes.sql` is missing a newline at the end of the file. Not sure if some editors would "complain" about that.

dkrieger-pvtl:
Overall, this PR was intended to be a "better place than a document", rather than a final answer.

> * Thinking out loud here, would these be better placed in the user facing documentation for gpupgrade until we automate things? I could see having them in two places (the documentation and gpupgrade repo) causing two sources of truth and causing drift.

I like having these in version control, but then we'd have to somehow distribute them to clients.  I'm also not sure if there's a higher standard(like having tests) if we document a script to users.  These were intended to be run by Support during support calls, for now.  At least until we resolve the issue of how to have pg_upgrade generate a drop script vs. producing a txt file of objects to drop.

> * To simplify things have we considered combining them into a single drop script "_for now_" separated with comments. Just offering it up for discussion.
> * Would it be possible to combine the three scripts to drop constraints into a single file and simply execute the SQL in order?

Good idea for the constraint dropping scripts.  

We could run them all in a single script, too, but the idea of separating running these scripts and having users scan the output before actually dropping objects might be confusing if we combine them.

> * I find the `script_to_` prefix in the filenames unnecessary.

Yeah, it's verbose, but to me `drop_...` indicates the script actually drops the objects, which it doesn't.  Users might be confused.  Do you have a suggestion for naming?

> * I am not sure the some of the SQL comments are adding much over the name of the file. For example, `generates a sql script to drop gphdfs roles in the cluster` and `script_to_drop_gphdfs_roles.sql`.

Ok.

> * Github is showing that `scripts/drop/script_to_recreate_partition_indexes.sql` is missing a newline at the end of the file. Not sure if some editors would "complain" about that.

I'll add a newline.

scripts/drop/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+* script_to_drop_constraints_step_1.sql
+* script_to_drop_constraints_step_2.sql
+* script_to_drop_constraints_step_3.sql
+
+Running
+=======
+
+This will only output the tuple results(not the number of rows or column headers)
+
```
bhuvnesh2703, line 35: could we have a wrapper script which iterates over all the databases and spits out these different SQL into a database specific file -> script_to_drop_constraints_step_1_<dbname>.sql>?

scripts/drop/script_to_drop_gphdfs_roles.sql
```diff
@@ -0,0 +1,11 @@
+-- Copyright (c) 2017-2020 VMware, Inc. or its affiliates
```
bhuvnesh2703, line 1: is there a dependency between dropping the gpdhfs tables first and then altering the role or it doesn't matter? if it matters we should put that in documentation,

scripts/drop/script_to_drop_constraints_step_1.sql
```diff
@@ -0,0 +1,16 @@
+-- Copyright (c) 2017-2020 VMware, Inc. or its affiliates
+-- SPDX-License-Identifier: Apache-2.0
+
+-- Step 1 of 3
+-- generates a script to Drop foreign/unique/primary key constraints from partitioned tables.
+--    Order is significant. Remove constraints on root partition table before
+--    non-partition tables so that we can cascade deleting constraints from
+--    child partition tables.
+SELECT 'ALTER TABLE ' || nspname || '.' || relname || ' DROP CONSTRAINT ' || conname || ' CASCADE;'
```
bhuvnesh2703, line 9: do we know if we can create the DDL to create these constraints for all those script_to_drop_constraints_step_n.sql script?

scripts/drop/README.md
kalensk: This is a good point - making it easier for them to apply these scripts across all databases. I think this would be great to have.

scripts/drop/script_to_drop_gphdfs_roles.sql
dkrieger-pvtl: I believe they are independent.  You need the permissions to write the HDFS tables but once a table has been created, I don't think there needs to be any role in the database that can actually read it.

kalensk:
> Yeah, it's verbose, but to me `drop_...` indicates the script actually drops the objects, which it doesn't. Users might be confused. Do you have a suggestion for naming?

I was thinking of using the term "generate" or "create". Something like `generate_drop_gphdfs_roles.sql". Being able to keep the names short yet descriptive is helpful since customers may have to manually type it out by hand.

scripts/drop/README.md
```diff
@@ -0,0 +1,36 @@
...
+Notes
+======
+
+These generating scripts can be run in any order BUT THE RESULTING DROP SCRIPTS
+SHOULD BE RUN IN THIS ORDER:
+
+* script_to_drop_constraints_step_1.sql
+* script_to_drop_constraints_step_2.sql
+* script_to_drop_constraints_step_3.sql
```
kalensk, line 29: If possible, combining them into a single script would remove the need for this README note.

PR #342, opened 2020-06-02
Fix dump diffs

bhuvnesh2703:
This PR contains couple commits which handles the below..
1. Replacing of single quotes in WITH clause of partition tables. This required some refactoring too.
2. Formatting View or Rule DDL stmt
3. Remove START WITH <num> line from Sequences


jchampio:
General comment: the new implementations are complex enough that I think it'd be helpful to pull them out of the monolithic "main" file and into their own source files. Then they can be plugged into the main loop with whatever API makes sense.

dgkimura:
Curious, and have a couple ignorant questions..

Is the crux of the diffs because we are comparing pg_dump version GPDB 5 output with pg_dump version GPDB 6 output?  If that's the case, then if we could use pg_dump version GPDB 6 to dump both the source and target clusters then would most of these diffs go away?

dgkimura:
> > Curious, and have a couple ignorant questions..
> > Is the crux of the diffs because we are comparing pg_dump version GPDB 5 output with pg_dump version GPDB 6 output? If that's the case, then if we could use pg_dump version GPDB 6 to dump both the source and target clusters then would most of these diffs go away?
> 
> We use GPDB6 binaries to dump both source and target clusters, https://github.com/greenplum-db/gpupgrade/blob/master/ci/scripts/upgrade-cluster.bash#L15

Ah, thanks for sharing!  I assume GPDB 6 is now set in stone so we can't fix that there?  Mostly trying to wiggle out of maintaining our own mini-parser.

ci/scripts/filters/partition_tables.go
```diff
@@ -0,0 +1,29 @@
...
+
+func FormatWithClauseIfExisting(line string) string {
+	result := partitionTableWithClauseRegex.FindAllStringSubmatch(line, -1)
+	if result == nil {
+		return line
+	}
+	groups := result[0]
+	stringWithTableReplacement := "WITH (tablename='" + groups[2] + "'"
+	// replace all occurrences of single quotes
+	stringWithoutSingleQuotes := singleQuoteRegex.ReplaceAllString(groups[3], "")
```
jchampio, line 26: Regex-less string replacement is probably better handled by the `strings` package; e.g.
```go
stringWithoutSingleQuotes := strings.ReplaceAll(groups[3], "'", "")
```
And then more complicated cases can be handled with `strings.Replacer`.

ci/scripts/filters/partition_tables.go
```diff
@@ -0,0 +1,29 @@
...
+	partitionTableWithClauseRegex = regexp.MustCompile(`(.+)WITH \(tablename='(.[^,]*?)', (.*) \)(\,*)`)
+}
+
+func FormatWithClauseIfExisting(line string) string {
+	result := partitionTableWithClauseRegex.FindAllStringSubmatch(line, -1)
+	if result == nil {
+		return line
+	}
+	groups := result[0]
+	stringWithTableReplacement := "WITH (tablename='" + groups[2] + "'"
```
jchampio, line 24: It looks like `groups[2]` isn't modified; why do we need that capture group? It doesn't look like there's a test case that modifies the table name, either.

ci/scripts/filters/view_and_rule.go
```diff
@@ -0,0 +1,82 @@
...
+	`public.d_xpect_setup d_xpect_setup_1`:        `public.d_xpect_setup`,
+	`testtable(\d+)_1\.`:                          `testtable${1}.`,
+	`qp_misc_rio.testtable(\d+) testtable(\d+)_1`: `qp_misc_rio.testtable${1}`,
+	`cte cte_1`:                                   `cte`,
+	`cte_1\.`:                                     `cte.`,
+}
+
+func init() {
+	spaceBetweenOpenBracketAndCharRegex = regexp.MustCompile(`\(\s`)
+	spaceBetweenCloseBracketRegex = regexp.MustCompile(`\)\s\)`)
```
jchampio, line 41: Do we have to handle only one space or tab, or does this need to be expanded?

(Or, if we don't have to handle tabs, can it be simplified to a `strings` replacement?)

ci/scripts/filters/view_and_rule.go
```diff
@@ -0,0 +1,82 @@
...
+	for regex, replacement := range viewReplacementPatterns {
+		viewReplacementRegex = append(viewReplacementRegex, &replacer{
+			Regex:       regexp.MustCompile(regex),
+			Replacement: replacement,
+		})
+	}
+}
+
+func StartFormattingViewOrRuleDdlStmtIfExisting(buf []string, line string) bool {
+	return len(buf) > 0 && ruleOrViewCommentRegex.MatchString(strings.Join(buf[:], " ")) && ruleOrViewCreateRegex.MatchString(line)
```
jchampio, line 54: The
```go
len(buf) > 0
```
check is interesting. We don't want to handle views/rules that don't have comments?

ci/scripts/filters/view_and_rule.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+func FormatViewOrRuleDdl(allTokens []string) string {
+	var line string
+	if allTokens[1] == "RULE" {
+		line = strings.Join(allTokens[:], " ")
```
jchampio, line 60: `allTokens` by itself should be fine; I usually only see `[:]` when converting an array to a slice. Is there another reason for it that I'm missing?

ci/scripts/filters/view_and_rule.go
```diff
@@ -0,0 +1,82 @@
...
+	} else {
+		line = strings.Join(allTokens[:4], " ") + "\n" + strings.Join(allTokens[4:], " ")
```
jchampio, line 62: The magic `[:4]` here needs some explanation.

ci/scripts/filters/view_and_rule.go
```diff
@@ -0,0 +1,82 @@
...
+		for _, r := range viewReplacementRegex {
+			line = r.replace(line)
+		}
+	}
+	line = spaceBetweenOpenBracketAndCharRegex.ReplaceAllString(line, `(`)
+	line = spaceBetweenCloseBracketRegex.ReplaceAllString(line, `))`)
+	return line
+}
+
+func BuildViewOrRuleDdl(line string, allTokens *[]string) (string, bool) {
```
jchampio, line 72: Same comment here on the API for "slice building". I think returning a new slice rather than updating an in/out pointer-to-slice makes things easier to understand on the caller's level.

jchampio:
@dgkimura 

> I assume GPDB 6 is now set in stone so we can't fix that there? Mostly trying to wiggle out of maintaining our own mini-parser.

I think we tried to investigate a fix already 😞 but I'll let Bhuvnesh confirm.

ci/scripts/filters/filter/filter.go
```diff
@@ -59,23 +52,22 @@ func init() {
...
 		blockRegexes = append(blockRegexes, regexp.MustCompile(pattern))
 	}
-	for regex, replacement := range replacementPatterns {
-		replacements = append(replacements, &replacer{
-			Regex:       regexp.MustCompile(regex),
-			Replacement: replacement,
-		})
+}
+
+func writeBufAndLine(out io.Writer, buf *[]string, line string) {
```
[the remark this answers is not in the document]
bhuvnesh2703: sure..updated

ci/scripts/filters/filter/filter.go
```diff
@@ -30,25 +30,18 @@ import (
...
 var lineRegexes []*regexp.Regexp
 var blockRegexes []*regexp.Regexp
-var replacements []*replacer
 
 func init() {
 	// linePatterns remove exactly what is matched, on a line-by-line basis.
 	linePatterns := []string{
-		"ALTER DATABASE .+ SET gp_use_legacy_hashops TO 'on';",
+		`ALTER DATABASE .+ SET gp_use_legacy_hashops TO 'on';`,
+		`START WITH \d`,
```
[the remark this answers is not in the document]
bhuvnesh2703: As of now in the dump there was only those cases where the START WITH and INITIAL VALUE are same. The way to do that would be to read the consecutive lines and if there is INITIAL VALUE with the same number as that of START WITH, we can remove that.

ci/scripts/filters/partition_tables.go
bhuvnesh2703: sure..

ci/scripts/filters/view_and_rule.go
bhuvnesh2703: nopes nothing special, changed to strings.Join(allTokens, " ")

ci/scripts/filters/view_and_rule.go
bhuvnesh2703: added comment

ci/scripts/filters/view_and_rule.go
bhuvnesh2703: there would be no views/rule which does not have the header like below.
```
--
-- Name: spi_view; Type: VIEW; Schema: public; Owner: gpadmin
--
```

ci/scripts/filters/view_and_rule.go
bhuvnesh2703: yes can be repalced..will do..

ci/scripts/filters/partition_tables.go
bhuvnesh2703: i modified it now..

ci/scripts/filters/filter/filter.go
```diff
@@ -91,14 +83,29 @@ func Filter(in io.Reader, out io.Writer) {
...
 
+		if formattingViewOrRuleDdlStmt || filters.StartFormattingViewOrRuleDdlStmtIfExisting(buf, line) {
+			formattingViewOrRuleDdlStmt = true
+			completeDdl, finishedFormatting := filters.BuildViewOrRuleDdl(line, &allTokens)
+			if finishedFormatting {
+				writeBufAndLine(out, &buf, completeDdl)
+				formattingViewOrRuleDdlStmt = false
+				allTokens = nil
+			}
+			continue nextline
```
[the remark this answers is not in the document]
bhuvnesh2703: I have catered the suggestion for Rule 3, have not done RULE 4. but let's sync up on it.

bhuvnesh2703:
Please review the commit which applies feedback..

bhuvnesh2703:
> @dgkimura
> 
> > I assume GPDB 6 is now set in stone so we can't fix that there? Mostly trying to wiggle out of maintaining our own mini-parser.
> 
> I think we tried to investigate a fix already 😞 but I'll let Bhuvnesh confirm.

Even though we call the new pg_dumpall, it creates the definition of the objects by invoking the code of the existing cluster, which makes these difference popup, and we can't really do anything there unless we make 5x and 6x code same (which we are not planning)

ci/scripts/filters/view_and_rule.go
```diff
@@ -0,0 +1,82 @@
...
+	for regex, replacement := range viewReplacementPatterns {
+		viewReplacementRegex = append(viewReplacementRegex, &viewReplacer{
+			Regex:       regexp.MustCompile(regex),
+			Replacement: replacement,
+		})
+	}
+}
+
+func IsViewOrRuleDdl(buf []string, line string) bool {
+	return len(buf) > 0 && ruleOrViewCommentRegex.MatchString(strings.Join(buf[:], " ")) && ruleOrViewCreateRegex.MatchString(line)
```
jchampio, line 51: `buf[:]` -> `buf`

ci/scripts/filters/filter/filter.go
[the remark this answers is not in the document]
jchampio: A `TODO` was added so we can keep an eye on it.

PR #343, opened 2020-06-09
initialize supports using a config file

kalensk:
This adds support to initialize to accept a configuration file. The default configuration to be shipped is found in the gpupgrade_config file.

We error out if the file flag and command line flags other than verbose are used together. Only one method or the other should be used.

Our configuration syntax and style is similar to upstream, but differs
in the following ways:
- an equal sign is required  between parameter names and values
- values are passed directly with no quoting needed.
- throws an error when a parameter is specified multiple times

[Test pipeline](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:configFile)

_Note:_ There will be follow up stories for updating global help and initialize help CLI design.


cli/commands/commands.go
```diff
@@ -587,6 +630,24 @@ func isLinkMode(input string) (bool, error) {
 	return false, fmt.Errorf("Invalid input %q. Please specify either %s.", input, strings.Join(choices, " or "))
 }
 
+func addFlags(cmd *cobra.Command, flags map[string]string) error {
+	for name, value := range flags {
+		flag := cmd.Flag(name)
+		if flag == nil {
+			return xerrors.Errorf("%q not found", name)
```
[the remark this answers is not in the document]
jchampio: `parameter %q not supported`, for me, does not add additional context compared to `%q not found`. What other options can you think of; what would be helpful to someone who receives this error? How do they fix the problem?

(I'm intentionally not saying what I would do because I'd like to avoid pushing my ideas as The Way to Do It.)

gpupgrade_config
```diff
@@ -0,0 +1,57 @@
...
+# ----------------------------
+#
+# This file consists of lines of the form:
+#
+#   name = value
+#
+# The "=" is optional. Whitespace may be used.
+# Comments are introduced with "#" anywhere on a line. 
+#
+# The commented-out settings shown in this file represent the default values.
```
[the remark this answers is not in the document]
jchampio: Well, that's a good question. To be honest my preference is that we write our own version of that comment from scratch.

We have copy-pasted a large documentation comment from Postgres, essentially. So we either need to properly give attribution and adhere to the requirements (license) Postgres has given us for using their documentation, in which we include the copyright and disclaimers, or we write our own.

cli/commands/commands.go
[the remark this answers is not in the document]
kalensk: I could see the following being more useful for customers: `The configuration parameter "unknown" was not found in the list of supported parameters: is-set, name, port, unset-flag.` (Note: the listed parameters are from testing and not the actual ones.)

Which would result in the following updated code since there does not appear to be an "easy" way to get a list of all the flag names.
```go
flag := cmd.Flag(name)
if flag == nil {
    var names []string
    cmd.Flags().VisitAll(func(flag *pflag.Flag) {
        names = append(names, flag.Name)
    })
    return xerrors.Errorf("The configuration parameter %q was not found in the list of supported parameters: %s.", name, strings.Join(names, ", "))
}

```

cli/commands/commands.go
```diff
@@ -324,7 +326,50 @@ func initialize() *cobra.Command {
...
+			// If the file flag is set check that no other flags are specified
+			// other than verbose.
+			if cmd.Flag("file").Changed {
+				var err error
+				cmd.Flags().Visit(func(flag *pflag.Flag) {
+					if flag.Name != "file" && flag.Name != "verbose" {
+						err = errors.New("The file flag cannot be used with any other flag.")
+					}
+				})
+				return err
```
bhuvnesh2703, line 346: as of now we get the below output on screen while running `gpupgrade initialize --file test --source-bindir /test` (excluding the help which is shown)
```
The file flag cannot be used with any other flag.
github.com/greenplum-db/gpupgrade/cli/commands.initialize.func1.1
	/Users/bhuvneshchaudhary/go/src/github.com/greenplum-db/gpupgrade/cli/commands/commands.go:343
github.com/spf13/pflag.(*FlagSet).Visit
	/Users/bhuvneshchaudhary/go/pkg/mod/github.com/spf13/pflag@v1.0.0/flag.go:306
github.com/greenplum-db/gpupgrade/cli/commands.initialize.func1
	/Users/bhuvneshchaudhary/go/src/github.com/greenplum-db/gpupgrade/cli/commands/commands.go:341
github.com/greenplum-db/gpupgrade/cli/commands.addHelpToCommand.func1
	/Users/bhuvneshchaudhary/go/src/github.com/greenplum-db/gpupgrade/cli/commands/commands.go:867
github.com/spf13/cobra.(*Command).execute
	/Users/bhuvneshchaudhary/go/pkg/mod/github.com/spf13/cobra@v0.0.1/command.go:687
github.com/spf13/cobra.(*Command).ExecuteC
	/Users/bhuvneshchaudhary/go/pkg/mod/github.com/spf13/cobra@v0.0.1/command.go:783
github.com/spf13/cobra.(*Command).Execute
	/Users/bhuvneshchaudhary/go/pkg/mod/github.com/spf13/cobra@v0.0.1/command.go:736
main.main
	/Users/bhuvneshchaudhary/go/src/github.com/greenplum-db/gpupgrade/cmd/gpupgrade/main.go:31
runtime.main
	/usr/local/Cellar/go/1.14.2_1/libexec/src/runtime/proc.go:203
runtime.goexit
	/usr/local/Cellar/go/1.14.2_1/libexec/src/runtime/asm_amd64.s:1373
```
can we not show the stack?

cli/commands/commands.go
```diff
@@ -418,12 +463,10 @@ After executing, you will need to finalize.`)
 		},
 	}
 
+	subInit.Flags().StringVarP(&file, "file", "f", "", "gpupgrade_config file")
```
bhuvnesh2703, line 466: do we need a better help message? this seems cryptic..

cli/commands/commands_test.go
```diff
@@ -129,3 +133,77 @@ func TestIsLinkMode(t *testing.T) {
...
+
+		// verify string flags
+		if name != flags["name"] {
+			t.Errorf("got %q want %q", name, flags["name"])
+		}
+
+		// verify int flags
+		expectedPort, err := strconv.Atoi(flags["port"])
+		if err != nil {
+			t.Errorf("Atoi returned error: %+v", err)
```
bhuvnesh2703, line 168: can we mention something which tells that port value can't be translated to integer .. something which can help take action as user?

cli/commands/commands.go
kalensk: Sure, I can updated it to `the configuration file to use`. Let me know if you prefer a different wording. Since we override the help in all sorts of ways I don't think I am even sure how to display that usage or if its possible.

cli/commands/commands_test.go
kalensk: I think Im going to leave this one as is for now since its a test and the line number will provided on failure.

cli/commands/commands.go
kalensk: I don't think so without doing some heavy lifting. In [main.go](https://github.com/greenplum-db/gpupgrade/blob/4cf9bc6ccee086288c7a77309e441540048bc001/cmd/gpupgrade/main.go#L31-L35)
```go
err = root.Execute()
if err != nil && err != daemon.ErrSuccessfullyDaemonized {
    // Use v to print the stack trace of an object errors.
    fmt.Printf("\n%+v\n", err)
    os.Exit(1)
}
```

Let me know if I am missing something.

cli/commands/commands.go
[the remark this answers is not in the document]
jchampio: Nice, that's much better. Let's try that on for size and see what Nirali and Peter think when they run acceptance.

cli/commands/parse_config_test.go
```diff
@@ -0,0 +1,168 @@
...
+			config:      `name = value"`,
+			parameter:   "name",
+			expected:    `value"`,
+		},
+		{
+			description: "parses values when missing ending quote",
+			config:      `name = "value`,
+			parameter:   "name",
+			expected:    `"value`,
+		},
```
jchampio, line 87: Thank you for adding these quoting cases -- now that we've decided not to do quoting, though, I'm not sure they're necessary anymore. If we were to add a quoting ruleset later, I think they would need to be rewritten anyway.

cli/commands/parse_config_test.go
```diff
@@ -0,0 +1,168 @@
...
+				t.Errorf("ParseConfig returned error: %+v", err)
+			}
+
+			value, ok := flags[c.parameter]
+			if c.shouldBeExcluded && ok {
+				t.Errorf("expected paramter %q to not be present", c.parameter)
+			}
+
+			if value != c.expected {
+				t.Errorf("got value %q want %q", value, c.expected)
```
jchampio, line 129: This is better, thanks.

Additional suggestion, take it or leave it: `"flags[%q] = %q, want %q", c.parameter, value, c.expected`

cli/commands/parse_config_test.go
```diff
@@ -0,0 +1,168 @@
...
+			config:      `name = "value`,
+			parameter:   "name",
+			expected:    `"value`,
+		},
+		{
+			description:      "ignores empty lines",
+			config:           `   \n  \t`,
+			parameter:        "name",
+			expected:         ``,
+			shouldBeExcluded: true,
```
jchampio, line 93: Good addition.

test/args.bats
```diff
@@ -18,3 +34,33 @@ load helpers
...
+		target-bindir = /my/new/bin/dir
+		source-master-port = ${PGPORT}
+		disk-free-ratio = 0
+		stop-before-cluster-creation = true
+	EOF
+
+    gpupgrade initialize --verbose --file "$config_file"
+
+    run gpupgrade config show --target-bindir
+    echo $output
```
jchampio, line 59: Left over from debugging, maybe?

cli/commands/parse_config.go
```diff
@@ -0,0 +1,81 @@
...
+	if err := scanner.Err(); err != nil {
+		return nil, xerrors.Errorf("scanning config: %w", err)
+	}
+
+	return flags, nil
+}
+
+// parseLine allows one parameter per line with an optional equal sign between
+// name and value. Surrounding whitespace is insignificant and blank lines are
+// ignored. Hash marks (#) designate the remainder of the line as a comment.
```
jchampio, line 50: The doc comment here should be updated; blank lines and comments are no longer handled by `parseLine()`.

cli/commands/commands.go
bhuvnesh2703: ok, we can do it in a separate story..

cli/commands/commands_test.go
bhuvnesh2703: sure.. thats fine.

cli/commands/parse_config.go
```diff
@@ -0,0 +1,81 @@
...
+func parseLine(line string) (parameter, error) {
+	parts := strings.SplitN(line, "=", 2)
+	if !strings.Contains(line, "=") {
```
jchampio, line 53: nit/optimization: if `len(parts) == 1`, then `line` did not contain `=`.

cli/commands/commands.go
bhuvnesh2703: thats right too, for this story, we can skip it as you already plan to do followup stories on help etc..

cli/commands/parse_config.go
```diff
@@ -0,0 +1,81 @@
...
+		parts = strings.Fields(line)
+	}
+
+	name := strings.TrimSpace(parts[0])
+	if len(parts) != 2 {
+		return parameter{}, xerrors.Errorf("no value found for parameter %q", name)
+	}
```
jchampio, line 60: The error message seems to match the case where `len(parts) < 2`. If `len(parts) > 2` then we have a configuration line like
```
name value with spaces
```
and I think the current error would be confusing. (I think we should still error out in the above situation, though, and add a test case for it.)

cli/commands/parse_config.go
```diff
@@ -0,0 +1,81 @@
...
+
+	// Replace config file parameters with underscores to dashes such that
+	// the correct flag can be found.
+	name = strings.ReplaceAll(name, "_", "-")
+
+	value := strings.TrimSpace(parts[1])
+	if !hasEnclosingQuotes(value) {
```
jchampio, line 67: IIUC, the current approach leads to
```
name = "value with # hash"
```
being parsed as having a value that _includes_ the double quotes, which is probably not what a user would be going for.

I think the `hasEnclosingQuotes` logic should be removed entirely. Without a quoting ruleset we cannot support the `#` character in configuration values, and I think that's okay (if suboptimal) for beta.

cli/commands/commands.go
```diff
@@ -324,7 +326,50 @@ func initialize() *cobra.Command {
...
+				}
+				defer func() {
+					if cErr := configFile.Close(); cErr != nil {
+						err = multierror.Append(err, cErr).ErrorOrNil()
+					}
+				}()
+
+				flags, err := ParseConfig(configFile)
+				if err != nil {
+					return err
```
jchampio, line 365: Wrapping the error in context would be helpful here; to show that any errors are due to the configuration file that was parsed. E.g. `in file gpupgrade_config: parameter "source-bindir" declared more than once`

A future PR could then add a line number, perhaps.

cli/commands/parse_config.go
kalensk: updated.

cli/commands/parse_config.go
kalensk: This has changed now that we no longer allow parameters without equal signs.

cli/commands/parse_config.go
kalensk: This has changed now that we no longer allow parameters without equal signs.

cli/commands/commands.go
kalensk: done.

test/args.bats
kalensk: fixed.

cli/commands/parse_config.go
kalensk: This has changed now that we no longer allow parameters without equal signs.

cli/commands/parse_config.go
```diff
@@ -0,0 +1,69 @@
...
+	return flags, nil
+}
+
+// parseLine allows one parameter per line with a required equal sign between
+// the name and value. Comments begin with an "#" and can begin anywhere on the
+// line.
+func parseLine(line string) (parameter, error) {
+	parts := strings.SplitN(line, "=", 2)
+	if len(parts) != 2 {
+		return parameter{}, xerrors.Errorf("parameter %q is not of the form name = value", line)
```
bhuvnesh2703, line 57: instead of parameter, shouldn't it be `entry` %q is not of the form 'name = value'

kalensk: Thanks for asking. Parameter is the term used by [upstream](https://www.postgresql.org/docs/current/config-setting.html) which we should also use  throughout this PR and docs to be consistent and avoid confusion.

PR #344, opened 2020-06-09
test/helpers: log() to stdout rather than TAP

jchampio:
Logging failure modes to the TAP stream corrupts the BATS "pretty" format output, which makes it difficult to tell which test has failed during e.g. Concourse runs. This was a nice idea, but it'll take much more work to make it actually useful in practice.

The tradeoff is that the failure message is now buried in the stdout with everything else. Perhaps we can find other ways to make this better.

### Current problem

TAP mode:
```bash
$ ./test/checks.bats -t
1..1
# all is lost
not ok 1 initialize prints disk space on failure
# (from function `fail' in file test/helpers.bash, line 16,
#  in test file test/checks.bats, line 81)
#   `fail "all is lost"' failed
```
Pretty mode. Note the loss of the test name as our TAP output overwrites it, and the problem gets worse for multiline failure messages:
```bash
$ ./test/checks.bats
    all is lostrints disk space on failure                                                                        1/1
 ✗ initialize prints disk space on failure
   (from function `fail' in file test/helpers.bash, line 16,
    in test file test/checks.bats, line 81)
     `fail "all is lost"' failed
```

### After the fix
```bash
$ ./test/checks.bats
 ✗ initialize prints disk space on failure
   (from function `fail' in file test/helpers.bash, line 22,
    in test file test/checks.bats, line 81)
     `fail "all is lost"' failed
   all is lost
```


jchampio:
Thanks!

PR #345, opened 2020-06-10
kill-services: try hub default port if configFile cannot be accessed

dkrieger-pvtl:
** the first commit and the fixup_fixup contain different implementations.  I intend you to review the last one **

kill-services is used internally for testing only.  Typically, a bats
test will call kill-services in both the setup() and teardown().
However, in setup() where kill-services is called, the configFile for
the current run is not set.  But if the hub is already running(say,
from a previous failed test or the developer has been running it),
kill-services will fail.  It should pass.

This fixes this case.  It is typically not currently seen because
kill-services exits quickly if the hub is not running.

Several possible fixes were considered:
- adding a flag `--default-port` to kill-services. 
    - this was dismissed as kill-services is supposed to kill all running hubs
    - in the future, we might need to support multiple upgrades per host, but that's for later
- doing nothing(force the developer to manually kill the hub)
    - this is extra work that frustrated me(which is why I filed this PR)

### DIFF
--- a/cli/commands/commands.go
+++ b/cli/commands/commands.go
@@ -163,25 +163,44 @@ func connTimeout() time.Duration {
 	return time.Duration(duration * float64(time.Second))
 }
 
-// connectToHub() performs a blocking connection to the hub, and returns a
-// CliToHubClient which wraps the resulting gRPC channel. Any errors result in
-// an os.Exit(1). This reads the hub's persisted configuration for the current
-// port. NOTE: This overloads the hub's persisted configuration with that of the
+// This reads the hub's persisted configuration for the current
+// port.  If tryDefault is true and the configuration file does not exist,
+// it will use the default port.  This might be the case if the hub is
+// still running, even though the state directory, which contains the
+// hub's persistent configuration, has been deleted.
+// Any errors result in an os.Exit(1).
+// NOTE: This overloads the hub's persisted configuration with that of the
 // CLI when ideally these would be separate.
-func connectToHub() idl.CliToHubClient {
+func getHubPort(tryDefault bool) int {
 	conf := &hub.Config{}
 	err := hub.LoadConfig(conf, upgrade.GetConfigFile())
-	if err != nil {
+
+	var pathError *os.PathError
+	if tryDefault && xerrors.As(err, &pathError) {
+		conf.Port = upgrade.DefaultHubPort
+	} else if err != nil {
 		gplog.Error("failed to retrieve hub port due to %v", err)
 		os.Exit(1)
 	}
 
+	return conf.Port
+}
+
+// calls connectToHubOnPort() using the port defined in the configuration file
+func connectToHub() idl.CliToHubClient {
+	return connectToHubOnPort(getHubPort(false))
+}
+
+// connectToHubOnPort() performs a blocking connection to the hub based on the
+// passed in port, and returns a CliToHubClient which wraps the resulting gRPC channel.
+// Any errors result in a call to os.Exit(1).
+func connectToHubOnPort(port int) idl.CliToHubClient {
 	// Set up our timeout.
 	ctx, cancel := context.WithTimeout(context.Background(), connTimeout())
 	defer cancel()
 
 	// Attempt a connection.
-	address := "localhost:" + strconv.Itoa(conf.Port)
+	address := "localhost:" + strconv.Itoa(port)
 	conn, err := grpc.DialContext(ctx, address, grpc.WithInsecure(), grpc.WithBlock())
 	if err != nil {
 		// Print a nicer error message if we can't connect to the hub.
@@ -497,7 +516,7 @@ func revert() *cobra.Command {
 			}
 
 			s := commanders.Substep(idl.Substep_STOP_HUB_AND_AGENTS)
-			err = stopHubAndAgents()
+			err = stopHubAndAgents(false)
 			s.Finish(&err)
 
 			if err != nil {
@@ -640,12 +659,12 @@ var killServices = &cobra.Command{
 			return nil
 		}
 
-		return stopHubAndAgents()
+		return stopHubAndAgents(true)
 	},
 }
 
-func stopHubAndAgents() error {
-	_, err := connectToHub().StopServices(context.Background(), &idl.StopServicesRequest{})
+func stopHubAndAgents(tryDefaultPort bool) error {
+	_, err := connectToHubOnPort(getHubPort(tryDefaultPort)).StopServices(context.Background(), &idl.StopServicesRequest{})
 	if err != nil {
 		errCode := grpcStatus.Code(err)
 		errMsg := grpcStatus.Convert(err).Message()
--- a/cli/commands/commands_test.go
+++ b/cli/commands/commands_test.go
@@ -6,6 +6,10 @@ package commands
 import (
 	"reflect"
 	"testing"
+
+	"github.com/greenplum-db/gpupgrade/hub"
+	"github.com/greenplum-db/gpupgrade/testutils"
+	"github.com/greenplum-db/gpupgrade/upgrade"
 )
 
 func TestParsePorts(t *testing.T) {
@@ -128,4 +132,46 @@ func TestIsLinkMode(t *testing.T) {
 			}
 		})
 	}
+}
+
+func TestGetHubPort(t *testing.T) {
+	t.Run("correctly pulls the port from the stored config", func(t *testing.T) {
+		stateDir := testutils.GetTempDir(t, "")
+		defer testutils.MustRemoveAll(t, stateDir)
+
+		// set GPUPGRADE_HOME to the stateDir to provide a home for the config file
+		resetEnv := testutils.SetEnv(t, "GPUPGRADE_HOME", stateDir)
+		defer resetEnv()
+
+		// save the expected port value to the conf file
+		expected := 12345
+		server := hub.New(&hub.Config{Port: expected}, nil, stateDir)
+		err := server.SaveConfig()
+		if err != nil {
+			t.Errorf("got unexpected error %#v", err)
+		}
+
+		// looks up port from config file
+		port := getHubPort(false)
+		if port != expected {
+			t.Errorf("got %d expected %d", port, expected)
+		}
+
+		// still looks up port from config file whn default port is allowed
+		port = getHubPort(true)
+		if port != expected {
+			t.Errorf("got %d expected %d", port, expected)
+		}
+
+	})
+
+	t.Run("uses default port if the config file does not exist", func(t *testing.T) {
+		expected := upgrade.DefaultHubPort
+
+		port := getHubPort(true)
+		if port != expected {
+			t.Errorf("got %d expected %d", port, expected)
+		}
+	})
+
 }
--- a/test/services.bats
+++ b/test/services.bats
@@ -26,6 +26,10 @@ teardown() {
 
     gpupgrade kill-services
     rm -r "$STATE_DIR"
+
+    if [ -n "${TMP_DIR}" ]; then
+        rm -r "${TMP_DIR}"
+    fi
 }
 
 @test "kill-services actually stops hub and agents" {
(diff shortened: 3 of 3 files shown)


dkrieger-pvtl:
> **1)** I think the code would be cleaner with the following implementation or so:

I agree the nested ifs is ugly, but there is a functional reason.  The code you suggest unconditionally connects to the hub.  `connectToHub()` is also called by `revert`.  For `revert`, we probably do want to error out if we cannot look up the hub port from the config file, as that represents some kind of broken state on our part.  But, for `kill-services`, we just want to kill the hub.

So it's really a question of how we want to design the interface; then we can clean the code.

Thoughts?

dkrieger-pvtl:
> **2)** I would much prefer unit tests for `connectToHub` over an additional BATs test.

Ideally I would too, but looking at what's involved(and that we don't currently have tests other than BATs), I decided to stick with the current usage.

Thoughts?

kalensk:
If you really need the `retryOnDefaultPort` bool doesn't something like the following work while eliminating the need for three nested ifs and repeating gplog?

```go
func stopHubAndAgents(retryOnDefaultPort bool) error {
	conf := &hub.Config{}
	err := hub.LoadConfig(conf, upgrade.GetConfigFile())
	var pathErr *os.PathError
	if retryOnDefaultPort && xerrors.As(err, &pathErr) {
		// If config.json does not exist try the default port
		conf.Port = upgrade.DefaultHubPort
	}

	if err != nil && !xerrors.As(err, &pathErr) {
		gplog.Error("failed to retrieve hub port due to %v", err)
		os.Exit(1)
	}
```

I am not sure this is any better, but another approach might be to have `conenctToHub()` and a `connectToHubOnAddress(address string)`. The `stopHubAndAgents` would have the logic on if failing to open the file is allowed and creating the correct address string with port. The downside is this repeats the `LoadConfig` logic, but that may be ok since its fairly straightforward?

dkrieger-pvtl:
> If you really need the `retryOnDefaultPort` bool doesn't something like the following work while eliminating the need for three nested ifs and repeating gplog?

The trick is we have three cases to consider for err != nil:
   tryDefaultPort && pathError: use default port
   tryDefaultPort && !pathError : error out
   !tryDefaultPort : error out

Perhaps this is cleaner as you suggest?  This uses the fact that xerrors.As(err, &pathErr) will return `false` if err is nil.

```go
    func stopHubAndAgents(retryOnDefaultPort bool) error {
	conf := &hub.Config{}
	err := hub.LoadConfig(conf, upgrade.GetConfigFile())

	var pathErr *os.PathError
        useDefaultPort = retryOnDefaultPort && xerrors.As(err, &pathErr)  
	if useDefaultPort {
		// If config.json does not exist try the default port
		conf.Port = upgrade.DefaultHubPort
	} else if err != nil {
		gplog.Error("failed to retrieve hub port due to %v", err)
		os.Exit(1)
	}
```

or this:


```go
    func stopHubAndAgents(retryOnDefaultPort bool) error {
	conf := &hub.Config{}
	if err := hub.LoadConfig(conf, upgrade.GetConfigFile()); err != nil {

	    var pathErr *os.PathError
            useDefaultPort = retryOnDefaultPort && xerrors.As(err, &pathErr)  
	    if useDefaultPort {
		   // If config.json does not exist try the default port
	   	  conf.Port = upgrade.DefaultHubPort
	    } else {
		gplog.Error("failed to retrieve hub port due to %v", err)
		os.Exit(1)
	}
```

> Another approach might be to have `conenctToHub()` and a `connectToHubOnAddress(address string)`. The `stopHubAndAgents` would have the logic on if failing to open the file is allowed and creating the correct address string with port. The downside is this repeats the `LoadConfig` logic, but that may be ok since its fairly straightforward?

I'd rather keep the logic in `connectToHub` as that's really where the "business logic" should be IMHO.

dkrieger-pvtl:
@kalensk I pushed a fixup that separates out obtaining the hub port from the connection attempt.  It adds more changes, as the decision to exit if the config file cannot be found is now done in each command.  But IMHO "hidden exits" are fairly gross.  I think the "business logic" should be as high up as possible.

kalensk:
There are definitely different ways of solving this and Im open to the ones that lead to easier to follow code. My main concern was the readability and maintainability of the initial approach with multi nested if's and branching, etc. Whatever approach taken should try try and meet these goals is all.

dkrieger-pvtl:
> There are definitely different ways of solving this and Im open to the ones that lead to easier to follow code. My main concern was the readability and maintainability of the initial approach with multi nested if's and branching, etc. Whatever approach taken should try try and meet these goals is all.

Yes, there are many ways to do this... @kalensk  please take review the latest...it's basically the original method but with the port lookup in a separate function.  The exit() if the hub lookup fails is still in that method since the overall code is cleaner.  I'm deferring testing for the exit() condition, but I think that's fine for now.

kalensk:
To mitigate from the majority of callers having to call `getHubPort(false)` what about two connectToHub methods?


```go
func connectToHub() idl.CliToHubClient {
	getHubPort(false)
}

func connectToHubOnPort(port int) idl.CliToHubClient {
   // connect code using port parameter
}

func stopHubAndAgents(tryDefaultPort bool) error {
	_, err := connectToHubOnPort(getHubPort(true)).StopServices...
}
```

cli/commands/commands_test.go
```diff
@@ -129,3 +134,46 @@ func TestIsLinkMode(t *testing.T) {
...
+			defer func() {
+				if err := os.Setenv("GPUPGRADE_HOME", old); err != nil {
+					t.Errorf("got unexpected error %#v", err)
+				}
+			}()
+		}
+		err := os.Setenv("GPUPGRADE_HOME", stateDir)
+		if err != nil {
+			t.Errorf("got unexpected error %#v", err)
+		}
```
kalensk, line 154: This setup could maybe using the following helpers?

```go
			dir := testutils.GetTempDir(t, "")
			defer testutils.MustRemoveAll(t, dir)

			resetEnv := testutils.SetEnv(t, "GPUPGRADE_HOME", filepath.Join(dir, ".gpupgrade"))
			defer resetEnv()

			err := commanders.CreateStateDir()
			if err != nil {
				t.Errorf("CreateStateDir returned error: %+v", err)
			}
```

cli/commands/commands_test.go
```diff
@@ -129,3 +134,46 @@ func TestIsLinkMode(t *testing.T) {
...
+	})
+
+	t.Run("uses default port if the config file does not exist", func(t *testing.T) {
+		expected := upgrade.DefaultHubPort
+
+		port := getHubPort(true)
+		if port != expected {
+			t.Errorf("got %d expected %d", port, expected)
+		}
```
kalensk, line 176: I really like the addition of the unit test! Thanks :)

nit: Consider combining this test into the the other such that `getHubPort(true)` can be exercised when the config has a non-default port populated?

test/services.bats
```diff
@@ -41,6 +41,28 @@ teardown() {
...
+    # stop them
+    gpupgrade kill-services
+
+    # make sure that they are down
+    ! process_is_running "[g]pupgrade hub"
+    ! process_is_running "[g]pupgrade agent"
+
+    # move the gpupgrade back so that teardown() will work
+    mv "${TMP_DIR}/gpupgrade" "${STATE_DIR}"
+
```
kalensk, line 63: Do we have to remove `$TEMP_DIR` at the end of the test?

dkrieger-pvtl:
> To mitigate from the majority of callers having to call `getHubPort(false)` what about two connectToHub methods?

Done.  The code is cleaner now.  Thanks.

cli/commands/commands_test.go
dkrieger-pvtl: I added an additional line to the first test to make sure the config file port is used whether or not the default port is used as a fallback.  The second test is different from the first...it tests that the default port is used when the config file does not exist.

cli/commands/commands_test.go
dkrieger-pvtl: Thanks...I didn't notice those other testutils functions.  The `commanders.CreateStateDir()` isn't needed, though, as you still need to  create a tmp dir name.  Instead, I just create the tempDir ahead of time.

I also corrected testutils.SetEnv() to only restore the original env value if it was set in the first place.

test/services.bats
dkrieger-pvtl: Done.

testutils/test_utils.go
```diff
@@ -135,17 +135,19 @@ func VerifyRename(t *testing.T, source, target string) {
...
 
 	return func() {
-		err := os.Setenv(envar, old)
-		if err != nil {
-			t.Fatalf("setting %s environment variable to %s", envar, old)
+		if reset {
+			err := os.Setenv(envar, old)
+			if err != nil {
+				t.Fatalf("setting %s environment variable to %s", envar, old)
+			}
```
kalensk, line 150: Can you explain a bit what was occurring to drive out this change? What issue were you running into? I am assuming if `GPUPGRADE_HOME` was unset it was setting it to empty string which was causing subsequent runs of the standalone binary to cause issues? But Im curious to see what was actually encountered.

If this change is really need, I would consider the following for clarity:
```go
return func() {
	if !reset {
		return
	}

	err := os.Setenv(envar, old)
	if err != nil {
		t.Fatalf("setting %s environment variable to %s", envar, old)
	}
}
```

cli/commands/commands.go
```diff
@@ -163,25 +163,39 @@ func connTimeout() time.Duration {
...
+		conf.Port = upgrade.DefaultHubPort
+	} else if err != nil {
 		gplog.Error("failed to retrieve hub port due to %v", err)
 		os.Exit(1)
 	}
 
+	return conf.Port
+}
+
+// connectToHubOnPort() performs a blocking connection to the hub, and returns a
```
kalensk, line 186: nit: update to connectToHub

cli/commands/commands.go
```diff
@@ -163,25 +163,39 @@ func connTimeout() time.Duration {
 	return time.Duration(duration * float64(time.Second))
 }
 
-// connectToHub() performs a blocking connection to the hub, and returns a
-// CliToHubClient which wraps the resulting gRPC channel. Any errors result in
-// an os.Exit(1). This reads the hub's persisted configuration for the current
-// port. NOTE: This overloads the hub's persisted configuration with that of the
+// This reads the hub's persisted configuration for the current
+// port.  If tryDefault is true, it will fall back to the default port.
+// Any errors result in an os.Exit(1).
```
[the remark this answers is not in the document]
dkrieger-pvtl: Done.

testutils/test_utils.go
dkrieger-pvtl: See https://gist.github.com/dkrieger-pvtl/97ee7158970283213b5cf9debfc6a219.  Yes, it's the difference between an unset environment variable and one that is set to "".  There's also a bug I found in my change...I need to call `os.Unsetenv()` if the original env value isn't there.

I didn't hit this bug, but found it by inspection....the gist in this comment shows how it could be hit.

kalensk: Since there was no direct issue or impact and merely code inspection, would it be ok to save this change for a different PR?

dkrieger-pvtl: Sure.

kalensk: FWIW, with the original code (so far...) I am having trouble producing a situation where that would be an issue. I'll keep poking a bit more.

dkrieger-pvtl: I remove this commit from this PR...now in https://github.com/greenplum-db/gpupgrade/pull/350

cli/commands/commands.go
dkrieger-pvtl: corrected.

cli/commands/commands.go
```diff
@@ -163,25 +163,41 @@ func connTimeout() time.Duration {
 	return time.Duration(duration * float64(time.Second))
 }
 
-// connectToHub() performs a blocking connection to the hub, and returns a
-// CliToHubClient which wraps the resulting gRPC channel. Any errors result in
-// an os.Exit(1). This reads the hub's persisted configuration for the current
-// port. NOTE: This overloads the hub's persisted configuration with that of the
+// This reads the hub's persisted configuration for the current
+// port.  If tryDefault is true and the configuration file does not exist,
+// it will use the default port. Any errors result in an os.Exit(1).
```
kalensk, line 168: nit: Consider adding a sentence explaining why or when `getHubPort` would need to return when the config file does not exist. That is something like: `This can happen when the state directory has been removed before the gupgrade processes have been shut down.`

dkrieger-pvtl: Done.

PR #346, opened 2020-06-11
remove greenplum.OutStreams in favor of steps.OutStreams

kalensk:
There were two identical OutStream types which caused some confusion and
issues. For example, using a function pointer to pass a step.OutStreams
into a function that accepted a greenplum.OutStreams.

[Test pipeline](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:outstreams)

### DIFF
--- a/greenplum/cluster.go
+++ b/greenplum/cluster.go
@@ -13,6 +13,7 @@ import (
 	"github.com/pkg/errors"
 	"golang.org/x/xerrors"
 
+	"github.com/greenplum-db/gpupgrade/step"
 	"github.com/greenplum-db/gpupgrade/upgrade"
 )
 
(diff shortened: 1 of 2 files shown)


greenplum/cluster.go
```diff
@@ -276,11 +277,11 @@ func (c *Cluster) GetDirForContent(contentID int) string {
 	return c.Primaries[contentID].DataDir
 }
 
-func (c *Cluster) Start(stream OutStreams) error {
+func (c *Cluster) Start(stream step.OutStreams) error {
```
bhuvnesh2703, line 280: is there any impact to the way we put the error on screen with this change?

kalensk: Should be a seamless change that is only internal since the two interfaces are identical: [stream.go](https://github.com/greenplum-db/gpupgrade/blob/4cf9bc6ccee086288c7a77309e441540048bc001/step/stream.go#L15-L18) and [runner.go](https://github.com/greenplum-db/gpupgrade/blob/4cf9bc6ccee086288c7a77309e441540048bc001/greenplum/runner.go#L20-L23)

PR #347, opened 2020-06-11
ci: rename old -> source and new -> target

kalensk:
For consistency and clarity rename all references in the ci directory from old to source, and new to target.

There are a lot of nuances and details here, so let me know if I missed something!

[Test Pipeline](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:rename)


ci/scripts/upgrade-cluster.bash
```diff
@@ -100,13 +100,13 @@ time ssh mdw bash <<EOF
 EOF
 
 # TODO: how do we know the cluster upgraded?  5 to 6 is a version check; 6 to 6 ?????
-#   currently, it's sleight of hand...old is on port $MASTER_PORT then new is!!!!
-#   perhaps use the controldata("pg_controldata $MASTER_DATA_DIR") system identifier?
+#   currently, it's sleight of hand...source is on port $MASTER_PORT then target is!!!!
+#   perhaps use the contrsourceata("pg_contrsourceata $MASTER_DATA_DIR") system identifier?
```
dkrieger-pvtl, line 104: This line should not be changed....the substring `old` in `controldata` aka `contr-old-data` should not be changed to `source`.

kalensk: Thanks! I fixed that last controldata line.

PR #348, opened 2020-06-11
Cleanup Sysutils - Remove GetUser and GetHost and Move TryEnv

kalensk:
`sys_utils.go` and `db.go` need some work, but this PR is hopefully a small step in the right direction.

This PR has two commits:
1) Removes `GetUser` and `GetHost` as they were not giving us much over the direct system calls.
2) Move `TryEnv` and its tests to db.go since it is only used in one place.

[Test Pipeline](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:sysutils)

### DIFF
--- a/db/db.go
+++ b/db/db.go
@@ -4,20 +4,32 @@
 package db
 
 import (
-	"github.com/greenplum-db/gpupgrade/utils"
+	"os/user"
 
 	"github.com/greenplum-db/gp-common-go-libs/dbconn"
+	"github.com/greenplum-db/gp-common-go-libs/gplog"
 	_ "github.com/lib/pq" //_ import for the side effect of having postgres driver available
+
+	"github.com/greenplum-db/gpupgrade/utils"
 )
 
 // TODO: do we need this code anymore?
 func NewDBConn(masterHost string, masterPort int, dbname string) *dbconn.DBConn {
-	currentUser, _, _ := utils.GetUser()
-	username := utils.TryEnv("PGUSER", currentUser)
+	currentUser, err := utils.System.CurrentUser()
+	if err != nil {
+		gplog.Error("Failed to look up current user: %s", err)
+		currentUser = &user.User{}
+	}
+	username := utils.TryEnv("PGUSER", currentUser.Username)
+
 	if dbname == "" {
 		dbname = utils.TryEnv("PGDATABASE", "")
 	}
-	hostname, _ := utils.GetHost()
+
+	hostname, err := utils.System.Hostname()
+	if err != nil {
+		gplog.Error("Failed to look up hostname: %s", err)
+	}
 	if masterHost == "" {
 		masterHost = utils.TryEnv("PGHOST", hostname)
 	}
--- a/db/db.go
+++ b/db/db.go
@@ -4,6 +4,7 @@
 package db
 
 import (
+	"os"
 	"os/user"
 
 	"github.com/greenplum-db/gp-common-go-libs/dbconn"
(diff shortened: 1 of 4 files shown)


db/db.go
```diff
@@ -4,22 +4,31 @@
...
 
 // TODO: do we need this code anymore?
 func NewDBConn(masterHost string, masterPort int, dbname string) *dbconn.DBConn {
-	currentUser, _, _ := utils.GetUser()
-	username := utils.TryEnv("PGUSER", currentUser)
+	user, err := utils.System.CurrentUser()
+	if err != nil {
+		gplog.Error("Failed to look up current user: %s", err)
+	}
+	username := tryEnv("PGUSER", user.Username)
```
dkrieger-pvtl, line 20: This code is not the same as the original...if `CurrentUser()` fails with an error, `user.Username` will panic.  Before, you'd use the result from `tryEnv("PGUSER","")`.

db/db.go
```diff
@@ -34,3 +43,11 @@ func NewDBConn(masterHost string, masterPort int, dbname string) *dbconn.DBConn
 		Version:  dbconn.GPDBVersion{},
 	}
 }
+
+func tryEnv(key string, defaultValue string) string {
+	val := utils.System.Getenv(key)
```
dkrieger-pvtl, line 48: This code should use [LookupEnv](https://golang.org/pkg/os/#LookupEnv) as follows:
```go
val, ok := utils.System.LookupEnv(key)              //need to add this to system functions
if !ok {
   return defaultValue
}

return val
```

There is a difference between an empty env variable and a non-existent one.  It's pedantic here, but I think still work doing so no one else cuts-and-pastes it.

db/db_test.go
```diff
@@ -53,5 +54,38 @@ func TestNewDBConn(t *testing.T) {
 			t.Errorf("got host %q want %q", conn.DBName, expected)
 		}
 	})
+}
+
+func resetSystemFunctions() {
```
dkrieger-pvtl, line 59: not needed if you rewrite the tests...see below.

db/db_test.go
```diff
@@ -53,5 +54,38 @@ func TestNewDBConn(t *testing.T) {
...
+	utils.System = utils.InitializeSystemFunctions()
+}
+
+func TestUserUtils(t *testing.T) {
+	t.Run("tryEnv returns environment variables", func(t *testing.T) {
```
dkrieger-pvtl, line 64: I'd prefer you generate a temporary environment variable name, set it directly, and use that.  There's no need to use dependency injection here.

db/db_test.go
```diff
@@ -53,5 +54,38 @@ func TestNewDBConn(t *testing.T) {
...
+			return expected
+		}
+
+		actual := tryEnv("VAR", "default")
+		if actual != expected {
+			t.Errorf("got %q want %q", actual, expected)
+		}
+	})
 
+	t.Run("tryEnv returns the default value when an environmental variable does not exist", func(t *testing.T) {
```
dkrieger-pvtl, line 78: As above, use the direct function call; there's no need for dependency injection.

kalensk: done.

db/db_test.go
kalensk: done.

db/db.go
kalensk: Updated.

db/db.go
kalensk: I updated to set `currentUser = &user.User{}` when there is an error.

db/db.go
```diff
@@ -4,20 +4,29 @@
...
 	_ "github.com/lib/pq" //_ import for the side effect of having postgres driver available
+
+	"github.com/greenplum-db/gpupgrade/utils"
 )
 
 // TODO: do we need this code anymore?
 func NewDBConn(masterHost string, masterPort int, dbname string) *dbconn.DBConn {
-	currentUser, _, _ := utils.GetUser()
-	username := utils.TryEnv("PGUSER", currentUser)
+	user, err := utils.System.CurrentUser()
```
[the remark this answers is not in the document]
kalensk: Sounds like a great followup PR if they do indeed overlap correctly. I do not want to tackle that for this current PR as it expands the scope a bit too much for this simple cleanup task.

I think the overall goal is to migrate away from the dbconn package entirely and simplify our sql connection. Perhaps something like how we do for update_catalog.go https://github.com/greenplum-db/gpupgrade/blob/e8fe74c707dbf2db37b0b465b0d6eec8afa2715c/hub/update_catalog.go#L59

PR #349, opened 2020-06-11
Set the value of bytea_output to escape

bhuvnesh2703:
On greenplum 6, the default value of bytea representation is hex, this
causes differences while comparing with the dump from a 5 cluster.
This commit changes the value of the GUC before taking dump to clear out
diffs due to bytea representation.

### DIFF
--- a/ci/scripts/upgrade-cluster.bash
+++ b/ci/scripts/upgrade-cluster.bash
@@ -5,6 +5,27 @@
 
 set -eux -o pipefail
 
+is_GPDB5() {
+    local gphome=$1
+    version=$(ssh mdw "$gphome"/bin/postgres --gp-version)
+    [[ $version =~ ^"postgres (Greenplum Database) 5." ]]
+}
+
+# set the database gucs
+# 1. bytea_output: by default for bytea the output format is hex on GPDB 6,
+#    so change it to escape to match GPDB 5 representation
+configure_gpdb_gucs() {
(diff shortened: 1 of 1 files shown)


ci/scripts/upgrade-cluster.bash
```diff
@@ -5,6 +5,28 @@
...
+EOF
+}
+
+is_GPDB5() {
+    local gphome=$1
+    local version=$("$gphome"/bin/postgres --gp-version)
+
+    [[ $version =~ ^"postgres (Greenplum Database) 5." ]]
+}
+
```
kalensk, line 29: Sourcing helpers at the top can avoid this duplication.

```sh
dirpath=$(dirname "${0}")
source "${dirpath}/../../test/helpers.bash"
```

I am running a [test pipeline](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:kalen_bytea-diffs) to ensure I have the path correct, etc.

kalensk: Sorry, I had the path wrong. Try:

```sh
source gpupgrade_src/test/helpers.bash
```

and at the bottom of helpers.bash
```sh
is_GPDB5() {
    local gphome=$1
    local version=$("$gphome"/bin/postgres --gp-version)

    [[ $version =~ ^"postgres (Greenplum Database) 5." ]]
}
```

PR #351, opened 2020-06-15
Bats support multi version upgrade

dgkimura:
This introduces GPHOME_SOURCE and GPHOME_TARGET environment variables.
Default both values to GPHOME so that developers don't have to work
harder for the common case of testing X-to-X upgrades. For X-to-Y
upgrades developer will have to set source and target variables.

https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:bats-support-multi-version-upgrade


test/helpers.bash
```diff
@@ -163,3 +177,96 @@ archive_dir() {
...
+    fi
+}
+
+# restore_cluster brings a cluster back to a known state before upgrade. It
+# uses rsync to account for an issue in GPDB5 where the standby and mirrrors
+# become out of sync and fail to start, thus causing gpstart to return non-zero
+# exit code.
+restore_cluster() {
+    if is_GPDB5 "$GPHOME_SOURCE"; then
+        for var in "${rsync_pairs[@]}"; do IFS="|"; set -- $var;
```
kalensk, line 254: 1) It seems like we should either pass in `rsync_pars` explicitly, or rename it to all caps to signify it's a global.

2) We might want to add a check at the beginning of this function to ensure that it is set and not empty and if so bail out early saying to run `setup_restore_cluster`?

test/finalize.bats
```diff
@@ -23,55 +23,59 @@ teardown() {
...
-        delete_finalized_cluster $NEW_CLUSTER
+        delete_finalized_cluster $GPHOME_TARGET $NEW_CLUSTER
     fi
 
     gpupgrade kill-services
 
+    for FUNCTION in "${TEARDOWN_FUNCTIONS[@]}"; do
+        $FUNCTION
+    done
+
```
kalensk, line 34: Instead of this pattern can we use the existing pattern? This would require some changes, but it seems like it might be cleaner to me? Just a thought.

```sh
if [ -n "$RYSNC_PAIRS" ]; then
    restore_cluster $RSYNC_PAIRS
fi
```

test/finalize.bats
```diff
@@ -121,13 +125,16 @@ upgrade_cluster() {
         #   That is a more accurate representation if the standby is running and
         #   in sync, since gpstate might simply check if the process is running.
         local new_datadir=$(gpupgrade config show --target-datadir)
-        local actual_standby_status=$(gpstate -d "${new_datadir}")
+        local actual_standby_status=$(source "${GPHOME_TARGET}/greenplum_path.sh" && gpstate -d "${new_datadir}")
```
kalensk, line 128: I suppose `local actual_standby_status=$("${GPHOME_TARGET}/bin/gpstate" -d "${new_datadir}")` didn't just work due to python issues   :(

test/gpinitsystem.bats
```diff
@@ -26,13 +26,17 @@ teardown() {
...
 }
 
 @test "initialize runs gpinitsystem based on the source cluster" {
     # Store the data directories for each source segment by port.
-    run $PSQL -AtF$'\t' -p $PGPORT postgres -c "select port, datadir from gp_segment_configuration where role = 'p'"
+    if is_GPDB5 "$GPHOME_SOURCE"; then
+        run $PSQL -AtF$'\t' -p $PGPORT postgres -c "SELECT s.port, e.fselocation as datadir FROM gp_segment_configuration s JOIN pg_filespace_entry e ON s.dbid = e.fsedbid JOIN pg_filespace f ON e.fsefsoid = f.oid WHERE f.fsname = 'pg_system' and s.role='p'"
+    else
+        run $PSQL -AtF$'\t' -p $PGPORT postgres -c "select port, datadir from gp_segment_configuration where role = 'p'"
+    fi
```
kalensk, line 39: Would the helper function `get_segment_configuration` be useful here?

test/gpinitsystem.bats
```diff
@@ -90,15 +94,15 @@ teardown() {
...
     # We need to have enough ports available for the master, standby, and mirrors.
     # XXX as usual in these tests, we assume a standard demo cluster.
-    local expected_ports="15432,15434,15435,15436"
-    local mirror_ports="15437,15438,15439"
-    local standby_port=15433
-    local newport=15432
+    local expected_ports="35432,35434,35435,35436"
+    local mirror_ports="35437,35438,35439"
+    local standby_port=35433
+    local newport=35432
```
kalensk, line 100: Why the port change? Seems unrelated to this particular commit and PR? I'm assuming we want to test something other than the defaults?

test/pg_upgrade.bats
```diff
@@ -101,22 +105,25 @@ teardown() {
...
-        "$GPHOME"/bin/postgres -C gp_dbid -D $datadir
+        "$gphome"/bin/postgres -C gp_dbid -D $datadir
     done | sort | uniq | wc -l
 }
 
 @test "upgrade maintains separate DBIDs for each segment" {
-    local old_dbid_num=$(count_primary_gp_dbids $PGPORT)
+    setup_restore_cluster "--mode=copy"
+
+    local old_dbid_num=$(count_primary_gp_dbids $GPHOME_GPHOME_SOURCE $PGPORT)
```
kalensk, line 121: typo in `$GPHOME_GPHOME_SOURCE `? Should be `$GPHOME_SOURCE`?

test/pg_upgrade.bats
```diff
@@ -101,22 +105,25 @@ teardown() {
...
+    local gphome=$1
+    local port=$2
 
     for datadir in $($PSQL -At -p $port postgres -c "
         select datadir from gp_segment_configuration where role='p'
     "); do
-        "$GPHOME"/bin/postgres -C gp_dbid -D $datadir
+        "$gphome"/bin/postgres -C gp_dbid -D $datadir
     done | sort | uniq | wc -l
 }
```
kalensk, line 116: Do we want to use the `query_data_dirs` helper?

test/revert.bats
```diff
@@ -51,47 +51,65 @@ setup() {
...
     gpupgrade revert --verbose
 
-    pg_isready -q || fail "expected source cluster to be running on port ${PGPORT}"
-    ! pg_isready -qp ${target_master_port} || fail "expected target cluster to not be running on port ${target_master_port}"
+    if command -v "${GPHOME_SOURCE}"/bin/pg_isready > /dev/null; then
+        pg_isready -q || fail "expected source cluster to be running on port ${PGPORT}"
+        ! pg_isready -qp ${target_master_port} || fail "expected target cluster to not be running on port ${target_master_port}"
+    else
+        "${GPHOME_SOURCE}"/bin/psql postgres -qc "SELECT 1" &> /dev/null || fail "expected source cluster to be running on port ${PGPORT}"
+        # TODO: Check target_master_port
```
kalensk, line 78: Do we need to circle back on this TODO?

ci/template.yml
```diff
@@ -218,21 +210,31 @@ jobs:
...
+- name: {{.Name}}
   plan:
   - in_parallel:
     - get: gpupgrade_src
       trigger: true
-    - get: gpdb_src
-      resource: gpdb6_src
+    - get: gpdb_src_source
+      resource: gpdb{{.Source}}_src
+    - get: gpdb_src_target
```
dkrieger-pvtl, line 221: It appears the `gpdb_src_target` is only used as a trigger for this job.  Is there another way you can trigger "on changed" without actually pulling the resource?  I find its current idiom confusing.

From a high-level, we are running tests between `5.rpm -> 6.rpm` and `6.rpm -> 6.rpm` using the demo-cluster made from `gpdb_src_source` with the tests from `gpupgrade_src`.  That's because the rpm do not have the gpAux to allow us to make the demo cluster?

test/helpers.bash
dgkimura: > 1. It seems like we should either pass in `rsync_pars` explicitly, or rename it to all caps to signify it's a global.

Ah, thanks!  I missed that detail.  I think I was copying existing pattern from `master_and_primary_datadirs` variable. So I'll all caps both of them.

> 2. We might want to add a check at the beginning of this function to ensure that it is set and not empty and if so bail out early saying to run `setup_restore_cluster`?

It should only be set if source cluster is GPDB 5.  I could add an assert inside this conditional, but I'm not quite sure I see the benefit of exiting early here.  If it's not set then it'll exit the for loop early as a noop.

test/finalize.bats
dgkimura: I'm open to this suggestion, but want to make sure I follow.  Is it more clear because it explicitly calls restore_cluster instead of a "function pointer"?

If I go with this suggestion, I think I'll keep the details of $RSYNC_PAIRS encapsulated in the restore_cluster and setup_restore_cluster functions as I'm not sure that the caller should need to know about that.  How does that sound?

kalensk: Yeah, Im not a fan of all the indirection especially in BASH. I rather have test code lean on the side of more verbose at the benefit of added clarity than not.

I agree with keeping things encapsulated inside `setup_restore_cluster` and `restore_cluster`. Feel free to rename the variables and functions to whatever makes the most sense.

I could see something like the following:
```sh

teardown() {
    if [ -n "$DATA_DIRS" ]; then
        restore_cluster $DATA_DIRS
    fi
}

@test "some test" {
    DATA_DIRS=$(save_datadirs_to_restore_cluster "$LINK_MODE")
    # ...
}

```

ci/template.yml
kalensk: Looks like we don't even need `gpdb_src_target`? When first working on this PR I added that originally, but it no longer seems to be the case. Good catch.

test/helpers.bash
```diff
@@ -163,3 +177,96 @@ archive_dir() {
...
+
+    if is_GPDB5 "$GPHOME_SOURCE"; then
+        rsync_pairs=($(get_rsync_pairs $GPHOME_SOURCE $PGPORT))
+    fi
+
+    if [ "$mode" == "--mode=link" ]; then
+        master_and_primary_datadirs=($(query_datadirs $GPHOME_SOURCE $PGPORT "role = 'p'"))
+    else
+        master_and_primary_datadirs=
+    fi
```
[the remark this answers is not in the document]
dgkimura: Added comment above

test/gpinitsystem.bats
dgkimura: Added comment above

test/finalize.bats
dgkimura: Yup

test/gpinitsystem.bats
dgkimura: Good suggestion! And thanks!

ci/scripts/install-tests.bash
```diff
@@ -8,31 +8,93 @@ set -ex
 # Install BATS
 ./bats/install.sh /usr/local
 
-source gpdb_src/concourse/scripts/common.bash
-mkdir -p /usr/local/greenplum-db-devel
-tar -xzf bin_gpdb/*.tar.gz -C /usr/local/greenplum-db-devel
+source go/src/github.com/greenplum-db/gpupgrade/test/helpers.bash
```
bhuvnesh2703, line 11: is this gonna be like this on the ci ?

kalensk: The full path like that is a hold over from before we were using Go modules when everything had to be in the GOPATH. I can try updating it to `gpupgrade_src/test/helpers.bash`

ci/parser/parse_template.go
```diff
@@ -30,6 +31,14 @@ var sourceVersions = []string{"6", "5"}
 var targetVersions = []string{"6"}
 var centosVersions = []string{"6", "7"}
 
+type CheckJob struct {
```
dkrieger-pvtl, line 34: super nit: not part of this PR, but perhaps `var sourceVersions = []string{"6", "5"}` should read `var sourceVersions = []string{"5", "6"}` instead.

Makefile
```diff
@@ -98,14 +98,6 @@ override BUILD_FLAGS += -ldflags $(VERSION_LD_STR)
 install:
 	go install $(BUILD_FLAGS) github.com/greenplum-db/gpupgrade/cmd/gpupgrade
 
-# We intentionally do not depend on install here -- the point of installcheck is
-# to test whatever has already been installed. NOTE: installcheck.bats is not
-# located in the test directory so that it is not run during check.
-installcheck:
```
dkrieger-pvtl, line 104: Why are we deleting this?

ci/scripts/install-tests.bash
```diff
@@ -8,31 +8,90 @@ set -ex
...
-time make_cluster
+export GPHOME_SOURCE=/usr/local/greenplum-db-source
+export GPHOME_TARGET=/usr/local/greenplum-db-target
 
-time chown -R gpadmin:gpadmin go
+# setup gpadmin user
+# setup_gpadmin_user.bash expects gpdb_src, whereas we are using
+# gpdb_src_source since this is a X-to-Y upgrade test.
+mkdir gpdb_src
+time ./gpdb_src_source/concourse/scripts/setup_gpadmin_user.bash "centos"
```
bhuvnesh2703, line 20: why do we have `gpdb_src_source` ?

Makefile
dkrieger-pvtl: Answered out-of-band: we have several other tests that essentially do the same thing.  `finalize.bats` can be run instead of this.

ci/parser/parse_template.go
```diff
@@ -120,10 +136,34 @@ func init() {
...
+		AllVersions:       deduplicate(sourceVersions, targetVersions),
 		UpgradeJobs:       upgradeJobs,
 		LastTargetVersion: targetVersions[len(targetVersions)-1],
 		Versions:          versions,
+		CheckJobs:         checkJobs,
+	}
+}
+
+// deduplicate combines, sorts, and deduplicates two string slices.
+func deduplicate(a, b []string) []string {
```
dkrieger-pvtl, line 148: nit: I find using maps to be much cleaner, but it's up to you.

```go
func deduplicate(a, b []string) []string {
	m := make(map[string]bool)
	for _, v := range a {
		m[v] = true
	}
	for _, v := range b {
		m[v] = true
	}
	
	var r []string
	for k, _ := range m {
		r = append(r, k)
	}
	
	return r

}
```

ci/scripts/install-tests.bash
```diff
@@ -8,31 +8,90 @@ set -ex
...
+sudo ln -s /usr/local/greenplum-db-${version} "$GPHOME_TARGET"
+
+chown -R gpadmin:gpadmin "$GPHOME_TARGET"
+
+# create source cluster
+chown -R gpadmin:gpadmin gpdb_src_source/gpAux/gpdemo
+source "$GPHOME_SOURCE"/greenplum_path.sh
+
+# XXX: Need to update demo_cluster.sh to handle symlinks by using find -H.
+# A PR for this should be made.
```
dkrieger-pvtl, line 48: What does this line mean?  Perhaps it is stale?

ci/scripts/install-tests.bash
```diff
@@ -8,31 +8,90 @@ set -ex
...
+sed -i.bak -E 's/(^GPPATH=`find) (\$GPSEARCH -name .*$)/\1 -H \2/' gpdb_src_source/gpAux/gpdemo/demo_cluster.sh
+
+pushd gpdb_src_source/gpAux/gpdemo
+    time su gpadmin -c "make create-demo-cluster"
+    source gpdemo-env.sh
+popd
+
```
dkrieger-pvtl, line 55: nit: add a comment here, something like:

`drop all gphdfs roles in GPDB5, as gphdfs is obsoleted in GPDB6`

ci/scripts/install-tests.bash
kalensk: Creating a demo cluster when using installed RPMs that use symlinks (such as we do here in install-tests.bash) does not work without this fix. That is this line in [demo_clusther.sh fails](https://github.com/greenplum-db/gpdb/blob/3b2aed6e2313792e4d1dd3072d300f12947b69cb/gpAux/gpdemo/demo_cluster.sh#L213). This is a workaround for that.

test/helpers.bash
```diff
@@ -31,26 +36,41 @@ abort() {
 
 # skip_if_no_gpdb() will skip a test if a cluster's environment is not set up.
 skip_if_no_gpdb() {
-    [ -n "${GPHOME}" ] || skip "this test requires an active GPDB cluster (set GPHOME)"
-    [ -n "${PGPORT}" ] || skip "this test requires an active GPDB cluster (set PGPORT)"
+    [ -n "${GPHOME_SOURCE}" ] || skip "this test requires an active GPDB source cluster (set GPHOME or GPHOME_SOURCE)"
+    [ -n "${GPHOME_TARGET}" ] || skip "this test requires an active GPDB target cluster (set GPHOME or GPHOME_TARGET)"
+    [ -n "${PGPORT}" ] || skip "this test requires an active GPDB source cluster (set PGPORT)"
```
dkrieger-pvtl, line 41: I think we also want to add a test that `MASTER_DATA_DIRECTORY` is set here as well; it is used in execute.bats and finalize.bats.

kalensk: That makes sense to me.

I remember trying to add that in a previous PR, but @jchampio had some reservations that are escaping me at the moment.

test/helpers.bash
```diff
@@ -31,26 +36,41 @@ abort() {
 
 # skip_if_no_gpdb() will skip a test if a cluster's environment is not set up.
 skip_if_no_gpdb() {
-    [ -n "${GPHOME}" ] || skip "this test requires an active GPDB cluster (set GPHOME)"
-    [ -n "${PGPORT}" ] || skip "this test requires an active GPDB cluster (set PGPORT)"
+    [ -n "${GPHOME_SOURCE}" ] || skip "this test requires an active GPDB source cluster (set GPHOME or GPHOME_SOURCE)"
```
bhuvnesh2703, line 39: why set GPHOME?

test/helpers.bash
kalensk: Now I remembering, it was that gpupgrade should only know the `port` and `bindir` and that is it. It then can connect to the cluster and query for all the information. Thus, `MASTER_DATA_DIRECTORY` should not be needed as a dependency and adding that here breaks that contract.

test/helpers.bash
kalensk: Because we default to GPHOME is the user does not set GPHOME_SOURCE and GPHOME_TARGET.
```
# Default to GPHOME for both the source and target installations. These may be
# overridden manually for cross-version testing.
GPHOME_SOURCE=${GPHOME_SOURCE:-$GPHOME}
GPHOME_TARGET=${GPHOME_TARGET:-$GPHOME}
```

bhuvnesh2703: ok.. so basically if testing 6-6 we somehow want it to work. thanks for explaining.

test/checks.bats
```diff
@@ -78,7 +78,7 @@ are_equivalent_within_tolerance() {
 @test "initialize prints disk space on failure" {
     require_gnu_stat
 
-    datadir=$(psql postgres -Atc "select datadir from gp_segment_configuration where role='p' and content=-1")
+    local datadir=$(query_datadirs $GPHOME_SOURCE $PGPORT "role='p' and content=-1")
```
dkrieger-pvtl, line 81: In general, the [variable declaration should be on a separate line from the assignment](https://github.com/koalaman/shellcheck/wiki/SC2155).  Otherwise, failures here won't cause the test to stop.  

I don't see any terrible consequences from the assignments you added this way, and the current code has many more transgressions, so I don't see this change(and others in the code) as a requirement for this PR.

In IntelliJ, I use [shellcheck](https://github.com/koalaman/shellcheck) which finds these issues.

PR #352, opened 2020-06-15
cluster: add an HasAllMirrorsAndStandby() check

dkrieger-pvtl:
HasAllMirrorsAndStandby() on a cluster returns true if the master
has a standby and if each primary has a mirror.
It returns false otherwise.

This is a utility function that will be useful in upcoming features.

### DIFF
--- a/greenplum/cluster.go
+++ b/greenplum/cluster.go
@@ -105,6 +105,16 @@ func (c *Cluster) HasMirrors() bool {
 	return false
 }
 
+func (c *Cluster) HasAllMirrorsAndStandby() bool {
+	for content := range c.Primaries {
+		if _, ok := c.Mirrors[content]; !ok {
+			return false
+		}
+	}
+
+	return true
+}
+
 // XXX This does not provide mirror hostnames yet.
 func (c *Cluster) GetHostnames() []string {
 	hostnameMap := make(map[string]bool)
--- a/greenplum/cluster_test.go
+++ b/greenplum/cluster_test.go
@@ -364,3 +364,86 @@ func TestSelectSegments(t *testing.T) {
 	}
 
 }
+
+func TestHasAllMirrorsAndStandby(t *testing.T) {
+	t.Run("returns true on full cluster", func(t *testing.T) {
+		segs := []greenplum.SegConfig{
+			{ContentID: -1, Role: "p"},
+			{ContentID: -1, Role: "m"},
+			{ContentID: 0, Role: "p"},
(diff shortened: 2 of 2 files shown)


greenplum/cluster.go
```diff
@@ -105,6 +105,17 @@ func (c *Cluster) HasMirrors() bool {
 	return false
 }
 
+// Returns true if each content has a standby/mirror
```
kalensk, line 108: minor nit: Now that the function name is more verbose I think the comment is not needed.

greenplum/cluster_test.go
```diff
@@ -364,3 +364,86 @@ func TestSelectSegments(t *testing.T) {
...
+			{ContentID: 2, Role: "p"},
+			{ContentID: 2, Role: "m"},
+		}
+		cluster, err := greenplum.NewCluster(segs)
+		if err != nil {
+			t.Fatalf("creating new cluster: %+v", err)
+		}
+
+		if !cluster.HasAllMirrorsAndStandby() {
+			t.Errorf("expected a full cluster")
```
kalensk, line 386: minor nit: I would just use the verbose terminology. Something like, "expected cluster that has all mirrors and standby". Similar for the other assertion below.

greenplum/cluster_test.go
```diff
@@ -364,3 +364,86 @@ func TestSelectSegments(t *testing.T) {
...
+			{ContentID: 0, Role: "p"},
+			{ContentID: 0, Role: "m"},
+			{ContentID: 1, Role: "p"},
+			{ContentID: 1, Role: "m"},
+			{ContentID: 2, Role: "p"},
+			{ContentID: 2, Role: "m"},
+		}
+		cluster, err := greenplum.NewCluster(segs)
+		if err != nil {
+			t.Fatalf("creating new cluster: %+v", err)
```
kalensk, line 382: minor nit: The convention is using the method name with returned error phrase, so `"NewCluster returned error %+v`. Similar for the other assertion below.

greenplum/cluster.go
dkrieger-pvtl: Done. Thanks.

greenplum/cluster_test.go
dkrieger-pvtl: Done.

greenplum/cluster_test.go
dkrieger-pvtl: Done.

PR #353, opened 2020-06-16
add new stream methods for general use

dkrieger-pvtl:
This PR moves two stream types(DevNull and BufferedStreams) to a common
location so it is generally useful.

These will be useful in future work.

### DIFF
--- a/step/stream.go
+++ b/step/stream.go
@@ -4,7 +4,9 @@
 package step
 
 import (
+	"bytes"
 	"io"
+	"io/ioutil"
 	"sync"
 
 	"github.com/greenplum-db/gp-common-go-libs/gplog"
@@ -20,6 +22,35 @@ type OutStreams interface {
 type OutStreamsCloser interface {
 	OutStreams
 	Close() error
+}
+
+// DevNullStream provides an implementation of OutStreams that drops
+//   all writes to it.
+var DevNullStream = devNullStream{}
+
+type devNullStream struct{}
+
+func (_ devNullStream) Stdout() io.Writer {
+	return ioutil.Discard
+}
+
+func (_ devNullStream) Stderr() io.Writer {
+	return ioutil.Discard
+}
+
+// BufferedStreams provides an implementation of OutStreams that stores
+//   all writes to underlying bytes.Buffer objects.
+type BufferedStreams struct {
+	StdoutBuf bytes.Buffer
+	StderrBuf bytes.Buffer
+}
+
+func (s *BufferedStreams) Stdout() io.Writer {
+	return &s.StdoutBuf
+}
+
+func (s *BufferedStreams) Stderr() io.Writer {
+	return &s.StderrBuf
 }
 
 // multiplexedStream provides an implementation of OutStreams that safely
--- a/agent/delete_directories.go
+++ b/agent/delete_directories.go
@@ -9,6 +9,7 @@ import (
 	"github.com/greenplum-db/gp-common-go-libs/gplog"
 
 	"github.com/greenplum-db/gpupgrade/idl"
+	"github.com/greenplum-db/gpupgrade/step"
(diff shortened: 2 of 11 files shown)


kalensk:
The DevNull stream is already in [sys_utils.go](https://github.com/greenplum-db/gpupgrade/blob/8f6369056c5d0a406a179587d5ca2acfcb80fdd3/utils/sys_utils.go#L135-L146) and the BufferedStreams in [upgrade_master_test.go](https://github.com/greenplum-db/gpupgrade/blob/8f6369056c5d0a406a179587d5ca2acfcb80fdd3/hub/upgrade_master_test.go#L276-L290).

Let's use those existing streams rather than duplicating.

dkrieger-pvtl:
> The DevNull stream is already in [sys_utils.go](https://github.com/greenplum-db/gpupgrade/blob/8f6369056c5d0a406a179587d5ca2acfcb80fdd3/utils/sys_utils.go#L135-L146) and the BufferedStreams in [upgrade_master_test.go](https://github.com/greenplum-db/gpupgrade/blob/8f6369056c5d0a406a179587d5ca2acfcb80fdd3/hub/upgrade_master_test.go#L276-L290).
> 
> Let's use those existing streams rather than duplicating.

Done; I've refactored to pull the current implementation into `step/stream.go`.

kalensk:
While are here with this PR do we want to "bite the bullet" and also consolidate all the other streams to `stream.go` or `stream_test.go` (for those implementations only used by tests). I know its more work, but now seems like now would be a great time.

- `StdStream` in [sys_utils.go](https://github.com/greenplum-db/gpupgrade/blob/21dced291e1347680dd9a15ea8d0171e78f21eb4/utils/sys_utils.go#L148-L163) could maybe go to `stream.go`
- `devNullSpy` in [directories_test.go](https://github.com/greenplum-db/gpupgrade/blob/21dced291e1347680dd9a15ea8d0171e78f21eb4/upgrade/directories_test.go#L316-L326) could maybe go into `stream_test.go`
- `DevNullWithClose` in [step_test.go](https://github.com/greenplum-db/gpupgrade/blob/21dced291e1347680dd9a15ea8d0171e78f21eb4/step/step_test.go#L351-L371) could maybe go into `stream_test.go`
- `failingStreams` in [upgrade_master_test.go](https://github.com/greenplum-db/gpupgrade/blob/21dced291e1347680dd9a15ea8d0171e78f21eb4/hub/upgrade_master_test.go#L291-L303) could maybe go into `stream_test.go`
- `DevNullWithClose` in [step_test.go](https://github.com/greenplum-db/gpupgrade/blob/21dced291e1347680dd9a15ea8d0171e78f21eb4/step/step_test.go#L351-L371) could maybe go into `stream_test.go`

dkrieger-pvtl:
> While are here with this PR do we want to "bite the bullet" and also consolidate all the other streams to `stream.go` or `stream_test.go` (for those implementations only used by tests). I know its more work, but now seems like now would be a great time.

I'd prefer not to, as my main task here is to implement revert on link-mode execute.  The other streams can be added in incrementally.  

We can add this as a (perhaps never done?) item of tech debt.

step/stream.go
```diff
@@ -22,6 +24,35 @@ type OutStreamsCloser interface {
...
+	return ioutil.Discard
+}
+
+func (_ devNullStream) Stderr() io.Writer {
+	return ioutil.Discard
+}
+
+// BufferedStreams provides an implementation of OutStreams that stores
+//   all writes to underlying bytes.Buffer objects.
+type BufferedStreams struct {
```
kalensk, line 43: super minor nit and I realize that this is a hold over from before where `BufferedStreams ` is plural and `DevNullStream` is not, would be nice to be consistent, but not strictly "needed".

PR #355, opened 2020-06-16
consolidate all test streams to testutils/stream.go

kalensk:
Riffing off https://github.com/greenplum-db/gpupgrade/pull/354 this consolidates all test streams into `testutils/stream.go`.

[Test pipeline](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:test_streams)

### DIFF
--- a/hub/common_test.go
+++ b/hub/common_test.go
@@ -32,15 +32,6 @@ func ResetRsyncExecCommand() {
 	execCommandRsync = nil
 }
 
-// failingWriter is an io.Writer for which all calls to Write() return an error.
-type failingWriter struct {
-	err error
-}
-
-func (f *failingWriter) Write(_ []byte) (int, error) {
-	return 0, f.err
-}
-
 // MustCreateCluster creates a utils.Cluster and calls t.Fatalf() if there is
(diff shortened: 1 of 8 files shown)


dkrieger-pvtl:
I like the idea of consolidating the various OutStreams implementations in a single file.  Since each _test file is its own binary, I think we do have to use a non-test file to do this across tests.

I would consider two changes:
1). consider if we want to add tests for "non-trivial" streams here
2). (for another PR) add a test(via a Makefile target, perhaps) that makes sure the no symbols from `testutils` makes it into our production binary.  We could use `go tool objdump -s testutils gpupgrade` and check if the output is empty or not.  I filed a story for this.

dkrieger-pvtl:
(I am not sure of the Draft PR workflow; I'd expect Approve here to mean approval to move to a full PR, but it appears not to denote that.). Anyway, I am fine with this moving to a full PR; I didn't review the code in detail yet.

bhuvnesh2703:
LGTM, But can we have them not being imported into main code base or somehow excluded from compilation of main code?

PR #360, opened 2020-06-19
provide a general purpose Rsync function

dkrieger-pvtl:
We call the rsync system command in many places in our code(at least 5).  The implementations differ slightly, and add a lot of boilerplate code.  In addition, the different commands hide a central feature: how to return stdout, stderr and errors from rsync.

Refactoring to a [functional options](https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis) approach simplifies the interface.  

It also rationalizes our handling of stdout, stderr and errors.  Generally, hub calls to rsync use a stream, so the caller can see stderr.  Agent calls do not, so we need to capture the error internally in Rsync().


utils/rsync/rsync.go
```diff
@@ -0,0 +1,148 @@
...
+	// when no streams are specified, capture stderr for the error message
+	stream := step.BufferedStreams{}
+	cmd.Stderr = stream.Stderr()
+	if opts.useStream {
+		cmd.Stdout = opts.stream.Stdout()
+		cmd.Stderr = opts.stream.Stderr()
+	}
+
+	gplog.Info("running Rsync as %s", cmd.String())
+	fmt.Println(cmd)
```
bhuvnesh2703, line 50: is it a left over fmt.Println?

utils/rsync/rsync_test.go
```diff
@@ -0,0 +1,331 @@
...
+			rsync.WithOptions([]string{"--archive", "--verbose"}...),
+			rsync.WithStream(streams),
+		}
+		if err := rsync.Rsync(opts...); err != nil {
+			t.Errorf("Rsync() returned error %+v", err)
+		}
+
+		actual := streams.StdoutBuf.String()
+		if !strings.Contains(actual, filename) {
+			t.Errorf("expected stdout to contain filename \"%s\" but has \"%s\"", filename, actual)
```
bhuvnesh2703, line 148: why not use %q to quote?

utils/rsync/rsync.go
```diff
@@ -0,0 +1,148 @@
...
+		for _, excludedFile := range files {
+			options.excludedFiles = append(options.excludedFiles, "--exclude", excludedFile)
+		}
+	}
+}
+
+func WithStream(stream step.OutStreams) Option {
+	return func(options *optionList) {
+		options.stream = stream
+		options.useStream = true
```
bhuvnesh2703, line 127: do we really need useStream var? or we can rely if stream is nil or not?

utils/rsync/rsync_test.go
```diff
@@ -0,0 +1,331 @@
...
+		if !xerrors.As(err, &exitError) {
+			t.Errorf("expected err to wrap ExitError")
+		}
+		if !strings.Contains(rsyncError.Error(), expected) {
+			t.Errorf("got %v, expected substring %s", err.Error(), expected)
+		}
+	})
+}
+
+func pathExists(path string) bool {
```
dgkimura, line 328: Caveat, I'm a Golang novice.  ;)

Seems like this function is being used as an assert?  Why not wrap `tErrorf()` inside the function, like..

```go
func assertPathExists(t *testing.T) (path string) bool {
	if err == nil {
		t.Errorf("file %q should exist, but it does not", path)
	}
}
func assertPathDoesNotExist(t *testing.T) (path string) bool {
	err := assertPathExists(path)
	if err == nil {
		t.Errorf("file %q should not exist, but it does", path)
	}
}
```

Also curious why this function lives here, but `MustReadFile ()` and `MustWriteToFile()` live in test_utils.go?

hub/copy_master.go
```diff
@@ -37,21 +37,21 @@ func Copy(streams step.OutStreams, destinationDir string, sourceDirs, hosts []st
...
 
-			result := Result{}
-			cmd.Stdout = &result.stdout
-			cmd.Stderr = &result.stderr
+			options := []rsync.Option{
+				rsync.WithSources(sourceDirs...),
+				rsync.WithRemoteHost(hostname),
+				rsync.WithDestination(destinationDir),
+				rsync.WithOptions("--archive", "--compress", "--delete", "--stats"),
+				rsync.WithStream(stream),
```
kalensk, line 47: Why can't we pass in the `streams` parameter directly? That is, this function `Copy` takes in a stream of type `step.Outstreams` which we should use.

utils/rsync/rsync_test.go
```diff
@@ -0,0 +1,331 @@
...
+	})
+
+	t.Run("when no input stream is provided, it returns an RsyncError that wraps an ExitError", func(t *testing.T) {
+		source := testutils.GetTempDir(t, "source")
+		defer testutils.MustRemoveAll(t, source)
+
+		destination := testutils.GetTempDir(t, "destination")
+		defer testutils.MustRemoveAll(t, destination)
+
+		testutils.MustWriteToFile(filepath.Join(source, "filename.txt"), []byte("hi"), t)
```
kalensk, line 297: nit: no need for `"hi"` here.

testutils/test_utils.go
```diff
@@ -122,6 +122,26 @@ func MustCreateDataDirs(t *testing.T) (string, string, func(*testing.T)) {
 	}
 }
 
+func MustReadFile(path string, t *testing.T) []byte {
```
kalensk, line 125: Can these be updated to follow the convention of the other `Must` functions where `t *testing.T` is the first parameter?

testutils/test_utils.go
```diff
@@ -122,6 +122,26 @@ func MustCreateDataDirs(t *testing.T) (string, string, func(*testing.T)) {
...
+
+	buf, err := ioutil.ReadFile(path)
+	if err != nil {
+		t.Fatalf("error reading file %q: %v", path, err)
+	}
+
+	return buf
+}
+
+func MustWriteToFile(path string, contents []byte, t *testing.T) {
```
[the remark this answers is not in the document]
kalensk: Yes, its so you can have 1 line in the tests and not several lines to check the error. Makes things more readable when there is a lot of test setup and things.

utils/rsync/rsync.go
```diff
@@ -0,0 +1,148 @@
...
+
+type Option func(*optionList)
+
+func WithSources(srcs ...string) Option {
+	return func(options *optionList) {
+		options.sources = append(options.sources, srcs...)
+	}
+}
+
+func WithRemoteHost(host string) Option {
```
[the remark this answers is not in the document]
kalensk: I think I like it in that it makes the host explicit, and easy when reviewing the calling code where the files are going.

utils/rsync/rsync.go
kalensk: No, not necessary as you pointed out, but it does make the following code a bit more readable/explicit. Personal preference I suppose.

```
if opts.useStream {
		cmd.Stdout = opts.stream.Stdout()
		cmd.Stderr = opts.stream.Stderr()
	}
```

utils/rsync/rsync_test.go
```diff
@@ -0,0 +1,331 @@
...
+	"github.com/greenplum-db/gpupgrade/testutils"
+	"github.com/greenplum-db/gpupgrade/utils"
+	"github.com/greenplum-db/gpupgrade/utils/rsync"
+)
+
+func TestRsync(t *testing.T) {
+	testhelper.SetupTestLogger()
+
+	if _, err := exec.LookPath("rsync"); err != nil {
+		t.Fatalf("tests require rsync (%v)", err)
```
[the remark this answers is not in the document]
kalensk: Long term that is probably the better way to go, but for now its a quick sanity check in case developers don't have rsync on their machines and to give them a useful error.

utils/rsync/rsync_test.go
kalensk: One could do that. There is definitely a balance between too much abstraction and code verbosity. I am not sure it is saving us much, whereas `MustReadFile` is used in test setup code which is nice to abstract to keep the focus on what is being tested.

The function is also in [upgrade/directories.go](https://github.com/greenplum-db/gpupgrade/blob/9a6af4e8eb1a643d120b75039ce679b3c3db6ad3/upgrade/directories.go#L163-L166). I have a preference to simply call it here in this test rather than repeating it.

hub/copy_master.go
dkrieger-pvtl: Copy() calls rsync on each host.  If we passed in the input parameter `streams` directly, we'd end up possibly mixing the stdout/stderr from each rsync call to each host, as they are all happening concurrently.  The current behavior collects each rsync's output separately and then streams them back to the caller in order once all calls have completed.  So each rsync's stdout/stderr is sent sequentially.

kalensk: > The current behavior collects each rsync's output separately and then streams them back to the caller in order once all calls have completed. So each rsync's stdout/stderr is sent sequentially.

I feel like I am missing something. The current implementation and this PR seem the same to me in this regard. Have we tried simply passing in `rsync.WithStream(streams)` with the `step.Outstreams`?

kalensk: With the following change using `step.Outstreams` I see the following output during execute in verbose mode. No need to instantiate a new `BufferedStreams`.

```go
options := []rsync.Option{
    rsync.WithSources(sourceDirs...),
    rsync.WithRemoteHost(hostname),
    rsync.WithDestination(destinationDir),
    rsync.WithOptions("--archive", "--compress", "--delete", "--stats"),
    rsync.WithStream(streams),
}

err := rsync.Rsync(options...)
if err != nil {
    err = xerrors.Errorf("copying source %q to destination %q on host %s: %w", sourceDirs, destinationDir, hostname, err)
}
result := Result{err: err}
results <- &result
}()
```

```
Starting COPY_MASTER...

Copying master catalog to primary segments...                      [IN PROGRESS]

Number of files: 1007
Number of files transferred: 975
Total file size: 123883787 bytes
Total transferred file size: 123883787 bytes
Literal data: 123883787 bytes
Matched data: 0 bytes
File list size: 13402
File list generation time: 0.006 seconds
File list transfer time: 0.000 seconds
Total bytes sent: 3504084
Total bytes received: 21662

sent 3504084 bytes  received 21662 bytes  2350497.33 bytes/sec
total size is 123883787  speedup is 35.14
Copying master catalog to primary segments...                      [COMPLETE]
```

utils/rsync/rsync.go
dkrieger-pvtl: good catch. thanks.

utils/rsync/rsync_test.go
dkrieger-pvtl: fixed, thanks.

utils/rsync/rsync.go
dkrieger-pvtl: Maybe it's a matter of personal preference. 

I find it clearer to use a boolean explicitly than compare to nil; the code is more readable IMHO.

hub/copy_master.go
dkrieger-pvtl: This example is only on a single host; we discussed.

utils/rsync/rsync_test.go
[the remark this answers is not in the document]
dkrieger-pvtl: I'm not sure of the history of this line; the prior behavior was to _skip_ these tests if "rsync" wasn't installed.  But I thought it best to _fail_ these tests so we don't have a silent failure.

That said, the [no-install test](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:rsync_support_prefactor_2/jobs/noinstall-tests/builds/9) doesn't have rsync; we are investigating.

utils/rsync/rsync_test.go
dkrieger-pvtl: The `Must` functions are meant for setup functions that should never fail...creating an empty directory, writing to a new file, etc.  

We could add `assert`-style functions but I think the logic is clearer in-line. 

I don't want to couple in `upgrade` here...we could have a common function in a utility package.

utils/rsync/rsync_test.go
dkrieger-pvtl: Done.

testutils/test_utils.go
dkrieger-pvtl: Done.

utils/rsync/rsync_test.go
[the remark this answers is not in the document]
dkrieger-pvtl: We have added a commit that installs rsync on the no-install instance.  One of the rsync tests uses a remote host of `localhost`, which the no-install instance is not setup for.  So we modified that test: better to run most things in the no-install tests than nothing.

We will file a follow up chore to test the options of rsync, instead of using rsync itself.

PR #363, opened 2020-06-22
ci: update lint job

kalensk:
See individual commits.

[Test pipeline](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:lint-job)


ci/template.yml
```diff
@@ -195,7 +195,24 @@ jobs:
...
     trigger: true
   - task: lint
-    file: gpupgrade_src/ci/tasks/lint.yml
+    config:
+      platform: linux
+      image_resource:
+        type: docker-image
+        source:
+          repository: golangci/golangci-lint
+          tag: 'v1.27'
```
dkrieger-pvtl, line 204: Why stick to this version and not the latest version of `golangci`?

kalensk: Good point! Updated to remove the tag field, which I think uses the latest.

kalensk: Seems like it from comparing the image sha's and in the concorse docs:
> tag: Optional. The tag to track. Defaults to latest.

https://github.com/concourse/docker-image-resource

PR #366, opened 2020-06-24
add step enum and step.HasRun()

kalensk:
This PR does several things:
- Update MustWriteToFile and MustReadFile take and return a string.
- Add Step enum
- add step.HasRun to determine if a substep was already run. This is useful in the revert after execute in link mode when we only want to restore the master and primaries once upgrade primaries substep has started.

[Test Pipeline](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:read-status)


testutils/test_utils.go
```diff
@@ -122,21 +122,21 @@ func MustCreateDataDirs(t *testing.T) (string, string, func(*testing.T)) {
...
-	return buf
+	return string(buf)
 }
 
-func MustWriteToFile(t *testing.T, path string, contents []byte) {
+func MustWriteToFile(t *testing.T, path string, contents string) {
 	t.Helper()
 
-	err := ioutil.WriteFile(path, contents, 0644)
+	err := ioutil.WriteFile(path, []byte(contents), 0644)
```
dgkimura, line 139: Probably doesn't matter, but I noticed that all the tests you converted to use this function had previously set file permissions to 600 or 640, none were 644.  So just making sure that change is intentional.

step/step_test.go
```diff
@@ -249,6 +249,108 @@ func TestStepRun(t *testing.T) {
...
+			dir := testutils.GetTempDir(t, "")
+			defer testutils.MustRemoveAll(t, dir)
+
+			resetEnv := testutils.SetEnv(t, "GPUPGRADE_HOME", dir)
+			defer resetEnv()
+
+			path := filepath.Join(dir, "status.json")
+			testutils.MustWriteToFile(t, path, "{}")
+			store := step.NewFileStore(path)
+			err := store.Write(idl.Step_INITIALIZE, idl.Substep_SAVING_SOURCE_CLUSTER_CONFIG, idl.Status_RUNNING)
```
dgkimura, line 282: Should `store.Write()`use `c.status` as last argument? Otherwise the loop in this test seems unnecessary.

step/file_store_test.go
```diff
@@ -163,8 +163,8 @@ func TestFileStore(t *testing.T) {
 		}
 
 		key := substep.String()
-		if raw[section][key] != status.String() {
-			t.Errorf("status[%q][%q] = %q, want %q", section, key, raw[section][key], status.String())
+		if raw[section.String()][key] != status.String() {
```
dgkimura, line 166: Nit: If I followed the types correctly, `section` is an `int`.  Instead of making these calls to `.String()` would it be better to just change the map type of raw so is..

```go
raw := make(map[int]map[string]string)
```

And then just update the string format of the `t.Errorf`?

testutils/test_utils.go
kalensk: Let's try 600 until more access is needed.

step/step_test.go
kalensk: Good catch!

step/file_store_test.go
kalensk: Agreed! That would be nice, but I don't think it is easily doable with than more work than its worth.


When using an `int` I get:
`decoding statuses: json: cannot unmarshal number FINALIZE into Go value of type idl.Step`

When using the enum `idl.Step` I get:
`cannot use section (type idl.Step) as type int in map index`

PR #367, opened 2020-06-24
revert after link-mode execute completes

dkrieger-pvtl:
During revert, recover the original source cluster by rsync'ing the
    untouched source cluster's standby and mirrors to the master and
    primaries.  This has the effect of reverting any changes to the
    original source cluster that may have occurred as a result of running
    gpupgrade execute or via actions of the user for testing the target
    cluster.

    We also perform this rsync for copy mode.  This overcomes a design flaw
    that would otherwise cause the source cluster mirrors to not come up.


hub/revert.go
```diff
@@ -34,6 +35,10 @@ func (s *Server) Revert(_ *idl.RevertRequest, stream idl.CliToHub_RevertServer)
 		}
 	}()
 
+	if !s.Source.HasAllMirrorsAndStandby() {
```
kalensk, line 38: I think we should add `s.UseLinkMode` to this check.

dkrieger-pvtl: Let's leave this for a follow up PR.  I'll add this to the main story.   There are other use cases that this current unqualified check protects against: a source cluster with only some mirrors, with no standby, etc.  That will require some head scratching to make sure it should be ok and some tests to validate.

test/revert.bats
```diff
@@ -70,10 +83,48 @@ setup() {
...
 
+    # Check to make sure the new cluster matches the old one.
+    local new_config=$(get_segment_configuration "${GPHOME_SOURCE}")
+    [ "$old_config" = "$new_config" ] || fail "actual config: $new_config, wanted: $old_config"
+
     isready || fail "expected source cluster to be running on port ${PGPORT}"
     ! isready "${GPHOME_TARGET}" ${target_master_port} || fail "expected target cluster to not be running on port ${target_master_port}"
 }
 
+@test "reverting after execute in link mode succeeds" {
```
bhuvnesh2703, line 94: i would prefer having a common functions which runs it. the database validation can still remain the same, just that function should have a if based code to validate the difference in rsync (link mode) (marker file copied from mirrors to primary) vs gprecoverseg (copy mode) (where marker file are not copied from mirrors to primaries)..
@kalensk has started the story to perform gprecoverseg, so i would leave on you and him to decide if you want to do it in this PR or the one with gprecoverseg functionality.

hub/restore_data_directories.go
```diff
@@ -0,0 +1,115 @@
...
+
+var Options = []string{"--archive", "--compress", "--stats"}
+
+var Excludes = []string{
+	"pg_hba.conf", "postmaster.opts", "postgresql.auto.conf", "internal.auto.conf",
+	"gp_dbid", "postgresql.conf", "backup_label.old", "postmaster.pid", "recovery.conf",
+}
+
+func RestoreMasterAndPrimaries(stream step.OutStreams, agentConns []*Connection, source *greenplum.Cluster) error {
+	if !source.HasAllMirrorsAndStandby() {
```
bhuvnesh2703, line 28: i think the caller should call this function when it has mirrors and standby? instead of calling it and failing. Also, that ensures that if there are no mirrors or standby we do whatever we can do in revert.

hub/revert.go
bhuvnesh2703: i would include link mode check here, and if we don't have mirrors or primaries, we do whatever we could do to cleanup and not call rsync and return error. 
Also, combining this check for mirrors and primaries also makes it tough for the case where you have mirrors but no standby.  For copy if there are no primaries and mirrors we should go do our best in cleanup.. 
i am open to do this in a followup story, but let's create that story and get it in soon.

agent/rsync.go
```diff
@@ -0,0 +1,64 @@
...
+	"github.com/greenplum-db/gp-common-go-libs/gplog"
+	"github.com/hashicorp/go-multierror"
+
+	"github.com/greenplum-db/gpupgrade/idl"
+	"github.com/greenplum-db/gpupgrade/upgrade"
+	"github.com/greenplum-db/gpupgrade/utils/rsync"
+)
+
+func (s *Server) Rsync(ctx context.Context, in *idl.RsyncRequest) (*idl.RsyncReply, error) {
+	gplog.Info("agent received request to rsync data directories from source mirrors to primaries")
```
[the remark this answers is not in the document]
dkrieger-pvtl: Good point...I am not aware of a good terminology for this that isn't awkward.  I could say "from source standby/mirror to master/primary", but I'd rather say something like "from source replica to primary" but I am not aware of a common name for this.

What do you recommend?

hub/revert.go
```diff
@@ -52,6 +57,25 @@ func (s *Server) Revert(_ *idl.RevertRequest, stream idl.CliToHub_RevertServer)
 		})
 	}
 
+	// Restoring the source master and primaries is only needed if upgrading the
+	// primaries had started.
+	// TODO: For now we use if the source master is not running to determine this.
+	running, err = s.Source.IsMasterRunning(st.Streams())
```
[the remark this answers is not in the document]
dkrieger-pvtl: The source cluster is stopped at the end of execute, so it should be stopped here.  We are only doing end-of-execute/end-of-initialize revert right now.  So technically this check is just to make sure we are at the end of execute and not the end of initialize.  

But, yes, follow up stories that allow revert before execute completes should check the substep status.

hub/restore_data_directories.go
```diff
@@ -0,0 +1,115 @@
...
+	"github.com/hashicorp/go-multierror"
+	"github.com/pkg/errors"
+
+	"github.com/greenplum-db/gpupgrade/greenplum"
+	"github.com/greenplum-db/gpupgrade/idl"
+	"github.com/greenplum-db/gpupgrade/step"
+	"github.com/greenplum-db/gpupgrade/utils/rsync"
+)
+
+var Options = []string{"--archive", "--compress", "--stats"}
```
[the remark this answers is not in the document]
dkrieger-pvtl: Yes, these are meant to be public as they are used in test packages(like agent_test in agent/rsync_test.go) as well.

I thought of explicitly inlining these strings in test and having these variables private, but didn't want to pin the implementation of the code to use these strings.  End-to-end testing should catch using the wrong options/excludes.

agent/rsync.go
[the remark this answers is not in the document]
bhuvnesh2703: `source to destination` is good enough if you are not putting in what directories?

test/revert.bats
dkrieger-pvtl: Quickly looking at these two tests(`reverting after execute in link mode succeeds` and `reverting after execute in copy mode succeeds`), it looks like combining the code might not be cleaner.   I'd prefer to defer to recoverseg, though, because of this.

hub/revert.go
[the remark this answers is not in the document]
bhuvnesh2703: ideally if this story was to do revert after execute completed, we should check the status from config.json first, and then double check if required using running, but sure, lets do in a followup story.

hub/restore_data_directories.go
dkrieger-pvtl: Yes, it is also called there(at the top if Revert())....it is there twice because they have different semantic purposes.  The one at the top of Revert() is right now a placeholder for the structure of clusters we allow for revert, which for the purposes of this PR is just clusters with standby+mirrors.  Subsequent PRs will relax this.  The one inside this function protects this function from trying to rsync a non-existent dir.

hub/revert.go
dkrieger-pvtl: This suggestion needs to await the copy-mode recoverseg PR since otherwise, in copy mode, we'll still end up rsync'ing non-existent standby/master dirs. We use rsync instead of recoverseg right now.

bhuvnesh2703:
LGTM, but there are some conflicting files.

test/revert.bats
kalensk: Yeah the upcoming recoverseg PR handles this.

PR #368, opened 2020-06-24
Don't print stack traces on errors 

kalensk:
Only print stack traces on panics, and simply return errors. Since stack traces for errors (rather than panics) are not very helpful and can overwhelm the user.

We also log errors to help with debugging. However, we do this at the debug level so that its not printed twice to stdout/stderr.

Also, only print usage for unknown flag errors.

[Test pipeline](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:errors)

### DIFF
--- a/cmd/gpupgrade/main.go
+++ b/cmd/gpupgrade/main.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"os"
 	"runtime/debug"
+	"strings"
 
 	"github.com/greenplum-db/gp-common-go-libs/gplog"
 	_ "github.com/lib/pq"
@@ -26,9 +27,17 @@ func main() {
 	gplog.InitializeLogging("gpupgrade_cli", logdir)
 
 	root := commands.BuildRootCommand()
+	// Silence usage since Cobra prints usage for all errors rather than just
+	// "unknown flag" errors.
+	root.SilenceUsage = true
 
 	err = root.Execute()
 	if err != nil && err != daemon.ErrSuccessfullyDaemonized {
+		if strings.HasPrefix(err.Error(), "unknown flag") {
+			cmd := os.Args[1]
+			fmt.Println(commands.Help[cmd])
+		}
+
 		// We use gplog.Debug instead of Error so the error is not displayed
 		// twice to the user in the terminal.
 		gplog.Debug("%+v", err)
--- a/cmd/gpupgrade/main.go
+++ b/cmd/gpupgrade/main.go
@@ -26,12 +26,12 @@ func main() {
 	gplog.InitializeLogging("gpupgrade_cli", logdir)
 
 	root := commands.BuildRootCommand()
-	root.SilenceErrors = true // we'll print these ourselves
 
 	err = root.Execute()
 	if err != nil && err != daemon.ErrSuccessfullyDaemonized {
-		// Use v to print the stack trace of an object errors.
-		fmt.Printf("\n%+v\n", err)
+		// We use gplog.Debug instead of Error so the error is not displayed
+		// twice to the user in the terminal.
+		gplog.Debug("%+v", err)
 		os.Exit(1)
 	}
 }
--- a/cli/commands/commands.go
+++ b/cli/commands/commands.go
@@ -57,6 +57,7 @@ import (
 )
 
 var (
+	Help           map[string]string
 	InitializeHelp string
 	ExecuteHelp    string
 	FinalizeHelp   string
@@ -102,6 +103,12 @@ func init() {
 		idl.Substep_ARCHIVE_LOG_DIRECTORIES,
 		idl.Substep_START_SOURCE_CLUSTER,
 	})
+	Help = map[string]string{
+		"initialize": InitializeHelp,
+		"execute":    ExecuteHelp,
+		"finalize":   FinalizeHelp,
+		"revert":     RevertHelp,
+	}
 }
 
 func BuildRootCommand() *cobra.Command {


cmd/gpupgrade/main.go
```diff
@@ -26,12 +27,18 @@ func main() {
...
-	root.SilenceErrors = true // we'll print these ourselves
+	// Silence usage since Cobra prints usage for all errors rather than just
+	// "unknown flag" errors.
+	root.SilenceUsage = true
 
 	err = root.Execute()
 	if err != nil && err != daemon.ErrSuccessfullyDaemonized {
-		// Use v to print the stack trace of an object errors.
-		fmt.Printf("\n%+v\n", err)
+		if strings.HasPrefix(err.Error(), "unknown flag") {
```
kalensk, line 36: This isn't ideal since cobra or really pflags may change, but so far I couldn't figure out a better solution. And printing the usage for every error _really_ distracts and makes it hard to figure out what went wrong.

cmd/gpupgrade/main.go
```diff
@@ -26,12 +27,18 @@ func main() {
...
+			cmd := os.Args[1]
+			fmt.Println(commands.Help[cmd])
+		}
+
+		gplog.Debug(err.Error())
```
dgkimura, line 41: How do others feel about logging the full stack trace?

kalensk: Upon some more reflection stack traces in the logs might be nice to speed up debugging. We could change it to: `gplog.Debug("%+v", err)`. 

Unfortunately using `gplog.Error` always prints to stderr, so we see the error twice, hence the use of `gplog.Debug`.

bhuvnesh2703: yes, i would expect the stack traces to be in log.

bhuvnesh2703:
I ran gpupgrade initialize, and in the log it reported the below
```
20200625:13:55:57 gpupgrade_cli:bchaudhary:Bhuvneshs-MBP.attlocal.net:004524-[ERROR]:-initialize: substep "CHECK_UPGRADE": 2 errors occurred:
        * exit status 1
        * upgrade primary segment on host Bhuvneshs-MBP.attlocal.net: rpc error: code = Unknown desc = upgrading primaries: 3 errors occurred:
        * check primary on host Bhuvneshs-MBP.attlocal.net with content 0: exit status 1
        * check primary on host Bhuvneshs-MBP.attlocal.net with content 2: exit status 1
        * check primary on host Bhuvneshs-MBP.attlocal.net with content 1: exit status 1



20200625:13:55:57 gpupgrade_cli:bchaudhary:Bhuvneshs-MBP.attlocal.net:004519-[DEBUG]:-initialize create cluster:
    github.com/greenplum-db/gpupgrade/cli/commands.initialize.func2
        /Users/bchaudhary/go/src/github.com/greenplum-db/gpupgrade/cli/commands/commands.go:476
  - InitializeCreateCluster:
    github.com/greenplum-db/gpupgrade/cli/commanders.InitializeCreateCluster
        /Users/bchaudhary/go/src/github.com/greenplum-db/gpupgrade/cli/commanders/steps.go:88
  - rpc error: code = Unknown desc = substep "CHECK_UPGRADE": 2 errors occurred:
        * exit status 1
        * upgrade primary segment on host Bhuvneshs-MBP.attlocal.net: rpc error: code = Unknown desc = upgrading primaries: 3 errors occurred:
        * check primary on host Bhuvneshs-MBP.attlocal.net with content 0: exit status 1
        * check primary on host Bhuvneshs-MBP.attlocal.net with content 2: exit status 1
        * check primary on host Bhuvneshs-MBP.attlocal.net with content 1: exit status 1
```

This part kind of is duplicated ? 
```
        * exit status 1
        * upgrade primary segment on host Bhuvneshs-MBP.attlocal.net: rpc error: code = Unknown desc = upgrading primaries: 3 errors occurred:
        * check primary on host Bhuvneshs-MBP.attlocal.net with content 0: exit status 1
        * check primary on host Bhuvneshs-MBP.attlocal.net with content 2: exit status 1
        * check primary on host Bhuvneshs-MBP.attlocal.net with content 1: exit status 1
```
Thoughts?

bhuvnesh2703:
LGTM, otherwise.

kalensk:
> Thoughts?

I am not sure yet. I am thinking that might be a side effect of gplog, and not sure how much we can realistically do right now. I'll poke some more.

kalensk:
Thanks for trying it out!

The double message is currently expected based on the current code. Specifically, the first error originates from [`s.CheckUpgrade(stream, conns)`](https://github.com/greenplum-db/gpupgrade/blob/da291d6a519a750fd9268b20ac48553d97e7efcc/hub/initialize.go#L107) failing and a [`gplog.Error`](https://github.com/greenplum-db/gpupgrade/blob/da291d6a519a750fd9268b20ac48553d97e7efcc/hub/initialize.go#L72) is written in the `defer`. The second gets bubbled up all the way to main.go and written as [`gplog.Debug("%+v", err)`](https://github.com/greenplum-db/gpupgrade/pull/368/files#).

cmd/gpupgrade/main.go
dkrieger-pvtl: Another option would be to have a list of commands that are allowed:
```go
allowed := []strings{"initialize","execute","finalize",...}
...
var goodCmd
for _, allow := range allowed {
    if strings.Contains(err.Error,allow) {
         goodCmd = true
   }
}

if !goodCmd {
   //print help message
}
```

At least then you are coupling to something we have more control over.  But then we have to maintain that list.

I don't see a strong preference either way.

kalensk:
> Another option would be to have a list of commands that are allowed:

If I am understanding correctly, that would still print the usage for all errors rather than only for "unknown flag" errors, since its not checking what type of error is returned, rather than the command is a known one. Errors can be returned when running a syntactically valid initialize command.

PR #369, opened 2020-06-25
Add initial infrastructure and scripts to complement pg_upgrade check failures

bhuvnesh2703:
This commit adds the following
1. SQL statements to generate DDL to modify Database objects which
   cannot be upgraded
2. Wrapper scripts to generate and execute the sql scripts.
3. Bats test to test if the scripts fixed the issues identified by
   pg_ugprade --checks

### DIFF
--- a/migration_scripts/pre-upgrade/execute_preupgrade_sql.bash
+++ b/migration_scripts/pre-upgrade/execute_preupgrade_sql.bash
@@ -1,0 +1,37 @@
+#!/usr/bin/env bash
+#
+# Copyright (c) 2017-2020 VMware, Inc. or its affiliates
+# SPDX-License-Identifier: Apache-2.0
+
+if [ "$#" -ne 3 ]; then
+    echo "Illegal number of parameters"
+    echo "Usage: $(basename $0) <GPHOME> <PGPORT> <INPUT_DIR>"
+    exit 1
+fi
+
+GPHOME=$1
+PGPORT=$2
+INPUT_DIR=$3
+
+main(){
+    local log_file="${INPUT_DIR}/migration_sql.log"
+
+    rm -f "$log_file"
+
+    cmd="find ${INPUT_DIR} -type f -name *.sql"
+    local files="$(eval "$cmd")"
+    if [ -z "$files" ]; then
+        echo "Executing command \"${cmd}\" returned no sql files. Exiting!" | tee -a "$log_file"
+        exit 1
+    fi
+
+    for file in ${files[*]}; do
+        local cmd="${GPHOME}/bin/psql -d postgres -p ${PGPORT} -f ${file}"
+        echo "Executing command: ${cmd}" | tee -a "$log_file"
+        ${cmd} | tee -a "$log_file"
+    done
+
+    echo "Check log file for execution details: $log_file"
+}
+
+main
--- a/migration_scripts/pre-upgrade/gen_alter_gphdfs_roles.sql
+++ b/migration_scripts/pre-upgrade/gen_alter_gphdfs_roles.sql
@@ -1,0 +1,11 @@
+-- Copyright (c) 2017-2020 VMware, Inc. or its affiliates
+-- SPDX-License-Identifier: Apache-2.0
+
+-- generates a sql script to drop gphdfs roles in the cluster
+SELECT 'ALTER ROLE '|| rolname || $$ NOCREATEEXTTABLE(protocol='gphdfs',type='readable'); $$
+FROM pg_roles
+WHERE rolcreaterexthdfs='t'
+UNION ALL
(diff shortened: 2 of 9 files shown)


test/check_fixups.bats
```diff
@@ -0,0 +1,86 @@
...
+        delete_finalized_cluster $GPHOME_TARGET $NEW_CLUSTER
+    fi
+
+    gpupgrade kill-services
+
+    restore_cluster
+
+    # reload old path and start
+    source "${GPHOME_SOURCE}/greenplum_path.sh"
+    gpstart -a
```
kalensk, line 40: Can we use `start_source_cluster` from helpers (the default args should be fine)?

test/check_fixups.bats
```diff
@@ -0,0 +1,86 @@
...
+    export GPUPGRADE_HOME="${STATE_DIR}/gpupgrade"
+
+    gpupgrade kill-services
+}
+
+teardown() {
+    # XXX Beware, BATS_TEST_SKIPPED is not a documented export.
+    if [ -n "${BATS_TEST_SKIPPED}" ]; then
+        return
+    fi
```
kalensk, line 28: I don't think this is needed (?), and should be replaced with `skip_if_no_gpdb`

test/check_fixups.bats
```diff
@@ -0,0 +1,86 @@
...
+    gpstart -a
+
+    $GPHOME_SOURCE/bin/psql -c "DROP ROLE IF EXISTS ${GPHDFS_USER}" -d template1
+    $GPHOME_SOURCE/bin/psql -c "DROP DATABASE IF EXISTS ${TEST_DBNAME}" -d template1
+}
+
+@test "preupgrade scripts should be able to generate sqls to modify non upgradeable objects and fix the pg_upgrade check errors" {
+    setup_restore_cluster "--mode=copy"
+
+    $PSQL -c "CREATE DATABASE $TEST_DBNAME;" -d template1
```
kalensk, line 49: Any reason to not connect to the `postgres` database instead of `template1` (similarly for all other calls)? Seems safer and more standard perhaps.

test/check_fixups.bats
```diff
@@ -0,0 +1,86 @@
...
+
+    run gpupgrade initialize \
+            --source-bindir="$GPHOME_SOURCE/bin" \
+            --target-bindir="$GPHOME_TARGET/bin" \
+            --source-master-port="${PGPORT}" \
+            --temp-port-range 6020-6040 \
+            --disk-free-ratio 0 \
+            --verbose
+
+    [ "$status" -ne 0 ] || fail "initialize to fail due to pg_upgrade checks: $output"
```
kalensk, line 60: would something like the following be more clear? `expected initialize to fail due to pg_upgrade check: $output`?

pgupgrade_helper_scripts/test/create_nonupgradable_objects.sql
```diff
@@ -0,0 +1,38 @@
+-- Copyright (c) 2017-2020 VMware, Inc. or its affiliates
+-- SPDX-License-Identifier: Apache-2.0
+DROP TABLE IF EXISTS regular CASCADE;
```
kalensk, line 3: super minor nit: A newline between the license and start of the script might make it a little easier to read.

pgupgrade_helper_scripts/preupgrade/gen_drop_constraint_step_1.sql
```diff
@@ -0,0 +1,16 @@
+-- Copyright (c) 2017-2020 VMware, Inc. or its affiliates
+-- SPDX-License-Identifier: Apache-2.0
+
+-- Step 1 of 3
```
kalensk, line 4: Might be out of scope for this PR, but have we tried simply consolidating all three drop constraint scripts into a single file such that there is no need for all three steps?

test/check_fixups.bats
bhuvnesh2703: done.

test/check_fixups.bats
bhuvnesh2703: sure.

pgupgrade_helper_scripts/preupgrade/gen_drop_constraint_step_1.sql
bhuvnesh2703: yes, i thought about it, and we might have to improve the queries as well as i have to exclude some cases which requires better queries.. So, i will file a story, and let's wrap it then.

test/check_fixups.bats
bhuvnesh2703: nothing specific, but template1 is also used generally and its the default database while you create other database. I use it being used in other utilities script as well. But, i would change it to postgres, no difference per say.

PR #371, opened 2020-06-26
revert: use gprecoverseg in copy mode

kalensk:
For performance on large clusters use gprecoverseg instead of rsync when
reverting in copy mode.

Restoring the mirrors is needed in copy mode on 5X since the source cluster is left in a bad state after execute. This is because running pg_upgrade on a primary results in a checkpoint that does not get replicated on the mirror. Thus, when the mirror is started it panics and a gprecoverseg or rsync is needed.

[Test pipeline](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:revert-gprecoverseg)


hub/restore_data_directories.go
```diff
@@ -51,6 +58,27 @@ func RestoreMasterAndPrimaries(stream step.OutStreams, agentConns []*Connection,
 	return mErr.ErrorOrNil()
 }
 
+// Restoring the mirrors is needed in copy mode on 5X since the source cluster
+// is left in a bad state after execute. This is because running pg_upgrade on
+// a primary results in a checkpoint that does not get replicated on the mirror.
+// Thus, when the mirror is started it panics and a gprecoverseg or rsync is needed.
+func RestoreMasterInCopyMode(stream step.OutStreams, cluster *greenplum.Cluster) error {
```
bhuvnesh2703, line 65: only the mirrors require gprecoverseg, standby master cannot be recovered using gprecoverseg -a. if standby master also need to brought back, gpinitstandby is required.

kalensk: Good catch. I renamed to `RestoreMirrors`.

hub/revert.go
```diff
@@ -118,12 +114,34 @@ func (s *Server) Revert(_ *idl.RevertRequest, stream idl.CliToHub_RevertServer)
...
+			if err != nil {
 				return xerrors.Errorf("starting source cluster: %w", err)
 			}
+
 			return nil
 		})
 	}
 
+	if !s.UseLinkMode {
+		st.Run(idl.Substep_RESTORE_MIRRORS, func(streams step.OutStreams) error {
```
bhuvnesh2703, line 141: what will happen to the master now, do we need another story for recovering master?

hub/restore_data_directories_test.go
```diff
@@ -74,6 +81,48 @@ func TestRestoreMasterAndPrimaries(t *testing.T) {
 		}
 	})
 
+	t.Run("restores the master in copy mode on GPDB5", func(t *testing.T) {
```
bhuvnesh2703, line 84: restore mirrors..

hub/restore_data_directories_test.go
```diff
@@ -74,6 +81,48 @@ func TestRestoreMasterAndPrimaries(t *testing.T) {
...
+			}
+		})
+
+		err := hub.RestoreMirrors(&testutils.DevNullWithClose{}, cluster)
+		if err != nil {
+			t.Errorf("unexpected err %#v", err)
+		}
+	})
+
+	t.Run("does not restore the master in copy mode on GPDB6 or higher", func(t *testing.T) {
```
bhuvnesh2703, line 103: mirrors

hub/restore_data_directories_test.go
```diff
@@ -144,6 +193,17 @@ func TestRestoreMasterAndPrimaries(t *testing.T) {
 		}
 	})
 
+	t.Run("errors when restoring the master fails in copy mode on GPDB5", func(t *testing.T) {
```
bhuvnesh2703, line 196: sorry, but why do we restore mirrors when master fails? if the upgrading primary step has failed then we should restore the mirros..

hub/restore_data_directories_test.go
kalensk: Updated.

hub/restore_data_directories_test.go
kalensk: Updated.

hub/restore_data_directories_test.go
kalensk: The test name just need to be renamed to mirrors like all the others. Its just checking that errors get returned when gprecoverseg fails.

hub/revert.go
kalensk: We do not need to recover the standby as explained below:

> This only exists on 5X primaries, due to the use of filerep for primary-standby replication.  On 5X, the master-standby uses WAL rep, and on 6X, all content pairs use Wal rep.  In summary, this issue is intrinsic to running a single-node tool on a GPDB cluster running on 5X.

See [pg_upgrade - mirror invalid checkpoints issue](https://docs.google.com/document/d/1lG1S-aCOpzxPl90-XN-sQ5Z92UZkmkC-5R5QNMvqTQA/edit#) Doc (Problem Statement section).

We are testing with a demo cluster that has a standby that are exercising this situation.

hub/restore_data_directories.go
dkrieger-pvtl: Right.  To my understanding, the standby doesn't need any special treatment since in 5X, the master-standby use walrep; the primary-mirror uses filerep which creates the problem.

dkrieger-pvtl:
The second commit message has a typo: `bats: combine revert after execite tests`...s/execite/execute.

cli/commanders/steps.go
```diff
@@ -54,6 +54,7 @@ var SubstepDescriptions = map[idl.Substep]substepText{
 	idl.Substep_ARCHIVE_LOG_DIRECTORIES:                  substepText{"Archiving log directories...", "Archive log directories"},
 	idl.Substep_RESTORE_SOURCE_MASTER_AND_PRIMARIES:      substepText{"Restoring source cluster data directories...", "Restore source cluster data directories"},
 	idl.Substep_START_SOURCE_CLUSTER:                     substepText{"Starting source cluster...", "Start source cluster"},
+	idl.Substep_RESTORE_MIRRORS:                          substepText{"Restoring source cluster mirrors...", "Restore source cluster mirrors"},
```
dkrieger-pvtl, line 57: I prefer the link-mode and the copy-mode revert-specific steps(that is, both `RESTORE_SOURCE_MASTER_AND_PRIMARIES` and `RESTORE_MIRRORS`) to have the same substep name. Maybe `RESTORE_SOURCE` or something similar.  We do different behaviors for copy and link mode but, from the standpoint of revert, they do the same semantic step.

hub/restore_data_directories.go
```diff
@@ -51,6 +58,27 @@ func RestoreMasterAndPrimaries(stream step.OutStreams, agentConns []*Connection,
 	return mErr.ErrorOrNil()
 }
 
+// Restoring the mirrors is needed in copy mode on 5X since the source cluster
+// is left in a bad state after execute. This is because running pg_upgrade on
+// a primary results in a checkpoint that does not get replicated on the mirror.
+// Thus, when the mirror is started it panics and a gprecoverseg or rsync is needed.
+func RestoreMirrors(stream step.OutStreams, cluster *greenplum.Cluster) error {
```
dkrieger-pvtl, line 65: As with the substep names, perhaps we want to have a single function that does recovery of the standby/mirrors.  This would be called in link mode where it currently is and in copy mode as in this PR.  Once we separate out the printing of a substep from its running, we could then eliminate the linkMode checks in Revert().

```go
func RestoreStandbyAndMirrors(stream step.OutStreams, cluster *greenplum.Cluster, agentConns []*Connection) error {
           if s.UseLinkMode {
                  return RestoreMasterAndPrimaries(stream, agentConns, cluster)
           }
           return RestoreMirrors(stream, cluster)
}
```

We'd probably want to rename `RestoreMasterandPrimaries` to something like `RsyncMasterAndPrimaries` and `RestoreMirrors` to `RecoversegMirrors`.  The intent here is to separate out the logical step from the action.  The current code risks having its intent a bit hard to follow.

hub/restore_data_directories.go
```diff
@@ -51,6 +58,27 @@ func RestoreMasterAndPrimaries(stream step.OutStreams, agentConns []*Connection,
...
+	if cluster.Version.AtLeast("6") {
+		return nil
+	}
+
+	gphome := filepath.Dir(path.Clean(cluster.BinDir)) //works around https://github.com/golang/go/issues/4837 in go10.4
```
dkrieger-pvtl, line 70: Is this still an issue in golang 1.14?

hub/restore_data_directories_test.go
```diff
@@ -74,6 +81,48 @@ func TestRestoreMasterAndPrimaries(t *testing.T) {
...
 	})
 
+	t.Run("restores mirrors in copy mode on GPDB5", func(t *testing.T) {
+		defer ResetRecoversegCmd()
+		hub.RecoversegCmd = exectest.NewCommandWithVerifier(hub.Success, func(utility string, args ...string) {
+			if utility != "bash" {
+				t.Errorf("got %q want bash", utility)
+			}
+
+			expected := []string{"-c", "source /source/greenplum_path.sh && /source/bin/gprecoverseg -a"}
```
dkrieger-pvtl, line 91: The source cluster path is `/source/bindir`, which is modified to `/source/bin` in the `RestoreMirrors` call.  Do you want to be pinning this behavior here in the test?

hub/revert.go
```diff
@@ -118,12 +114,34 @@ func (s *Server) Revert(_ *idl.RevertRequest, stream idl.CliToHub_RevertServer)
 
 	if !running {
 		st.Run(idl.Substep_START_SOURCE_CLUSTER, func(streams step.OutStreams) error {
-			if err := s.Source.Start(streams); err != nil {
+			err := s.Source.Start(streams)
+			var exitErr *exec.ExitError
+			if xerrors.As(err, &exitErr) {
+				// Since the cluster is in a "bad state" (see comment above
```
dkrieger-pvtl, line 120: This is only true in copy-mode for a 5X source cluster.  Perhaps a more accurate phrasing:

```
// In copy mode for a 5X source cluster, the source mirrors will fail to come up here.  We fix
//. this below in Substep_RESTORE_MIRRORS.  Ideally, we would check for this specific case,
//. but for now, use this weaker check.
// TODO: only allow a 5X cluster with down mirrors to proceed; otherwise error.  
```

test/revert.bats
```diff
@@ -61,38 +68,19 @@ teardown() {
...
-
-    # Check to make sure the new cluster matches the old one.
-    local new_config=$(get_segment_configuration "${GPHOME_SOURCE}")
-    [ "$old_config" = "$new_config" ] || fail "actual config: $new_config, wanted: $old_config"
-
-    isready || fail "expected source cluster to be running on port ${PGPORT}"
-    ! isready "${GPHOME_TARGET}" ${target_master_port} || fail "expected target cluster to not be running on port ${target_master_port}"
-}
+    # Save segment configuration
+    local old_config=$(get_segment_configuration "${GPHOME_SOURCE}")
```
dkrieger-pvtl, line 76: Please use the idiom:
```
local old_config
old_config=$(get_segment_configuration "${GPHOME_SOURCE}")
```

See https://github.com/koalaman/shellcheck/wiki/SC2155.

There are 5 places in this function that can use this fix.  The declarations could all be at the top of the function to make in cleaner.

test/revert.bats
```diff
@@ -61,38 +68,19 @@ teardown() {
...
 
-@test "reverting after execute in link mode succeeds" {
-    local target_master_port=6020
+    # Place marker files on mirrors
+    MARKER=source-cluster.MARKER
```
dkrieger-pvtl, line 79: This is a clever idea, but you are really pinning implementation here in a end-to-end test.  That is, that link-mode does an rsync and that copy-mode does not.  Is this really necessary?  Isn't recovery of the source cluster sufficient here?

I think a more appropriate check would be to validate that all segments(standby+mirrors) are up and in their preferred roles.  Note then that the only use of `mode` in this function would be in the call to `initiailze`, which is more appropriate in a end-to-end test.

bhuvnesh2703: @dkrieger-pvtl Unfortunately, we don't test the flow of revert.go, so just checking the state of the cluster will not tell if the flow was correct (rsync vs gprecoverseg). So, with the current state i would perfer this marker test file stuff. Or if you think there is a better way to check if rsync or gprecoverseg was called in the revert flow?

cli/commanders/steps.go
dkrieger-pvtl: substeps renamed to the same one.

hub/restore_data_directories.go
dkrieger-pvtl: This isn't so clean....instead, we renamed the underlying functions to reference Rsync or Recoverseg explicitly.

hub/revert.go
dkrieger-pvtl: rephrased.

test/revert.bats
dkrieger-pvtl: Right, I see the point.  And we are very unlikely to modify our use or rsync/recoverseg for recovery, so in practice pinning this implementation won't matter.  Right now, I cannot see a better way test the innards or revert, initialize, execute or finalize other than our bats end-to-end tests.

hub/restore_source_cluster.go
```diff
@@ -51,7 +58,28 @@ func RestoreMasterAndPrimaries(stream step.OutStreams, agentConns []*Connection,
...
 
-func RestoreMaster(stream step.OutStreams, standby greenplum.SegConfig, master greenplum.SegConfig) error {
+// Restoring the mirrors is needed in copy mode on 5X since the source cluster
+// is left in a bad state after execute. This is because running pg_upgrade on
+// a primary results in a checkpoint that does not get replicated on the mirror.
+// Thus, when the mirror is started it panics and a gprecoverseg or rsync is needed.
+func Recoverseg(stream step.OutStreams, cluster *greenplum.Cluster) error {
+	if cluster.Version.AtLeast("6") {
+		return nil
+	}
```
jchampio, line 68: IMO, `Recoverseg()` should perform recovery when it's called. Business logic as to _when_ we call that belongs in higher layers of code.

hub/revert.go
```diff
@@ -118,12 +114,33 @@ func (s *Server) Revert(_ *idl.RevertRequest, stream idl.CliToHub_RevertServer)
...
+
+			if err != nil {
 				return xerrors.Errorf("starting source cluster: %w", err)
 			}
+
 			return nil
 		})
 	}
 
+	if !s.UseLinkMode {
```
jchampio, line 139: E.g. I think the GPDB 6+ check belongs here; we have no reason to run a "restore source cluster" substep otherwise.

PR #374, opened 2020-06-30
revert: improve the output UI

dkrieger-pvtl:
Upon successful completion of revert, print out the original source version and the log archive directory.

This necessitated a few steps:
- pre-factor to rename upgrade.ID to unique.ID to avoid import cycles
- add a new datamap returned from Revert() to the UI to obtain the new data
- only print out cli-side Revert() messages in verbose mode

In addition, testing was adding for the similar function that obtains similar data at the end of execute and revert.


cli/commanders/steps.go
```diff
@@ -198,13 +199,45 @@ func extractTargetClusterInfo(dataMap map[string]string) (string, string, error)
...
 	if len(missingKeys) > 0 {
 		return "", "", xerrors.Errorf("did not receive the expected configuration values: %s", strings.Join(missingKeys, ", "))
 	}
 
 	return port, datadir, nil
 }
 
+func ExtractArchiveInfo(dataMap map[string]string) (string, string, error) {
+	version, ok := dataMap[idl.ResponseKey_source_version.String()]
+	if !ok {
```
dkrieger-pvtl, line 220: add a unit test for this case

cli/commanders/steps.go
```diff
@@ -198,13 +199,45 @@ func extractTargetClusterInfo(dataMap map[string]string) (string, string, error)
...
+		return "", "", xerrors.Errorf("no source version")
+	}
+
+	archiveDir, ok := dataMap[idl.ResponseKey_revert_log_archive_directory.String()]
+	if !ok {
```
dkrieger-pvtl, line 225: add a unit test for this case

cli/commanders/steps.go
```diff
@@ -198,13 +199,45 @@ func extractTargetClusterInfo(dataMap map[string]string) (string, string, error)
...
+	// make sure the returned datadir is actually a directory
+	f, err := os.Stat(datadir)
+	if err != nil {
+		return "", "", xerrors.Errorf("bad returned master data directory: %w", err)
+	}
+	if !f.IsDir() {
+		return "", "", xerrors.Errorf("returned master data directory is not a directory: %s", datadir)
+	}
+
 	if len(missingKeys) > 0 {
```
dkrieger-pvtl, line 211: add a unit test for this case

kalensk:
I may be missing something, but wouldn't it make more sense and be clearer to return a gRPC response struct which bundles all the info we need and pass that up to the CLI to extract the desired fields. Ideally a different response for each step - initialize, execute, finalize, or revert.

dkrieger-pvtl:
> It seems like the current approach is showing its drawbacks and limitations and indicating maybe a more robust solution is needed?

Yeah, I realized that as soon as I filed the PR.  The current approach has 1 way to do it for finalize/execute, but another for revert.  I'll try an approach where execute/finalize/revert can push the fields that they want, and then the CLI can pull the fields it needs.

dkrieger-pvtl:
I'll close this until I find something cleaner.

PR #375, opened 2020-07-01
upgrade: eliminate all `--` in upgradeID generated ids

dkrieger-pvtl:
GPDB has a bug that doesn't handle "--" in directory names in some
cases.  But gpupgrade generates a base-64 encoded string that goes into our target cluster's data directory names during execute, that can contain `--`.  It is easier to fix here right now, than in GPDB.  

We have two options:
1). replace `--` in a generated ID with some string, like `dd`.  This can work, and doesn't materially increase the chance of a collision.  The problem is that the String() and the underlying ID bytes are then not related by the simple base-64 scheme any more.

2). keep generating an ID until it has no `--`.  This works as long as there is enough entropy in a users random number generator.  I think it is pathological if not.  

3). I suppose a third option is to replace any `--` in the string and re-encode the value.

I choose option 2).

### DIFF
--- a/upgrade/id.go
+++ b/upgrade/id.go
@@ -8,6 +8,7 @@ import (
 	"encoding/base64"
 	"encoding/binary"
 	"fmt"
+	"strings"
 )
 
 // ID is a unique identifier for a cluster upgrade.
@@ -18,17 +19,26 @@ type ID uint64
 func NewID() ID {
 	var bytes [8]byte // 64 bits
 
-	// Use crypto/rand for this to avoid chicken-and-egg (i.e. what should we
-	// seed math/rand with?). This is more expensive, but we expect this to be
-	// called only once per upgrade anyway.
-	_, err := rand.Read(bytes[:])
-	if err != nil {
-		// TODO: should we fall back in this case? It will be system-dependent.
-		panic(fmt.Sprintf("unable to get random data: %+v", err))
-	}
+	for {
+		// Use crypto/rand for this to avoid chicken-and-egg (i.e. what should we
+		// seed math/rand with?). This is more expensive, but we expect this to be
+		// called only once per upgrade anyway.
+		_, err := rand.Read(bytes[:])
+		if err != nil {
+			// TODO: should we fall back in this case? It will be system-dependent.
+			panic(fmt.Sprintf("unable to get random data: %+v", err))
+		}
 
-	num := binary.LittleEndian.Uint64(bytes[:])
-	return ID(num)
+		num := binary.LittleEndian.Uint64(bytes[:])
+
+		// gpstart has a bug that doesn't handle "--" in directory names.
+		// Until that is resolved, we need to generate IDs without "--".
+		if strings.Contains(ID(num).String(), "--") {
+			continue
+		}
+
+		return ID(num)
+	}
 }
 
 // String returns an unpadded, filesystem-safe base64 encoding of the
--- a/upgrade/id_test.go
+++ b/upgrade/id_test.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"strings"
 	"testing"
 
 	"github.com/greenplum-db/gpupgrade/upgrade"
@@ -70,4 +71,16 @@ func TestIDCommand(_ *testing.T) {
 
 	fmt.Printf("%d", upgrade.NewID())
 	os.Exit(0)
+}
+
+// Make sure we are filtering out "--".  Empirical testing shows about 300 iterations
+//  are required to hit a "--", so we choose 10000 to ensure we'd catch an erroneous
+//  implementation.  This test takes 10ms on my laptop.
+func TestNoDoubleDash(t *testing.T) {
+	for i := 0; i < 10000; i++ {
+		id := upgrade.NewID()
+		if strings.Contains(id.String(), "--") {
+			t.Fatalf("id %s contains --", id)
+		}
+	}
 }


kalensk:
This _may_ be enough, but part of me would feel more confident at generating a new ID, checking for `--`, and if so generate a new ID and check again, etc. It would be great to hear some additional input from others on the overall approach.

dkrieger-pvtl:
> This _may_ be enough, but part of me would feel more confident at generating a new ID, checking for `--`, and if so generate a new ID and check again, etc. It would be great to hear some additional input from others on the overall approach.

This is a good idea; I'm not sure I have a particular preference either way, as we don't need to reverse the mapping from string to ID.  But this idea does keep an ID cogent.

dkrieger-pvtl:
@kalensk  I've implemented two different strategies that are viable. Preferences?

jchampio:
I think my vote is for option 2.

upgrade/id_test.go
```diff
@@ -71,3 +72,15 @@ func TestIDCommand(_ *testing.T) {
 	fmt.Printf("%d", upgrade.NewID())
 	os.Exit(0)
 }
+
+// Make sure we are filtering out "--".  Empirical testing shows about 300 iterations
+//  are required to hit a "--", so we choose 10000 to ensure we'd catch an erroneous
+//  implementation.  This test takes 10ms on my laptop.
+func TestNoDoubleDash(t *testing.T) {
```
bhuvnesh2703, line 79: is it possible to mock the output of num := binary.LittleEndian.Uint64(bytes[:]) and have a -- to check we never come to continue and a case where we come to continue instead of this loop?
```
		if strings.Contains(ID(num).String(), "--") {
			continue
		}
```

upgrade/id.go
```diff
@@ -18,17 +19,26 @@ type ID uint64
...
+		if err != nil {
+			// TODO: should we fall back in this case? It will be system-dependent.
+			panic(fmt.Sprintf("unable to get random data: %+v", err))
+		}
+
+		num := binary.LittleEndian.Uint64(bytes[:])
+
+		// gpstart has a bug that doesn't handle "--" in directory names.
+		// Until that is resolved, we need to generate IDs without "--".
+		if strings.Contains(ID(num).String(), "--") {
```
kalensk, line 36: Thanks for the comment. I like it.

upgrade/id_test.go
kalensk: I think spending some time now and seeing if this is doable would very worthwhile, especially in the long term. 

(Or if there is some other way to test. I'll see if I can think of any other ideas.)

dkrieger-pvtl: The test already does this, statistically.  The probability of the NewID() returning with a string representation with "--" is about (1/64)*(1/64)*(18) or about once every 225 generations.  So in a loop of 10,000, we should hit this 50 times or so.  So, we get a "--" 50 times and no "--" about 9950 times.

If you look at the code coverage running this test, you see that the continue is hit.

If you experiment with the implementation and remove the "--" check, you can see the statistics and know the path coverage will give you what you are requesting logically.

That's why I didn't mock out any part of NewID().   As further evidence, you know that the NewID() loop terminates as, if it didn't, `gpupgrade initialize` would hang forever. 

So, the fact that `gpugprade initialize` doesn't hang and code coverage shows `continue` is hit shows that we do, in fact, cover both paths in this test.

kalensk: I less worried about the coverage, but test time and _guaranteeing_ we are testing what we want to be testing.

That is, finding a way that guarantees we get `--` not only hardens the test which ensures what we are testing, but also reduces time. I am advocating that a little time spent now hardening the test, will pay off dividends from here on out.

dkrieger-pvtl: > That is, finding a way that guarantees we get `--` not only hardens the test which ensures what we are testing, but also reduces time

The test takes about 8ms to run and always hits a "--" eventually.  It _does_ hit "--" about 0.3% of the time as the first hit, but the loop runs a second time in that case, so we always run both paths.  The entire `make unit` takes about 5000ms on my local machine.  At least, if the random number generator isn't broken, which we can assume to be true, these facts should remain true on any test hardware.

On my local machine:

`go test -run "TestNoDouble" -count 1000` takes 7 seconds, quite consistent over 3 runs

Modifying the implementation to detect another string, "xx", which is just as probable as "--":
```go
 func TestNoDoubleDash(t *testing.T) {
        for i := 0; i < 10000; i++ {
                id := upgrade.NewID()
-               if strings.Contains(id.String(), "--") {
-                       t.Fatalf("id %s contains --", id)
+               if strings.Contains(id.String(), "xx") {
+                       fmt.Println("xx:", i,"end")
+                       break
                }
        }
 }
```

go test -run "TestNoDouble" -count 1000 | grep  xx | wc -l

returns 1000 for three runs.  It hits "xx" about 3/1000 times.  

So:
- we always hit the "--" case 
- our test is "pretty fast"; we have 115 top level tests(`go test -list "Test" ./... | grep -v ok | grep Test | wc -l`), so we can afford to add 7ms to our test case.
- it is a reasonable assumption that any system running our tests takes about this much time.

The reason I like the current test is that it does not require modifying the implementation.  That is the ideal case.  Adding function-based dependency injection does not make things cleaner.

dkrieger-pvtl: I implemented the suggestion, but note it is more appropriate to inject the random byte generator, not the base64 conversion.  That way, we keep everything consistent(we aren't making up a new byte conversion, just the byte generator, which is how the ids are really generated).  See [id.go](https://github.com/dkrieger-pvtl/gpupgrade/blob/upgradeID_no_doubledash_2/upgrade/id.go) and [id_test.go](https://github.com/dkrieger-pvtl/gpupgrade/blob/upgradeID_no_doubledash_2/upgrade/id_test.go).

The only technical advantage I see is that we can now test the panic case(bad random number generator).  We likely don't really want to pin this behavior(we probably want some fallback, like generating 8 bytes from the system time), but with this function generation, we can at least test that we get the currently expected behavior.

dkrieger-pvtl: After team discussion, the "vote is split", so I am opting for the non-function injection solution.  It is cleaner, and the only technical advantage is that we can test the case of the random number generator failing.  This isn't likely to ever be hit, and the added complication of the implementation doesn't seem to justify the less clean code.

upgrade/id_test.go
```diff
@@ -71,3 +72,15 @@ func TestIDCommand(_ *testing.T) {
 	fmt.Printf("%d", upgrade.NewID())
 	os.Exit(0)
 }
+
+// Make sure we are filtering out "--".  Empirical testing shows about 300 iterations
```
jchampio, line 76: If I did my math right, the actual probability is 0.22%. Should be around 1 in 450 or so. The probability of not hitting one in 10000 iterations is 1 in 3.7 billion.

PR #377, opened 2020-07-01
UI: improve revert output

dkrieger-pvtl:
Upon successful completion of revert, print out the original source version and the log archive directory.

This necessitated a few steps:

- move `GetArchiveDirectoryName` to upgrade so that it can use the upgrade.ID; it naturally belongs there anyway.
- add a few more items to the dataMap returned from Revert() to the UI to obtain the new data
- only print out cli-side Revert() messages in verbose mode
- (bonus) make the output of execute consistent with the other steps by removing a newline. 

There was some discussion as to the correct approach, and the following were considered as well:
1). provide a package that makes storing/retrieving values from the datamap: https://github.com/dkrieger-pvtl/gpupgrade/tree/execute_revert_UI_datamap
2). add step-specific protobuf fields instead of a datamap: https://github.com/dkrieger-pvtl/gpupgrade/tree/execute_revert_UI_proto

Both of those were thought to be too heavyweight for this.


cli/commanders/steps.go
```diff
@@ -163,46 +157,26 @@ func Finalize(client idl.CliToHubClient, verbose bool) error {
 	return nil
 }
 
-func Revert(client idl.CliToHubClient, verbose bool) error {
+func Revert(client idl.CliToHubClient, verbose bool) (string, string, error) {
```
kalensk, line 160: I think creating a `RevertResponse` type to bundle these fields would be clearer.

hub/server.go
```diff
@@ -466,3 +467,15 @@ func MakeTargetClusterMessage(target *greenplum.Cluster) *idl.Message {
...
+	data := make(map[string]string)
+	data[idl.ResponseKey_source_version.String()] = source.Version.VersionString
+	data[idl.ResponseKey_revert_log_archive_directory.String()] = archiveDir
+
+	return &idl.Message{
+		Contents: &idl.Message_Response{
+			Response: &idl.Response{Data: data},
+		},
+	}
+}
```
kalensk, line 481: I just realized we do have a similar helper for execute and finalize.

I guess I am on the fence on if this helper is getting us much, and find the indirection not worth it. If we are extracting it for testability, I don't see the advantages over either not testing or using BATs.

Simply inline with the following (feel free to play with the message formatting) is clearer to me:
```go
data := map[string]string{
    idl.ResponseKey_source_version.String():               s.Source.Version.VersionString,
    idl.ResponseKey_revert_log_archive_directory.String(): archiveDir,
}

err = stream.Send(&idl.Message{Contents: &idl.Message_Response{
    Response: &idl.Response{Data: data}},
})
if err != nil {
    return xerrors.Errorf("send revert response message: %w", err)
}
```

test/revert.bats
```diff
@@ -62,8 +66,8 @@ teardown() {
         exit 1
     fi
 
-    # check that the archived log directory was created within the last 3 minutes
-    if [[ -z $(find "${HOME}/gpAdminLogs/gpupgrade-"* -type d -cmin -3) ]]; then
+    # check that the archived log directory corresponds to this tests upgradeID
+    if [[ -z $(find "${HOME}/gpAdminLogs/gpupgrade-${upgradeID}-"* -type d) ]]; then
```
kalensk, line 70: nit: move `-type d` to right after find for readability.

hub/revert.go
```diff
@@ -91,12 +93,13 @@ func (s *Server) Revert(_ *idl.RevertRequest, stream idl.CliToHub_RevertServer)
...
 			return err
 		}
-		newDir := filepath.Join(filepath.Dir(oldDir), utils.GetArchiveDirectoryName(time.Now()))
+		newDir := filepath.Join(filepath.Dir(oldDir), utils.GetArchiveDirectoryName(s.UpgradeID, time.Now()))
 		if err = utils.System.Rename(oldDir, newDir); err != nil {
 			if utils.System.IsNotExist(err) {
 				gplog.Debug("log directory %s not archived, possibly due to multi-host environment. %+v", newDir, err)
 			}
 		}
+		archiveDir = newDir
```
kalensk, line 102: Any reason why we can't just rename newDir above archiveDir?

hub/revert.go
```diff
@@ -21,6 +21,8 @@ import (
 )
 
 func (s *Server) Revert(_ *idl.RevertRequest, stream idl.CliToHub_RevertServer) (err error) {
+	var archiveDir string
```
kalensk, line 24: nit: I have a preference for putting this right above the step where it is initialized. I suppose I could see how others might prefer it at the very top like the old C style.

hub/server.go
dkrieger-pvtl: I did that at first but it looked a bit out of place within Execute(), Revert() and Finalize()...I'd like to keep that code "clean" and seeing the struct fields was a bit jarring in that context.  But I'm open to preferences.

hub/revert.go
dkrieger-pvtl: It reads cleaner to me to have a separate variable, so it's clear that `archiveDir` is in scope for the response but `newDir` is only local to the code block.  Otherwise, when reading the code I have to look all over to see where else `archiveDir` is modified.

hub/revert.go
dkrieger-pvtl: I'm open...somehow to my eye its cleaner to see `archiveDir` at the top and then used at the bottom.  That was one motivation for my (abandoned) attempts to use a function setter to populate response.

dkrieger-pvtl:
> I find the `upgradeID` work very confusing based on the overall goal of this PR since it seems unrelated? Or perhaps can you explain more about this line `utils.GetArchiveDirectoryName(s.UpgradeID, time.Now())` as that might clear up some of my confusion?

The corresponding story specifies the UIs to print it after a successful revert:

`The gpupgrade logs can be found on the master and segment hosts in:
~/gpAdminLogs/gpupgrade-<upgradeID>-<some timestamp>`

We could consider not having the upgrade ID, but it makes sense to me: the `<some timestamp>` is the time the archive was logged, which is essentially during the revert call.  Having the upgradeID gives us a clean logical way to identify which archived logs correspond to which gpugprade run(via the upgrade ID).  It is also useful for the bats test to check for the archiveDir(see [bats test](https://github.com/dkrieger-pvtl/gpupgrade/blob/execute_revert_UI/test/revert.bats#L70) modified in this PR).

In other words, without the upgradeID in the log archive directory, it seems harder to find the logs corresponding to an upgrade.  In fact, you could imagine removing the timestamp, but that seems useful if a user has a small number of gpupgrade runs to look at(which is true for our customers).

cli/commanders/steps.go
dkrieger-pvtl: I'm on the fence; it's pretty easy to read the code as is to me.  I'll try it and see what it looks like.

hub/server.go
dkrieger-pvtl: Also, I think the below is our preferred formatting.  Even inlining it so that the map is initialized inside the protobuf looks like an eyeful to me.

```go
	data := map[string]string) {
	    data[idl.ResponseKey_target_port.String()] : strconv.Itoa(target.MasterPort())
	    data[idl.ResponseKey_target_master_data_directory.String()] }target.MasterDataDir()
        }  
	msg :=  &idl.Message{
		Contents: &idl.Message_Response{
			Response: &idl.Response{Data: data},
		},
	}

        err = stream.Send(msg)
        if err != nil {
            return xerrors.Errorf("send revert response message: %w", err)
        }
```

kalensk:
I guess I would like a bit more discussion on what exactly we would like the archive directory to be since `/Users/gpadmin/gpAdminLogs/gpupgrade-PmuBSyufs-A-2020-07-01T13:57` seems a bit much to me. Why have both the timestamp and ID when one of those identifiers should be enough? Perhaps just the upgradeID and using the systems directory timestamp in the metadata is enough?

hub/revert.go
kalensk: Hehe, for me its the exact opposite. I find that way less clear  :)

dkrieger-pvtl:
> I guess I would like a bit more discussion on what exactly we would like the archive directory to be since `/Users/gpadmin/gpAdminLogs/gpupgrade-PmuBSyufs-A-2020-07-01T13:57` seems a bit much to me. Why have both the timestamp and ID when one of those identifiers should be enough? Perhaps just the upgradeID and using the system's timestamp in the metadata is enough?

I think both are useful...upgradeID for easily logically connecting an upgrade run's logs...timestamp for searching for the approximate time of the revert run.  The filesystems creation/modification time are subject to change upon copying, transferring to other machines, etc.  So the timestamp in the filename seems good too.

jchampio:
> /Users/gpadmin/gpAdminLogs/gpupgrade-PmuBSyufs-A-2020-07-01T13:57

I have no problem with the above approach; it's verbose, but if that turns out to be a problem in practice then we can make it smaller fairly easily.

> The upgradeID is in the upgrade package, which imports other packages.

What's the exact cycle? I want to make sure we're breaking up the right package.

dkrieger-pvtl:
> > The upgradeID is in the upgrade package, which imports other packages.
> 
> What's the exact cycle? I want to make sure we're breaking up the right package.

With this change you get the below.  You see it's basically utils importing upgrade(for upgrade.ID) and then upgrade imports utils(or step).  The inclusion of step I think is from the generic streams definitions.  Those should also likely be moved to their own package, as their aren't dependent on any other of our packages.  But that doesn't need to be solved for this PR.

```
  |2.3.1| Davids-MacBook-Pro-2 in ~/go/src/github.com/greenplum-db/gpupgrade
±  |bats_add_tablespaces → dkrieger ↑1 {93} U:2 ?:3 ✗| → git diff
diff --git a/hub/revert.go b/hub/revert.go

-               newDir := filepath.Join(filepath.Dir(oldDir), utils.GetArchiveDirectoryName(time.Now()))
+               newDir := filepath.Join(filepath.Dir(oldDir), utils.GetArchiveDirectoryName(s.UpgradeID, time.Now()))

diff --git a/utils/sys_utils.go b/utils/sys_utils.go

-func GetArchiveDirectoryName(t time.Time) string {
-       return t.Format("gpupgrade-2006-01-02T15:04")
+func GetArchiveDirectoryName(id upgrade.ID, t time.Time) string {
+       return fmt.Sprintf("x-%s-%s", id.String(), t.Format("gpupgrade-2006-01-02T15:04")
 }

```

```
can't load package: import cycle not allowed
package github.com/greenplum-db/gpupgrade/agent
	imports github.com/greenplum-db/gpupgrade/greenplum
	imports github.com/greenplum-db/gpupgrade/step
	imports github.com/greenplum-db/gpupgrade/utils
	imports github.com/greenplum-db/gpupgrade/upgrade
	imports github.com/greenplum-db/gpupgrade/step
can't load package: import cycle not allowed
package github.com/greenplum-db/gpupgrade/agent
	imports github.com/greenplum-db/gpupgrade/greenplum
	imports github.com/greenplum-db/gpupgrade/step
	imports github.com/greenplum-db/gpupgrade/utils
	imports github.com/greenplum-db/gpupgrade/upgrade
	imports github.com/greenplum-db/gpupgrade/utils
go build -o gpupgrade -gcflags="all=-N -l" -ldflags "-X github.com/greenplum-db/gpupgrade/cli/commands.UpgradeVersion=0.3.0+dev.152.g4ebacc11" github.com/greenplum-db/gpupgrade/cmd/gpupgrade
import cycle not allowed
package github.com/greenplum-db/gpupgrade/cmd/gpupgrade
	imports github.com/greenplum-db/gpupgrade/cli/commands
	imports github.com/greenplum-db/gpupgrade/agent
	imports github.com/greenplum-db/gpupgrade/greenplum
	imports github.com/greenplum-db/gpupgrade/step
	imports github.com/greenplum-db/gpupgrade/utils
	imports github.com/greenplum-db/gpupgrade/upgrade
	imports github.com/greenplum-db/gpupgrade/step
import cycle not allowed
package github.com/greenplum-db/gpupgrade/cmd/gpupgrade
	imports github.com/greenplum-db/gpupgrade/cli/commands
	imports github.com/greenplum-db/gpupgrade/agent
	imports github.com/greenplum-db/gpupgrade/greenplum
	imports github.com/greenplum-db/gpupgrade/step
	imports github.com/greenplum-db/gpupgrade/utils
	imports github.com/greenplum-db/gpupgrade/upgrade
	imports github.com/greenplum-db/gpupgrade/utils
make: *** [build] Error 1
```

jchampio:
Why does `greenplum` depend on `step`, and `utils` depend on `upgrade` for that matter? There's no conceptual reason for that to be the case. And it looks like we recently tried to consolidate `OutStreams` into one place -- we had reasons for not doing that to begin with; this is one of them.

This is something I'm willing to put my foot down over, because significant work and discussion has been put into the motion towards better package organization, and it looks like we're starting to undo that work. Let's organize the packages so that their contents and dependencies make logical sense, rather than working around cycles with one-off, single-use packages.

dkrieger-pvtl:
> Why does `greenplum` depend on `step`, and `utils` depend on `upgrade` for that matter? There's no conceptual reason for that to be the case. 

`greenplum` uses `steps.OutStreams`.  That has been true for at least 4 months now.  `utils` depends on `upgrade` is new in this PR, and that's because the `GetArchiveDirectoryName` is in `utils/sys_utils.go`.  I agree that it likely should not be in there, and moving it to `upgrade/directories.go` is a better option; I'll move it there instead of modifying the package of upgrade.

> And it looks like we recently tried to consolidate `OutStreams` into one place -- we had reasons for not doing that to begin > with; this is one of them.

Yes, for sharing common OutStreams, we consolidated the streams in one place.  Otherwise, we end up creating one-off streams in many packages.  We can discuss.

> 
> This is something I'm willing to put my foot down over, because significant work and discussion has been put into the motion towards better package organization, and it looks like we're starting to undo that work. Let's organize the packages so that their contents and dependencies make logical sense, rather than working around cycles with one-off, single-use packages.

I agree for this PR that the `upgrade.ID` can stay where it is, and I can move `GetArchiveDirectoryName`.  For the others, I'd prefer to handle that in another PR.

jchampio:
> greenplum uses steps.OutStreams. That has been true for at least 4 months now.

2647b010ce64ff4e4fcd5b13b8e7b13a4bd484d6 was committed two weeks ago.

> utils depends on upgrade is new in this PR, and that's because the GetArchiveDirectoryName is in utils/sys_utils.go. I agree that it likely should not be in there, and moving it to upgrade/directories.go is a better option; I'll move it there instead of modifying the package of upgrade.

Sounds good.

bhuvnesh2703:
> > I guess I would like a bit more discussion on what exactly we would like the archive directory to be since `/Users/gpadmin/gpAdminLogs/gpupgrade-PmuBSyufs-A-2020-07-01T13:57` seems a bit much to me. Why have both the timestamp and ID when one of those identifiers should be enough? Perhaps just the upgradeID and using the system's timestamp in the metadata is enough?
> 
> I think both are useful...upgradeID for easily logically connecting an upgrade run's logs...timestamp for searching for the approximate time of the revert run. The filesystems creation/modification time are subject to change upon copying, transferring to other machines, etc. So the timestamp in the filename seems good too.

can we have the format alike YYYYMMDDHHMM or even include SS,  `T13:57` this looks weird?

cli/commanders/steps.go
dkrieger-pvtl: I tried it; I like it.  Done.

test/revert.bats
dkrieger-pvtl: This doesn't work...the `find` expression containing options must come after the path:
```
#> find /tmp/foo -type d
/tmp/foo

#> find -type d /tmp/foo
find: illegal option -- t
usage: find [-H | -L | -P] [-EXdsx] [-f path] path ... [expression]
       find [-H | -L | -P] [-EXdsx] -f path [path ...] [expression]

```

cli/commands/commands.go
```diff
@@ -581,14 +581,33 @@ func revert() *cobra.Command {
...
+
+The gpupgrade logs can be found on the master and segment hosts in: 
+%s
+
+NEXT ACTIONS
+------------
+If you would like to re-run gpupgrade, run "gpupgrade initialize" again.
+
+If you no longer want to proceed with gpupgrade, you must recreate 
+any tables, indexes, and/or roles that were dropped in order to pass
```
bhuvnesh2703, line 609: just say any objects, or any objects like tables, indexes etc

dkrieger-pvtl: This is boilerplate from PM...there will be a follow up story to tweak all our UI outputs.

dkrieger-pvtl:
> can we have the format like YYYYMMDDHHMM or even include SS, `T13:57` this looks weird?

There will be a follow up story for dealing with UI tweaks like this.

hub/server.go
dkrieger-pvtl: Done...take a look and see if you like it.  I found a compact style.

cli/commands/commands.go
```diff
@@ -581,14 +581,33 @@ func revert() *cobra.Command {
...
 			s.Finish(&err)
 			if err != nil {
 				gplog.Error(err.Error())
 				return err
 			}
 
+			fmt.Println()
 			fmt.Println("Revert completed successfully.")
+			fmt.Printf(`
+The source cluster has reverted to %s.
```
bhuvnesh2703, line 599: Technically the source cluster is always of the version with which it started.
Revert completed successfully.
The source cluster is ready to use, or something like this instead of The source cluster has reverted to?

hub/revert.go
dkrieger-pvtl: Done...I went with your suggestion.

hub/revert.go
dkrieger-pvtl: Done.

cli/commands/commands.go
dkrieger-pvtl: This is boilerplate from PM...there will be a follow up story to tweak all our UI outputs.

But I find it clearer to be explicit for the user to remind them...as they'll have to change their env to run it(MASTER_DATA_DIRECTORY, etc).

cli/commanders/steps.go
```diff
@@ -163,46 +162,26 @@ func Finalize(client idl.CliToHubClient, verbose bool) error {
...
 
 	stream, err := client.Revert(context.Background(), &idl.RevertRequest{})
 	if err != nil {
 		gplog.Error(err.Error())
-		return err
+		return &RevertResponse{}, err
 	}
 
-	_, err = UILoop(stream, verbose)
+	dataMap, err := UILoop(stream, verbose)
```
kalensk, line 176: nit: maybe renaming to `respMap` is more accurate. I find `dataMap` not very descriptive.

hub/revert.go
```diff
@@ -142,5 +143,13 @@ func (s *Server) Revert(_ *idl.RevertRequest, stream idl.CliToHub_RevertServer)
 		})
 	}
 
+	message := &idl.Message{Contents: &idl.Message_Response{Response: &idl.Response{Data: map[string]string{
+		idl.ResponseKey_source_version.String():               s.Source.Version.VersionString,
+		idl.ResponseKey_revert_log_archive_directory.String(): archiveDir,
+	}}}}
+	if err := stream.Send(message); err != nil {
+		return xerrors.Errorf("could not send revert response message: %w", err)
```
kalensk, line 151: nit: I totally realized this was not introduced here and was from before. Consider something like `sending response message: %w`

I think keeping error messages short and in the present tense helps with readability. That is, trying to avoid long error messages kind of like:
`failed to x: failed to y: failed to create new store: the error`
in favor or something like:
`x: y: new store: the error`

See:
https://github.com/golang/go/wiki/CodeReviewComments#error-strings

cli/commanders/steps.go
dkrieger-pvtl: I went with `response`; it's used as a map right afterwards, so there's no need to say it's a map.

hub/revert.go
dkrieger-pvtl: thanks, done.

PR #378, opened 2020-07-01
bash-completion: fix completion on Mac by updating cobra

jchampio:
cobra 1.0.0 has fixed an annoying bug when tab-completing with MacOS' older Bash version:

    bash: flaghash[${flagname}]: bad array subscript

Tab-completion now works as expected.


jchampio:
(There's a `go.mod` failure in the CI that I can't repro on my machine yet; I'm looking into it.)

PR #381, opened 2020-07-06
Refine migration scripts to drop non-upgradeable objects

bhuvnesh2703:
- Don't include unique and primary key while dropping
  from root partitions
- Unique and primary key constraints where the constraint name is not
  equal to relation name must only be dropped on non-partitioned name
- Drop statement of indexes on partitioned table should not create drop
  statement for the indexes created due to primary or unique key
  constraint
- Modify the test to ensure the correctness

Unique or primary key constraints cannot be dropped from partitioned
tables because
- Drop constraint on the root table does not cascade to child
  partitions
- Operation to drop the child partitions constraint directly is
  prohibited
The problematic table must be dropped for fixing the pg_upgrade check
error.


migration_scripts/test/create_nonupgradable_objects.sql
```diff
@@ -10,7 +10,8 @@ CREATE TABLE pt_with_index (a int references regular(a), b int, c int, d int)
     PARTITION BY RANGE(b)
         (
         PARTITION pt1 START(1),
-        PARTITION pt2 START(2) END (3)
+        PARTITION pt2 START(2) END (3),
+        PARTITION pt3 START(3) END (4)
```
dkrieger-pvtl, line 14: The comment `-- create partitioned table with foreign, unique and primary key constraints` does not appear to match the resulting table here.  There is a foreign key constraint on (a) but there is no unique or primary key constraint here.

migration_scripts/test/create_nonupgradable_objects.sql
```diff
@@ -19,6 +20,7 @@ CREATE INDEX ptidxc_prt_2 on pt_with_index_1_prt_pt2(c);
 INSERT INTO pt_with_index SELECT i, i%2+1, i, i FROM generate_series(1,10)i;
 
 -- create tables where the index relation name is not equal primary/unique key constraint name
```
dkrieger-pvtl, line 22: It might be clearer to add another comment line:
`-- we do so by squatting on the default name of the primary/unique constraint name with a TYPE`.

migration_scripts/test/create_nonupgradable_objects.sql
```diff
@@ -37,3 +40,23 @@ INSERT INTO table_with_primary_constraint VALUES(2, 2);
 
 -- create role with gphdfs readable and writable privileges
 CREATE ROLE gphdfs_user CREATEEXTTABLE(protocol='gphdfs', type='writable') CREATEEXTTABLE(protocol='gphdfs', type='readable');
+
+
+-- create partitioned tables where the index relation name is not equal primary/unique key constraint name
```
dkrieger-pvtl, line 45: This appears to be true only for the root partition.  This should be pointed out in this comment.

migration_scripts/pre-upgrade/gen_drop_fk_constraint.sql
```diff
@@ -1,16 +1,17 @@
...
--- generates a script to Drop foreign/unique/primary key constraints from partitioned tables.
---    Order is significant. Remove constraints on root partition table before
---    non-partition tables so that we can cascade deleting constraints from
---    child partition tables.
+-- generates a script to Drop foreign key constraints from root partitioned tables.
+-- Note: 
+-- 1. Foreign key constraints are not allowed to be added on to the child partition.
+-- 2. Primary and Foreign Key constraints cannot be dropped on the child partitions directly,
+-- also dropping them on the root partition does not cascade the drop of such constraints, as its
+-- not tracked in the catalog. So, we don't touch such constraint.
```
jchampio, line 9: How would a user dig themselves out of this situation?

migration_scripts/pre-upgrade/gen_drop_partition_indexes.sql
```diff
@@ -0,0 +1,41 @@
...
+   JOIN pg_catalog.pg_namespace n ON (n.oid = c.relnamespace)
+   UNION SELECT n.nspname,
+                partitiontablename AS relname
+   FROM pg_catalog.pg_partitions p
+   JOIN pg_catalog.pg_class c ON (p.partitiontablename = c.relname)
+   JOIN pg_catalog.pg_namespace n ON (n.oid = c.relnamespace))
+SELECT 'DROP INDEX '|| schemaname ||'.'||indexname||';'
+FROM partitions 
+JOIN pg_catalog.pg_indexes ON (partition_table = tablename AND partition_schema = schemaname)
+AND (indexname, schemaname, tablename) NOT IN (SELECT constraint_rel,base_relation_schema,base_relation FROM unique_primary_constraints);
```
jchampio, line 41: What is this `AND` linked to? Can you tack an `AND` onto the end of a `JOIN` without an explicit `WHERE`?

migration_scripts/pre-upgrade/gen_drop_fk_constraint.sql
bhuvnesh2703: For Primary and Foreign key constraint, the user will have to
1. Create a new table with the same schema
2. Load the from old table
3. Drop the old table
4. Rename the new table

I was thinking that this should be in documentation where we mention about the checks and possible ways to resolve them.

dkrieger-pvtl: And 5. recreate the Primary and Foreign key constraints on the new table.

migration_scripts/pre-upgrade/gen_drop_fk_constraint.sql
```diff
@@ -1,16 +1,17 @@
...
 SELECT 'ALTER TABLE ' || nspname || '.' || relname || ' DROP CONSTRAINT ' || conname || ' CASCADE;'
```
dkrieger-pvtl, line 10: Since we are only dealing for FOREIGN KEY constraints here, what is the use of `CASCADE`?  The fk constraint is not backed by any indices, so I am thinking it can be removed.

migration_scripts/pre-upgrade/gen_drop_fk_constraint.sql
```diff
@@ -1,16 +1,17 @@
...
 FROM pg_constraint cc
     JOIN
     (SELECT DISTINCT c.oid, n.nspname, c.relname
      FROM pg_catalog.pg_partition p
         JOIN pg_catalog.pg_class c ON (p.parrelid = c.oid)
         JOIN pg_catalog.pg_namespace n ON (n.oid = c.relnamespace)) as sub ON sub.oid=cc.conrelid
-WHERE cc.contype IN ('f', 'u', 'p');
+WHERE cc.contype IN ('f');
```
dkrieger-pvtl, line 17: nit: this might be clearer as `WHERE cc.contype='f'`.

migration_scripts/pre-upgrade/gen_drop_fk_constraint.sql
```diff
@@ -1,16 +1,17 @@
...
+-- 2. Primary and Foreign Key constraints cannot be dropped on the child partitions directly,
+-- also dropping them on the root partition does not cascade the drop of such constraints, as its
+-- not tracked in the catalog. So, we don't touch such constraint.
 SELECT 'ALTER TABLE ' || nspname || '.' || relname || ' DROP CONSTRAINT ' || conname || ' CASCADE;'
 FROM pg_constraint cc
     JOIN
     (SELECT DISTINCT c.oid, n.nspname, c.relname
      FROM pg_catalog.pg_partition p
         JOIN pg_catalog.pg_class c ON (p.parrelid = c.oid)
         JOIN pg_catalog.pg_namespace n ON (n.oid = c.relnamespace)) as sub ON sub.oid=cc.conrelid
```
dkrieger-pvtl, line 16: Overall question: there is `pg_partition` which appears to show only the root table and `pg_partitions` which shows the entire structure of the partition tree.   But, even so, the `parlevel` in both of these is `0`, i.e. it only shows the root.  Is this always true in 5 and 6?  I am wondering if we should also have a check that `p.parlevel=0` to make sure we are only operating on the root partition.

migration_scripts/pre-upgrade/gen_drop_fk_constraint.sql
jchampio: > I was thinking that this should be in documentation where we mention about the checks and possible ways to resolve them.

@Nirali13 @litzell @lenapivotal FYI

migration_scripts/pre-upgrade/gen_drop_partition_indexes.sql
bhuvnesh2703: it becomes part of the join filter.. between 
`SELECT constraint_rel,base_relation_schema,base_relation FROM unique_primary_constraints)` and other part of the query. basically it becomes the join condition (ON x and y...)

migration_scripts/pre-upgrade/gen_drop_fk_constraint.sql
bhuvnesh2703: if i create a multi level table parlevel will may be 0,1..etc 
pg_partition only contains the root parrelid, so we don't need parlevel..

test/migration_scripts.bats
```diff
@@ -64,6 +64,11 @@ teardown() {
     $PREUPGRADE_SCRIPTS_DIR/generate_preupgrade_sql.bash $GPHOME_SOURCE $PGPORT $PREUPGRADE_DIR
     $PREUPGRADE_SCRIPTS_DIR/execute_preupgrade_sql.bash $GPHOME_SOURCE $PGPORT $PREUPGRADE_DIR
 
+    # the migration script should not remove primary / unique key constraints on partitioned tables, so
+    # remove them manually by dropping the table as they can't be dropped.
+    $GPHOME_SOURCE/bin/psql -d $TEST_DBNAME -p $PGPORT -c "DROP TABLE table_with_unique_constraint_p;"
+    $GPHOME_SOURCE/bin/psql -d $TEST_DBNAME -p $PGPORT -c "DROP TABLE table_with_primary_constraint_p;"
+
     gpupgrade initialize \
```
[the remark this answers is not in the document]
bhuvnesh2703: i donot wanted to be very specific, that's why kept Checking.. If you see the flow of the test.
1. gpupgrade initialize should fail, we check that
2. we also verify that there are as Checking ** fatal line in the log file to ensure it failed during pg_upgrade check
3. we again verify the status.json for the state of the step.. to ensure that its failed

keeping it loose in text allows any test to be added before it which can fail, and i don't need to worry about the specific test which failed..

migration_scripts/pre-upgrade/gen_drop_fk_constraint.sql
lenapivotal: @jchampio  I will create a doc story to ensure we track it. thank you

migration_scripts/pre-upgrade/gen_drop_fk_constraint.sql
```diff
@@ -0,0 +1,32 @@
+-- Copyright (c) 2017-2020 VMware, Inc. or its affiliates
+-- SPDX-License-Identifier: Apache-2.0
+
+-- generates a script to Drop foreign key constraints from root partitioned tables.
```
dkrieger-pvtl, line 4: nit: s/Drop/drop/

migration_scripts/pre-upgrade/gen_drop_partition_indexes.sql
```diff
@@ -0,0 +1,99 @@
+-- Copyright (c) 2017-2020 VMware, Inc. or its affiliates
+-- SPDX-License-Identifier: Apache-2.0
+
+-- generates a script to DROP partition indexes
```
dkrieger-pvtl, line 4: nit:
`-- generates a script to DROP partition indexes` 
is cleaner as
`-- generates a script to drop partition indexes that do not correspond to unique or primary constraints`

migration_scripts/pre-upgrade/gen_drop_primary_unique_constraint.sql
```diff
@@ -0,0 +1,58 @@
+-- Copyright (c) 2017-2020 VMware, Inc. or its affiliates
+-- SPDX-License-Identifier: Apache-2.0
+
+-- Generate a script to to DROP unique/primary key constraints from
```
dkrieger-pvtl, line 4: nit: s/DROP/drop

migration_scripts/pre-upgrade/gen_drop_primary_unique_constraint.sql
```diff
@@ -0,0 +1,58 @@
...
+-- non-partitioned tables. Exclude all the root and child partitions as we
+-- cannot drop unique/primary constraints on the child partition tables
+WITH CTE as
```
dkrieger-pvtl, line 7: please add a comment:
`--cte to get oids of all tables that are not partition tables`.

migration_scripts/pre-upgrade/gen_drop_primary_unique_constraint.sql
```diff
@@ -0,0 +1,58 @@
...
+      pg_class
+   WHERE
+      oid NOT IN
+      (
+         SELECT DISTINCT
+            parrelid
+         FROM
+            pg_partition
+         UNION ALL
+         SELECT
```
dkrieger-pvtl, line 22: You have:
```sql
UNION ALL
SELECT
```

Don't you just want unique oids with no duplicates, i.e.

```sql
UNION
SELECT DISTINCT
```
?

test/migration_scripts.bats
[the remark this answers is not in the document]
dkrieger-pvtl: In a follow up PR, please place these drops in a script under migration_scripts/test.  These "belong" with the sql, not with the bats test.

PR #383, opened 2020-07-08
Various minor fixups

kalensk:
See individual commits.

[Pipeline](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:fixups)


upgrade/directories_test.go
```diff
@@ -226,7 +226,7 @@ func TestArchiveSource(t *testing.T) {
 
 		testutils.VerifyRename(t, source, target)
 
-		testutils.VerifyLogDoesNotContain(t, testlog, "Source directory does not exist")
+		testutils.VerifyStringDoesNotContain(t, string(testlog.Contents()), "Source directory does not exist")
```
jchampio, line 229: This is the only commit that I don't think is worth it. If a test wants generic string checking, it can call `strings.Contains()` itself. In the meantime, these tests have a reason (if shortlived) for wanting to check logfile contents.

jchampio:
(First four commits get +1 from me.)

upgrade/directories_test.go
kalensk: Sure. I can remove this commit.

kalensk: I replaced the last commit with one that just adds the missing `t.helper()`.

PR #384, opened 2020-07-08
ci: add documentation to update the GPDB rpms in the CI pipeline.

dkrieger-pvtl:
I think this will be useful.  It may need to updated from time to time but as it forms the SOURCE and TARGET clusters we use, I think it is important to document.

### DIFF
--- a/ci/README.md
+++ b/ci/README.md
@@ -4,15 +4,20 @@ This document is for developers when they consider changes to the pipeline. It
 is currently not complete, but sections will be added as need be.
 
 ## SOURCE and TARGET clusters for our tests
-The SOURCE and TARGET versions are currently hand-curated.  We pull a specific version of
-rpm from [Pivnet](https://network.pivotal.io/products/pivotal-gpdb/#/releases/683946/file_groups/2659) or 
+The SOURCE and TARGET versions are currently hand-curated.  Note that two rpm files need to be updated: `rhel6` and `rhel7`.
+
+We pull a specific version of the
+rpms from [Pivnet](https://network.pivotal.io/products/pivotal-gpdb/#/releases/683946/file_groups/2659) or 
 from its corresponding gcp instance path and place it ourselves into both the _prod_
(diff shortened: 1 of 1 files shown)


ci/README.md
```diff
@@ -0,0 +1,39 @@
...
+)
+  - This is used for the prod instance.  
+  - Note that as soon as you drop a new release here, any triggered pipeline in prod will automatically pull the latest semver
+  of the rpm.
+  
+The _dev_ and _prod_ buckets each have two subdirectories: "greenplum-db-5" and "greenplum-db-6".  You copy the
+corresponding version(5/6) into the corresponding directory.  The changes are automatically picked up by each 
+pipeline when it is triggered.
+
+The recommended workflow:
```
jchampio, line 31: We'd talked about getting rid of the manual-copy workflow as soon as we got RPATH support in 5 and 6. Has that changed?

kalensk: My understanding is that releng is planning on providing us and other teams a bucket to pull release candidate RPMs. Thus, simply updating our pipeline to pull from that new bucket should be all that is needed.

dkrieger-pvtl: According to our discussion yesterday with @bradfordb-vmware, the rpms are not uploaded to an automatically accessible location.  They are available on PivNet, or in its staging location on GCP.  Releng is working to push both 5X and 6X rpms to an automatically accessible location(i.e. a GCP bucket).

Then there's the issue of when we want to pull new rpms into our pipelines.  There's always the chance that a new rpm will break existing tests, so I see two options:

1) "go for it"...just use the latest rpms in both `cm` and `prod`.  If we think the risk that using new rpms is low(which I think is true), we can use this, keeping in mind that any failures might be due to this.

2). "curated"...test the latest rpms in `cm`, that is, automatically pull the latest rpms in `cm`.  If it passes, hand-pull(perhaps with a script) that into `master`.

Note we don't have a workflow currently to test a new rpm other than to affect all of `cm`, but given that the risk of failure is low, that's probably ok.

dkrieger-pvtl:
This process will be changed shortly...pulling this PR and placing its content in an [internal Tracker Story](https://www.pivotaltracker.com/story/show/173734458)

PR #386, opened 2020-07-09
bats: add tablespaces to finalize.bats for testing coverage

dkrieger-pvtl:
Create a tablespace in a 5X cluster for upgrade. Note that, since pg_upgrade does fails to upgrade tablespaces in 6-6 mode, we only add a tablespace in GPDB5.

As a side effect of the changes made, we have a minor additional prefactor comment:  use correct psql binary in get_segment_configration.

The code change in this PR was straightforward, but much of the work was in realizing that the GPUPGRADE_TARGET rpm in our CI pipeline was outdated.  Hence, it did not contain a fix to 6X_STABLE that caused this addition to fail on the CI until the rpm version was updated.  Both prod and dev have been updated to use `greenplum-db-6.9.0-rhel7-x86_64.rpm` as their target cluster rpm.

### DIFF
--- a/test/finalize.bats
+++ b/test/finalize.bats
@@ -9,8 +9,6 @@ load finalize_checks
 setup() {
     skip_if_no_gpdb
 
-    PSQL="$GPHOME_SOURCE/bin/psql -X --no-align --tuples-only postgres"
-
     setup_state_dir
 
     gpupgrade kill-services
@@ -34,6 +32,13 @@ teardown() {
     source "${GPHOME_SOURCE}/greenplum_path.sh"
     gpstart -a
 
+    # delete tablespace data added to the source cluster
+    if is_GPDB5 "${GPHOME_SOURCE}"; then
+        delete_tablespace_data
+    fi
+
+    # delete the state_dir, which also contains the tablespace filesystem
+    cleanup_state_dir
 }
 
 upgrade_cluster() {
@@ -71,6 +76,11 @@ upgrade_cluster() {
                gpstart -a
         fi
 
+        # upgrade on 6-6 does not work due to a bug in pg_upgrade
+        if is_GPDB5 "$GPHOME_SOURCE"; then
+            create_tablespace_with_table
+        fi
+
         gpupgrade initialize \
             --source-gphome="$GPHOME_SOURCE" \
             --target-gphome="$GPHOME_TARGET" \
@@ -82,6 +92,10 @@ upgrade_cluster() {
 
         gpupgrade execute --verbose
         gpupgrade finalize --verbose
+
+        if is_GPDB5 "$GPHOME_SOURCE"; then
+            check_tablespace_data
+        fi
 
         NEW_CLUSTER="$MASTER_DATA_DIRECTORY"
 
@@ -95,7 +109,7 @@ upgrade_cluster() {
             for datadir in "${datadirs[@]}"; do
                 local archive=$(archive_dir "$datadir")
                 rm -rf ${archive}
-                mv ${datadir}_backup ${archive}
+                mv "${datadir}"_backup "${archive}"
             done
         else
             validate_data_directories "EXISTS" "${datadirs}"
@@ -143,6 +157,12 @@ setup_state_dir() {
     export GPUPGRADE_HOME="${STATE_DIR}/gpupgrade"
 }
 
+cleanup_state_dir() {
+    if [ -n "$STATE_DIR" ]; then
+        rm -r "$STATE_DIR"
+    fi
+}
+
 get_standby_status() {
     local standby_status=$1
     echo "$standby_status" | grep 'Standby master state'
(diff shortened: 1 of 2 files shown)


test/finalize.bats
```diff
@@ -143,6 +157,12 @@ setup_state_dir() {
     export GPUPGRADE_HOME="${STATE_DIR}/gpupgrade"
 }
 
+cleanup_state_dir() {
+    if [ -n "$STATE_DIR" ]; then
+        rm -r "$STATE_DIR"
+    fi
+}
+
```
kalensk, line 165: Any particular reason for this code motion in this particular commit? Seems unrelated or not needed for directly testing tablesapces?

test/finalize.bats
```diff
@@ -170,3 +190,52 @@ validate_data_directories() {
...
+    "${GPHOME_SOURCE}"/bin/psql -d postgres -c "DROP TABLE IF EXISTS batsTable;"
+    "${GPHOME_SOURCE}"/bin/psql -d postgres -c "DROP TABLESPACE IF EXISTS batsTbsp;"
+    "${GPHOME_SOURCE}"/bin/psql -d postgres -c "DROP FILESPACE IF EXISTS batsFS;"
+}
+
+check_tablespace_data() {
+    local row_count
+    row_count=$("$GPHOME_TARGET"/bin/psql -d postgres -Atc "SELECT COUNT(*) FROM batsTable;")
+    if (( row_count != 100 )); then
+        fail "Table batsTable only has $row_count rows; should have 100"
```
kalensk, line 239: I think the error message can give a bit more context in that its trying to verify tablesaces:
`failed verifying tablespaces. batsTable got $rows want 100`

test/finalize.bats
```diff
@@ -170,3 +190,52 @@ validate_data_directories() {
             [ ! -f "${datadir}/${marker_file}" ] || fail "unexpected ${marker_file} marker file in target datadir: $datadir"
         done
 }
+
+
+
+# This is 5X-only due to a bug in pg_upgrade for 6-6
```
kalensk, line 196: Do we have any further details on the bug that would be useful breadcrumbs to expand this comment?

test/finalize.bats
```diff
@@ -170,3 +190,52 @@ validate_data_directories() {
...
+				$(hostname):2:${TABLESPACE_ROOT}/p1/demoDataDir0
+				$(hostname):3:${TABLESPACE_ROOT}/p2/demoDataDir1
+				$(hostname):4:${TABLESPACE_ROOT}/p3/demoDataDir2
+				$(hostname):5:${TABLESPACE_ROOT}/m1/demoDataDir0
+				$(hostname):6:${TABLESPACE_ROOT}/m2/demoDataDir1
+				$(hostname):7:${TABLESPACE_ROOT}/m3/demoDataDir2
+				$(hostname):8:${TABLESPACE_ROOT}/m/standby
+EOF
+
+    gpfilespace --config "${TABLESPACE_CONFIG}"
```
kalensk, line 220: I guess this doesn't need a subshell that sources GPHOME_SOURCE...Does it make sense to fully qualify the path for safety - `$GPHOME_SOURCE/bin/gpfilesapce`?

test/finalize.bats
```diff
@@ -170,3 +190,52 @@ validate_data_directories() {
...
+    "${GPHOME_SOURCE}"/bin/psql -d postgres -c "CREATE TABLE batsTable(a int) TABLESPACE batsTbsp;"
+    "${GPHOME_SOURCE}"/bin/psql -d postgres -c "INSERT INTO batsTable SELECT i from generate_series(1,100)i;"
+}
+
+# This is 5X-only
+delete_tablespace_data() {
+    "${GPHOME_SOURCE}"/bin/psql -d postgres -c "DROP TABLE IF EXISTS batsTable;"
+    "${GPHOME_SOURCE}"/bin/psql -d postgres -c "DROP TABLESPACE IF EXISTS batsTbsp;"
+    "${GPHOME_SOURCE}"/bin/psql -d postgres -c "DROP FILESPACE IF EXISTS batsFS;"
+}
```
kalensk, line 233: Might be more readable to combine these into a single statement or a single multi-line PSQL command?

test/finalize.bats
```diff
@@ -34,6 +32,13 @@ teardown() {
     source "${GPHOME_SOURCE}/greenplum_path.sh"
     gpstart -a
 
+    # delete tablespace data added to the source cluster
+    if is_GPDB5 "${GPHOME_SOURCE}"; then
+        delete_tablespace_data
+    fi
+
+    # delete the state_dir, which also contains the tablepsace filesystem
```
bhuvnesh2703, line 40: tablepsace - spelling

test/finalize.bats
```diff
@@ -170,3 +190,52 @@ validate_data_directories() {
             [ ! -f "${datadir}/${marker_file}" ] || fail "unexpected ${marker_file} marker file in target datadir: $datadir"
         done
 }
+
+
+
+# This is 5X-only due to a bug in pg_upgrade for 6-6
+create_tablespace_with_table() {
+    # the tablespace directory will get deleted when the STATE_DIR does in teardown()
```
bhuvnesh2703, line 198: ## the tablespace directory will get deleted when the STATE_DIR `is deleted` in teardown()

test/finalize.bats
```diff
@@ -34,6 +32,13 @@ teardown() {
     source "${GPHOME_SOURCE}/greenplum_path.sh"
     gpstart -a
 
+    # delete tablespace data added to the source cluster
+    if is_GPDB5 "${GPHOME_SOURCE}"; then
+        delete_tablespace_data
+    fi
```
kalensk, line 38: I think we would want to make this if statement more robust in case someone adds another test to finalize or something. That is, we only want to delete tablespaces if they have been created, and not just if its GPDB 5X.

test/finalize.bats
```diff
@@ -170,3 +190,52 @@ validate_data_directories() {
...
         done
 }
+
+
+
+# This is 5X-only due to a bug in pg_upgrade for 6-6
+create_tablespace_with_table() {
+    # the tablespace directory will get deleted when the STATE_DIR does in teardown()
+    TABLESPACE_ROOT="${STATE_DIR}"/testfs
+    TABLESPACE_CONFIG="${TABLESPACE_ROOT}"/"fs.txt"
```
bhuvnesh2703, line 200: nit: not required in quotes, "fs.txt"

test/finalize.bats
```diff
@@ -170,3 +190,52 @@ validate_data_directories() {
...
+
+    # create the directories required to implement our filespace
+    mkdir -p "${TABLESPACE_ROOT}"
```
bhuvnesh2703, line 203: since using -p don't need this mkdir, and you can combine them (but upto u)

test/finalize.bats
dkrieger-pvtl: I thought about this, but the `delete_tablespace_data` only drops objects `IF EXISTS` so it does no harm.  Setting a global variable seemed uglier.  This could also be added to a teardown array of functions.

test/finalize.bats
dkrieger-pvtl: This is new code, not code motion.  We weren't deleting the state directory(erroneously) before.

test/finalize.bats
dkrieger-pvtl: I avoided any more details here, because they are not relevant for this bats test.  Once 6-6 upgrade of tablespaces works, we can remove the branching on 5X in this test.

dkrieger-pvtl: I could say instead
`# This is 5X-only because pg_upgrade currently does not work for 6-6`

test/finalize.bats
dkrieger-pvtl: Great point...updated to reference the absolute path as you suggest.  I don't think we specify, but I presumed that since the SOURCE cluster must be running upon entry that its environment was sourced at the start of the function.

test/finalize.bats
dkrieger-pvtl: Yes, done in both cases.  But I use a here doc to make sure each line is in a separate transactions.  See [psql -c documentation](https://www.postgresql.org/docs/9.4/app-psql.html).  Indeed, in `delete_tablespace_data`, the table needs to be dropped in a transaction prior to dropping the tablespace.

test/finalize.bats
dkrieger-pvtl: I also added source'ing GPHOME_SOURCE's greenplum_path.sh to reduce any ordering dependencies.  In other words to make sure we are calling what we are thinking of calling.

PR #387, opened 2020-07-09
bats: use temporary ports that are not in the ephemeral port range

dkrieger-pvtl:
The bats test "initialize accepts a port range" was using temporary
ports for the target cluster of 35432 and above.  This is in the
ephemeral port range, and we were getting test failures.  We
presume this is due to another task in our job grabbing the
ephemeral port.

### DIFF
--- a/test/gpinitsystem.bats
+++ b/test/gpinitsystem.bats
@@ -94,12 +94,12 @@ teardown() {
 @test "initialize accepts a port range" {
     # We need to have enough ports available for the master, standby, and
     # mirrors. As usual in these tests, we assume a standard demo cluster.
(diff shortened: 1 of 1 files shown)


dkrieger-pvtl:
> What's the ephemeral range on our CI machines?

Note sure exactly but @bradfordb-vmware implied it starts at `32768`.

dkrieger-pvtl:
For additional context, each Concourse job(`6-6-install-tests`) has many tasks(`install-tests`).  Each task is in its own container, so has its own private port range.  So we must be conflicting with something we are running in that task that is grabbing the ephemeral port.  (if that's indeed the issue, which I suspect it is).

PR #389, opened 2020-07-10
test/initialize: codify TEARDOWN_FUNCTIONS for general use

jchampio:
The `TEARDOWN_FUNCTIONS` pattern is now pulled into `teardown_helpers.bash` for general use by tests. It's similar in use to Go's `defer` keyword:

    register_teardown some_teardown_function
    register_teardown echo hello

    run_teardowns # echoes 'hello', then runs some_teardown_function

Teardown functions are executed in LIFO order.

Changes from previous behavior:
- Teardowns are executed in reverse order of registration.
- Teardowns may have arguments: there's no need to create a single-line function wrapper if it's more convenient to register the teardown command inline.

### DIFF
--- a/test/initialize.bats
+++ b/test/initialize.bats
@@ -4,6 +4,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 load helpers
+load teardown_helpers
 
 setup() {
     skip_if_no_gpdb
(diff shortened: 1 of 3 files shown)


test/initialize.bats
```diff
@@ -28,32 +29,29 @@ setup() {
...
 
-set_target_cluster_var_for_teardown() {
-    TARGET_CLUSTER="$(gpupgrade config show --target-datadir)"
+    run_teardowns
 }
 
-teardown_target_cluster() {
-    delete_target_datadirs $TARGET_CLUSTER
+delete_target_on_teardown() {
+    register_teardown delete_target_datadirs "$(gpupgrade config show --target-datadir)"
```
kalensk, line 47: I might be in the minority, but it might be easier for me to follow if this was inlined?

dkrieger-pvtl: I can go either way, but I tend to prefer single-line statements to be inlined.

PR #391, opened 2020-07-14
add ExecuteRPC helper function

kalensk:
This extracts the common loop that executes gRPC requests from the hub to the agent.

This is a small step forward and hopefully will allow easier iteration. Future work may include simplifying existing tests.

[Pipeline](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:rpc)

### DIFF
--- a/hub/rpc_test.go
+++ b/hub/rpc_test.go
@@ -1,0 +1,74 @@
+// Copyright (c) 2017-2020 VMware, Inc. or its affiliates
+// SPDX-License-Identifier: Apache-2.0
+
+package hub_test
+
+import (
+	"errors"
+	"reflect"
+	"sort"
+	"testing"
+
+	"github.com/hashicorp/go-multierror"
+	"golang.org/x/xerrors"
+
+	"github.com/greenplum-db/gpupgrade/hub"
+)
+
+func TestExecuteRPC(t *testing.T) {
+	t.Run("executes multiple requests", func(t *testing.T) {
+		agentConns := []*hub.Connection{
+			{nil, nil, "mdw", nil},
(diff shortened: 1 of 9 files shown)


hub/server.go
```diff
@@ -144,47 +144,23 @@ func (s *Server) StopServices(ctx context.Context, in *idl.StopServicesRequest)
 // TODO: add unit tests for this; this is currently tricky due to h.AgentConns()
 //    mutating global state
 func (s *Server) StopAgents() error {
-	// FIXME: s.AgentConns() fails fast if a single agent isn't available
-	//    we need to connect to all available agents so we can stop just those
-	_, err := s.AgentConns()
-	if err != nil {
-		return err
-	}
```
jchampio, line 152: This introduces a functional change during code motion (`s.AgentConns()` vs `s.agentConns`); I don't think it should be part of this PR.

hub/rpc_test.go
```diff
@@ -0,0 +1,48 @@
...
+	"errors"
+	"testing"
+
+	"github.com/hashicorp/go-multierror"
+	"golang.org/x/xerrors"
+
+	"github.com/greenplum-db/gpupgrade/hub"
+)
+
+func TestExecuteRPC(t *testing.T) {
```
jchampio, line 16: I think these tests should verify that each passed connection gets an invocation.

hub/server.go
kalensk: Sure, added it back.

hub/rpc_test.go
kalensk: Added a test. Take a look and let me know if you have other ideas.

PR #392, opened 2020-07-16
Rsync master

bhuvnesh2703:
The commit contains the following
1. Rename Remote to Destination
2. Add option to pass in source host
3. Use rsync standbyhost:standbydir masterdir patter to rsync the master data files.


utils/rsync/rsync.go
```diff
@@ -24,14 +25,27 @@ var rsyncCommand = exec.Command
 func Rsync(options ...Option) error {
 	opts := newOptionList(options...)
 
+	if opts.hasSourceHost && opts.hasDestinationHost {
```
dkrieger-pvtl, line 28: This is a limitation of `rsync`, and this rule is really owned by the rsync implementation.  I wonder if we should delegate to rsync and just return that error to the user.

utils/rsync/rsync.go
```diff
@@ -24,14 +25,27 @@ var rsyncCommand = exec.Command
...
-	if opts.hasRemoteHost {
-		dstPath = opts.remoteHost + ":" + opts.destination
+	if opts.hasDestinationHost {
+		dstPath = opts.destinationHost + ":" + opts.destination
+	}
+
+	srcPath := opts.sources
+	if opts.hasSourceHost {
+		// only one path is allowed with host in source path
+		if len(opts.sources) != 1 {
```
dkrieger-pvtl, line 40: This is a limitation of `rsync`, and this rule is really owned by the rsync implementation.  I wonder if we should delegate to rsync and just return that error to the user.

bhuvnesh2703: before i access the 0th element, i would like to ensure that there is something in the array, so i will keep this check, because i don't know when or if multiple files would be supported with remote host. 
the other way would be to check when setting it whether its has only one path.. but for now i would keep it this way. unless there is any major objection.

utils/rsync/rsync.go
bhuvnesh2703: i understand the business role concern which you mentioned. I am okay to remove it, i don't have a strong preference here. but will wait for others to comment on it before deciding to keep it or remove it.

utils/rsync/rsync.go
dkrieger-pvtl: You can adjust the code to match whether we want to decide the business rules ourselves or to let `rsync` decide.  If the former, the check is fine.  If not, you'd form an array of `["host:src1","host:src2", ...]`

bhuvnesh2703:
@dkrieger-pvtl take a look, i have applied the feedback and added a test.

utils/rsync/rsync.go
bhuvnesh2703: i removed it.

utils/rsync/rsync.go
kalensk: For this particular case, we are checking that devs are passing in sane options to rsync. Its not the same situation as in the other comment.

utils/rsync/rsync_test.go
```diff
@@ -323,6 +337,63 @@ func TestRsync(t *testing.T) {
...
+			}
+
+			expectedDestination := targetDir
+			if expectedDestination != args[1] {
+				t.Errorf("got %q, want %q", args[1], expectedDestination)
+			}
+		})
+
+		rsync.SetRsyncCommand(cmd)
+
```
dkrieger-pvtl, line 365: Please add `defer rsync.ResetRsyncCommand()`...otherwise, all tests after this on in this Test will be the mock rsync and not the original one.

utils/rsync/rsync.go
bhuvnesh2703: Had a syncup with @dkrieger-pvtl and explained my reasoning for keep this.
- When the user passes one host, what does the user want -> remotehost:path1 remotehost:path2 or remotehost:path1 path2, i don't know, this API is little loose for that.. 
so for now, we do what we know and is allowed.

utils/rsync/rsync_test.go
bhuvnesh2703: added..

PR #393, opened 2020-07-16
deleteStateDirectory: don't check contents before deleting state directory

bhuvnesh2703:
State directory on the agents does not contain any content until
pg_upgrade check is run in initialize, but we support revert both after
initialize and execute. If initialize failed before pg_upgrade check,
there will be no contents in the state directory, so there is currently
no marker file which validates the state directory.

Also, before this commit, the check used to verify the config.json and
status.json files on the agent before deleting the state directory, but
these files don't exist on agents.

### DIFF
--- a/agent/delete_directories.go
+++ b/agent/delete_directories.go
@@ -24,7 +24,9 @@ func (s *Server) DeleteStateDirectory(ctx context.Context, in *idl.DeleteStateDi
 		return &idl.DeleteStateDirectoryReply{}, err
 	}
 
-	err = deleteDirectories([]string{s.conf.StateDir}, upgrade.StateDirectoryFiles, hostname, step.DevNullStream)
+	// pass an empty []string to avoid check for any pre-existing files,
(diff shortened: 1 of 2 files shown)


agent/delete_directories.go
```diff
@@ -24,7 +24,7 @@ func (s *Server) DeleteStateDirectory(ctx context.Context, in *idl.DeleteStateDi
 		return &idl.DeleteStateDirectoryReply{}, err
 	}
 
-	err = deleteDirectories([]string{s.conf.StateDir}, upgrade.StateDirectoryFiles, hostname, step.DevNullStream)
+	err = deleteDirectories([]string{s.conf.StateDir}, []string{}, hostname, step.DevNullStream)
```
dkrieger-pvtl, line 27: Please add a comment here for our future selves. Something like:
`// do not check for any pre-existing files: this call might come in before any stateDir files are created`

bhuvnesh2703: comment added

PR #394, opened 2020-07-16
startstop: pass master data directory variable as environment

bhuvnesh2703:
gpstart requires MASTER_DATA_DIRECTORY environment to be set prior to
invocation else it fails when it check if filespaces are configured in
GPDB 5. This should be fixed in GPDB too, but passing the environment
via gpupgrade for now allows to run gpstart successfully.

### DIFF
--- a/greenplum/cluster.go
+++ b/greenplum/cluster.go
@@ -297,7 +297,7 @@ func (c *Cluster) GetDirForContent(contentID int) string {
 }
 
 func (c *Cluster) Start(stream step.OutStreams) error {
-	return runStartStopCmd(stream, c.GPHome, fmt.Sprintf("gpstart -a -d %[1]s", c.MasterDataDir()))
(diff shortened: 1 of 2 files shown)


jchampio:
On further thought, the tests will fail if we do that because `greenplum.Runner` has no way to control the internal `exec.Cmd` in use. That should be fixed and then we can tackle my suggestion, in the future.

greenplum/cluster.go
```diff
@@ -297,7 +297,7 @@ func (c *Cluster) GetDirForContent(contentID int) string {
 }
 
 func (c *Cluster) Start(stream step.OutStreams) error {
-	return runStartStopCmd(stream, c.GPHome, fmt.Sprintf("gpstart -a -d %[1]s", c.MasterDataDir()))
+	return runStartStopCmd(stream, c.GPHome, fmt.Sprintf("gpstart -a -d %[1]s", c.MasterDataDir()), fmt.Sprintf("env MASTER_DATA_DIRECTORY=%[1]s", c.MasterDataDir()))
```
jchampio, line 300: Minor readability things:
- you shouldn't need `env`
- `%[1]s` can just be `%s`

PR #396, opened 2020-07-17
Refactor archive log dir

bhuvnesh2703:
This PR contains 2 commit
1. Add the move utility
2. Use the move utility in archive log method in revert and related test cases.

### DIFF
--- a/agent/archive_log_directory_test.go
+++ b/agent/archive_log_directory_test.go
@@ -5,9 +5,13 @@ package agent_test
 
 import (
 	"context"
-	"errors"
 	"os"
+	"os/exec"
+	"os/user"
+	"path/filepath"
 	"testing"
+
+	"github.com/greenplum-db/gpupgrade/testutils"
 
 	"github.com/greenplum-db/gp-common-go-libs/testhelper"
 	"golang.org/x/xerrors"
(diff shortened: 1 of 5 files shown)


utils/sys_utils.go
```diff
@@ -96,3 +106,13 @@ func GetLogDir() (string, error) {
 func GetTablespaceDir() string {
 	return filepath.Join(GetStateDir(), "tablespaces")
 }
+
+func Move(src string, dst string) error {
```
jchampio, line 110: This needs a chunk of documentation to explain when and why we should prefer it over `os.Rename`.

utils/sys_utils_test.go
```diff
@@ -0,0 +1,49 @@
...
+	t.Run("move fails", func(t *testing.T) {
+		src := ""
+
+		dst := testutils.GetTempDir(t, "")
+		defer os.RemoveAll(dst)
+
+		err := utils.Move(src, dst)
+		if err == nil {
+			t.Errorf("unexpected error %#v", err)
+		}
```
jchampio, line 36: This check can be subsumed by the later `xerrors.As()` check, since `nil` will not match a `utils.MoveError`.

utils/sys_utils.go
```diff
@@ -96,3 +106,13 @@ func GetLogDir() (string, error) {
...
+	cmd := exec.Command("mv", src, dst)
+	output, err := cmd.CombinedOutput()
+	if err != nil {
+		err = MoveError{ErrorText: string(output), Err: err}
```
jchampio, line 114: If you use `cmd.Output()` instead, the stderr stream will be automatically captured. It might let you get rid of `MoveError` entirely.

utils/sys_utils_test.go
```diff
@@ -0,0 +1,49 @@
...
+
+		var moveError utils.MoveError
+		if !xerrors.As(err, &moveError) {
+			t.Errorf("got error %#v, wanted type %T", err, moveError)
+		}
+
+		expected := "mv: : No such file or directory\n"
```
jchampio, line 43: I think this string is likely to be platform-specific; it's based on whatever `strerror()` is returning. This applies to the other failure test in this PR.

utils/sys_utils_test.go
```diff
@@ -0,0 +1,49 @@
...
+		src := testutils.GetTempDir(t, "")
+		defer os.RemoveAll(src)
+
+		dst := testutils.GetTempDir(t, "")
+		defer os.RemoveAll(dst)
+
+		err := utils.Move(src, dst)
+		if err != nil {
+			t.Errorf("unexpected error %#v", err)
+		}
```
jchampio, line 24: Probably want to check the contents of `src` to make sure `dst` was moved into it.

agent/archive_log_directory_test.go
```diff
@@ -17,55 +20,61 @@ import (
...
-		calls := 0
-
-		utils.System.Rename = func(oldPath, newPath string) error {
-			calls++
-
-			if oldPath != oldLogDir {
-				t.Errorf("got %q want %q", oldPath, oldLogDir)
-			}
+		dt := time.Now().Format("01022006150405")
+		homeDir := filepath.Join("tmp", dt)
```
jchampio, line 51: nit: there's a TempDir pattern already in use for this sort of thing.

utils/sys_utils_test.go
bhuvnesh2703: done.

PR #398, opened 2020-07-17
test/migration_scripts: use rsync's --archive rather than -r

jchampio:
`--archive` matches what we really want: preserve everything we can about the original directory tree.


jchampio:
[Pipeline in progress as a sanity check.](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:dev:rsync-archive)

PR #399, opened 2020-07-20
copy-mode revert now syncs the standby properly

dkrieger-pvtl:
upgrade: add the dbid of the standby to the master in UpgradeMaster

When executing a 5-6 pg_upgrade, the source master should know the dbid of its
standby.  This is in general the correct semantic.

Furthermore, depending on how we do revert after execute, the standby may
never sync its WAL with the master and exit abruptly:

1). rsync based revert (currently done for link-mode)
    Setting the standby-dbid isn't strictly necessary but is still corret.
2). gprecoverseg based revert (currently done for copy-mode)
    Setting the standby-dbib is needed to sync the standby to the master.

For context, see:
https://groups.google.com/a/greenplum.org/d/msg/gpdb-dev/RunORQtf0X4/reX311xIAwAJ

revert.bats tests have been modified to validate that the standby is in sync with the master after an execute revert.

NOTE: This solution sets the master's standby dbid using `gpupgrade`.  We could instead do this in `pg_upgrade`...the master could query `gp_segment_configuration` and pass the `-o -x <standby_dbid>` directly in `start_postmaster`.  This is the "correct solution" in that the fix is really a property of pg_upgrade, not gpupgrade.  However, since this fix is only needed in gpdb-5 and apparently its only use is gpupgrade, we deem it sufficient to fix in `gpugprade`.   It is also a specialized use: it is only needed if one "reverts" after a pg_upgrade.  That said, placing the fix into `pg_upgrade` is straightforward enough.  This also introduces a merge conflict with upstream; the fix would also need to be placed into master for the 5to7 gpupgrade, likely.

### DIFF
--- a/hub/upgrade_master.go
+++ b/hub/upgrade_master.go
@@ -66,12 +66,20 @@ func UpgradeMaster(args UpgradeMasterArgs) error {
 		upgrade.WithWorkDir(wd),
 		upgrade.WithOutputStreams(tee, args.Stream.Stderr()),
 	}
+
 	if args.CheckOnly {
 		options = append(options, upgrade.WithCheckOnly())
 	}
 
 	if args.UseLinkMode {
 		options = append(options, upgrade.WithLinkMode())
+	}
+
+	// When upgrading from 5 the master must be provided with its standby's dbid to allow WAL to sync.
+	if args.Source.Version.Before("6") {
+		if args.Source.HasStandby() {
+			options = append(options, upgrade.WithOldOptions(fmt.Sprintf("-x %d", args.Source.Standby().DbID)))
+		}
 	}
 
 	err = upgrade.Run(pair, options...)
--- a/hub/upgrade_master_test.go
+++ b/hub/upgrade_master_test.go
@@ -15,6 +15,7 @@ import (
 	"syscall"
 	"testing"
 
+	"github.com/greenplum-db/gp-common-go-libs/dbconn"
 	"github.com/greenplum-db/gp-common-go-libs/testhelper"
 	"golang.org/x/xerrors"
 
@@ -117,6 +118,7 @@ func TestUpgradeMaster(t *testing.T) {
 
 	source := MustCreateCluster(t, []greenplum.SegConfig{
 		{ContentID: -1, Port: 5432, DataDir: "/data/old", DbID: 1, Role: "p"},
+		{ContentID: -1, Port: 5433, DataDir: "/data/standby", DbID: 2, Role: "m"},
 	})
 	source.GPHome = "/usr/local/source"
 
@@ -226,6 +228,63 @@ func TestUpgradeMaster(t *testing.T) {
 		stderr := stream.StderrBuf.String()
 		if stderr != StreamingMainStderr {
 			t.Errorf("got stderr %q, want %q", stderr, StreamingMainStderr)
+		}
+	})
+
+	t.Run("sets the standby dbid on the master if the GPDB version is 5", func(t *testing.T) {
+		execCmd := exectest.NewCommandWithVerifier(Success, func(command string, args ...string) {
+			expected := "--old-options -x 2"
+			if !strings.Contains(strings.Join(args, " "), expected) {
+				t.Errorf("did not find %q in the args %q", expected, args)
+			}
+		})
+		SetExecCommand(execCmd)
+		defer ResetExecCommand()
+
+		rsync.SetRsyncCommand(exectest.NewCommand(Success))
+		defer rsync.ResetRsyncCommand()
+
+		source.Version = dbconn.NewVersion("5.28.0")
+
+		err := UpgradeMaster(UpgradeMasterArgs{
+			Source:      source,
+			Target:      target,
+			StateDir:    tempDir,
+			Stream:      step.DevNullStream,
+			CheckOnly:   false,
+			UseLinkMode: false,
+		})
+		if err != nil {
+			t.Errorf("returned error %+v", err)
+		}
+	})
+
+	t.Run("does not set the standby dbid on the master if the GPDB version is 6", func(t *testing.T) {
+		execCmd := exectest.NewCommandWithVerifier(Success, func(command string, args ...string) {
+			for _, arg := range args {
+				if arg == "--old-options" {
+					t.Errorf("expected --old-options to not be in args %q", args)
+				}
+			}
+		})
+		SetExecCommand(execCmd)
+		defer ResetExecCommand()
+
+		rsync.SetRsyncCommand(exectest.NewCommand(Success))
+		defer rsync.ResetRsyncCommand()
+
+		source.Version = dbconn.NewVersion("6.10.0")
+
+		err := UpgradeMaster(UpgradeMasterArgs{
+			Source:      source,
+			Target:      target,
+			StateDir:    tempDir,
+			Stream:      step.DevNullStream,
+			CheckOnly:   false,
+			UseLinkMode: false,
+		})
+		if err != nil {
+			t.Errorf("returned error %+v", err)
 		}
 	})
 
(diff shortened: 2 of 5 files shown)


hub/upgrade_master.go
```diff
@@ -74,6 +75,13 @@ func UpgradeMaster(args UpgradeMasterArgs) error {
 		options = append(options, upgrade.WithLinkMode())
 	}
 
+	// passing the standby's dbid to the master allows the standby to sync its WAL log after copy-mode revert
```
jchampio, line 78: Can you make this comment reflect the "5->6+ only" nature of this requirement? Something like the internal pg_upgrade context comments is what I'm looking for.

hub/upgrade_master_test.go
```diff
@@ -229,6 +231,75 @@ func TestUpgradeMaster(t *testing.T) {
 		}
 	})
 
+	t.Run("sets the standby dbid on the master if the GPDB version is 5", func(t *testing.T) {
+		execCmd := exectest.NewCommandWithVerifier(Success, func(command string, args ...string) {
+			for i, arg := range args {
+				if arg == "--old-options" {
+					if len(args) >= i+1 && args[i+1] == "-x 2" {
```
jchampio, line 238: Check me -- is the `len(args) >= i+1` an off-by-one? If `len(args) == i+1` then `args[i+1]` will crash.

Elsewhere we use the `flag` package to quickly parse these args lists in the tests.

test/revert.bats
```diff
@@ -172,3 +175,49 @@ test_revert_after_execute() {
...
+        local RESULTS
+        RESULTS=$("$PSQL" -AXtf$'\t' postgres -c "SELECT sent_location, replay_location from pg_stat_replication")
+        IFS='|' read -r sent replay <<< "$RESULTS"
+
+        if [[ -z "$RESULTS" ]] && ! is_source_standby_running; then
+            break
+        elif [[ "$sent" == "$replay" ]]; then
+            INSYNC=0
+            break
+        fi
```
jchampio, line 198: nit: I think if you `SELECT sent_location == replay_location`, you can have the server tell you `INSYNC` directly without needing to do the `read -r`/comparison in Bash.

test/revert.bats
```diff
@@ -172,3 +175,49 @@ test_revert_after_execute() {
...
+        if [[ $content == "-1" ]] && [[ $role == "m" ]]; then
+            standbyDir=$datadir
+        fi
+    done <<< "$config"
+
+    if ! "${GPHOME_SOURCE}"/bin/pg_ctl status -D "$standbyDir" > /dev/null; then
+        return 1
+    fi
+    return 0
+}
```
jchampio, line 223: missing newline

test/revert.bats
```diff
@@ -172,3 +175,49 @@ test_revert_after_execute() {
...
+    while read -r content role hostname port datadir; do
+        if [[ $content == "-1" ]] && [[ $role == "m" ]]; then
+            standbyDir=$datadir
+        fi
+    done <<< "$config"
+
+    if ! "${GPHOME_SOURCE}"/bin/pg_ctl status -D "$standbyDir" > /dev/null; then
+        return 1
+    fi
+    return 0
```
jchampio, line 222: nit: I think `return 0` is redundant here.

hub/upgrade_master_test.go
dkrieger-pvtl: Good catch...changed to making sure len(args) is at least 2 greater than or equal to.

For `flag`, it seems that the [flag API](https://golang.org/pkg/flag/) requires all flags passed as args in `flags.Parse(args)` to be defined, which seems less readable/maintainable than the current code.  In particular, the test would then need to know a priori which args are _required_, like `source`.

hub/upgrade_master.go
dkrieger-pvtl: Comment changed to `	// for 5->6+ pg_upgrade, the master must be provided with its standby's dbid to allow WAL to sync`.

test/revert.bats
dkrieger-pvtl: Done...it is cleaner now.  Note it is `sent_location = replay_location`(`=` and not `==`).

test/revert.bats
dkrieger-pvtl: DONE.

test/revert.bats
dkrieger-pvtl: DONE.

hub/upgrade_master_test.go
```diff
@@ -229,6 +231,75 @@ func TestUpgradeMaster(t *testing.T) {
...
+	t.Run("sets the standby dbid on the master if the GPDB version is 5", func(t *testing.T) {
+		execCmd := exectest.NewCommandWithVerifier(Success, func(command string, args ...string) {
+			for i, arg := range args {
+				if arg == "--old-options" {
+					if len(args) >= i+2 && args[i+1] == "-x 2" {
+						return
+					}
+				}
+			}
+			t.Errorf("did not find --old-options -x 2 in the args")
```
kalensk, line 243: I find the following significantly more readable:

```
expected := "--old-options -x 2"
if !strings.Contains(strings.Join(args, " "), expected) {
    t.Errorf("args %q do not contain %q", args, expected)
}
```

test/revert.bats
```diff
@@ -172,3 +175,45 @@ test_revert_after_execute() {
...
+is_source_standby_running() {
+    local config
+    config=$(get_segment_configuration "${GPHOME_SOURCE}")
+
+    local standbyDir
+    while read -r content role hostname port datadir; do
+        if [[ $content == "-1" ]] && [[ $role == "m" ]]; then
+            standbyDir=$datadir
+        fi
+    done <<< "$config"
```
kalensk, line 214: Can `query_datadirs` with a where clause help us here? Something like:

```
local standby_datadir
standby_datadir=$(query_datadirs $GPHOME_SOURCE $PGPORT "content = '-1' AND role = 'm'")
```


nit: Our convention is to use snake case for variable names, so `standbyDir` would be `standby_dir`.

test/revert.bats
```diff
@@ -172,3 +175,45 @@ test_revert_after_execute() {
...
+        elif [[ "$INSYNC" == "t" ]]; then
+            break
+        fi
+        # otherwise, standby is not in sync yet, but might sync later.
+
+        sleep 5
+    done
+
+    [[ $INSYNC == "t" ]]
+}
```
kalensk, line 202: I find using the current time and `until` loop a bit complex and unnecessary. What about using a simple duration and counter such as:

```
local duration=600  # Wait 10 minutes (600 seconds) for the standby to disappear or be in sync
local poll=5        # Poll every 5 seconds

while (( $duration > 0 )); do
    # check the standby...
    
    sleep $poll
    (( duration = duration - poll ))
done
```

There is probably a simpler way of doing `while (( $duration > 0 ))`

hub/upgrade_master_test.go
dkrieger-pvtl: This works and would continue to work even if "-x" and "2" were passed in as separate args.  The current code explicitly requires one arg "-x 2" and would fail if the internal implementation decided to pass args to `--old-options` via an array of strings vs a single string.

So for this, the current implementation makes fewer assumptions and is more robust.

test/revert.bats
dkrieger-pvtl: Done...much cleaner.

test/revert.bats
dkrieger-pvtl: Sure.  Done.

hub/upgrade_master_test.go
dkrieger-pvtl: Done.

hub/upgrade_master_test.go
dkrieger-pvtl: Converted to a string. See https://github.com/greenplum-db/gpupgrade/pull/399#discussion_r459674264.

hub/upgrade_master.go
dkrieger-pvtl: Comment changed.

PR #400, opened 2020-07-23
test/migration_scripts: prevent rsyncs to/from root

jchampio:
We're hitting intermittent issues where `$datadir_root` is blank, which is causing `rsync` to destroy data. Although it's not yet clear why the existing checks aren't good enough -- perhaps `realpath` is failing intermittently -- add code to bail out, at point of call, if either variable is blank.

### DIFF
--- a/test/migration_scripts.bats
+++ b/test/migration_scripts.bats
@@ -65,7 +65,10 @@ backup_source_cluster() {
     datadir_root="$(realpath "$MASTER_DATA_DIRECTORY"/../..)"
 
     gpstop -af
-    rsync --archive "$datadir_root"/ "$backup_dir"/
+    # TODO: Find out why in some cases the variables used in rsync are empty/not-set
+    # which causes deletion of the the root directory. Once we have identified,
+    # do the necessary refactoring
+    rsync --archive "${datadir_root:?}"/ "${backup_dir:?}"/
     gpstart -a
 }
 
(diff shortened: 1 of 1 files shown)


test/migration_scripts.bats
```diff
@@ -65,7 +65,7 @@ backup_source_cluster() {
     datadir_root="$(realpath "$MASTER_DATA_DIRECTORY"/../..)"
 
     gpstop -af
-    rsync --archive "$datadir_root"/ "$backup_dir"/
+    rsync --archive "${datadir_root:?}"/ "${backup_dir:?}"/
```
kalensk, line 68: Can you explain the `:?` construct? So far from the man page that is for includes/excludes, but that is not what is goin gon here. Is the `?` a BASH construct to match a single character. But then what is the `:`?

jchampio: This is a parameter expansion: https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html

> If parameter is null or unset, the expansion of word (or a message to that effect if word is not present) is written to the standard error and the shell, if it is not interactive, exits. Otherwise, the value of parameter is substituted.

dkrieger-pvtl: To be explicit, the proposed construct fails fast:

```bash
#!/usr/bin/env bats

teardown() {
  restore_source_cluster
}
restore_source_cluster() {
    ls "${DFDF:?}"
    echo XXXX >&3
}
@test "experiment" {}
```

produces:
```
bats -t erase.bats
1..1
not ok 1 experiment
# (from function `restore_source_cluster' in file erase.bats, line 7,
#  from function `teardown' in test file erase.bats, line 4)
#   `restore_source_cluster' failed
# /var/folders/ww/tpm1msfd54dccn7sv5x6p33m0000gn/T/bats.26300.src: line 7: DFDF: parameter null or not set
```

PR #401, opened 2020-07-24
test/execute: sanitize directory removal

jchampio:
Make sure the target master datadir really looks like a target master datadir before removal. Additionally apply some shellcheck recommended fixes to harden the implementation.

### DIFF
--- a/test/execute.bats
+++ b/test/execute.bats
@@ -133,7 +133,8 @@ ensure_hardlinks_for_relfilenode_on_master_and_segments() {
         --disk-free-ratio 0 \
         --verbose
 
-    local datadir="$(gpupgrade config show --target-datadir)"
+    local datadir
+    datadir="$(gpupgrade config show --target-datadir)"
     NEW_CLUSTER="${datadir}"
 
     # Initialize creates a backup of the target master data dir, during execute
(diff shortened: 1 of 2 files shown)


test/execute.bats
```diff
@@ -133,15 +133,17 @@ ensure_hardlinks_for_relfilenode_on_master_and_segments() {
...
     NEW_CLUSTER="${datadir}"
 
     # Initialize creates a backup of the target master data dir, during execute
     # upgrade master steps refreshes the content of the target master data dir
     # with the existing backup. Remove the target master data directory to
     # ensure that initialize created a backup and upgrade master refreshed the
     # target master data directory with the backup.
-    rm -rf "${datadir}"/*
+    abort_unless_target_master "${datadir}"
+    rm -rf "${datadir:?}"/*
```
bhuvnesh2703, line 146: do we need this now, since abort_unless_target_master should fail, or abort_unless_target_master will not terminate the execution because of the way bats does it and will need `datadir:?`

jchampio: After the events of this week, I prefer having both, even though at the moment they are redundant:
- using `:?` ensures that even if something accidentally disables the abort code, we will still not blow away your root directory, and
- `abort_unless_target_master` gives a much friendlier error message than `:?`.

test/helpers.bash
```diff
@@ -59,16 +59,25 @@ start_source_cluster() {
     isready || (source "$GPHOME_SOURCE"/greenplum_path.sh && "${GPHOME_SOURCE}"/bin/gpstart -a)
 }
 
+# Sanity check that the passed directory looks like a valid master data
+# directory for a target cluster. Intended to be called right before deleting
+# said directory.
+abort_unless_target_master() {
```
dkrieger-pvtl, line 65: nit: in golang, we use `must` to indicate functions that will panic/exit when they cannot satisfy their contracts.  So we could name this `must_be_target_master()`.

jchampio: We've chosen in the past to use `action_[if_not/unless]_condition` rather than `require_condition` for the Bash helpers. I'm on board to change it but not in this PR.

jchampio:
(re-running the pipeline to double-check, given all the other findings over the last week)

PR #403, opened 2020-07-25
write gpupgrade files atomically

kalensk:
Use renamio to atomically write `config.json`, `gpinisystem_config`, and `tablespaces.txt` files atomically.

The general pattern is:
```golang
// Use renameio to ensure atomicity when writing
file, err :=  renameio.TempFile("", upgrade.GetConfigFile())
if err != nil {
    return err
}
defer func() {
    if cErr := file.Cleanup(); cErr != nil {
        err = multierror.Append(err, cErr).ErrorOrNil()
    }
}()

// write contents to file ...
_, err = file.WriteString(contents)
if err != nil {
    return xerrors.Errorf("writing file: %w", err)
}

return file.CloseAtomicallyReplace()
```

However, the purpose of this draft PR is to solicit comments and ideas on how to write unit tests for the failure cases **_if_** it's worth doing so. That is, testing that we are correctly returning errors if the following functions fail: `renameio.TempFile`, `file.Cleanup`, `file.CloseAtomicallyReplace()`, and writing contents to the file. I think the cost and effort to write these tests might not be worth it. Also, we already have precedence for [not writing these tests](https://github.com/greenplum-db/gpupgrade/blob/master/step/file_store_test.go) for these cases for [file_store.go](https://github.com/greenplum-db/gpupgrade/blob/0484ee47d9b211374a81591ad8b3fa0422244a57/step/file_store.go#L108-L124) which also uses renameio.

For the final implementation I will probably pull the above pattern into a single helper function. Ideas on where I should put it or does `sys_utils.go` work?

### DIFF
--- a/greenplum/tablespace.go
+++ b/greenplum/tablespace.go
@@ -9,12 +9,14 @@ import (
 	"path/filepath"
 	"strconv"
 
+	"github.com/google/renameio"
+	"github.com/hashicorp/go-multierror"
+
 	"github.com/greenplum-db/gp-common-go-libs/dbconn"
 	"github.com/pkg/errors"
 	"golang.org/x/xerrors"
 
 	"github.com/greenplum-db/gpupgrade/idl"
-	"github.com/greenplum-db/gpupgrade/utils"
 )
 
 const tablespacesQuery = `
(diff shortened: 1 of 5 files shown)


greenplum/tablespace.go
```diff
@@ -141,14 +143,24 @@ func TablespacesFromDB(conn *dbconn.DBConn, tablespacesFile string) (Tablespaces
...
+	}()
+
 	if err := tablespaceTuples.Write(file); err != nil {
 		return nil, xerrors.Errorf("populate tablespace mapping file: %w", err)
 	}
 
+	if err := file.CloseAtomicallyReplace(); err != nil {
+		return nil, err
+	}
+
```
dkrieger-pvtl, line 164: We do not need to atomically create this file.  The tablespace file is a new file that will be created on demand each time this runs; there is no other consumer of this file that might see a partially created tablespace file.

hub/init_target_cluster.go
```diff
@@ -106,14 +107,24 @@ func CreateInitialInitsystemConfig(targetMasterDataDir string) ([]string, error)
 	return gpinitsystemConfig, nil
 }
 
-func WriteInitsystemFile(gpinitsystemConfig []string, gpinitsystemFilepath string) error {
-	gpinitsystemContents := []byte(strings.Join(gpinitsystemConfig, "\n"))
+func WriteInitsystemFile(gpinitsystemConfig []string, path string) error {
```
dkrieger-pvtl, line 110: Same comment as with the tablespaces file...this file creation need not be atomic.  The current `ioutil.WriteFile` is sufficient and much simpler.

kalensk:
Based on the above feedback opened PR https://github.com/greenplum-db/gpupgrade/pull/410

PR #404, opened 2020-07-27
Update Revert verbiage

kalensk:
See individual commits and their associated stories for updates to revert verbiage.

[Test pipeline](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:revert_verbiage)


cli/commands/commands.go
```diff
@@ -600,18 +602,18 @@ func revert() *cobra.Command {
...
 ------------
-If you would like to re-run gpupgrade, run "gpupgrade initialize" again.
+To restart the upgrade, run "gpupgrade initialize" again.
 
-If you no longer want to proceed with gpupgrade, you must recreate 
-any tables, indexes, and/or roles that were dropped in order to pass 
-the pg_upgrade checks.`, response.Version, response.ArchiveDir)
+To use the reverted cluster, you must recreate any tables, indexes, and/or 
+roles that were dropped or altered to pass the pg_upgrade checks.
+`, response.Version, response.ArchiveDir)
```
dkrieger-pvtl, line 616: nit: This is correct as per the story, but I wonder if we want some verbiage similar to what we do at the end of a successful execute: `The source cluster is now running. The PGPORT is %s and the MASTER_DATA_DIRECTORY is %s.`. You could say the user of course knows this, but they will need to reset their PGPORT/MASTER_DATA_DIRECTORY back to use the source cluster.

kalensk: Updated.

cli/commanders/steps.go
```diff
@@ -119,15 +121,21 @@ func Execute(client idl.CliToHubClient, verbose bool) error {
 	message := fmt.Sprintf(`
 Execute completed successfully.
 
-The target cluster is now running. The PGPORT is %s and the MASTER_DATA_DIRECTORY is %s.
+The target cluster is now running. The PGPORT is %s and the 
+MASTER_DATA_DIRECTORY is %s
```
dkrieger-pvtl, line 125: Note the spec has a `.` at the end of this sentence, but I think it's clearer to not have one, so a user can cut-and-paste.

kalensk: Good eye! Right, I pointed that out to Peter and we chose to omit the period for the reasons you listed, but neglected to update the spec. I'll do so now.

PR #405, opened 2020-07-28
test: correctly return from run_teardowns

jchampio:
The teardown function in BATS runs without `set -e` protection. Explicitly return on failures.

### DIFF
--- a/test/teardown_helpers.bats
+++ b/test/teardown_helpers.bats
@@ -46,3 +46,13 @@ EOF
     run_teardowns
     (( ! was_run )) || fail "teardown was re-run incorrectly"
 }
+
+@test "teardown failures are reported even without set -e" {
+    local status=0
+
+    register_teardown false
+
+    # Chaining commands with || will disable set -e.
+    run_teardowns || status=$?
+    [ "$status" -eq 1 ] || fail "status was $status; expected 1"
+}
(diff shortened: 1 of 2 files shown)


test/teardown_helpers.bats
```diff
@@ -46,3 +46,13 @@ EOF
...
+
+@test "teardown failures are reported even without set -e" {
+    local status=0
+
+    register_teardown false
+
+    # Chaining commands with || will disable set -e.
+    run_teardowns || status=$?
+    [ "$status" -eq 1 ] || fail "status was $status; expected 1"
+}
```
dkrieger-pvtl, line 58: Please add a test that ensures `run_teardown` fails fast.  Something like this:
```bash
@test "teardown failures fail-fast and are reported even without set -e" {
    local was_run=0
    local status=0

    # the first registered teardown will run last
    register_teardown eval was_run=1
    register_teardown false

    # Chaining commands with || disables set -e.
    run_teardowns || status=$?
    [ "$status" -eq 1 ] || fail "status was $status; expected 1"
    [ "$was_run" -eq 0 ] || fail "was_run was $was_run; expected 0"
}
```

jchampio: I can do that in a follow-up; I agree it's important. Do you have any issues with the current code?

jchampio: (In this case, we have additional work depending on the fix.)

dkrieger-pvtl: Sure, a follow up is fine.  The current code and test LGTM, I just wanted to make sure we pinned the behavior of failing fast.

PR #408, opened 2020-07-29
revert tablespaces

kalensk:
This adds the ability to revert user defined tablespaces. This includes rsync'ing tablespaces from the mirror to the primary in link mode. And removing the user tablespaces from the target cluster.

**_Note:_** A team discussion and a followup PR is needed before delivering the story to ensure the rsync and delete tablespace functions on the agent are safe.

[Pipeline](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:dev:revert-tablespaces)


greenplum/cluster.go
```diff
@@ -37,8 +37,9 @@ type Cluster struct {
 	// check for key existence.
 	Mirrors map[int]SegConfig
 
-	GPHome  string
-	Version dbconn.GPDBVersion
+	GPHome         string
+	Version        dbconn.GPDBVersion
+	CatalogVersion string
```
jchampio, line 42: No other code in `Cluster` makes use of the `CatalogVersion`; I think this belongs somewhere else. Perhaps the server config object.

upgrade/directories.go
```diff
@@ -210,3 +212,68 @@ func DeleteDirectories(directories []string, requiredPaths []string, streams ste
 
 	return mErr.ErrorOrNil()
 }
+
+// Tablespace Directory Structure
```
dkrieger-pvtl, line 216: Use something like: `// Tablespace Directory Structure when GPDB-5 is upgraded to a GPDB-6`

The structure of a GPDB-6 tablespace created directly in 6(vs. an upgraded one) is different.

upgrade/directories.go
```diff
@@ -210,3 +212,68 @@ func DeleteDirectories(directories []string, requiredPaths []string, streams ste
...
+//
+//  GPDB 5X:  DIR/<fsname>/<datadir>/<tablespace_oid>/<db_oid>/<relfilenode>
+//  GPDB 6X:  DIR/<fsname>/<datadir>/<tablespace_oid>/<db_oid>/GPDB_6_<catalog_version>/<db_oid>/<relfilenode>
+func DeleteTablespaceDirectories(streams step.OutStreams, dirs []string) error {
+	err := DeleteDirectoriesFunc(dirs, []string{}, streams)
+	if err != nil {
+		return err
+	}
+
+	// For example, the 6X tablesapce
```
dkrieger-pvtl, line 250: s/tablesapce/tablespace/

upgrade/directories.go
```diff
@@ -210,3 +212,68 @@ func DeleteDirectories(directories []string, requiredPaths []string, streams ste
...
+	//    /filespace/demoDataDir0/16386/1/GPDB_6_301908232
+	// has been deleted above. Now check that it's parent directory
```
dkrieger-pvtl, line 252: s/it's/its/

upgrade/directories.go
```diff
@@ -210,3 +212,68 @@ func DeleteDirectories(directories []string, requiredPaths []string, streams ste
...
+	// can also be deleted by ensuring that its contents do not overlap with
+	// the tablespace of 5X.
+	for _, dir := range dirs {
+		parent := filepath.Dir(dir)
```
dkrieger-pvtl, line 256: Please `path.Clean(dir)` prior to calling `Dir` due to a long-standing golang bug.

https://play.golang.org/p/QsG7Yxxe0Ss

```
package main

// show bug: /a/b
// but this works: /a
// a fix: /a

import (
	"fmt"
	"path"
	"path/filepath"
)

func main() {
	fmt.Println("show bug:", filepath.Dir("/a/b/"))
	fmt.Println("but this works:", filepath.Dir("/a/b"))
	fmt.Println("a fix:", filepath.Dir(path.Clean("/a/b/")))
}
```

upgrade/directories.go
```diff
@@ -210,3 +212,68 @@ func DeleteDirectories(directories []string, requiredPaths []string, streams ste
...
+			return nil
+		}
+
+		// If the directory is empty it 'only' contained the target cluster
+		// tablespace and is safe to delete.
+		if err := utils.System.Remove(parent); err != nil {
+			return err
+		}
+	}
+
```
dkrieger-pvtl, line 277: Note this block does not handle the case of multiple passed in `dirs` having the same parent directory.  That is because `os.Remove` fails if its input directory does not exist.  This function is called by both the master and primaries to delete datadirs on the target cluster, and by construction they do not hit this case(since their last two path elements are `<dbid>/GPDB_6_<catalog_version>` and hence don't overlap).

However, this function is generic, and could have been called `DeleteDirectoriesIncludingParentIfEmpty` or some such.  This is an artifact of the code factoring that has this as a standalone function.  

Possible soluitions: using `os.RemoveAll` will resolve this, and we've just checked above that the directory is, in fact, empty.  Note that the function ultimately called through `DeleteDirectoriesFunc()` in this function here calls `os.RemoveAll` on the subdirectory of `parent`.  Or you can argue that this function is specific to tablespaces and just put a comment in, but perhaps that's not ideal.

upgrade/directories.go
kalensk: Yeah if needed, we should clean the directory when it is first created, so any downstream callers don't need to worry about it.

dkrieger-pvtl: It's probably better to clean it at point of use(e.g. here) since that's where the bug in the golang library is...

idl/cli_to_hub.proto
```diff
@@ -75,14 +75,15 @@ enum Substep {
...
     UPGRADE_MIRRORS = 20;
-    DELETE_PRIMARY_DATADIRS = 21;
-    DELETE_MASTER_DATADIR = 22;
-    DELETE_SEGMENT_STATEDIRS = 23;
-    STOP_HUB_AND_AGENTS = 24;
-    DELETE_MASTER_STATEDIR = 25;
-    ARCHIVE_LOG_DIRECTORIES = 26;
-    RESTORE_SOURCE_CLUSTER = 27;
-    START_SOURCE_CLUSTER = 28;
+    DELETE_TABLESPACES = 21;
```
jchampio, line 78: For the future -- there's no reason to keep this list sorted. The run order is not static, and keeping it "in order" makes the diff bigger.

(Adding to the end also trains muscle memory for a typical protobuf use case, where it's important to only add new features to the end so that existing clients remain compatible. That's not necessary here but I think it's worth pointing out.)

hub/delete_data_directories.go
```diff
@@ -48,3 +58,73 @@ func deleteDataDirectories(agentConns []*Connection, cluster *greenplum.Cluster,
 
 	return ExecuteRPC(agentConns, request)
 }
+
+func DeleteTablespaceDirectories(streams step.OutStreams, agentConns []*Connection, target *greenplum.Cluster, tablespaces greenplum.Tablespaces) error {
```
jchampio, line 62: I'm having trouble reviewing this file. I think the call stack is
```
hub.DeleteTablespaceDirectories
-> hub.DeleteTablespaceDirectoriesOnMaster
  -> hub.DeleteTablespaceDirectoriesFunc (is an alias of)
     upgrade.DeleteTablespaceDirectories
    -> upgrade.DeleteDirectoriesFunc (is an alias of)
       upgrade.DeleteDirectories
-> hub.DeleteTablespaceDirectoriesOnPrimaries
  -> idl.AgentClient.DeleteTablespaceDirectories (via RPC)
    -> agent.Server.DeleteTablespaceDirectories
      -> upgrade.DeleteTablespaceDirectories
        -> upgrade.DeleteDirectoriesFunc (is an alias of)
           upgrade.DeleteDirectories
```
The global injections make it hard to prove test correctness (see next comment).

upgrade/directories_test.go
```diff
@@ -444,6 +448,125 @@ func TestDeleteDirectories(t *testing.T) {
...
+		}
+		defer func() {
+			utils.System.Remove = os.Remove
+		}()
+
+		err := upgrade.DeleteTablespaceDirectories(&testutils.DevNullWithClose{}, []string{tablespaceDir})
+		if err != expected {
+			t.Errorf("got %#v want %#v", err, expected)
+		}
+	})
```
jchampio, line 542: For a concrete example: this test replaces `os.Remove` globally in order to test a single error path, where the call to delete the directories themselves succeeds, but then the call to remove the parent fails. Since `upgrade.DeleteTablespaceDirectoriesFunc` wasn't replaced, we actually go through its `os.RemoveAll` call live. If `upgrade.DeleteTablespaceDirectories` changes to use `os.Remove` instead of `os.RemoveAll` at some point, this test will continue to pass, but we'll be testing something completely different.

I think these sorts of smells point to a need for better structure. Many possible alternatives exist (direct dependency injection, gathering of all injection seams into a single place, testing against a live filesystem).

hub/restore_source_cluster.go
```diff
@@ -56,6 +56,29 @@ func RsyncMasterAndPrimaries(stream step.OutStreams, agentConns []*Connection, s
...
 }
 
+func RsyncMasterAndPrimariesTablespaces(stream step.OutStreams, agentConns []*Connection, source *greenplum.Cluster, tablespaces greenplum.Tablespaces) error {
+	var wg sync.WaitGroup
+	errs := make(chan error, 2)
+
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		errs <- RsyncMasterTablespaces(stream, source.MasterHostname(), tablespaces[greenplum.MasterDbid], tablespaces[source.Standby().DbID])
```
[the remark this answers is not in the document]
jchampio: I agree that we should remove the `MasterDbid` constant; we know exactly what the master's DBID is.

upgrade/directories.go
```diff
@@ -210,3 +212,68 @@ func DeleteDirectories(directories []string, requiredPaths []string, streams ste
...
+//   │           ├── 12094
+//   │           │   ├── 16384
+//   │           │   └── PG_VERSION
+//   │           └── 2
+//   │               └── GPDB_6_301908232
+//   │                   └── 12812
+//   │                       └── 16389
+//
+//  GPDB 5X:  DIR/<fsname>/<datadir>/<tablespace_oid>/<db_oid>/<relfilenode>
+//  GPDB 6X:  DIR/<fsname>/<datadir>/<tablespace_oid>/<db_oid>/GPDB_6_<catalog_version>/<db_oid>/<relfilenode>
```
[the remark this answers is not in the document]
dkrieger-pvtl: Yes...the GPDB 6X path should read `DIR/<fsname>/<datadir>/<tablespace_oid>/<dbID>/GPDB_6...`

upgrade/directories.go
jchampio: > Yeah if needed, we should clean the directory when it is first created, so any downstream callers don't need to worry about it.

This doesn't scale particularly well. Adopting this approach means that all functions need meticulous precondition documentation ("you *must* have cleaned all input paths") and close review of all new callers and entry points. I find it's more maintainable to make the bottom layer bulletproof.

upgrade/directories.go
jchampio: I don't think there's any value in a "generic" remove-a-list-of-directories-and-maybe-their-parents-too function. This is coupled to the tablespace use case.

dkrieger-pvtl: Sure.  Under that interpretation, a comment here would suffice to state our assumption that the `dirs` collection don't contain any two directories under the same parent.

hub/init_target_cluster.go
```diff
@@ -189,3 +195,38 @@ func GetMasterSegPrefix(datadir string) (string, error) {
...
+		if strings.HasPrefix(line, prefix) {
+			version = strings.TrimSpace(strings.TrimPrefix(line, prefix))
+		}
+	}
+
+	if err := scanner.Err(); err != nil {
+		return "", xerrors.Errorf("scanning pg_controldata: %w", err)
+	}
+
+	return version, nil
```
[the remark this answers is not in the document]
dkrieger-pvtl: The [catalog version](https://github.com/greenplum-db/gpdb/blob/master/src/include/catalog/catversion.h#L48) is well-defined and I think we can count on it not changing format.  So we should check stable things: that it is 9 characters long, and that it starts with a `3`.

idl/cli_to_hub.proto
kalensk: Yup, definitely good to note.

My thinking and preference was to put them in "order" for readability and clarity until we have finished implemented all major steps. That is, revert is the last major step to almost be done implementing.

hub/delete_data_directories_test.go
```diff
@@ -155,3 +160,202 @@ func TestDeleteSegmentDataDirs(t *testing.T) {
...
+			return nil
+		}
+
+		err := hub.DeleteTargetTablespacesOnMaster(step.DevNullStream, target, masterTablespaces, "301908232")
+		if err != nil {
+			t.Errorf("DeleteTargetTablespacesOnMaster returned error %+v", err)
+		}
+	})
+
+	t.Run("deletes tablespace directories on the primaries", func(t *testing.T) {
```
dkrieger-pvtl, line 209: nit: maybe this should be called `"deletes tablespace directories on the primaries but not master"`.

hub/delete_data_directories_test.go
```diff
@@ -155,3 +160,202 @@ func TestDeleteSegmentDataDirs(t *testing.T) {
...
+				Location:    "/tmp/testfs/master/demoDataDir-1/16387",
+				UserDefined: 1,
+			},
+			1663: {
+				Location:    "/data/qddir/demoDataDir-1",
+				UserDefined: 0,
+			},
+		}
+
+		hub.DeleteTablespaceDirectoriesFunc = func(streams step.OutStreams, dirs []string) error {
```
dkrieger-pvtl, line 191: If this test directly operated on temp directories(vs. using function injection), we could eliminate the added code complexity of the dependency injection function `hub.DeleteTablespaceDirectoriesFunc`.

hub/revert.go
```diff
@@ -69,11 +77,19 @@ func (s *Server) Revert(_ *idl.RevertRequest, stream idl.CliToHub_RevertServer)
...
+			if err := RsyncMasterAndPrimaries(stream, s.agentConns, s.Source); err != nil {
+				return err
+			}
+
+			return RsyncMasterAndPrimariesTablespaces(stream, s.agentConns, s.Source, s.Tablespaces)
 		})
 	}
 
 	if len(s.Config.Target.Primaries) > 0 {
+		st.Run(idl.Substep_DELETE_TABLESPACES, func(streams step.OutStreams) error {
```
dkrieger-pvtl, line 89: nit: do we want to run this as the last function in this block?  The substep above does, and the mismatch causes me to wonder if there's a reason for the difference in ordering here.

upgrade/directories.go
```diff
@@ -210,3 +213,83 @@ func DeleteDirectories(directories []string, requiredPaths []string, streams ste
...
+	}
+
+	return nil
+}
+
+func TablespacePath(tablespaceLocation string, dbID int, majorVersion uint64, catalogVersion string) string {
+	return filepath.Join(
+		tablespaceLocation,
+		strconv.Itoa(dbID),
+		fmt.Sprintf("GPDB_%d_%s", majorVersion, catalogVersion),
```
dkrieger-pvtl, line 293: comment: we refactoring to this as the name for `GPDB_6_301908232` wasn't clear for this.  Note that `/GPDB_6_301908232` is known as the `tablespace_directory_suffix` in the [pg_upgrade code](https://github.com/greenplum-db/gpdb/blob/6X_STABLE/contrib/pg_upgrade/tablespace.c#L129).

hub/init_target_cluster.go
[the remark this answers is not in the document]
jchampio: > So we should check stable things: that it is 9 characters long, and that it starts with a 3.

IMO this is the sort of thing that comes back to bite us when assumptions change. I don't think gpupgrade should be in the business of telling `pg_controldata` what is and isn't a valid catalog version.

hub/restore_source_cluster.go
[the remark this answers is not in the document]
dkrieger-pvtl: Done.

hub/init_target_cluster.go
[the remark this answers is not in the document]
dkrieger-pvtl: Fine...we discussed leaving it up to testing of `pg_controldata`....

upgrade/directories.go
```diff
@@ -210,3 +212,68 @@ func DeleteDirectories(directories []string, requiredPaths []string, streams ste
...
+//   │           │   ├── 16384
+//   │           │   └── PG_VERSION
+//   │           └── 2
+//   │               └── GPDB_6_301908232
+//   │                   └── 12812
+//   │                       └── 16389
+//
+//  GPDB 5X:  DIR/<fsname>/<datadir>/<tablespace_oid>/<db_oid>/<relfilenode>
+//  GPDB 6X:  DIR/<fsname>/<datadir>/<tablespace_oid>/<db_oid>/GPDB_6_<catalog_version>/<db_oid>/<relfilenode>
+func DeleteTablespaceDirectories(streams step.OutStreams, dirs []string) error {
```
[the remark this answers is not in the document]
dkrieger-pvtl: Done with New.

hub/delete_data_directories.go
dkrieger-pvtl: We removed the Func....

agent/delete_directories.go
```diff
@@ -30,3 +30,10 @@ func (s *Server) DeleteDataDirectories(ctx context.Context, in *idl.DeleteDataDi
 	err := DeleteDirectoriesFunc(in.Datadirs, upgrade.PostgresFiles, step.DevNullStream)
 	return &idl.DeleteDataDirectoriesReply{}, err
 }
+
+func (s *Server) DeleteTablespaceDirectories(ctx context.Context, in *idl.DeleteTablespaceRequest) (*idl.DeleteTablespaceReply, error) {
+	gplog.Info("got a request to delete tablespace directories from the hub")
+
+	err := upgrade.DeleteNewTablespaceDirectories(step.DevNullStream, in.GetDirs())
```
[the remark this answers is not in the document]
dkrieger-pvtl: Fine to keep.

upgrade/directories.go
```diff
@@ -210,3 +213,83 @@ func DeleteDirectories(directories []string, requiredPaths []string, streams ste
...
+		// NOTE: Each directory passed in has a different parent.
+		if err := os.Remove(parent); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func TablespacePath(tablespaceLocation string, dbID int, majorVersion uint64, catalogVersion string) string {
```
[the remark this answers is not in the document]
dkrieger-pvtl: Done.

upgrade/directories_test.go
```diff
@@ -444,6 +444,204 @@ func TestDeleteDirectories(t *testing.T) {
...
+			t.Errorf("expected parent dbid directory %q to be deleted", parent)
+		}
+	})
+
+	t.Run("does not delete parent dbid directory if its not empty", func(t *testing.T) {
+		tablespaceDir, parent := mustMakeTablespaceDir(t)
+		defer testutils.MustRemoveAll(t, parent)
+
+		// create a tablespace relfilenode in parent tablespaceDir
+		path := filepath.Join(filepath.Dir(tablespaceDir), "12812")
```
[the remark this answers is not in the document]
dkrieger-pvtl: Done.

upgrade/directories_test.go
```diff
@@ -444,6 +444,204 @@ func TestDeleteDirectories(t *testing.T) {
...
+
+		if !upgrade.PathExists(parent) {
+			t.Errorf("expected parent dbid directory %q to not be deleted", parent)
+		}
+	})
+
+	t.Run("errors when tablespace directory can't be deleted", func(t *testing.T) {
+		tablespaceDir, parent := mustMakeTablespaceDir(t)
+		defer func() {
+			err := os.Chmod(parent, 0700)
```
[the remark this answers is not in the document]
dkrieger-pvtl: Done...used a test-wide variable.

upgrade/directories_test.go
```diff
@@ -444,6 +444,204 @@ func TestDeleteDirectories(t *testing.T) {
...
 
+func TestDeleteTablespaceDirectories(t *testing.T) {
+	utils.System.Hostname = func() (s string, err error) {
+		return "", nil
+	}
+	defer func() {
+		utils.System.Hostname = os.Hostname
+	}()
+
+	t.Run("deletes tablespace directories", func(t *testing.T) {
```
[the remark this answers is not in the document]
dkrieger-pvtl: Done.

agent/rsync_test.go
```diff
@@ -105,7 +105,7 @@ func TestRsync(t *testing.T) {
 			{Source: source, Destination: destination},
 		}}
 
-		_, err := server.Rsync(context.Background(), request)
+		_, err := server.RsyncDataDirectories(context.Background(), request)
```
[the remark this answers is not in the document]
dkrieger-pvtl: Done.

hub/restore_source_cluster.go
```diff
@@ -56,6 +56,33 @@ func RsyncMasterAndPrimaries(stream step.OutStreams, agentConns []*Connection, s
...
+	var wg sync.WaitGroup
+	errs := make(chan error, 2)
+
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		errs <- RsyncMasterTablespaces(stream, source.StandbyHostname(), tablespaces[source.Master().DbID], tablespaces[source.Standby().DbID])
+	}()
+
+	errs <- RsyncPrimariesTablespace(agentConns, source, tablespaces)
```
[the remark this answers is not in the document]
dkrieger-pvtl: Done.

hub/delete_data_directories_test.go
dkrieger-pvtl: DOne.

hub/delete_data_directories_test.go
dkrieger-pvtl: Done.

hub/revert.go
dkrieger-pvtl: DOne.

upgrade/directories.go
```diff
@@ -210,3 +213,83 @@ func DeleteDirectories(directories []string, requiredPaths []string, streams ste
...
+		// tablespace and is safe to delete.
+		// NOTE: Each directory passed in has a different parent.
+		if err := os.Remove(parent); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
```
dkrieger-pvtl, line 288: nit: perhaps we want to call this `NewTablespacePath`, or something similar, as this represents a tablespace path on the _target_ cluster.  Note that the input parameter _tablespaceLocation_ includes the tablespaceOID, which took me a while to figure out from our population of tablespaces in initialize.

upgrade/directories_test.go
```diff
@@ -444,6 +444,172 @@ func TestDeleteDirectories(t *testing.T) {
...
+
+		if upgrade.PathExists(tablespaceDir) {
+			t.Errorf("expected directory %q to be deleted", tablespaceDir)
+		}
+
+		if upgrade.PathExists(dbId) {
+			t.Errorf("expected parent dbid directory %q to be deleted", dbId)
+		}
+	})
+
```
dkrieger-pvtl, line 486: Consider adding a [multiple tablespace test](https://gist.github.com/dkrieger-pvtl/c6d8fd525e2502794163a0ced8f8d588).

hub/delete_data_directories_test.go
```diff
@@ -155,3 +163,220 @@ func TestDeleteSegmentDataDirs(t *testing.T) {
...
+				16386: {
+					Location:    "/tmp/testfs/master/demoDataDir-1/16386",
+					UserDefined: 1,
+				},
+				16387: {
+					Location:    "/tmp/testfs/master/demoDataDir-1/16387",
+					UserDefined: 1,
+				},
+				1663: {
+					Location:    "/data/qddir/demoDataDir-1",
```
dkrieger-pvtl, line 245: add a comment `// system tablespace locations do not include the tablespace oid`.

ci/tasks/noinstall-tests.yml
```diff
@@ -23,10 +23,16 @@ run:
...
     ./bats/install.sh /usr/local
 
-    export GOFLAGS="-mod=readonly" # do not update dependencies during build
-    cd gpupgrade_src
-    make
-    make check --keep-going
+    adduser  --disabled-password --gecos "" --ingroup tty --shell /bin/bash gpadmin
+    chmod -R a+w gpupgrade_src
+
+    su gpadmin -c '
```
jchampio, line 31: Please keep `set -ex` operation in the subshell. Yesterday, we saw examples of tests running after the build failed.

agent/delete_directories.go
```diff
@@ -30,3 +30,10 @@ func (s *Server) DeleteDataDirectories(ctx context.Context, in *idl.DeleteDataDi
 	err := DeleteDirectoriesFunc(in.Datadirs, upgrade.PostgresFiles, step.DevNullStream)
 	return &idl.DeleteDataDirectoriesReply{}, err
 }
+
+func (s *Server) DeleteTablespaceDirectories(ctx context.Context, in *idl.DeleteTablespaceRequest) (*idl.DeleteTablespaceReply, error) {
+	gplog.Info("received request to delete tablespace directories from the hub")
+
+	err := upgrade.DeleteTablespaceDirectories(step.DevNullStream, in.GetDirs())
```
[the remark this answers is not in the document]
dkrieger-pvtl: Decided to defer this.  There are arguments both ways: a single function encapsulates everything we are doing with tablespaces for deletion but two functions more closely aligns this to deleting directories on the agents.

ci/tasks/noinstall-tests.yml
dkrieger-pvtl: Curious...I had thought that `set` set some shell variables that were inherited in the subshell, but no:
```bash
#!/usr/bin/env bash

set -ex

bash <<EOF
  # you need the "set -ex" here or THERE does get printed
  set -ex
  asdfasf
  echo THERE
EOF
```

dkrieger-pvtl, line 31: It's probably worth adding a comment explaining why we use gpadmin. Something like:
`# running as root complicates writing tests due to root's ability to violate file permissions`

hub/restore_source_cluster_test.go
```diff
@@ -35,16 +35,79 @@ func TestRsyncMasterAndPrimaries(t *testing.T) {
...
+		{DbID: 2, ContentID: -1, Hostname: "standby", DataDir: "/data/standby", Role: greenplum.MirrorRole},
+		{DbID: 3, ContentID: 0, Hostname: "sdw1", DataDir: "/data/dbfast1/seg1", Role: greenplum.PrimaryRole},
+		{DbID: 4, ContentID: 0, Hostname: "msdw1", DataDir: "/data/dbfast_mirror1/seg1", Role: greenplum.MirrorRole},
+		{DbID: 5, ContentID: 1, Hostname: "sdw2", DataDir: "/data/dbfast2/seg2", Role: greenplum.PrimaryRole},
+		{DbID: 6, ContentID: 1, Hostname: "msdw2", DataDir: "/data/dbfast_mirror2/seg2", Role: greenplum.MirrorRole},
 	})
 	cluster.GPHome = "/usr/local/greenplum-db"
 	cluster.Version = dbconn.NewVersion("5.0.0")
 
+	tablespaces := greenplum.Tablespaces{
```
dkrieger-pvtl, line 48: nit: system defined tablespaces(`UserDefined: 0`) do not include the tablespace oid as their last path element...i.e. `/tmp/m/qddir/1664` should be `/tmp/m/qddir`.  This does not have a material impact as we ignore user defined tablespaces, but keeping this consistent improves code readability.

hub/restore_source_cluster_test.go
```diff
@@ -82,6 +145,51 @@ func TestRsyncMasterAndPrimaries(t *testing.T) {
...
+			}
+
+			source := args[3]
+			expected := "standby:/tmp/m/standby/1663/"
+			if source != expected {
+				t.Errorf("got source %q want %q", source, expected)
+			}
+
+			destination := args[4]
+			expected = "/tmp/m/qddir/1663"
```
dkrieger-pvtl, line 181: as above, the `1663` should not be here.

PR #409, opened 2020-07-29
test/finalize: move to teardown infrastructure

jchampio:
The existing teardown logic was extremely difficult to follow. We already had the backup/restore pattern in use for `migration_scripts`, so co-opt it here (and improve it by moving to the new `teardown_helpers` API).

### DIFF
--- a/test/finalize.bats
+++ b/test/finalize.bats
@@ -4,14 +4,19 @@
 # SPDX-License-Identifier: Apache-2.0
 
 load helpers
+load teardown_helpers
 load finalize_checks
 
 setup() {
     skip_if_no_gpdb
 
-    setup_state_dir
+    STATE_DIR=$(mktemp -d /tmp/gpupgrade.XXXXXX)
+    register_teardown rm -r "$STATE_DIR"
 
+    export GPUPGRADE_HOME="${STATE_DIR}/gpupgrade"
     gpupgrade kill-services
+
+    backup_source_cluster "$STATE_DIR"/backup
 }
 
 teardown() {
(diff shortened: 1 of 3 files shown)


test/finalize.bats
```diff
@@ -20,32 +25,46 @@ teardown() {
...
+}
 
-    # reload old path and start
-    source "${GPHOME_SOURCE}/greenplum_path.sh"
-    gpstart -a
+# XXX backup_source_cluster is a hack to work around the standby-revert bug.
+# Instead of relying on revert to correctly reset the state of the standby, copy
+# over the original cluster contents during teardown.
+#
+# Remove this and its companion ASAP.
```
jchampio, line 37: Do we want to promote this out of "remove this ASAP" to "maybe we start using it in more places"?

jchampio: (Revert tests are another place where a do-not-pass-Go backup/restore would be a good thing.)

test/finalize.bats
```diff
@@ -4,14 +4,19 @@
...
 load helpers
+load teardown_helpers
 load finalize_checks
 
 setup() {
     skip_if_no_gpdb
 
-    setup_state_dir
+    STATE_DIR=$(mktemp -d /tmp/gpupgrade.XXXXXX)
+    register_teardown rm -r "$STATE_DIR"
```
dkrieger-pvtl, line 14: use "${STATE_DIR:?}" to protect against a pathology in which STATE_DIR isn't set.

jchampio: `rm -r ""` isn't dangerous -- we've added that for constructions like `rm -rf "$dir"/`.

test/finalize.bats
bhuvnesh2703: Are we anytime planning to remove it? If not, sure we can promote it to start using... text

PR #410, opened 2020-07-29
atomically write config.json

kalensk:
Atomically write config.json. This includes extracting writing atomically to a helper.

For previous discussion see draft PR https://github.com/greenplum-db/gpupgrade/pull/403

[Pipeline](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:renameio2)

### DIFF
--- a/hub/server.go
+++ b/hub/server.go
@@ -4,6 +4,7 @@
 package hub
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
@@ -380,28 +381,12 @@ func (c *Config) Save(w io.Writer) error {
 
 // SaveConfig persists the hub's configuration to disk.
 func (s *Server) SaveConfig() (err error) {
-	// TODO: Switch to an atomic implementation like renameio. Consider what
-	// happens if Config.Save() panics: we'll have truncated the file
-	// on disk and the hub will be unable to recover. For now, since we normally
-	// only save the configuration during initialize and any configuration
-	// errors could be fixed by reinitializing, the risk seems small.
-	file, err := utils.System.Create(upgrade.GetConfigFile())
-	if err != nil {
-		return err
-	}
-	defer func() {
-		if cerr := file.Close(); cerr != nil {
-			cerr = xerrors.Errorf("closing hub configuration: %w", cerr)
-			err = multierror.Append(err, cerr).ErrorOrNil()
-		}
-	}()
-
-	err = s.Config.Save(file)
-	if err != nil {
-		return xerrors.Errorf("saving hub configuration: %w", err)
+	var buffer bytes.Buffer
+	if err = s.Config.Save(&buffer); err != nil {
+		return xerrors.Errorf("save config: %w", err)
 	}
 
-	return nil
+	return utils.AtomicallyWrite(upgrade.GetConfigFile(), buffer.Bytes())
 }
 
 func LoadConfig(conf *Config, path string) error {
--- a/hub/server_test.go
+++ b/hub/server_test.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"net"
 	"os"
+	"path/filepath"
 	"reflect"
 	"sort"
 	"strings"
@@ -14,7 +15,6 @@ import (
 	"time"
 
 	"github.com/greenplum-db/gp-common-go-libs/testhelper"
-	"github.com/hashicorp/go-multierror"
 	"github.com/pkg/errors"
 	"golang.org/x/xerrors"
 	"google.golang.org/grpc"
@@ -24,7 +24,7 @@ import (
 	"github.com/greenplum-db/gpupgrade/hub"
 	"github.com/greenplum-db/gpupgrade/testutils"
 	"github.com/greenplum-db/gpupgrade/testutils/mock_agent"
-	"github.com/greenplum-db/gpupgrade/utils"
+	"github.com/greenplum-db/gpupgrade/upgrade"
 )
 
 const timeout = 1 * time.Second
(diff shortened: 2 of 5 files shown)


utils/sys_utils_test.go
```diff
@@ -6,6 +6,8 @@ import (
 	"path/filepath"
 	"testing"
 
+	"github.com/greenplum-db/gpupgrade/upgrade"
```
dkrieger-pvtl, line 9: imports need to be sorted

hub/server_test.go
```diff
@@ -7,14 +7,16 @@ import (
...
 	"net"
 	"os"
+	"path/filepath"
 	"reflect"
 	"sort"
 	"strings"
 	"testing"
 	"time"
 
+	"github.com/greenplum-db/gpupgrade/upgrade"
```
dkrieger-pvtl, line 17: imports need to be sorted

utils/sys_utils.go
```diff
@@ -110,3 +113,23 @@ func Move(src string, dst string) error {
 
 	return err
 }
+
+func AtomicallyWrite(path string, data []byte) error {
```
dkrieger-pvtl, line 117: comment: I had thought that this interface reads more naturally with the data first, that is:

`AtomicallyWrite(data []byte, path string)`

but [ioutil.WriteFile](https://golang.org/pkg/io/ioutil/#WriteFile) has it in the current order.  So I guess that's the standard.

hub/server_test.go
kalensk: Fixed.

Any idea what the best settings for our project are in intelliJ. I recently reinstalled my operating system and my previous settings are gone. I just updated them now to the following, but I am not quite sure if they work.

```
Preferences > Editor > Code Style > Go > "Imports" tab

Sorting type = gofmt
And all of the following boxes checked:
- Group stdlib imports
- Move all stdlib imports in a single group
- Group current project imports
- Move all imports  in a single declaration
```

utils/sys_utils_test.go
kalensk: updated

utils/sys_utils.go
kalensk: Correct. That is what I have seen usually: `(path, data)` for example.

hub/server.go
```diff
@@ -380,28 +381,12 @@ func (c *Config) Save(w io.Writer) error {
...
-		}
-	}()
 
-	err = s.Config.Save(file)
-	if err != nil {
-		return xerrors.Errorf("saving hub configuration: %w", err)
-	}
-
-	return nil
+	return utils.AtomicallyWrite(upgrade.GetConfigFile(), buffer.Bytes())
```
jchampio, line 389: FYI: a downside of this approach is that we need to first buffer the contents in memory and then write them to disk, whereas the `renameio` API allows direct streaming via `io.Writer`. That's perfectly fine for this case, because the config file is small and this new API is intuitive, but keep it in mind if the use of `AtomicallyWrite()` starts to expand.

hub/server_test.go
dkrieger-pvtl: I also had `Sorting type = gofmt` but just switched it to `Sorting type = goimports`(setting 1).  The sorting example in that IntelliJ Preferences panel isn't clear to me.   That said, I also have a FileWatcher(Tools > File Watchers) with `goimports` set(setting 2).

That said, `Preferences > Languages& Frameworks > Go > Imports` has a checkbox `Add unambiguous imports on the fly` and `Optimize imports on the fly (for current project)`, both of which are checked for me.(setting 3)

As long as I don't add my own whitespace or imports, things always seem to "just work" with the three settings I've mentioned here.

Note that with these settings, the auto import formatting still respects intentionally added whitespace, ie. this doesn't resort.  
```go
import (
        ...

	"github.com/greenplum-db/gpupgrade/greenplum"
	"github.com/greenplum-db/gpupgrade/hub"
	"github.com/greenplum-db/gpupgrade/testutils"
	"github.com/greenplum-db/gpupgrade/testutils/mock_agent"
)
```

utils/sys_utils.go
```diff
@@ -110,3 +113,23 @@ func Move(src string, dst string) error {
...
+	// Use renameio to atomically write the file located at path.
+	file, err := renameio.TempFile("", path)
+	if err != nil {
+		return err
+	}
+	defer func() {
+		if cErr := file.Cleanup(); cErr != nil {
+			err = multierror.Append(err, cErr).ErrorOrNil()
```
jchampio, line 125: This `err` won't be returned; its scope ends at the top level of the function. You'll need to declare a named return value and make sure that `err` isn't rescoped. (I wonder if there's a lint option to catch this.)

utils/sys_utils_test.go
```diff
@@ -59,3 +60,36 @@ func TestMove(t *testing.T) {
...
+		}
+
+		contents := testutils.MustReadFile(t, path)
+		if contents != expected {
+			t.Errorf("wrote %#q want %q", contents, expected)
+		}
+	})
+
+	t.Run("errors when path does not exist", func(t *testing.T) {
+		path := "/does/not/exist"
```
jchampio, line 83: The wording and design of this test implies to me that `AtomicallyWrite` shouldn't create a file if it doesn't already exist, but I think the intent was to handle the case where the containing directory doesn't exist.

hub/server.go
dkrieger-pvtl: This was the original idea--to use the io.Writer method on the `file` returned by `renameio` functions.  But I found that code unintuitive.  And, yes, our config files are small now.  If we do this for large files(which these days I think means > 10MB or so), we'd likely need to revisit this.  Perhaps I am being unimaginative, but right now all I can think of for large files would be the need to change their directories(like we do in finalize), but that can be handled directly via `rename` as it doesn't suffer from the pathologies that `renameio` handles at the file level.

utils/sys_utils.go
dkrieger-pvtl: great catch....and surprising that the linter doesn't catch this.

PR #411, opened 2020-07-31
add global hub connection timeout flag

kalensk:
This adds a global hub connection timeout flag as a `PersistentFlag` to the root command which then gets applied to all sub commands (initialize, execute, finalize, revert, etc...).

[Pipeline](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:timeout)

### DIFF
--- a/cli/commands/commands.go
+++ b/cli/commands/commands.go
@@ -59,6 +59,7 @@ import (
 )
 
 var (
+	timeout        time.Duration
 	Help           map[string]string
 	InitializeHelp string
 	ExecuteHelp    string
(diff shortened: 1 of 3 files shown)


jchampio:
@Nirali13 is this a requested feature flag?

Nirali13:
> @Nirali13 is this a requested feature flag?

I am not sure what this feature is about. AFAIK, I don't recall discussing about hub connection timeout functionality.

cli/commands/commands.go
```diff
@@ -59,6 +59,7 @@ import (
 )
 
 var (
+	timeout        time.Duration
```
dkrieger-pvtl, line 62: As a global, I'd expect a variable called `timeout` to apply to all connections of some type, but it only applies to the hub.  So I'd recommend having a hub-specific name like `hubTimeout`.

kalensk:
Leaving this as is for now, as there is no pressing need.

PR #413, opened 2020-08-03
exectest: add standard success/failure Main implementations

jchampio:
Many of the current tests using `exectest` create just two `Main` implementations: one for success and one for failure. Since this pattern seems to be in wide use, pull it up into the `exectest` package itself, which now supplies two precreated `Main` implementations:

- `exectest.Success` exits with code 0
- `exectest.Failure` exits with code 1

### DIFF
--- a/testutils/exectest/mains_test.go
+++ b/testutils/exectest/mains_test.go
@@ -1,0 +1,40 @@
+package exectest_test
+
+import (
+	"errors"
+	"os/exec"
+	"testing"
+
+	. "github.com/greenplum-db/gpupgrade/testutils/exectest"
+)
+
+func TestBuiltinMains(t *testing.T) {
+	t.Run("Success()", func(t *testing.T) {
+		success := NewCommand(Success)("/unused/path")
+		err := success.Run()
+		if err != nil {
+			t.Errorf("exited with error %v", err)
+
+			var exitErr *exec.ExitError
+			if errors.As(err, &exitErr) {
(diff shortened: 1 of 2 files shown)


testutils/exectest/mains_test.go
```diff
@@ -0,0 +1,41 @@
...
+	t.Run("Failure()", func(t *testing.T) {
+		failure := NewCommand(Failure)("/unused/path")
+		err := failure.Run()
+		if err == nil {
+			t.Fatal("exited without an error")
+		}
+
+		var exitErr *exec.ExitError
+		if !xerrors.As(err, &exitErr) {
+			t.Fatalf("unexpected error %#v", err)
```
dkrieger-pvtl, line 35: nit: this test is unlikely to fail, but the error here is a type mismatch, so perhaps this error message is clearer: 
`t.Fatalf("error: %#v of wrong type: wanted type %T, got %T", err, exitErr, err)`
or just:
`t.Fatalf("wanted type %T, got %T", exitErr, err)`

kalensk: nit: Could also use `errors.As` over `xerrors.As`.

jchampio:
@dkrieger-pvtl I'm leaving that up to future commits.

testutils/exectest/mains_test.go
jchampio: Good points, will do. (This is cherry-picked from several months ago and I forgot to check for the use of more recent idioms.)

jchampio: Done.

PR #414, opened 2020-08-03
Remove `config set` subcommand

jchampio:
It's not useful anymore, and it's interfering with changes to `s.Target`.


test/config.bats
```diff
@@ -33,10 +29,7 @@ teardown() {
     fi
 }
 
-@test "configuration can be read after it is written" {
-    gpupgrade config set --target-gphome /usr/local/target
-    gpupgrade config set --source-gphome /usr/local/source
-
+@test "configuration can be read piece by piece" {
```
dkrieger-pvtl, line 32: For a follow-on story, it'd be good to test `--id` and `--target-datadir` as well here.

jchampio: That'd be ideal, yeah. `--target-datadir` would require waiting for a full target cluster creation, which would bloat this test file quite a bit.

PR #416, opened 2020-08-04
Use Initalize config to identify what directories to delete

bhuvnesh2703:
Revert may be executed even before the target cluster is created, so
it's better to rely on Initialize Config. This allows for cleanup of the
target cluster directories which could be a left over from a failed init
cluster during initialize

### DIFF
--- a/greenplum/segconfig.go
+++ b/greenplum/segconfig.go
@@ -17,6 +17,8 @@ type SegConfig struct {
 	Role      string
 }
 
+type SegConfigs []SegConfig
+
 const (
 	PrimaryRole = "p"
 	MirrorRole  = "m"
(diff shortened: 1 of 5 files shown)


greenplum/segconfig.go
```diff
@@ -84,3 +86,18 @@ func MustGetSegmentConfiguration(connection *dbconn.DBConn) []SegConfig {
 	gplog.FatalOnError(err)
 	return segConfigs
 }
+
+// SelectSegmentConfigs returns a list of all segments that match the given selector
+// function. Segments are visited in order of ascending content ID (primaries
+// before mirrors).
+func (s SegConfigs) SelectSegmentConfigs(selector func(*SegConfig) bool) SegConfigs {
```
jchampio, line 93: nit: `segs.Select()` would convey the same intent, I think, with less stutter.

PR #417, opened 2020-08-05
revert: always start the source cluster, if stopped

jchampio:
If the source has stopped for any reason after a failed revert, we still have to start it again. Move to `step.AlwaysRun()`.


jchampio:
Mis-pushed a branch, sorry -- one second while I rebase...

PR #418, opened 2020-08-06
verify tablespace directories

kalensk:
Verify tablespace directories before deleting and rsync'ing.

[Test pipeline](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:verify_ts)


kalensk:
Needs more investigation to verify when rsync'ing tablespace directories.

Currently, when rsync'ing we only pass the tablespace location directory rather than the full tablespace path such as `DIR/<fsname>/<datadir>/<tablespaceOid>/<dbOid>/<relfilenode>`. Thus, PG_VERSION is not found:
`invalid 5X source tablespace directory: missing /tmp/gpupgrade.N1mIVU/testfs/m1/demoDataDir0/16559/PG_VERSION`

PR #420, opened 2020-08-06
go.mod: make use of pgx explicit

dkrieger-pvtl:
We use the pgx module directly in our code as well as indirectly via
go-common-libs(as told by "go mod why github.com/jackc/pgx").  However
the current go.mod file lists pgx as an indirect dependency.

Include a direct import of pgx in the correct place, which for us is
the main function of the hub.

Co-authored-by: Jacob Champion <pchampion@vmware.com>

### DIFF
--- a/cmd/gpupgrade/main.go
+++ b/cmd/gpupgrade/main.go
@@ -11,6 +11,7 @@ import (
 	"strings"
 
 	"github.com/greenplum-db/gp-common-go-libs/gplog"
+	_ "github.com/jackc/pgx" // pgx is used directly in the hub via its global state
 	_ "github.com/lib/pq"
 
 	"github.com/greenplum-db/gpupgrade/cli"
--- a/go.mod
+++ b/go.mod
@@ -11,7 +11,7 @@ require (
 	github.com/google/renameio v0.1.0
 	github.com/greenplum-db/gp-common-go-libs v1.0.4
 	github.com/hashicorp/go-multierror v1.0.0
-	github.com/jackc/pgx v3.2.0+incompatible // indirect
+	github.com/jackc/pgx v3.2.0+incompatible
 	github.com/kballard/go-shellquote v0.0.0-20180428030007-95032a82bc51
 	github.com/lib/pq v1.3.0
 	github.com/onsi/gomega v1.7.1


kalensk:
> Include a direct import of pgx in the correct place, which for us is the main function of the hub.

Can you explain this more, and the construct of the import being done (ie: `_ "github.com/jackc/pgx"`).

I am assuming we use `pgx` throughout the hub - specifically, `initialize,go`, `update_catalog.go`, and `upgrade_mirrors.go` and adding the unused import to the hub main "_forces_" go.mod to include it as a direct dependency? Is this _actually_ how this works?

dkrieger-pvtl:
> > Include a direct import of pgx in the correct place, which for us is the main function of the hub.
> 
> Can you explain this a bit more, and the construct of the import being done (ie: `_ "github.com/jackc/pgx"`).
> 
> I am assuming we use `pgx` throughout the hub - specifically, `initialize,go`, `update_catalog.go`, and `upgrade_mirrors.go` and adding the unused import to the hub main "_forces_" go.mod to include it as a direct dependency? Is this _actually_ how this works?

Prior to this PR, pgx was listed as an indirect dependency; `go mod why` showed that it is being included by go-common-libs.  However, it is also used directly(a text search shows `"pgx"` is used directly in the three places in the hub you mentioned.)  The fact that it was working before is a "fluke"....since we include go-common-libs in hub _and_ it was coincidentally called prior to a use of pgx, it worked.

Yes, the `import _ github.com/jackc/pgx` [forces the initialization](https://golang.org/ref/spec#Import_declarations)  routines of pgx to be run, which is all we need.  It sets up some global handlers.  We only need to call it in once somewhere in the hub, but we know main is always called so we do it there.

cmd/gpupgrade/main.go
```diff
@@ -11,6 +11,7 @@ import (
 	"strings"
 
 	"github.com/greenplum-db/gp-common-go-libs/gplog"
+	_ "github.com/jackc/pgx"
```
kalensk, line 14: Let's add a comment to make it obvious that pgx is used in our code base, even though we are not using it explicitly here in main.go? Something like:
`// Include pgx as a direct import as its used throughout the hub.`

kalensk:
Thanks for the great clarification.

cmd/gpupgrade/main.go
dkrieger-pvtl: I think this is obvious from the semantics...that's the entire point of a `_` import.  If it's there, it means we import it for the initialization side effect.

dkrieger-pvtl: And if it's there at all, we must be using it directly.  (It's also in the go.mod file.)

dkrieger-pvtl: to make it more obvious, I added a comment.

PR #423, opened 2020-08-06
revert: if a directory does not exist its not error

bhuvnesh2703:
If a directory does not exist, it shouldn't be marked as an error.
Consider if delete was run on 2 segments, 1 finished and the other one
failed. Rerun of delete shouldn't error out because directory on the
host which finished successfully last time will not exist.

### DIFF
--- a/upgrade/directories.go
+++ b/upgrade/directories.go
@@ -175,34 +175,46 @@ func PathExists(path string) bool {
 	return err == nil
 }
 
+func verifyPathsExist(path string, files ...string) error {
+	var mErr *multierror.Error
+
+	for _, f := range files {
+		path := filepath.Join(path, f)
+		_, err := utils.System.Stat(path)
+		if err != nil {
+			mErr = multierror.Append(mErr, err)
+		}
+	}
+
+	return mErr.ErrorOrNil()
+}
+
 // Each directory in 'directories' is deleted only if every path in 'requiredPaths' exists
 // in that directory.
 func DeleteDirectories(directories []string, requiredPaths []string, streams step.OutStreams) error {
+	hostname, err := utils.System.Hostname()
+	if err != nil {
+		return err
+	}
+
 	var mErr *multierror.Error
 	for _, directory := range directories {
-		statError := false
-
-		for _, requiredPath := range requiredPaths {
-			filePath := filepath.Join(directory, requiredPath)
-			_, err := utils.System.Stat(filePath)
-			if err != nil {
-				mErr = multierror.Append(mErr, err)
-				statError = true
-			}
-		}
-
-		if statError {
-			continue
-		}
-
-		hostname, err := utils.System.Hostname()
-		if err != nil {
-			return err
-		}
-
+		gplog.Debug("Deleting directory: %q on host %q\n", directory, hostname)
 		_, err = fmt.Fprintf(streams.Stdout(), "Deleting directory: %q on host %q\n", directory, hostname)
 		if err != nil {
 			return err
+		}
+
+		if !PathExists(directory) {
+			fmt.Fprintf(streams.Stdout(), "directory: %q does not exist on host %q\n", directory, hostname)
+			gplog.Debug("Directory: %q does not exist on host %q\n", directory, hostname)
+			continue
+		}
+
+		err = verifyPathsExist(directory, requiredPaths...)
+		if err != nil {
+			mErr = multierror.Append(mErr, err)
+			continue
 		}
 
 		err = utils.System.RemoveAll(directory)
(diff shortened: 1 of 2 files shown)


upgrade/directories.go
```diff
@@ -178,10 +178,24 @@ func PathExists(path string) bool {
...
+		return err
+	}
+
 	var mErr *multierror.Error
 	for _, directory := range directories {
-		statError := false
+		_, err := utils.System.Stat(directory)
+		if os.IsNotExist(err) {
+			// directory may have been already removed during previous execution
+			fmt.Fprintf(streams.Stdout(), "directory: %q does not exist on host %q\n", directory, hostname)
```
dkrieger-pvtl, line 191: We do not stream back from the agents, so this message will not appear anywhere for them.  Perhaps we want to put this in the per-agent log file via `gplog` as well?  Note our usual scheme--stream for hub, message in err for agent--won't work here because this is an informational message, not an error.

upgrade/directories.go
```diff
@@ -195,11 +209,6 @@ func DeleteDirectories(directories []string, requiredPaths []string, streams ste
 			continue
 		}
 
-		hostname, err := utils.System.Hostname()
-		if err != nil {
-			return err
-		}
-
 		_, err = fmt.Fprintf(streams.Stdout(), "Deleting directory: %q on host %q\n", directory, hostname)
```
dkrieger-pvtl, line 212: same here: agent will never see this message.

upgrade/directories.go
```diff
@@ -178,10 +178,24 @@ func PathExists(path string) bool {
...
+		if os.IsNotExist(err) {
+			// directory may have been already removed during previous execution
+			fmt.Fprintf(streams.Stdout(), "directory: %q does not exist on host %q\n", directory, hostname)
+			continue
+		} else if err != nil {
+			mErr = multierror.Append(mErr, err)
+			continue
+		}
 
+		statError := false
```
dkrieger-pvtl, line 198: With the addition of the above loop,  I am finding this function hard to read. Maybe refactor the required dirs?

replace this block with:
```go
		err := VerifyDirectoriesExist(directory, requiredPaths...)
		if err != nil {
			continue
		}
```

```go
func VerifyDirectoriesExist(path string, files ...string) error {
	var mErr *multierror.Error

	for _, f := range files {
		path := filepath.Join(path, f)
		if err != nil {
		}
	}

	return mErr.ErrorOrNil()
}
```

upgrade/directories.go
kalensk: Note: That this method is shared across the master and agent, so even though the messages don't currently show up for the agents, they do for the master which is useful.

upgrade/directories.go
```diff
@@ -178,10 +178,24 @@ func PathExists(path string) bool {
...
 func DeleteDirectories(directories []string, requiredPaths []string, streams step.OutStreams) error {
+	hostname, err := utils.System.Hostname()
+	if err != nil {
+		return err
+	}
+
 	var mErr *multierror.Error
 	for _, directory := range directories {
-		statError := false
+		_, err := utils.System.Stat(directory)
```
dkrieger-pvtl, line 188: This block mixes two concerns: checking that the directory exists _and_ protecting the next check from running it not.  I think the code is cleaner(easier to read) if we just check for the existence of the directory.  If it is there, or if we could not so determine, the Verify check next is sufficient to tell us what went wrong.  So we do not end up checking this twice.

```
		// directory may have been already removed during previous execution
		if !PathExists(directory) {
			fmt.Fprintf(streams.Stdout(), "directory: %q does not exist on host %q\n", directory, hostname)
			continue
		}
```

dkrieger-pvtl: The two code suggestions above are summarized in this [gist](https://gist.github.com/dkrieger-pvtl/f49c51fc6d0de77f707ccf3b51294f12).

upgrade/directories_test.go
```diff
@@ -640,6 +687,44 @@ func TestDeleteTablespaceDirectories(t *testing.T) {
...
+			if upgrade.PathExists(dir.dbIdDir) {
+				t.Errorf("expected parent dbid directory %q to be deleted", dir.dbIdDir)
+			}
+		}
+
+		err = upgrade.DeleteNewTablespaceDirectories(step.DevNullStream, tsDirs)
+		if err != nil {
+			t.Errorf("unexpected error %+v", err)
+		}
+	})
```
dkrieger-pvtl, line 727: nit: This is the multiple tablespace version of the other test you added.  That's fine, but I am not sure what additional testing power this adds.  It's not necessary, but one potential test could be to simulate a partial failure by deleting all the tablespaces, adding a new one in, and then showing that only that new one is deleted successfully.  But that can wait until we test failures during revert.

PR #424, opened 2020-08-06
Rsync vs restore

bhuvnesh2703:
Note: This commit `revert: if a directory does not exist its not error` is part of a different PR https://github.com/greenplum-db/gpupgrade/pull/423

If the target cluster has not been started during execute, replacing pg_control.old to pg_control is sufficient for starting the source cluster. This commit incorporates this functionality.


hub/revert.go
```diff
@@ -78,13 +78,26 @@ func (s *Server) Revert(_ *idl.RevertRequest, stream idl.CliToHub_RevertServer)
...
+			st.Run(idl.Substep_RESTORE_SOURCE_CLUSTER, func(stream step.OutStreams) error {
+				if err := RsyncMasterAndPrimaries(stream, s.agentConns, s.Source); err != nil {
+					return err
+				}
+
+				return RsyncMasterAndPrimariesTablespaces(stream, s.agentConns, s.Source, s.Tablespaces)
+			})
+		} else {
+			// since the target cluster was not started, just restore pg_control.old
+			// to pg_control
```
dkrieger-pvtl, line 96: The justification for this is the [pg_upgrade documentation](https://www.postgresql.org/docs/9.4/pgupgrade.html).  Perhaps add `See https://www.postgresql.org/docs/9.4/pgupgrade.html` to this.

hub/revert.go
```diff
@@ -78,13 +78,26 @@ func (s *Server) Revert(_ *idl.RevertRequest, stream idl.CliToHub_RevertServer)
...
+			st.Run(idl.Substep_RESTORE_PGCONTROL, func(stream step.OutStreams) error {
+				return RestoreMasterAndPrimariesPgControl(s.agentConns, s.Source)
```
dkrieger-pvtl, line 98: I think we want to thread through the `streams` to provide detailed information in verbose mode to the user.  Though we haven't rethought our use of verbose mode, all the current functionality in general streams back data for the master.

upgrade/restore_pg_control.go
```diff
@@ -0,0 +1,39 @@
...
+		if err != nil {
+			return err
+		}
+
+		gplog.Debug("file %s exists", dst)
+		return nil
+	} else if err != nil {
+		return err
+	}
+
```
dkrieger-pvtl, line 36: I think the above logic is clearer if we just explicitly check for the condition we are interested in: that the rename has already been done.  The rename will fail if there's some other error.  So I'd suggest replaceing L23-33 with:
```go
    // the rename already happened before
    if !PathExists(src) && PathExists(dst) {
        gplog.Debug("file %q already renamed to %q", src, dst)
    }
```

hub/restore_source_cluster.go
```diff
@@ -29,6 +30,8 @@ var Excludes = []string{
 	"gp_dbid", "postgresql.conf", "backup_label.old", "postmaster.pid", "recovery.conf",
 }
 
+var RestoreMasterPgControl = upgrade.RestorePgControl
```
dkrieger-pvtl, line 33: This function injection is introduced to enable unit tests.  However, the unit tests that use this can, instead, use direct filesystem operations(temp directory with actual "pg_control" files).  The direct use of the function is clearer and easier to follow.

hub/revert.go
```diff
@@ -157,8 +174,29 @@ func (s *Server) Revert(_ *idl.RevertRequest, stream idl.CliToHub_RevertServer)
...
+			// if the primary upgrade substep was started, regardless of
+			primariesUpgraded, err := step.HasRun(idl.Step_EXECUTE, idl.Substep_UPGRADE_PRIMARIES)
+			if err != nil {
+				return err
+			}
+
+			if !primariesUpgraded {
+				return nil
+			}
+
```
dkrieger-pvtl, line 199: Is this optimization check useful relative to coding/testing/maintenance costs?  I am under the impression that an incremental recovery is relatively fast in any case.  The optimization above to prevent an unnecessary rsync certainly seems worthwhile.

dkrieger-pvtl: During discussion, we decided that this optimization is worth it was we already use it in checking that start can fail, so we do not even need to add an explicit test for it.

agent/restore_pg_control_test.go
```diff
@@ -0,0 +1,58 @@
...
+				t.Errorf("got error type %T, want %T", err, &os.PathError{})
+			}
+
+			if !strings.Contains(err.(*os.PathError).Path, dirs[i]) {
+				t.Errorf("got path %s, want %s", err.(*os.PathError).Path, dirs[i])
+			}
+		}
+	})
+
+	t.Run("finishes successfully", func(t *testing.T) {
```
jchampio, line 48: nit: I'm not sure this test gets us much. The `RestorePgControl` injection shouldn't even be called, if I'm reading the code correctly.

upgrade/restore_pg_control_test.go
```diff
@@ -0,0 +1,102 @@
...
+		if err != nil {
+			t.Errorf("expected file %s to exist, got error %#v", targetFile, err)
+		}
+	})
+
+	t.Run("re-run of RestorePgControl finishes successfully if RestorePgControl already succeeded before", func(t *testing.T) {
+		source := setupGlobalDir(t)
+		defer testutils.MustRemoveAll(t, source)
+
+		pgControlFile := filepath.Join(source, "global", "pg_control")
```
dkrieger-pvtl, line 62: nit: s/pgControlFile/dst.  This makes the intent of the test clearer.

agent/restore_pg_control_test.go
```diff
@@ -0,0 +1,58 @@
...
+		dirs := []string{"/tmp/test1", "/tmp/test2"}
+		_, err := server.RestorePrimariesPgControl(context.Background(), &idl.RestorePgControlRequest{Datadirs: dirs})
+		var mErr *multierror.Error
+		if !xerrors.As(err, &mErr) {
+			t.Errorf("error %#v does not contain type %T", err, mErr)
+			return
+		}
+
+		if len(dirs) != mErr.Len() {
+			t.Errorf("got error count %d, want %d", mErr.Len(), len(dirs))
```
jchampio, line 34: The code below this relies on the lengths to be the same in order to not panic, so this should probably be a `Fatalf`.

agent/restore_pg_control_test.go
```diff
@@ -0,0 +1,58 @@
...
+
+	"github.com/greenplum-db/gpupgrade/agent"
+	"github.com/greenplum-db/gpupgrade/idl"
+)
+
+func TestServer_RestorePrimariesPgControl(t *testing.T) {
+	testhelper.SetupTestLogger()
+	server := agent.NewServer(agent.Config{})
+
+	t.Run("bubbles up errors", func(t *testing.T) {
```
dkrieger-pvtl, line 24: It took me a bit to see that this test works by having no pg_control files.  Maybe rename the test to `bubbles up errors when no pg_control files exist`.

agent/restore_pg_control_test.go
```diff
@@ -0,0 +1,58 @@
...
+		dirs := []string{"/tmp/test1", "/tmp/test2"}
+		_, err := server.RestorePrimariesPgControl(context.Background(), &idl.RestorePgControlRequest{Datadirs: dirs})
+		var mErr *multierror.Error
+		if !xerrors.As(err, &mErr) {
+			t.Errorf("error %#v does not contain type %T", err, mErr)
+			return
+		}
```
kalensk, line 31: nit: I don't think we need the return, and can use a Fatal since if the types don't match we want to fail the test early, and we can slowly migrate away from xerrors to errors. Thus, the above section can be:

```
	if !errors.As(err, & mErr) {
			t.Fatalf("got error %#v want type %T", err, mErr)
		}
```

cli/commanders/steps.go
```diff
@@ -61,6 +61,8 @@ var SubstepDescriptions = map[idl.Substep]substepText{
 	idl.Substep_ARCHIVE_LOG_DIRECTORIES:                  substepText{"Archiving log directories...", "Archive log directories"},
 	idl.Substep_RESTORE_SOURCE_CLUSTER:                   substepText{"Restoring source cluster...", "Restore source cluster"},
 	idl.Substep_START_SOURCE_CLUSTER:                     substepText{"Starting source cluster...", "Start source cluster"},
+	idl.Substep_RESTORE_PGCONTROL:                        substepText{"Restoring source cluster...", "Restore source cluster"},
+	idl.Substep_RECOVERSEG_SOURCE_CLUSTER:                substepText{"Starting gprecoverseg...", "Start gprecoverseg"},
```
kalensk, line 65: 1) The Substep_RESTORE_PGCONTROL description is the same as Substep_RESTORE_SOURCE_CLUSTER. Shouldn't we differentiate them? That is, renaming to something like "Restoring source cluster pg_control file..."

2) I think the Substep_RECOVERSEG_SOURCE_CLUSTER description should be more precise with something like "Running gprecoverseg on the source cluster..."

hub/restore_source_cluster.go
```diff
@@ -213,3 +216,52 @@ func RsyncPrimariesTablespaces(agentConns []*Connection, source *greenplum.Clust
 
 	return ExecuteRPC(agentConns, request)
 }
+
+func RestorePrimariesPgControl(agentConns []*Connection, source *greenplum.Cluster) error {
```
kalensk, line 220: nit: Can `RestoreMasterAndPrimariesPgControl` come first in the file and then `RestoreMasterPgControl` followed by `RestorePrimariesPgControl`?  That is, rearrange the function placements within the file

I find it easier for me to read and follow when the entry point function comes first and then the other functions in order (when possible).

test/revert.bats
```diff
@@ -226,3 +226,108 @@ is_source_standby_running() {
...
+    "$PSQL" postgres --single-transaction -f - <<"EOF"
+        CREATE TABLE master_failure (a int, b int);
+        INSERT INTO master_failure SELECT i, i FROM generate_series(1,10)i;
+EOF
+
+    register_teardown "$PSQL" postgres -c "DROP TABLE IF EXISTS master_failure"
+
+    local file
+    file=$("$GPHOME_SOURCE"/bin/psql -d postgres -Atc "SELECT relfilenode FROM pg_class WHERE relname='master_failure';")
+    local dboid
```
jchampio, line 240: nit: this `local` can be combined with the previous one.

hub/revert.go
kalensk: Yes, please.

upgrade/restore_pg_control.go
```diff
@@ -0,0 +1,39 @@
...
+	"github.com/greenplum-db/gp-common-go-libs/gplog"
+
+	"github.com/greenplum-db/gpupgrade/utils"
+)
+
+// RestorePgControl restores pg_control.old to pg_control,
+// if pg_control.old is not existing, then pg_control
+// must exist, otherwise an error is thrown.
+func RestorePgControl(dir string) error {
```
jchampio, line 18: nit: can the documentation for this indicate that `dir` is a segment data directory, and not the directory containing the pg_control file? Alternatively, renaming it to `datadir` would make it clear.

test/revert.bats
```diff
@@ -226,3 +226,108 @@ is_source_standby_running() {
...
+    [ "$old_config" = "$new_config" ] || fail "actual config: $new_config, wanted: $old_config"
+
+    # ensure target cluster is down
+    ! isready "${GPHOME_TARGET}" ${target_master_port} || fail "expected target cluster to not be running on port ${target_master_port}"
+
+    is_source_standby_in_sync || fail "expected standby to eventually be in sync"
+}
+
+@test "reverting succeeds after copy-mode execute fails during master upgrade" {
+    setup_master_upgrade_failure
```
dkrieger-pvtl, line 326: This test would be clearer to me if this step were located between `gpugprade initialize` and `gpupgrade execute` in the above test, which I believe it can be.

test/revert.bats
```diff
@@ -226,3 +226,108 @@ is_source_standby_running() {
...
+        --temp-port-range ${target_master_port}-6040 \
+        --disk-free-ratio 0 \
+        --mode "$mode" \
+        --verbose 3>&-
+
+    # Execute should fail.
+    local status=0
+    gpupgrade execute --verbose || status=$?
+    [ "$status" -ne 0 ] || fail "expected execute to fail"
+
```
dkrieger-pvtl, line 289: I think this can be replaced with the equivalent and shorter:
```bash
    run gpupgrade execute --verbose
    [ "$status" -ne 0 ] || fail "expected execute to fail"
```

test/revert.bats
```diff
@@ -226,3 +226,108 @@ is_source_standby_running() {
...
+EOF
+
+    register_teardown "$PSQL" postgres -c "DROP TABLE IF EXISTS master_failure"
+
+    local file
+    file=$("$GPHOME_SOURCE"/bin/psql -d postgres -Atc "SELECT relfilenode FROM pg_class WHERE relname='master_failure';")
+    local dboid
+    dboid=$("$GPHOME_SOURCE"/bin/psql -d postgres -Atc "SELECT oid FROM pg_database WHERE datname='postgres';")
+    mv "$MASTER_DATA_DIRECTORY"/base/"$dboid"/"$file" "$MASTER_DATA_DIRECTORY"/base/"$dboid"/"$file".bkp
+    register_teardown mv "$MASTER_DATA_DIRECTORY"/base/"$dboid"/"$file".bkp "$MASTER_DATA_DIRECTORY"/base/"$dboid"/"$file"
```
jchampio, line 243: nit: probably easier/more maintainable to just combine all the quoted strings:
`"${MASTER_DATA_DIRECTORY}/base/${dboid}/${file}.bkp"`

upgrade/restore_pg_control_test.go
```diff
@@ -0,0 +1,102 @@
...
+	"github.com/greenplum-db/gpupgrade/upgrade"
+	"github.com/greenplum-db/gpupgrade/utils"
+)
+
+// Common setup required for multiple test cases
+func setupGlobalDir(t *testing.T) string {
+	source := testutils.GetTempDir(t, "")
+
+	globalDir := filepath.Join(source, "global")
+	err := utils.System.Mkdir(globalDir, 0755)
```
jchampio, line 25: Use `os` rather than `utils.System`; the latter is only helpful if you plan to inject an alternative implementation.

test/revert.bats
```diff
@@ -226,3 +226,108 @@ is_source_standby_running() {
...
+
+@test "reverting succeeds after copy-mode execute fails during master upgrade" {
+    setup_master_upgrade_failure
+    test_revert_after_execute_master_failure copy
+}
+
+@test "reverting succeeds after link-mode execute fails during master upgrade" {
+    setup_master_upgrade_failure
+    test_revert_after_execute_master_failure link
+}
```
dkrieger-pvtl, line 333: This added bats test has the most power in link-mode.  In copy-mode, you are testing existing code paths...do we want to keep the copy-mode version of this test?

test/revert.bats
```diff
@@ -226,3 +226,108 @@ is_source_standby_running() {
...
+
+    # Check to make sure the old cluster still matches
+    new_config=$(get_segment_configuration "${GPHOME_SOURCE}")
+    [ "$old_config" = "$new_config" ] || fail "actual config: $new_config, wanted: $old_config"
+
+    # ensure target cluster is down
+    ! isready "${GPHOME_TARGET}" ${target_master_port} || fail "expected target cluster to not be running on port ${target_master_port}"
+
+    is_source_standby_in_sync || fail "expected standby to eventually be in sync"
+}
```
dkrieger-pvtl, line 323: This test is clever.  It contains, however, much repeated code from `test_revert_after_execute()` that obscures the real intent of this test: that we do _not_ rsync.  I think removing all tests but that one here will make this intent clearer; the rest is tested above.

upgrade/restore_pg_control_test.go
```diff
@@ -0,0 +1,102 @@
...
+			if path != pgControlFile {
+				t.Errorf("got path %q, want %q", path, pgControlFile)
+			}
+
+			return nil, errors.New("permission denied")
+		}
+
+		err := upgrade.RestorePgControl(source)
+		if err == nil {
+			t.Errorf("expected error")
```
jchampio, line 99: nit: for a stronger test, you can store the error returned by `Stat()` and then check that `errors.Is(err, expected)`.

hub/revert.go
```diff
@@ -125,6 +142,11 @@ func (s *Server) Revert(_ *idl.RevertRequest, stream idl.CliToHub_RevertServer)
 		return DeleteStateDirectories(s.agentConns, s.Source.MasterHostname())
 	})
 
+	handleMirrorStartupFailure, err := s.mirrorStartupFailure()
```
jchampio, line 145: nit: when I'm skimming the source code, the name `mirrorStartupFailure()` doesn't contain an indication of what the method is doing. I know what it's doing because I wrote the code with you, but in a few months that might fade... How about `expectMirrorFailure()` or similar?

hub/restore_source_cluster_test.go
```diff
@@ -448,3 +448,188 @@ func TestRsyncMasterAndPrimaries(t *testing.T) {
...
+		}
+
+		err := hub.RestorePrimariesPgControl(agentConns, cluster)
+		if err != nil {
+			t.Errorf("unexpected err %#v", err)
+		}
+	})
+}
+
+func TestRestoreMasterAndPrimariesPgControl(t *testing.T) {
```
jchampio, line 538: Nit: it looks like this function's tests subsume the ones above it. Is `RestorePrimariesPgControl` worth exporting and testing separately?

test/revert.bats
bhuvnesh2703: help me understand why it will be clearer?
It is currently:
- Start with a state which is bad, do initialize and then execute..

agent/restore_pg_control_test.go
bhuvnesh2703: oh right, fixed it.

test/revert.bats
jchampio: > the real intent of this test: that we do not rsync.

That wasn't my intent for the test -- the intent was to prove that revert works fully. We currently have a test in there that also checks for an optimization, which is fine with me, but not the intent of the test.

test/revert.bats
jchampio: There's a new code path that needs to be tested -- during copy mode, if we fail during master, we don't need to perform segment recovery.

cli/commanders/steps.go
bhuvnesh2703: i agree, but for now, i will leave it as is and do a followup with peter and change it accordingly.

hub/revert.go
```diff
@@ -78,13 +78,30 @@ func (s *Server) Revert(_ *idl.RevertRequest, stream idl.CliToHub_RevertServer)
...
+				return RsyncMasterAndPrimariesTablespaces(stream, s.agentConns, s.Source, s.Tablespaces)
+			})
+		} else {
+			// since the target cluster was not started, just restore pg_control.old
+			// to pg_control
+			st.Run(idl.Substep_RESTORE_PGCONTROL, func(stream step.OutStreams) error {
+				return RestoreMasterAndPrimariesPgControl(s.agentConns, s.Source)
+			})
+		}
 	}
```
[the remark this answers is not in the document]
bhuvnesh2703: in the code snippet you shared, RestoreMasterAndPrimariesPgControl happens twice. But i am open to refactoring, however, i need to take feedback on the user experience with Peter and then refactor it accordingly. For now, keeping it as is.

test/revert.bats
jchampio: It's not equivalent -- it hides all output from `execute`, nixing the point of the `--verbose`. Be careful when using `run`.

bhuvnesh2703:
@dkrieger-pvtl @kalensk applied the feedback, pls take a look on fixups

hub/revert.go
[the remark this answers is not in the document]
dkrieger-pvtl: Sure...the main issue to me is whether we want to tell the user that we are rewriting the pg_control files explicitly.  The code structure is more of a nit.

test/revert.bats
dkrieger-pvtl: So, in other words, that rsync works fully when execute fails in this specific place, as opposed to after a successful execute as above.  That makes sense.  I think that's clear from what's there, I guess I wasn't paying attention enough to the function names.

kalensk: I agree with @dkrieger-pvtl in that revert.bats now has a lot of repeated assertions between `test_revert_after_execute ` and `test_revert_after_execute_master_failure`. This repetition makes it hard for me to follow what is going on and the intent of the revert tests as a whole, etc.

test/revert.bats
dkrieger-pvtl: I guess your are correct...we do not want to tweak the cluster after upgrade starts.  We want to start with a bad cluster.  Ideally we'd catch that during initialize, but we do not currently.    

The comment came because when I read `# Execute should fail.` above, it took me a bit to figure out why.  Perhaps changing the comment to `# Execute should fail...we called setup_master_upgrade_failure before this function` or something similar.

test/revert.bats
dkrieger-pvtl: And now that I understand the intent of the test clearly(test revert after an upgrade failure), it makes sense to test for copy and link mode.

agent/restore_pg_control.go
```diff
@@ -0,0 +1,28 @@
...
+	"context"
+
+	"github.com/hashicorp/go-multierror"
+
+	"github.com/greenplum-db/gpupgrade/idl"
+	"github.com/greenplum-db/gpupgrade/step"
+	"github.com/greenplum-db/gpupgrade/upgrade"
+)
+
+var RestorePgControl = upgrade.RestorePgControl
```
dkrieger-pvtl, line 16: Please remove this...it is no longer used other than in this file.

cli/commanders/steps.go
```diff
@@ -61,6 +61,8 @@ var SubstepDescriptions = map[idl.Substep]substepText{
 	idl.Substep_ARCHIVE_LOG_DIRECTORIES:                  substepText{"Archiving log directories...", "Archive log directories"},
 	idl.Substep_RESTORE_SOURCE_CLUSTER:                   substepText{"Restoring source cluster...", "Restore source cluster"},
 	idl.Substep_START_SOURCE_CLUSTER:                     substepText{"Starting source cluster...", "Start source cluster"},
+	idl.Substep_RESTORE_PGCONTROL:                        substepText{"Restoring source cluster...", "Restore source cluster"},
```
dkrieger-pvtl, line 64: Please make the strings notionally correct for now, with something like `Restoring source pg_control`.  We can iterate later but right now might really confuse us when debugging.

hub/restore_source_cluster_test.go
```diff
@@ -448,3 +451,131 @@ func TestRsyncMasterAndPrimaries(t *testing.T) {
...
+			{nil, sdw1, "sdw1", nil},
+			{nil, sdw2, "sdw2", nil},
+		}
+
+		err := hub.RestoreMasterAndPrimariesPgControl(step.DevNullStream, agentConns, cluster)
+		if err != nil {
+			t.Errorf("unexpected err %#v", err)
+		}
+	})
+}
```
dkrieger-pvtl, line 581: In both of the tests here, you can test using direct filesystem commands.  You can set up a tempdir and check the filesystem directly.  Note your tests would be easier to understand if you had separate functions for modifying the master vs the segments like the other tests that follow the "do concurrently on master and segments" pattern.

test/revert.bats
dkrieger-pvtl: Good point.

test/revert.bats
dkrieger-pvtl: We can use the current code as a general pattern: set up `gpupgrade execute` to fail at a specific substep and then run the general code pattern.  We will need some conditional checks but I think it will be more readable as we'll have a single function.  Basically, we run all the post-checks if we go all the way through, and if we intentionally fail at earlier substeps, we only run a subset of the checks.

But I think we can push this to another PR...we'll likely need a full story to test `execute` failing at/during each substep.

upgrade/restore_pg_control_test.go
```diff
@@ -0,0 +1,111 @@
...
+			}
+
+			return nil, expectedError
+		}
+
+		err := upgrade.RestorePgControl(source, step.DevNullStream)
+		if !errors.Is(err, expectedError) {
+			t.Errorf("got error type %T, want %T", err, expectedError)
+		}
+	})
```
dkrieger-pvtl, line 110: This test can use the filesystem directly to test this.  See  https://github.com/greenplum-db/gpupgrade/blob/master/upgrade/directories_test.go#L654 for one technique we use.

bhuvnesh2703: with the recent fixups, i had made the change wherever applicable expect for one of the use case, where i wanted stat to return an error which is anything other than not exist because chmod doesn't help there to get such an error. we can discuss and spend some time to find out other ways (we tried few)

hub/revert.go
```diff
@@ -78,13 +78,31 @@ func (s *Server) Revert(_ *idl.RevertRequest, stream idl.CliToHub_RevertServer)
...
+				if err := RsyncMasterAndPrimaries(stream, s.agentConns, s.Source); err != nil {
+					return err
+				}
+
+				return RsyncMasterAndPrimariesTablespaces(stream, s.agentConns, s.Source, s.Tablespaces)
+			})
+		} else {
+			// Since the target cluster was not started, just restore pg_control.old
+			// to pg_control.
+			// Refer: https://www.postgresql.org/docs/9.4/pgupgrade.html
```
dkrieger-pvtl, line 101: nit.  Instead of this, maybe:

```
			// Since the target cluster was not started, just restore pg_control.old
			// to pg_control.  See the "Reverting to old cluster" section of 
                        // https://www.postgresql.org/docs/9.4/pgupgrade.html
```

hub/revert.go
```diff
@@ -175,3 +198,25 @@ func (s *Server) Revert(_ *idl.RevertRequest, stream idl.CliToHub_RevertServer)
 
 	return st.Err()
 }
+
+// If revert was run after the step to upgrade primary, the mirror segments can fail to start
+// due to invalid checkpoint if the restoration of source primaries was not performed from the
+// corresponding mirrors.
```
dkrieger-pvtl, line 204: nit.  I know I'll find this comment cryptic in a month, though it makes sense right now.  Maybe be more explicit.  Something like:
```
// in 5X, running pg_upgrade on the primaries can cause the mirrors to receive an invalid
// checkpoint upon starting. There are two ways to resolve this: running 
// Substep_RESTORE_SOURCE_CLUSTER or running Substep_RECOVERSEG_SOURCE_CLUSTER.
// If the former hasn't run yet, then we do expect mirror failure upon start, so return true.
```

hub/revert.go
```diff
@@ -157,8 +180,8 @@ func (s *Server) Revert(_ *idl.RevertRequest, stream idl.CliToHub_RevertServer)
 		return nil
 	})
 
-	if !s.UseLinkMode {
-		st.Run(idl.Substep_RESTORE_SOURCE_CLUSTER, func(streams step.OutStreams) error {
+	if handleMirrorStartupFailure {
```
dkrieger-pvtl, line 183: nit: we only need to run this on a 5X source cluster, so we could conditionalize this.  At the very least, a TODO comment would be good here.

upgrade/restore_pg_control.go
```diff
@@ -0,0 +1,42 @@
...
+	src := filepath.Join(globalDir, "pg_control.old")
+	dst := filepath.Join(globalDir, "pg_control")
+
+	renamed, err := AlreadyRenamed(src, dst)
+	if err != nil {
+		return err
+	}
+
+	if renamed {
+		gplog.Debug("already renamed %q to %q", src, dst)
```
dkrieger-pvtl, line 31: nit: I think we also want to send this message to `streams.Stdout()`.

upgrade/restore_pg_control.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/greenplum-db/gp-common-go-libs/gplog"
+
+	"github.com/greenplum-db/gpupgrade/step"
+	"github.com/greenplum-db/gpupgrade/utils"
+)
+
+// RestorePgControl renames pg_control.old to pg_control,
+// if pg_control.old is not existing, then pg_control
+// must exist, otherwise an error is thrown.
+// The dir passed to RestorePgControl must be the segment data directory.
```
dkrieger-pvtl, line 19: nit: I find this comment a bit confusing.  Maybe:
```
// RestorePgControl renames the segment data directory file
// pg_control.old to pg_control so that the cluster can start.
// It is idempotent as it already checks if the rename has occurred.
```

nit: note the make sure that the pg_control files are part of a segment via the existence of the `global` directory.  That's probably fine for now.

upgrade/directories.go
```diff
@@ -168,15 +168,44 @@ func VerifyDataDirectory(path string) error {
 	return mErr.ErrorOrNil()
 }
 
+// TODO: Remove alreadyRenamed and use AlreadyRenamed
 func alreadyRenamed(archive, target string) bool {
 	return PathExists(archive) && !PathExists(target)
 }
 
+func AlreadyRenamed(src, dst string) (bool, error) {
```
dkrieger-pvtl, line 176: nit.  A comment would be nice since the caller of this function knows its purpose but its not clear from the implementation.   Maybe something like:
```
// AlreadyRenamed infers if a successful os.Rename(src, dst) has already occurred
// by making sure src does not exist but dst does.
```

hub/revert.go
bhuvnesh2703: handleMirrorStartupFailure already incorporates that.

PR #427, opened 2020-08-07
Move tablespace info to cluster

dkrieger-pvtl:
Currently, the tablespace information is in the Config struct, but it is semantically attached to the source cluster.  So this PR moves that information into the greenplum.Cluster struct.

This is a DRAFT PR, as I am not sure of the history of keeping that information where it currently is.

[test pipeline](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:move_tablespace_info_to_cluster)

### DIFF
--- a/greenplum/cluster.go
+++ b/greenplum/cluster.go
@@ -40,6 +40,8 @@ type Cluster struct {
 	GPHome  string
 	Version dbconn.GPDBVersion
 
+	// Tablespaces contains the tablespace in the database keyed by
+	// dbid and tablespace oid
 	Tablespaces                Tablespaces
 	TablespacesMappingFilePath string
 }
--- a/greenplum/cluster.go
+++ b/greenplum/cluster.go
@@ -39,6 +39,9 @@ type Cluster struct {
 
 	GPHome  string
 	Version dbconn.GPDBVersion
+
+	Tablespaces                Tablespaces
+	TablespacesMappingFilePath string
 }
 
 // ClusterFromDB will create a Cluster by querying the passed DBConn for
--- a/hub/copy_master.go
+++ b/hub/copy_master.go
@@ -86,7 +86,7 @@ func (s *Server) CopyMasterDataDir(streams step.OutStreams, destination string) 
 }
 
 func (s *Server) CopyMasterTablespaces(streams step.OutStreams, destinationDir string) error {
-	if s.Tablespaces == nil {
+	if s.Source == nil || s.Source.Tablespaces == nil {
 		return nil
 	}
 
@@ -94,7 +94,7 @@ func (s *Server) CopyMasterTablespaces(streams step.OutStreams, destinationDir s
 	sourcePaths := []string{s.Source.TablespacesMappingFilePath}
 
 	// include all the master tablespace directories
-	for _, tablespace := range s.Tablespaces.GetMasterTablespaces() {
+	for _, tablespace := range s.Source.Tablespaces.GetMasterTablespaces() {
 		if !tablespace.IsUserDefined() {
 			continue
 		}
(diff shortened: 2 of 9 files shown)


kalensk:
Makes sense to me. The code looks good to me as a straight change.

jchampio:
> Currently, the tablespace information is in the Config struct, but it is semantically attached to the source cluster.

I don't think this is a good enough reason to pull something into an "object". I think the hints that this is not something that really belongs as part of a `Cluster` are
- we only ever use the `Source`'s tablespace information
- nothing inside the `Cluster` needs the tablespace information; we're using the new field as a cache rather than something that's part of the object
- tablespace information is something that _belongs_ to a cluster, but not something that _makes up_ a cluster. There are other ways to express that relationship in code (one of them is in use currently), that keep the object's responsibilities focused. See aggregation vs. composition.

If we needed tablespace information for both clusters, I might see a case for this, but as-is I don't.

kalensk:
Another idea is to rename `s.Tablespaces` to `s.SourceTablespaces` to better differentiate between tablespaces on the source and target cluster.

dkrieger-pvtl:
> * tablespace information is something that _belongs_ to a cluster, but not something that _makes up_ a cluster. There are other ways to express that relationship in code (one of them is in use currently), that keep the object's responsibilities focused. See aggregation vs. composition.

I think tablespace information, in fact, _makes up_ a cluster.  The tablespace information is natively part of a cluster; it contains database tables.  It's just as important to the cluster as the primaries and mirrors.  The tablespace information does not exist independently of the cluster itself.  This argues for composition to me, that is, having it part of the cluster.  It is technically a cache, but so is much of the current Cluster--we store data there because the source cluster is not up always when we need that information.  

The fact that we do not need tablespace information for the target in the Cluster itself is an implementation detail. We _do_ in fact need tablespace information for the deleting the target tablespace directories in revert.  We just happen to construct those paths in a way that doesn't associate them with the Cluster right now. We _could_ do so, say, just after upgrading the primaries in execute.  Or we could do so in finalize.

The above is for the tablespace information itself.  The path to the file containing gpupgrade's _is_ an implementation detail of gpupgrade, and isn't part of the cluster itself.  So I could see that filename in Config, while the tablespace information  could be in Cluster.

Overall, the motivation for this PR is that I find the current tablespace code hard to read, and this helps me read the code as Cluster now models an actual Greenplum installation more closely.

jchampio:
> I think tablespace information, in fact, _makes up_ a cluster. The tablespace information is natively part of a cluster; it contains database tables.

When I say "cluster" I'm talking about the `Cluster` object API here. It's important to make that distinction. There are many, many things that comprise an actual GPDB cluster deployment that I would also argue against including in the `Cluster`.

When I say "makes up" I mean "the object is not useful to us unless it has that information". What you've constructed is basically an attached property -- the object itself doesn't care what that property contains, and none of its behavior changes. There are cases where that tradeoff makes sense (for example, I think `Cluster.Version` is a case where we got that tradeoff right), but I don't see it here yet.

> It is technically a cache, but so is much of the current Cluster--we store data there because the source cluster is not up always when we need that information.

The other cached data is critical to the `Cluster` API. If a `Cluster` didn't know how many primaries and mirrors it had, and where all the segments were, it would not be nearly as useful to us. That's not true of cached tablespace data -- `s.Target` is perfectly usable without it.

> The fact that we do not need tablespace information for the target in the Cluster itself is an implementation detail. We _do_ in fact need tablespace information for the deleting the target tablespace directories in revert.

I think it would be important to see an implementation that made use of that before deciding that having tablespace info on a target `Cluster` would help us out. As it is now, it's a pretty big violation of the Rule of Three (or even Two).

There are probably useful APIs that make use of cached tablespace data. But let's find and develop those APIs before attaching this data to `Cluster`.

> Overall, the motivation for this PR is that I find the current tablespace code hard to read

Does Kalen's suggestion solve that problem? `s.SourceTablespaces` seems just as readable as `s.Source.Tablespaces`.

dkrieger-pvtl:
There is enough disagreement that I'll pull this one for now and try a simple in-place name change of the field to see if that adds enough readability.

PR #428, opened 2020-08-11
remove gomega as a dependency

dkrieger-pvtl:
We remove gomega as a dependency. 

The only use was for the testlogger comparison routines.  So we replace it by:

1). defining our own synchronized buffer.
2). implementing our own test logger setup that uses that buffer.
3). using that test logger everywhere
4). removing the dependency.

Ideally, the SyncBuf would be in testutils directly, but due to pre-existing dependency cycles(testutils includes greenplum and upgrade, and greenplum includes upgrade) it is simplest for now to place the log code into its own package.

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -1,8 +1,6 @@
 cloud.google.com/go v0.26.0/go.mod h1:aQUYkXzVsufM+DwF1aE+0xfcU+56JwCaLick0ClmMTw=
 github.com/BurntSushi/toml v0.3.1/go.mod h1:xHWCNGjB5oqiDr8zfno3MHue2Ht5sIBksp03qcyfWMU=
 github.com/DATA-DOG/go-sqlmock v1.3.3/go.mod h1:f/Ixk793poVmq4qj/V1dPUg2JEAKC73Q5eFN3EC/SaM=
-github.com/DATA-DOG/go-sqlmock v1.4.0 h1:yxQ63CFIA8Sxkh0vqIofuNrsXl/LZ42TpeTLV4Nb5HM=
-github.com/DATA-DOG/go-sqlmock v1.4.0/go.mod h1:3TucWNLPFOLcHhha1CPp7Kis1UG2h/AqGROPyOeZzsM=
 github.com/DATA-DOG/go-sqlmock v1.4.1 h1:ThlnYciV1iM/V0OSF/dtkqWb6xo5qITT1TJBG1MRDJM=
 github.com/DATA-DOG/go-sqlmock v1.4.1/go.mod h1:f/Ixk793poVmq4qj/V1dPUg2JEAKC73Q5eFN3EC/SaM=
 github.com/OneOfOne/xxhash v1.2.2/go.mod h1:HSdplMjZKSmBqAxg5vPj2TmRDmfkzw+cTzAElWljhcU=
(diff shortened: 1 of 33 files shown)


utils/syncbuf/syncbuf.go
```diff
@@ -0,0 +1,44 @@
...
+	return s.buf.Read(b)
+}
+
+func (s *Syncbuf) Bytes() []byte {
+	s.Lock()
+	defer s.Unlock()
+
+	// copy contents into a buffer that does not expose our internals
+	var buf bytes.Buffer
+	buf.Write(s.buf.Bytes())
```
kalensk, line 41: I was looking at possibly using [`io.Copy`](https://golang.org/pkg/io/#Copy), but I suppose that does not get us much in this case.

utils/syncbuf/syncbuf_test.go
```diff
@@ -0,0 +1,75 @@
...
+				if err != nil {
+					t.Errorf("Write failed %+v", err)
+				}
+			}()
+		}
+
+		wg.Wait()
+
+		// Check that seq1 or seq2 are written in any order the correct number of times.
+		totalBytes := (len(seq1) + len(seq2)) * runs
```
kalensk, line 58: minor nit: Maybe rename to just `length` since the use of bytes implies this is a byte, whereas its actually the size as an integer.

kalensk:
Good change! Happy to see the removal of gomega. Hope to see the same for common-go-libs one day.

utils/syncbuf/syncbuf.go
dkrieger-pvtl: I agree; I do not see how this makes things cleaner for us.

utils/syncbuf/syncbuf_test.go
dkrieger-pvtl: I changed it to `expectedLength`, since that clearly expresses our intent.   As an aside, I'd read `totalBytes` as a total of something in units of bytes, not as type byte.

utils/syncbuf/syncbuf.go
```diff
@@ -32,6 +32,8 @@ func (s *Syncbuf) Read(b []byte) (n int, err error) {
 	return s.buf.Read(b)
 }
 
+// Bytes() returns the current contents of the Syncbuf, without
+//  modifying it.
```
jchampio, line 36: General comment -- can you remove the hanging indent(s)? No other godoc uses them, AFAICT. Full paragraphs separated by an empty line are typical and encouraged, so it doesn't really help readability.

go.sum
```diff
@@ -48,8 +48,6 @@ github.com/golang/glog v0.0.0-20160126235308-23def4e6c14b h1:VKtxabqXZkF25pY9ekf
 github.com/golang/glog v0.0.0-20160126235308-23def4e6c14b/go.mod h1:SBH7ygxi8pfUlaOkMMuAQtPIUF8ecWP5IEl/CR7VP2Q=
```
jchampio, line 48: Did `go mod tidy` clean up the sums?

testutils/testlog/log.go
```diff
@@ -0,0 +1,42 @@
...
+	"strings"
+	"testing"
+
+	"github.com/greenplum-db/gp-common-go-libs/gplog"
+	"github.com/greenplum-db/gpupgrade/utils/syncbuf"
+)
+
+func SetupLogger() (*syncbuf.Syncbuf, *syncbuf.Syncbuf, *syncbuf.Syncbuf) {
+	testStdout := syncbuf.New()
+	testStderr := syncbuf.New()
```
jchampio, line 13: Future cleanup: it seems like nothing ever uses these.

utils/syncbuf/syncbuf.go
dkrieger-pvtl: Done.

jchampio:
> greenplum includes upgrade

I thought we got rid of those sorts of cycles. When did this reappear?

go.sum
dkrieger-pvtl: It did somewhat.  I just re-ran it and it cleaned them up some more.

testutils/testlog/log.go
dkrieger-pvtl: Right, but I didn't want to modify this right now as we weren't sure of the long-term intent of this interface.  TODO comment added.

dkrieger-pvtl:
> > greenplum includes upgrade
> 
> I thought we got rid of those sorts of cycles. When did this reappear?

Back in May for PathExists.  Likely we want some check to make sure our dependencies stay sane.

PR #435, opened 2020-08-14
cli: don't log errors twice and return errors from connectToHub

kalensk:
Two CLI cleanup and/or possible pre-factor commits. 
- Don't log errors twice. Since main already logs errors including their stacktrace, so there is no need to log the same error in commands.go.
- Return errors from connectToHub. Now that all commands are using RunE and main logs errors, update connectToHub to return errors rather than logging.

[Pipeline](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:refactorCommands)

### DIFF
--- a/cli/commands/commands.go
+++ b/cli/commands/commands.go
@@ -536,12 +536,11 @@ func finalize() *cobra.Command {
 		Use:   "finalize",
 		Short: "finalizes the cluster after upgrade execution",
 		Long:  FinalizeHelp,
-		Run: func(cmd *cobra.Command, args []string) {
+		RunE: func(cmd *cobra.Command, args []string) error {
 			client := connectToHub()
 			response, err := commanders.Finalize(client, verbose)
 			if err != nil {
-				gplog.Error(err.Error())
-				os.Exit(1)
+				return err
 			}
 
 			message := fmt.Sprintf(`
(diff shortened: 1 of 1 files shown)


cli/commands/commands.go
```diff
@@ -216,13 +216,11 @@ func connectToHubOnPort(port int) idl.CliToHubClient {
 		// Print a nicer error message if we can't connect to the hub.
 		if ctx.Err() == context.DeadlineExceeded {
 			gplog.Error("could not connect to the upgrade hub (did you run 'gpupgrade initialize'?)")
-		} else {
-			gplog.Error(err.Error())
 		}
-		os.Exit(1)
+		return nil, err
```
dkrieger-pvtl, line 220: This PR has pretty minimal information printed to the console, basically:

`Error: <whatever error grpc.DialContext provides>`

and then the user will have to go to the log to figure out what is happening(and the user will need to know to go to the log).  We should provide more context.  Maybe something like wrapping the only error in this function with something like:

`xerrors.Errorf("hub connection to port %d: %w", port, err)`.

Note this is not new in this PR, but nonetheless is straightforward to add and I think quite useful.

kalensk: Good point. I added context to the error, so the user will now see something like: `Error: connecting to hub on port 7527: context deadline exceeded`

PR #437, opened 2020-08-14
Run revert.bats on CCP

jchampio:
Add jobs to the pipeline to run BATS tests against a CCP cluster, and rewrite pieces of `revert.bats` to run against an arbitrary cluster.

### DIFF
--- a/ci/parser/parse_template.go
+++ b/ci/parser/parse_template.go
@@ -72,6 +72,22 @@ func (j *UpgradeJob) BaseName() string {
 	return fmt.Sprintf("%s-to-%s%s", j.Source, j.Target, suffix)
 }
 
+type MultihostBatsJob struct {
+	Source, Target string
+	CentosVersion  string
+}
+
+func (j *MultihostBatsJob) Name() string {
+	return fmt.Sprintf("%s-centos-%s", j.BaseName(), j.CentosVersion)
+}
+
+// BaseName returns the pipeline job name without the operating system.
+// This is used as a tag in Concourse's serial group to limit similar jobs
+// between operating systems from running at once to avoid overloading Concourse.
+func (j *MultihostBatsJob) BaseName() string {
+	return fmt.Sprintf("%s-to-%s-%s", j.Source, j.Target, "multihost-bats")
+}
+
 type Version struct {
 	CentosVersion string
 	GPVersion     string
@@ -83,6 +99,7 @@ type Data struct {
 	LastTargetVersion string
 	Versions          []*Version
 	CheckJobs         []*CheckJob
+	MultihostBatsJobs []*MultihostBatsJob
 }
 
 var data Data
@@ -90,6 +107,7 @@ var data Data
 func init() {
 	var checkJobs []*CheckJob
 	var upgradeJobs []*UpgradeJob
+	var multihostBatsJobs []*MultihostBatsJob
 	for _, sourceVersion := range sourceVersions {
 		for _, targetVersion := range targetVersions {
 			checkJobs = append(checkJobs, &CheckJob{
@@ -98,6 +116,11 @@ func init() {
 			})
 			for _, centosVersion := range centosVersions {
 				upgradeJobs = append(upgradeJobs, &UpgradeJob{
+					Source:        sourceVersion,
+					Target:        targetVersion,
+					CentosVersion: centosVersion,
+				})
+				multihostBatsJobs = append(multihostBatsJobs, &MultihostBatsJob{
 					Source:        sourceVersion,
 					Target:        targetVersion,
 					CentosVersion: centosVersion,
@@ -141,6 +164,7 @@ func init() {
 		LastTargetVersion: targetVersions[len(targetVersions)-1],
 		Versions:          versions,
 		CheckJobs:         checkJobs,
+		MultihostBatsJobs: multihostBatsJobs,
 	}
 }
 
--- a/ci/scripts/bats-on-cluster.bash
+++ b/ci/scripts/bats-on-cluster.bash
@@ -1,0 +1,91 @@
+#!/bin/bash
+#
+# Copyright (c) 2017-2020 VMware, Inc. or its affiliates
+# SPDX-License-Identifier: Apache-2.0
+
+set -eux -o pipefail
+
+is_GPDB5() {
+    local gphome=$1
+    version=$(ssh mdw "$gphome"/bin/postgres --gp-version)
+    [[ $version =~ ^"postgres (Greenplum Database) 5." ]]
+}
+
+drop_gphdfs_roles() {
+  echo 'Dropping gphdfs role...'
+  ssh mdw "
+      set -x
+
+      source /usr/local/greenplum-db-source/greenplum_path.sh
+      export MASTER_DATA_DIRECTORY=/data/gpdata/master/gpseg-1
+
+      psql -d postgres <<SQL_EOF
+          CREATE OR REPLACE FUNCTION drop_gphdfs() RETURNS VOID AS \\\$\\\$
+          DECLARE
+            rolerow RECORD;
+          BEGIN
+            RAISE NOTICE 'Dropping gphdfs users...';
+            FOR rolerow IN SELECT * FROM pg_catalog.pg_roles LOOP
+              EXECUTE 'alter role '
+                || quote_ident(rolerow.rolname) || ' '
+                || 'NOCREATEEXTTABLE(protocol=''gphdfs'',type=''readable'')';
+              EXECUTE 'alter role '
+                || quote_ident(rolerow.rolname) || ' '
+                || 'NOCREATEEXTTABLE(protocol=''gphdfs'',type=''writable'')';
+              RAISE NOTICE 'dropping gphdfs from role % ...', quote_ident(rolerow.rolname);
+            END LOOP;
+          END;
+          \\\$\\\$ LANGUAGE plpgsql;
+
+          SELECT drop_gphdfs();
+
+          DROP FUNCTION drop_gphdfs();
+SQL_EOF
+  "
+}
+
+#
+# MAIN
+#
+
+# TODO: combine this or at least pull out common functions with upgrade-cluster.bash?
+
+# This port is selected by our CI pipeline
+MASTER_PORT=5432
+
+# We'll need this to transfer our built binaries over to the cluster hosts.
+./ccp_src/scripts/setup_ssh_to_cluster.sh
+
+# Cache our list of hosts to loop over below.
+mapfile -t hosts < cluster_env_files/hostfile_all
+
+# Install gpupgrade binary onto the cluster machines.
+chmod +x bin_gpupgrade/gpupgrade
+for host in "${hosts[@]}"; do
+    scp bin_gpupgrade/gpupgrade "gpadmin@$host:/tmp"
+    ssh centos@$host "sudo mv /tmp/gpupgrade /usr/local/bin"
+done
+
+# Install gpupgrade_src on mdw
+scp -rpq gpupgrade_src gpadmin@mdw:/home/gpadmin
+
+# Install bats on mdw
+scp -rpq bats centos@mdw:~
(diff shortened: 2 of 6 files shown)


test/revert.bats
```diff
@@ -38,26 +101,26 @@ teardown() {
...
+    for host in "${HOSTS[@]}"; do
+	! host_process_is_running "$host" "[g]pupgrade agent" || fail "expected agent to have been stopped on host ${host}"
+    done
 
     # target data directories are deleted
-    while read -r datadir; do
-        run stat "$datadir"
-        ! [ $status -eq 0 ] || fail "expected datadir ${datadir} to have been deleted"
+    while read -r host datadir; do
+        ssh "$host" "[ ! -d '$datadir' ]" || fail "expected datadir ${host}:${datadir} to have been deleted"
```
jchampio, line 117: SSH calls inside a `read` loop need to be using `-n` so that they don't accidentally consume the stdin that `read` is using.

test/revert.bats
```diff
@@ -38,26 +101,26 @@ teardown() {
...
+    target_hosts_dirs=$(jq -r '.Target.Primaries[] | .Hostname + " " + .DataDir' "${GPUPGRADE_HOME}/config.json")
     upgradeID=$(gpupgrade config show --id)
 
     gpupgrade revert --verbose
 
     # gpupgrade processes are stopped
     ! process_is_running "[g]pupgrade hub" || fail 'expected hub to have been stopped'
-    ! process_is_running "[g]pupgrade agent" || fail 'expected agent to have been stopped'
+    for host in "${HOSTS[@]}"; do
+	! host_process_is_running "$host" "[g]pupgrade agent" || fail "expected agent to have been stopped on host ${host}"
```
jchampio, line 112: s/tab/space/

ci/scripts/bats-on-cluster.bash
```diff
@@ -0,0 +1,96 @@
...
+
+set -eux -o pipefail
+
+is_GPDB5() {
+    local gphome=$1
+    version=$(ssh mdw "$gphome"/bin/postgres --gp-version)
+    [[ $version =~ ^"postgres (Greenplum Database) 5." ]]
+}
+
+drop_gphdfs_roles() {
```
jchampio, line 14: Would using the migration scripts directly be easier to maintain?

test/tablespace_helpers.bash
```diff
@@ -5,29 +5,47 @@
...
+    mkdir "$TABLESPACE_ROOT"
+    echo "filespace:batsFS" > "$TABLESPACE_CONFIG"
+
+    local host dbid datadir
+    while read -r host dbid datadir; do
+	tablespace_dir="${TABLESPACE_ROOT}/${datadir}"
+
+	ssh -n "$host" mkdir -p "$(dirname "$tablespace_dir")"
+	echo "${host}:${dbid}:${tablespace_dir}" >> "$TABLESPACE_CONFIG"
+    done <<< "$entries"
```
dkrieger-pvtl, line 44: nit: this shows up as follows in my IntelliJ Editor, even though it shows correctly in github.  It looks like there is a single tab-as-eight-spaces here, though it appears the file uses 4 spaces as the standard indent elsewhere.  Perhaps change this to eight-spaces?
```
    while read -r host dbid datadir; do
  tablespace_dir="${TABLESPACE_ROOT}/${datadir}"

  ssh -n "$host" mkdir -p "$(dirname "$tablespace_dir")"
  echo "${host}:${dbid}:${tablespace_dir}" >> "$TABLESPACE_CONFIG"
    done <<< "$entries"
```

test/revert.bats
```diff
@@ -219,10 +279,10 @@ is_source_standby_in_sync() {
...
-    local standby_datadir
-    standby_datadir=$(query_datadirs "$GPHOME_SOURCE" "$PGPORT" "content = '-1' AND role = 'm'")
+    local host datadir
+    read -r host datadir <<<"$(query_host_datadirs "$GPHOME_SOURCE" "$PGPORT" "content = '-1' AND role = 'm'")"
 
-    if ! "${GPHOME_SOURCE}"/bin/pg_ctl status -D "$standby_datadir" > /dev/null; then
+    if ! ssh "$host" "${GPHOME_SOURCE}"/bin/pg_ctl status -D "$datadir" > /dev/null; then
         return 1
     fi
 }
```
dkrieger-pvtl, line 288: Note that when we update to master, there will be multi-host updates to the new revert-after-master-upgrade-failure bats tests added in this file.

test/revert.bats
```diff
@@ -26,6 +25,70 @@ teardown() {
...
+    export GPUPGRADE_HOME="${STATE_DIR}/gpupgrade"
+
+    # The hosts list will contain the master host, which we already created a
+    # directory for. That's okay; mkdir -p will ignore it. We still need the
+    # teardown.
+    for host in "${hosts[@]}"; do
+        ssh "$host" mkdir -p "$STATE_DIR"
+        register_teardown ssh "$host" rm -r "$STATE_DIR"
+    done
+}
```
dkrieger-pvtl, line 42: For a future PR...upgrade currently assumes that the cli, hub and agents run on hosts which have the same state dir.  From the upgrade code, this is derived either from HOME or GPUPGRADE_HOME.

However, there's no guarantee a priori that the STATE_DIR generated on the master host is not otherwise used on the segment hosts.  Ideally, we'd generate the STATE_DIR locally on each host.  But the current upgrade code does not allow for that.

This is a pedantic point, as we generate the CCP cluster for this test, and it is unlikely that the STATE_DIR we generate is otherwise used on the segment hosts.

test/tablespace_helpers.bash
jchampio: Already fixed; see above.

test/revert.bats
jchampio: > This is a pedantic point, as we generate the CCP cluster for this test, and it is unlikely that the STATE_DIR we generate is otherwise used on the segment hosts.

Yeah, I don't get the feeling that this is worth spending cycles on.

ci/scripts/bats-on-cluster.bash
```diff
@@ -0,0 +1,96 @@
...
+# We'll need this to transfer our built binaries over to the cluster hosts.
+./ccp_src/scripts/setup_ssh_to_cluster.sh
+
+# Cache our list of hosts to loop over below.
+mapfile -t hosts < cluster_env_files/hostfile_all
+
+export GPHOME_SOURCE=/usr/local/greenplum-db-source
+export GPHOME_TARGET=/usr/local/greenplum-db-target
+export PGPORT="$MASTER_PORT"
+export MASTER_DATA_DIRECTORY=/data/gpdata/master/gpseg-1
```
kalensk, line 65: Are these used or can we remove them since I see them used down below on lines 88-91 right before calling revert.bats.

ci/scripts/bats-on-cluster.bash
```diff
@@ -0,0 +1,96 @@
...
+          DROP FUNCTION drop_gphdfs();
+SQL_EOF
+  "
+}
+
+#
+# MAIN
+#
+
+# TODO: combine this or at least pull out common functions with upgrade-cluster.bash?
```
kalensk, line 51: Curious to see if this been tried. 'm a big advocate for doing so.

test/revert.bats
```diff
@@ -26,6 +25,70 @@ teardown() {
...
+setup_state_dirs() {
+    local hosts=("$@")
+
+    # Create a temporary state directory locally, on the master segment.
+    STATE_DIR=`mktemp -d /tmp/gpupgrade.XXXXXX`
+    export GPUPGRADE_HOME="${STATE_DIR}/gpupgrade"
+
+    # The hosts list will contain the master host, which we already created a
+    # directory for. That's okay; mkdir -p will ignore it. We still need the
+    # teardown.
```
kalensk, line 37: Im having a hard time following this first sentence, can you explain it or possibly rephrase.

test/revert.bats
```diff
@@ -219,10 +279,10 @@ is_source_standby_in_sync() {
 }
 
 is_source_standby_running() {
-    local standby_datadir
-    standby_datadir=$(query_datadirs "$GPHOME_SOURCE" "$PGPORT" "content = '-1' AND role = 'm'")
+    local host datadir
```
kalensk, line 282: nit: Rename to `standby` to make the following lines more readable since the query is specifically extracting the standby, its easier for me to follow and keep track when I see standby instead of host.

ci/scripts/bats-on-cluster.bash
dkrieger-pvtl: Probably...though we currently don't have a way to run just one start script.

ci/scripts/bats-on-cluster.bash
dkrieger-pvtl: Not yet...it was faster to create a new file, but once this goes in, we can likely have a "setup-cluster.helper" file that both use.

test/revert.bats
dkrieger-pvtl: Likely a better phrasing is:
```
# This repeats the creation of the STATE_DIR on the master host done above,
# but since "mkdir -p" is idempotent, that is ok.  We still need the master host teardown.
```

jchampio: The master host will be contained in `${hosts[@]}`. So we will perform a `mkdir` for it even though it was already handled by the `mktemp -d`. Is there a wording you have in mind that would make things easier?

ci/scripts/bats-on-cluster.bash
jchampio: > though we currently don't have a way to run just one start script.

We shouldn't have to; only one should be generated.

jchampio:
Rebased onto master; new code due to the rebase is in a squash-commit at the end.

test/revert.bats
kalensk: What @dkrieger-pvtl  had or another take is perhaps: 

```
Create STATE_DIR on the master by looping through all hosts as a convince even though they are unused.
```

ci/scripts/bats-on-cluster.bash
jchampio: We should be able to remove them.

jchampio: Done.

test/revert.bats
jchampio: Done; let us know what you think.

test/revert.bats
jchampio: Done.

ci/scripts/bats-on-cluster.bash
jchampio: This is an easy change to make in the future, and we have stories blocked on this PR, so let's defer.

ci/parser/parse_template.go
```diff
@@ -72,6 +72,22 @@ func (j *UpgradeJob) BaseName() string {
...
+	return fmt.Sprintf("%s-centos-%s", j.BaseName(), j.CentosVersion)
+}
+
+// BaseName returns the pipeline job name without the operating system.
+// This is used as a tag in Concourse's serial group to limit similar jobs
+// between operating systems from running at once to avoid overloading Concourse.
+func (j *MultinodeBatsJob) BaseName() string {
+	return fmt.Sprintf("%s-to-%s-%s", j.Source, j.Target, "multinodeBats")
+}
+
```
kalensk, line 90: Super weird to me to see "multinode" bats in the CI pipeline rather than the more common/idiomatic multi-host. Any reasons why multinode was chosen rather than `5-to-6-multihost-bats-centos-7`?

dkrieger-pvtl: No particular reason...either `multinode` or `multihost` is descriptive enough for me.

kalensk: Sure, I am thinking multihost would be more accurate, especially for PMs and other higher level people that look at our pipelines.

dkrieger-pvtl: Done.

PR #438, opened 2020-08-14
pull commanders.Substep up into commands.go

kalensk:
Unify the way we run CLI only substeps in commands.go. This will be useful when passing verbose into commanders.Substep.

[Pipeline](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:commandSubsteps)

### DIFF
--- a/cli/commands/commands.go
+++ b/cli/commands/commands.go
@@ -411,7 +411,9 @@ func initialize() *cobra.Command {
 			fmt.Println("Initialize in progress.")
 			fmt.Println()
 
+			s := commanders.Substep(idl.Substep_CREATING_DIRECTORIES)
 			err = commanders.CreateStateDir()
+			s.Finish(&err)
 			if err != nil {
 				return xerrors.Errorf("create state directory: %w", err)
 			}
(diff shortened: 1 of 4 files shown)


kalensk:
Can't quite use `defer s.Finish(&err)` as is otherwise the status gets printed after the completed text (see below). Going to need to re-arrange some things.

```
Initialize in progress.

Saving source cluster configuration...                             [COMPLETE]
Starting gpupgrade agent processes...                              [COMPLETE]
Generating target cluster configuration...                         [COMPLETE]
Creating target cluster...                                         [COMPLETE]
Stopping target cluster...                                         [COMPLETE]
Backing up target master...                                        [COMPLETE]
Running pg_upgrade checks...                                       [COMPLETE]

Initialize completed successfully.

NEXT ACTIONS
------------
To proceed with the upgrade, run "gpupgrade execute"
followed by "gpupgrade finalize".

To return the cluster to its original state, run "gpupgrade revert".
Checking disk space...                                             [COMPLETE]
Starting gpupgrade hub process...                                  [COMPLETE]
Creating directories...                                            [COMPLETE]

```

cli/commands/commands.go
```diff
@@ -411,7 +411,9 @@ func initialize() *cobra.Command {
 			fmt.Println("Initialize in progress.")
 			fmt.Println()
 
+			s := commanders.Substep(idl.Substep_CREATING_DIRECTORIES)
 			err = commanders.CreateStateDir()
+			s.Finish(&err)
```
jchampio, line 416: Is anything still using the `defer` form? If not, the `Finish()` method no longer needs to take a `*error`.

kalensk: Doesn't look like anything else does. Will save for a later PR.

PR #440, opened 2020-08-17
log step and substep timing

kalensk:
Log step and substep duration. Currently, for substeps that are run on the CLI timing is printed using gplog rather than the step log as is done for the other substeps.

[Pipeline](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:timing)

### DIFF
--- a/cli/commanders/steps.go
+++ b/cli/commanders/steps.go
@@ -13,6 +13,7 @@ import (
 	"golang.org/x/xerrors"
 
 	"github.com/greenplum-db/gpupgrade/idl"
+	"github.com/greenplum-db/gpupgrade/utils/stopwatch"
 )
 
 type ExecuteResponse struct {
@@ -39,6 +40,13 @@ type receiver interface {
 type substepText struct {
 	OutputText string
 	HelpText   string
+}
+
+type substep struct {
+	name    idl.Substep
+	text    substepText
+	verbose bool
+	timer   *stopwatch.Stopwatch
 }
 
 var SubstepDescriptions = map[idl.Substep]substepText{
@@ -262,7 +270,7 @@ func FormatStatus(status *idl.SubstepStatus) string {
 	return Format(line.OutputText, status.Status)
 }
 
-// Format is also exported for ease of testing (see FormatStatus). Use Substep
+// Format is also exported for ease of testing (see FormatStatus). Use NewSubstep
 // instead.
 func Format(description string, status idl.Status) string {
 	indicator, ok := indicators[status]
@@ -273,13 +281,19 @@ func Format(description string, status idl.Status) string {
 	return fmt.Sprintf("%-67s%-13s", description, indicator)
 }
 
-// Substep prints out an "in progress" marker for the given substep description,
+// NewSubstep prints out an "in progress" marker for the given substep description,
 // and returns a struct that can be .Finish()d (in a defer statement) to print
 // the final complete/failed state.
-func Substep(step idl.Substep) *substepText {
+func NewSubstep(step idl.Substep, verbose bool) *substep {
 	substepText := SubstepDescriptions[step]
 	fmt.Printf("%s\r", Format(substepText.OutputText, idl.Status_RUNNING))
-	return &substepText
+
+	return &substep{
+		name:    step,
+		text:    substepText,
+		verbose: verbose,
+		timer:   stopwatch.Start(),
+	}
 }
 
 // Finish prints out the final status of the substep; either COMPLETE or FAILED
@@ -287,17 +301,27 @@ func Substep(step idl.Substep) *substepText {
 // error rather than error to make it possible to defer:
 //
 //    func runSubstep() (err error) {
-//        s := Substep("Doing something...")
+//        s := NewSubstep("Doing something...")
 //        defer s.Finish(&err)
 //
 //        ...
 //    }
 //
-func (s *substepText) Finish(err *error) {
+func (s *substep) Finish(err *error) {
 	status := idl.Status_COMPLETE
 	if *err != nil {
 		status = idl.Status_FAILED
 	}
 
-	fmt.Printf("%s\n", Format(s.OutputText, status))
+	fmt.Printf("%s\n", Format(s.text.OutputText, status))
+
+	LogDuration(s.name.String(), s.verbose, s.timer.Stop())
+}
+
+func LogDuration(operation string, verbose bool, timer *stopwatch.Stopwatch) {
+	msg := operation + " took " + timer.String()
+	if verbose {
+		fmt.Println("\n" + msg)
+	}
+	gplog.Debug(msg)
 }
--- a/cli/commands/commands.go
+++ b/cli/commands/commands.go
@@ -56,6 +56,7 @@ import (
 	"github.com/greenplum-db/gpupgrade/step"
 	"github.com/greenplum-db/gpupgrade/upgrade"
 	"github.com/greenplum-db/gpupgrade/utils"
+	"github.com/greenplum-db/gpupgrade/utils/stopwatch"
 )
 
 var (
@@ -411,7 +412,15 @@ func initialize() *cobra.Command {
 			fmt.Println("Initialize in progress.")
 			fmt.Println()
 
-			s := commanders.Substep(idl.Substep_CREATING_DIRECTORIES)
+			timer := stopwatch.Start()
+			operation := strings.Title(strings.ToLower(idl.Step_INITIALIZE.String()))
+			defer func() {
+				if err != nil {
+					commanders.LogDuration(operation, verbose, timer.Stop())
+				}
+			}()
+
+			s := commanders.NewSubstep(idl.Substep_CREATING_DIRECTORIES, verbose)
 			err = commanders.CreateStateDir()
 			s.Finish(&err)
 			if err != nil {
@@ -423,7 +432,7 @@ func initialize() *cobra.Command {
 				return xerrors.Errorf("create initial cluster configs: %w", err)
 			}
 
-			s = commanders.Substep(idl.Substep_START_HUB)
+			s = commanders.NewSubstep(idl.Substep_START_HUB, verbose)
 			err = commanders.StartHub()
 			s.Finish(&err)
 			if err != nil {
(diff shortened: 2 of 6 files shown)


kalensk:
Note that simply always printing the timing in a defer in commands.go is tricky since it will print _after_ the completion text which is not desired. I also wanted the timing to print even when a step fails.

I am planning to refactor the CLI substeps to make things cleaner. See WIP [substep branch](https://github.com/kalensk/gpupgrade/tree/substep).

utils/stopwatch/stopwatch_test.go
```diff
@@ -0,0 +1,68 @@
...
+	t.Run("outputs correctly formatted duration", func(t *testing.T) {
+		startTime := mustParseDuration(t, "-3h26m8s")
+		timer := stopwatch.NewTime(time.Now().Add(startTime))
+
+		duration := timer.Stop()
+		expected := mustParseDuration(t, "3h26m")
+		if duration != expected {
+			t.Errorf("got %q want %q", duration, expected)
+		}
+	})
```
kalensk, line 23: Do we _really_ need this test? If not, we can remove `NewTime()`.

cli/commanders/steps.go
```diff
@@ -276,10 +284,16 @@ func Format(description string, status idl.Status) string {
 // Substep prints out an "in progress" marker for the given substep description,
 // and returns a struct that can be .Finish()d (in a defer statement) to print
 // the final complete/failed state.
-func Substep(step idl.Substep) *substepText {
+func Substep(step idl.Substep, verbose bool) *substep {
```
jchampio, line 287: You'd mentioned not liking the collision with `Substep()` -- the function should probably more accurately be called `newSubstep()`, if that helps.

cli/commands/commands.go
```diff
@@ -500,9 +511,17 @@ func execute() *cobra.Command {
...
+		RunE: func(cmd *cobra.Command, args []string) (err error) {
 			cmd.SilenceUsage = true
 
+			timer := stopwatch.Start()
+			operation := strings.Title(strings.ToLower(idl.Step_EXECUTE.String()))
+			defer func() {
+				if err != nil {
+					timer.StopAndLogDuration(operation, verbose)
+				}
+			}()
```
jchampio, line 523: I think it's going to be difficult to maintain this defer logic as-is. We should try hard to only have a single call to `Stop()`.

utils/stopwatch/stopwatch.go
```diff
@@ -0,0 +1,63 @@
...
+
+	return duration
+}
+
+func (s *Stopwatch) StopAndLogDuration(operation string, verbose bool) {
+	s.Stop()
+
+	msg := operation + " took " + s.Stop().String()
+	if verbose {
+		fmt.Println("\n" + msg)
```
jchampio, line 60: Though I understand why this exists, I think further coupling between the top-level UI concerns and internal business logic should be avoided. We're already having trouble tracking down where extra or missing newlines come from.

Put another way, the `AndLogDuration` portion of this API seems tightly coupled to one particular use case, rather than a general-purpose utility. Can the timer simply store the duration for the UI to print appropriately?

cli/commands/commands.go
```diff
@@ -411,7 +412,15 @@ func initialize() *cobra.Command {
...
 			fmt.Println()
 
-			s := commanders.Substep(idl.Substep_CREATING_DIRECTORIES)
+			timer := stopwatch.Start()
+			operation := strings.Title(strings.ToLower(idl.Step_INITIALIZE.String()))
+			defer func() {
+				if err != nil {
+					timer.StopAndLogDuration(operation, verbose)
+				}
+			}()
```
jchampio, line 421: This is a pretty heavy invocation, and it's duplicated verbatim three times. Is there a UI-specific API to be pulled out that would simplify things?

utils/stopwatch/stopwatch.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+
+	if duration.Hours() < time.Minute.Seconds() {
+		return duration.Round(time.Minute)
+	}
+
+	return duration
+}
+
+func (s *Stopwatch) StopAndLogDuration(operation string, verbose bool) {
```
dkrieger-pvtl, line 55: This couples this low-level, utility module to our logging infrastructure.  This belongs elsewhere.  Since we currently only use this in `commanders` and `commands`, and `commanders` is the more specific package, I'd just include this there.

utils/stopwatch/stopwatch.go
```diff
@@ -0,0 +1,63 @@
...
+
+func (s *Stopwatch) Stop() time.Duration {
+	return Round(time.Since(s.startTime))
+}
+
+// Round returns a duration rounded to the next lowest unit of precision to
+// prevent excessive precision from being displayed when printed. Currently,
+// only hours, minutes, and seconds are supported. For example, hours are
+// rounded to the nearest minute, minutes to the nearest second, and seconds to
+// the nearest millisecond.
```
jchampio, line 38: I don't think this matches the implementation.

Durations less than one second will be rounded to milliseconds.
Durations less than one minute will be rounded to seconds.
Durations less than 60 hours (was that intended?) will be rounded to minutes.
All other durations will not be rounded at all.

utils/stopwatch/stopwatch.go
```diff
@@ -0,0 +1,63 @@
...
+func Round(duration time.Duration) time.Duration {
+	if duration.Seconds() < time.Second.Seconds() {
```
jchampio, line 40: `time.Second.Seconds()` is identical to `1` in all cases.

utils/stopwatch/stopwatch.go
```diff
@@ -0,0 +1,63 @@
...
+		return duration.Round(time.Millisecond)
+	}
+
+	if duration.Minutes() < time.Minute.Seconds() {
+		return duration.Round(time.Second)
+	}
+
+	if duration.Hours() < time.Minute.Seconds() {
```
jchampio, line 48: This seems incorrect but I'm not sure what was intended (see above).

utils/stopwatch/stopwatch.go
```diff
@@ -0,0 +1,63 @@
...
+// NewTime is used for unit testing.
+func NewTime(start time.Time) *Stopwatch {
+	return &Stopwatch{
+		startTime: start,
+	}
+}
+
+func (s *Stopwatch) Stop() time.Duration {
+	return Round(time.Since(s.startTime))
+}
```
dkrieger-pvtl, line 32: This utility allows `Stop` to be called multiple times, with different results which isn't what a physical stopwatch does...it does start, stop, and lap(elapsed since last lap or start).  Based on our usage, the term stopwatch is fine, but I think the implementation should latch the duration between start and the first call to stop.   We do not want the user to call `Stop` multiple times and get different numbers....that would be a `Lap` function.

utils/stopwatch/stopwatch_test.go
jchampio: You could export the `startTime` member (`watch.Start`, maybe?) and get rid of the constructor.

But the test is still going to be inherently flaky, I think, since `Stop()` can't be controlled. Does the stopwatch need to mix the rounding behavior (a UI concern) with the timing logic (a business concern)?

cli/commands/commands.go
kalensk: I thought that was what my [substep branch](https://github.com/kalensk/gpupgrade/tree/substep).

kalensk:
> You could export the startTime member (watch.Start, maybe?) and get rid of the constructor.
> 
> But the test is still going to be inherently flaky, I think, since Stop() can't be controlled. Does the stopwatch need to mix 
> the rounding behavior (a UI concern) with the timing logic (a business concern)?

Updated so Stop() simply stops and String() does the rounding. I am hesitant to export startTime to prevent callers from misusing it in production.

utils/stopwatch/stopwatch_test.go
kalensk: Updated so `Stop()` simply stops and `String()` does the rounding. I am hesitant to export `startTime` to prevent callers from misusing it in production.

utils/stopwatch/stopwatch.go
kalensk: Sure, extracted to `commanders.LogDuration()`

utils/stopwatch/stopwatch.go
kalensk: Removed those lines and updated the comment to just be:
```
// round returns a duration rounded to the next lowest unit to prevent excessive
// precision from being displayed to enable pretty printing.
```

utils/stopwatch/stopwatch.go
kalensk: Updated.

utils/stopwatch/stopwatch.go
kalensk: Updated the implementation to:
```go
func round(duration time.Duration) time.Duration {
	switch {
	case duration.Seconds() < 1:
		return duration.Round(time.Millisecond)

	case duration.Minutes() < 60:
		return duration.Round(time.Second)

	default:
		return duration.Round(time.Minute)
	}
}
```

Let me know if that works for you.

cli/commands/commands.go
kalensk: @dkrieger-pvtl and I tried wrapping it in a function and it is fairly difficult to read to us.

cli/commands/commands.go
kalensk: Sorry, I should have been more clear. We could create a `step.Run` type of framework for the `RunE` functions of initialize, execute, finalize, and revert to separate the business logic from the UI. This could be done later.

utils/stopwatch/stopwatch.go
jchampio: Much better, thanks.

dkrieger-pvtl:
Note: there will need to be a follow-up PR to add `--print-timing` to the calls to `pg_upgrade`. (see https://github.com/greenplum-db/gpdb/pull/10661).  This will need to be coordinated with the test pipelines as they need to be updated to use a version of gpdb incorporating this change as `pg_upgrade --unknown-flag` will fail...

utils/stopwatch/stopwatch_test.go
jchampio: > I am hesitant to export startTime to prevent callers from misusing it in production.

I'd argue that `NewTime()` already enables shenanigans.

dkrieger-pvtl:
Note(no action needed): For substeps that occur incrementally(like the rsync in revert), if that substep errors out and is then restarted(via rerunning revert), the substep timing will include only that amount needed for that run.  So if the total rsync takes 5 hours, but failed after 3 hours, is fixed, and the re-run takes 2 hours, the user will have to know to sum those to know the total time for the substep.  That is, we currently print the time _of this run_ of the substep, not how long the substep took during this upgrade process.  That's an issue for later, though.  I filed an issue for us to track this.

PR #441, opened 2020-08-18
Revert: add tests for primary upgrade failure

dkrieger-pvtl:
bats: add primary upgrade failure tests to revert.bats for execute

Similar to the master upgrade failure, force a primary upgrade failure during execute.  Do this by creating a table and moving
its relfilenode out of the way.

We did not need to add any code; the existing code works when pg_upgrade on a segment works.


test/revert.bats
```diff
@@ -343,8 +370,10 @@ test_revert_after_execute_master_failure() {
 
     # Execute should fail.
     local status=0
-    gpupgrade execute --verbose || status=$?
+    run gpupgrade execute --verbose
```
dkrieger-pvtl, line 373: There's a concern that this will swallow the verbose output; if there's a failure below(say in revert), we won't see that output....maybe do:
```bash
echo "$output"
```
also?

test/revert.bats
```diff
@@ -343,8 +370,10 @@ test_revert_after_execute_master_failure() {
 
     # Execute should fail.
     local status=0
```
dkrieger-pvtl, line 372: remove this unused variable.

test/revert.bats
dkrieger-pvtl: Done.

PR #442, opened 2020-08-18
revert: match implementation with design goals

jchampio:
Continue to evolve and clean up the revert implementation. Each commit is self-contained.

### DIFF
--- a/hub/revert.go
+++ b/hub/revert.go
@@ -115,27 +115,6 @@ func (s *Server) Revert(_ *idl.RevertRequest, stream idl.CliToHub_RevertServer) 
 		})
 	}
 
-	var archiveDir string
-	st.Run(idl.Substep_ARCHIVE_LOG_DIRECTORIES, func(_ step.OutStreams) error {
-		// Archive log directory on master
-		oldDir, err := utils.GetLogDir()
-		if err != nil {
-			return err
-		}
-		archiveDir = filepath.Join(filepath.Dir(oldDir), upgrade.GetArchiveDirectoryName(s.UpgradeID, time.Now()))
-
-		gplog.Debug("moving directory %q to %q", oldDir, archiveDir)
-		if err = utils.Move(oldDir, archiveDir); err != nil {
-			return err
-		}
-
-		return ArchiveSegmentLogDirectories(s.agentConns, s.Config.Target.MasterHostname(), archiveDir)
-	})
-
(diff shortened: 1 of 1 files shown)


hub/revert.go
```diff
@@ -122,26 +89,31 @@ func (s *Server) Revert(_ *idl.RevertRequest, stream idl.CliToHub_RevertServer)
...
-		oldDir, err := utils.GetLogDir()
-		if err != nil {
-			return err
-		}
-		archiveDir = filepath.Join(filepath.Dir(oldDir), upgrade.GetArchiveDirectoryName(s.UpgradeID, time.Now()))
+	if s.UseLinkMode {
+		// In any of the three cases of source cluster recovery in link-mode pg_upgrade, it is correct
+		// to restore the pg_control file.  See the "Reverting to old cluster" section of
+		// https://www.postgresql.org/docs/9.4/pgupgrade.html
+		st.Run(idl.Substep_RESTORE_PGCONTROL, func(streams step.OutStreams) error {
```
kalensk, line 96: minor nit: When I first read this I wasn't sure what "In any of the three cases" referred to. Maybe something like:

> In link-mode it is always correct to restore the pg_control file in any of the three cases of source cluster recovery as outlined in the "Reverting to old cluster" section of
https://www.postgresql.org/docs/9.4/pgupgrade.html

hub/revert.go
```diff
@@ -69,42 +69,9 @@ func (s *Server) Revert(_ *idl.RevertRequest, stream idl.CliToHub_RevertServer)
...
-			// https://www.postgresql.org/docs/9.4/pgupgrade.html
-			st.Run(idl.Substep_RESTORE_PGCONTROL, func(streams step.OutStreams) error {
-				return RestoreMasterAndPrimariesPgControl(streams, s.agentConns, s.Source)
-			})
-		}
-	}
-
+	// The target cluster's data directories and tablespaces must be removed.
+	// TODO: combine the primary/master data directory deletion into a single
+	// substep implementation, to match the prevailing patterns.
```
kalensk, line 74: Would it be possible to do this soon, such as right after this PR or so? I would find that very useful for the long-term maintenance and readability of the code.

jchampio: Added a chore; it's a decent amount of unrelated churn.

hub/revert.go
jchampio: Updated with additional context.

PR #443, opened 2020-08-19
revert: add idempotence regression tests

jchampio:
_This is a draft because the log directory archival step is not yet idempotent and causes a legitimate failure in the multihost pipeline._

Setting the DebugIdempotence flag in the hub configuration will cause the hub to fail every revert substep exactly once. The corresponding end-to-end test simply loops in a call to revert, checking that any errors are due to the idempotence framework, until the step succeeds.

This tests only a subset of idempotence -- namely, that each individual substep (and all of its `AlwaysRun` predecessors) can be re-run after completion -- but in practice this catches a number of common bugs.

### DIFF
--- a/hub/revert.go
+++ b/hub/revert.go
@@ -49,7 +49,12 @@ func (s *Server) Revert(_ *idl.RevertRequest, stream idl.CliToHub_RevertServer) 
 		return xerrors.Errorf("connect to gpupgrade agent: %w", err)
 	}
 
-	archiveDir, err := s.revert(st)
+	var runner SubstepRunner = st
+	if s.DebugIdempotence {
+		runner = idempotenceRunner{st}
+	}
+
+	archiveDir, err := s.revert(runner)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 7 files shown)


test/revert.bats
```diff
@@ -428,3 +428,54 @@ test_revert_after_execute_pg_upgrade_failure() {
...
+
+        if [ "$status" -eq 0 ]; then
+            break # revert is finished
+        fi
+
+        # We expect all failures to be forced by the idempotence tests; anything
+        # else suggests that idempotence is broken for some substep.
+        [[ $output = *"forced error for idempotence testing"* ]] || \
+                fail "unexpected failure during revert"
+    done
```
dkrieger-pvtl, line 472: We should check that we are actually running with the idempotence runner here.  If `enable_idempotence_debugging` were disabled or did the wrong thing, then this test would be none the wiser.  Maybe simply checking that we hit the `"forced error for idempotence testing"` error at least once?

jchampio: Good idea.

hub/revert.go
```diff
@@ -67,7 +72,12 @@ func (s *Server) Revert(_ *idl.RevertRequest, stream idl.CliToHub_RevertServer)
 	return st.Err()
 }
 
-func (s *Server) revert(st *step.Step) (string, error) {
+type SubstepRunner interface {
+	Run(idl.Substep, func(step.OutStreams) error)
+	AlwaysRun(idl.Substep, func(step.OutStreams) error)
+}
```
kalensk, line 78: Having the `SubstepRunner` interface here is confusing for me compared to seeing it in `step.go` or similar. Can we move it?

kalensk (approved):
Looks great to me as a big step forward in testing idempotence.  Can you remind me if this is possible to merge and start to use, or what is blocked on starting to use this framework?

Minor comment about placement of `SubstepRunner`, and David's comment on hardening the test.

kalensk:
Closing for now as lots of work needs to be done before merging as outlined in [stop.md](https://gist.github.com/jchampio/386efc1ccf0e582a1aace694df6300be). We also have stories to use this framework. The branch as been preserved on origin at [/dev/idempotence-tests](https://github.com/greenplum-db/gpupgrade/tree/dev/idempotence-tests).

PR #444, opened 2020-08-20
upgrade: enable pg_upgrade print timings via environment variable

dkrieger-pvtl:
Use this environment variable to tell pg_upgrade to include timing information
in its returned output.  We would prefer to use the flag `--print-timing` but that
would require coordination with a new release of GPDB-6 as well as a bump in the
minimum version of GPDB-6 required for gpupgrade.  Once we have bumped the minimum
version of GPDB-6 past a release that includes the new print-timing code, we can
migrate to the flag.  See https://github.com/greenplum-db/gpdb/pull/10661.

Note that we are awaiting feedback on this technique in  https://github.com/greenplum-db/gpdb/pull/10661 but in the meantime, we can review this PR.


upgrade/run_test.go
```diff
@@ -146,7 +147,7 @@ func TestRun(t *testing.T) {
 		}
 	})
 
-	t.Run("unsets PGPORT and PGHOST", func(t *testing.T) {
+	t.Run("unsets PGPORT and PGHOST and sets __GPDB_PGUPGRADE_PRINT_TIMING__", func(t *testing.T) {
```
jchampio, line 150: The structural change to the test here is big enough that I think this should be its own test. `EnvironmentMain` is easy to use.

dkrieger-pvtl: Added a separate test.

PR #446, opened 2020-08-21
exectest: add NewCommandMock for mock-style tests

jchampio:
Writing integration-level tests with `exectest` is currently difficult. Clients must write low-level command verifiers, most of which end up looking like an assertion that the arguments match some slice of strings, and this obfuscates the test purpose. Additionally, if the unit under test needs to call more than one external command, the only way to control them independently is to create multiple injection points, which starts to unravel quickly.

These sorts of tests are typically less interested in exact control over the Command; rather, they want to perform setups like "all calls to gpstart should fail" or "if pgrep returns no results, bash should be executed with this set of arguments." This is the domain of mock frameworks, and we already have one of those.

Add a new type of `exectest` pattern: `NewCommandMock()` returns both an `exec.Command` implementation, and a `gomock` object to spy on and control that implementation. The `greenplum` package has been migrated to this approach; see the commit message for details.


testutils/exectest/mock.go
```diff
@@ -0,0 +1,77 @@
...
+//
+// This code should be read, "We expect the command `bash -c 'false || true'` to
+// be created, followed by the command `echo hello`."
+//
+//     command, mock := exectest.NewCommandMock(ctrl)
+//
+//     mock.EXPECT().Command("bash", gomock.Any()).
+//         AnyTimes().
+//         Return(exectest.Failure)
+//     mock.EXPECT().Command("rsync", gomock.Any()).
```
bhuvnesh2703, line 38: the example should contain echo instead of rsync?

testutils/exectest/mock.go
```diff
@@ -0,0 +1,77 @@
...
+//         AnyTimes()
+//
+// This code should be read, "We expect any number of commands invoking bash and
+// rsync, with any arguments, in any order. All calls to bash should fail with
```
bhuvnesh2703, line 42: echo instead of rsync?

testutils/exectest/mock_test.go
```diff
@@ -0,0 +1,149 @@
...
+		if err != nil {
+			t.Errorf("running echo: %+v", err)
+		}
+
+		stdout := string(out)
+		if stdout != printedStdout {
+			t.Errorf("echo printed %q, want %q", out, printedStdout)
+		}
+	})
+
```
kalensk, line 77: Can you explain this test and expectations at a high level? It appears the mock's have different return expectations than the actual commands being run. Thus, the mocks are not being met, correct?

That is I would expect the following to fail, but it does not. Clearly my understanding is off and could use some additional explanation.
```go
gomock.InOrder(
    mock.EXPECT().Command("bash", gomock.Any()).
        Return(exectest.Failure),
    mock.EXPECT().Command("echo", gomock.Any()).
        Return(PrintingMain),
)

cmd := command("bash", "-c", "false || true")
cmd.Run()

cmd = command("echo", "hello")
```

testutils/exectest/mock.go
jchampio: It's not meant to be a mirror of the example above it, but a completely new one.

testutils/exectest/mock.go
jchampio: See above.

greenplum/start_or_stop_cluster_test.go
```diff
@@ -144,30 +144,17 @@ func TestStartOrStopCluster(t *testing.T) {
...
+		defer ctrl.Finish()
+
+		mock, cleanup := greenplum.MockExecCommand(ctrl)
+		defer cleanup()
+
+		script := "source /usr/local/source/greenplum_path.sh " +
+			"&& MASTER_DATA_DIRECTORY=" + masterDataDir + " /usr/local/source/bin/gpstop -a -d " + masterDataDir
+
+		mock.EXPECT().Command("pgrep", []string{"-F", masterPidFile})
+		mock.EXPECT().Command("bash", []string{"-c", script})
```
kalensk, line 157: Does checking the order matter, and if not why not?

testutils/exectest/mock_test.go
jchampio: Right, I think the confusion is probably due to the `gomock` DSL and our abuse of it here.

In `gomock`, the `Return()` method specifies a mock behavior, not an expectation. I.e. when you say
```go
mock.EXPECT().Add(1, 2).Return(10)
```
you're not saying "I expect `Add(1, 2)` to return `10`", you're saying "I expect `Add()` to be called with the arguments `(1, 2)` and I want the mock to return `10` when that happens".

In the same way,
```go
mock.EXPECT().Command("bash", gomock.Any()).
        Return(exectest.Failure)
```
does not mean "all calls to bash must fail"; it means "I expect a bash command to be invoked and I want the mock framework to run the `exectest.Failure` implementation when it is".

Does that help clarify?

greenplum/start_or_stop_cluster_test.go
jchampio: In a since it does matter, but it also pins implementation behavior that I didn't feel was important to verify in this test, since there are other tests that ensure that a successful `pgrep` causes the `bash` call to be omitted.

testutils/exectest/mock_test.go
kalensk: Yes, thank you. I definitely had it a bit backwards.

PR #449, opened 2020-09-02
Remove the access to s.agentConns

bhuvnesh2703:
From the API perspective, the developer should only call s.AgentConns()
to get the agent connections. The access to s.agentConns makes it
confusing to chose, so remove the member agentConns from the struct
Server.


bhuvnesh2703:
Closing it for now as we will come back to it later during refactoring for agents.

PR #450, opened 2020-09-08
revert: DeleteMasterAndPrimaryDataDirectories

kalensk:
Combine deleting the master and primary data directories into a single function to match prevailing patterns and follow convention.

[Test pipeline](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:combine-delete-data-dirs)


hub/delete_data_directories.go
```diff
@@ -22,8 +22,27 @@ func DeleteMirrorAndStandbyDataDirectories(agentConns []*Connection, cluster *gr
 	return deleteDataDirectories(agentConns, segs)
 }
 
-func DeletePrimaryDataDirectories(agentConns []*Connection, segConfigs greenplum.SegConfigs) error {
-	return deleteDataDirectories(agentConns, segConfigs)
+func DeleteMasterAndPrimaryDataDirectories(streams step.OutStreams, agentConns []*Connection, source InitializeConfig) error {
+	var wg sync.WaitGroup
+	errs := make(chan error, 2)
+
+	wg.Add(1)
```
jchampio, line 29: Since there is only ever one goroutine being coordinated, a `WaitGroup` is probably a little bit much. A simpler channel usage would work too (it can probably be further improved):

```go
func DeleteMasterAndPrimaryDataDirectories(...) error {
    masterErr := make(chan error)

    go func() {
        masterErr <- upgrade.DeleteDirectories(...)
    }()

    err := deleteDataDirectories(agentConns, source.Primaries)

    mErr := multierror.Append(err, <-masterErr)
    return mErr.ErrorOrNil()
}
```

hub/revert.go
```diff
@@ -69,19 +69,9 @@ func (s *Server) Revert(_ *idl.RevertRequest, stream idl.CliToHub_RevertServer)
...
-		st.Run(idl.Substep_DELETE_PRIMARY_DATADIRS, func(_ step.OutStreams) error {
-			return DeletePrimaryDataDirectories(s.agentConns, s.TargetInitializeConfig.Primaries)
-		})
-	}
-
-	if s.TargetInitializeConfig.Master.DataDir != "" {
-		st.Run(idl.Substep_DELETE_MASTER_DATADIR, func(streams step.OutStreams) error {
-			datadir := s.TargetInitializeConfig.Master.DataDir
-			return upgrade.DeleteDirectories([]string{datadir}, upgrade.PostgresFiles, streams)
+	if s.TargetInitializeConfig.Primaries != nil && s.TargetInitializeConfig.Master.DataDir != "" {
```
jchampio, line 72: For a future PR: it'd be nice if we knew whether the `TargetInitializeConfig` had actually been initialized without having to reach inside it.

jchampio (approved):

hub/revert.go
kalensk: Agreed!

hub/delete_data_directories.go
kalensk: I am assuming we do not have to close the `masterErr` chan?

jchampio: Correct.

PR #451, opened 2020-09-09
update version output

kalensk:
Overall this includes:
- Renaming UpgradeVersion to Version for clarity
- Update the version output to include the commit sha and release type
- Return empty values when version fields are not set. This has been approved by the PM since the designer happened to be out that day.

Soon we will have to finalize our overall release process, but for now we can run make with the RELEASE variable set.

[Test pipelines](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:dev:version)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,8 +12,12 @@ MODULE_NAME=gpupgrade
 #   note git hash for that version(might have been rebased, etc); call it GIT_HASH
 #   YOUR_BRANCH> git tag -a TAGNAME -m "version 0.1.1: add version" GIT_HASH
 #   YOUR_BRANCH> git push origin TAGNAME
-VERSION := $(shell git describe --tags --long| perl -pe 's/(.*)-([0-9]*)-(g[0-9a-f]*)/\1+dev.\2.\3/')
-VERSION_LD_STR="-X github.com/greenplum-db/$(MODULE_NAME)/cli/commands.Version=$(VERSION)"
+VERSION := $(shell git describe --tags --abbrev=0)
+COMMIT := $(shell git rev-parse --short --verify HEAD)
+RELEASE=Dev Build
+VERSION_LD_STR := -X 'github.com/greenplum-db/$(MODULE_NAME)/cli/commands.Version=$(VERSION)'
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -12,8 +12,12 @@ MODULE_NAME=gpupgrade
...
 #   YOUR_BRANCH> git tag -a TAGNAME -m "version 0.1.1: add version" GIT_HASH
 #   YOUR_BRANCH> git push origin TAGNAME
-GIT_VERSION := $(shell git describe --tags --long| perl -pe 's/(.*)-([0-9]*)-(g[0-9a-f]*)/\1+dev.\2.\3/')
-VERSION_LD_STR="-X github.com/greenplum-db/$(MODULE_NAME)/cli/commands.UpgradeVersion=$(GIT_VERSION)"
+VERSION := $(shell git describe --tags --abbrev=0)
+COMMIT := $(shell git rev-parse --short --verify HEAD)
+RELEASE=Dev Build
+VERSION_LD_STR := -X 'github.com/greenplum-db/$(MODULE_NAME)/cli/commands.Version=$(VERSION)'
+VERSION_LD_STR := $(VERSION_LD_STR) -X 'github.com/greenplum-db/$(MODULE_NAME)/cli/commands.Commit=$(COMMIT)'
+VERSION_LD_STR := $(VERSION_LD_STR) -X 'github.com/greenplum-db/$(MODULE_NAME)/cli/commands.Release=$(RELEASE)'
```
jchampio, line 20: Use `VERSION_LD_STR += ...` instead of manually appending.

cli/commands/version.go
```diff
@@ -5,17 +5,16 @@ package commands
 
 import "fmt"
 
-// This global var UpgradeVersion should have a value set at build time.
+// This global var Version should have a value set at build time.
```
jchampio, line 8: Looks like this comment needs to be updated? `These global variables should have values...` or otherwise.

jchampio (approved):
LGTM with comments.

cli/commands/version.go
kalensk: I updated the comment to `These variables are set during build time as specified in the Makefile.`, and changed from  `var Version = ""` to `var Version string`

Makefile
kalensk: Updated, thanks.

PR #455, opened 2020-09-18
Pass PGPORT to gprecoverseg

bhuvnesh2703:
gprecoverseg connects to default port 5432 if PGPORT variable is not set
in the environment before triggering gprecoverseg. This commit passes
PGPORT while invoking gprecoverseg command.


kalensk:
Happy to merge this if there is a time constraint and test pipeline passes. If we do , let's create a story first thing and prioritize migrating to using `greenplum.Run` and seeing if we can add a test.

kalensk (approved):
Looks good.

Running a quick sanity check might be a good idea.

PR #458, opened 2020-09-28
data_migration_scripts: ignore any environmental .psqlrc 

jchampio:
.psqlrc settings can really mess up the script generation/execution (e.g. enabling `\timing` in a .psqlrc leads to script failure). Ignore .psqlrc by passing the `-X` option to psql.

The attached end-to-end test works only for source clusters of 6X and onward, since it uses the `PSQLRC` envvar that showed up in the 9.x line.

The prefactoring commit in this PR moves the migration tests to the teardown framework.

### DIFF
--- a/test/migration_scripts.bats
+++ b/test/migration_scripts.bats
@@ -138,6 +138,39 @@ drop_unfixable_objects() {
     diff -U3 --speed-large-files "$MIGRATION_DIR"/before.sql "$MIGRATION_DIR"/after.sql
 }
 
+@test "migration scripts ignore .psqlrc files" {
+    # 5X doesn't support the PSQLRC envvar we need to avoid destroying the dev
+    # environment.
+    if is_GPDB5 "$GPHOME_SOURCE"; then
+        skip "GPDB 5 does not support alternative PSQLRC locations"
+    fi
+
+    $PSQL -c "CREATE DATABASE $TEST_DBNAME;" -d $DEFAULT_DBNAME
+    $PSQL -f "$SCRIPTS_DIR"/test/create_nonupgradable_objects.sql -d $TEST_DBNAME
+
+    MIGRATION_DIR=$(mktemp -d /tmp/migration.XXXXXX)
+    register_teardown rm -r "$MIGRATION_DIR"
+
+    # Set up psqlrc to kill any psql processes as soon as they're started.
+    export PSQLRC="$MIGRATION_DIR"/psqlrc
+    printf '\! kill $PPID\n' > "$PSQLRC"
+
+    "$SCRIPTS_DIR"/migration_generator_sql.bash "$GPHOME_SOURCE" "$PGPORT" "$MIGRATION_DIR"
+    drop_unfixable_objects
+    "$SCRIPTS_DIR"/migration_executor_sql.bash "$GPHOME_SOURCE" "$PGPORT" "$MIGRATION_DIR"/start
+
+    gpupgrade initialize \
+        --source-gphome="$GPHOME_SOURCE" \
+        --target-gphome="$GPHOME_TARGET" \
+        --source-master-port="${PGPORT}" \
+        --temp-port-range 6020-6040 \
+        --disk-free-ratio 0 \
+        --verbose
+    gpupgrade revert --verbose
+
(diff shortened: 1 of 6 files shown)


test/migration_scripts.bats
```diff
@@ -165,6 +114,8 @@ drop_unfixable_objects() {
     EXCLUSIONS+="-T table_with_name_as_second_column "
 
     MIGRATION_DIR=`mktemp -d /tmp/migration.XXXXXX`
```
dkrieger-pvtl, line 116: nit: while you're at it, consider replacing ``` `mktemp -d...` ``` with `$(mktemp -d...)`. See https://github.com/koalaman/shellcheck/wiki/SC2006.  There's one other such usage in this file that you can change as well.

dkrieger-pvtl (commented):
I think you also want to fix the use of psql in `data_migration_scripts/revert/recreate_gphdfs_external_tables.sh`...by adding a `-X` to the psql usage there as well.

kalensk:
> recreate_gphdfs_external_tables.sh

I am not seeing anything that calls this script. Is it still relevant? If so, what is the expected customer use case for using it especially given the `migration_generator_sql.bash` and `migration_executor_sql.bash` workflow?

dkrieger-pvtl:
> > recreate_gphdfs_external_tables.sh
> 
> I am not seeing anything that calls this script. Is it still relevant? If so, what is the expected customer use case for using it especially given the `migration_generator_sql.bash` and `migration_executor_sql.bash` workflow?

The script is called as part of `migration_generator_sql.bash`:

- generate a GPDB-5 database

- add an external table using the appropriate part of test/create_nonupgradeable_objects.sql:
```sql
CREATE OR REPLACE FUNCTION noop() RETURNS integer AS 'select 0' LANGUAGE SQL;
DROP PROTOCOL IF EXISTS gphdfs CASCADE;
CREATE PROTOCOL gphdfs (writefunc=noop, readfunc=noop);

CREATE EXTERNAL TABLE ext_gphdfs (name text)
	LOCATION ('gphdfs://example.com/data/filename.txt')
	FORMAT 'TEXT' (DELIMITER '|');
CREATE EXTERNAL TABLE "ext gphdfs" (name text) -- whitespace in the name
	LOCATION ('gphdfs://example.com/data/filename.txt')
	FORMAT 'TEXT' (DELIMITER '|');
```


- run `./migration_generator_sql.bash /usr/local/gpdb5 15432 erase`

- then `erase/revert/migration_postgres_recreate_gphdfs_external_tables.sql` contains the results of running this script.

So it runs, but I cannot show that this actually runs the `~/.psqlrc` file.  Note that the [8.3 psql doc](migration_postgres_recreate_gphdfs_external_tables.sql) on which GPDB-5 is based does not say that `psql -c` suppresses reading of the `~/.psqlrc` file, but [9.4 psql doc](https://www.postgresql.org/docs/9.4/app-psql.html) on which 6 is based, does.  However, I think that the 5 psql doc is incorrect as I can demonstrate that GPDB-5/psql does suppress reading of `~/.psqlrc`.

That said, it's probably good form to add the `-X` to the use of psql in `recreate_gphdfs_external_tables.sh`.

Note there also appears to be a bug in how we handle `.sh` script generation files: the entire `.sh` script gets inlined in the output file.  I'm looking into that.

kalensk:
As defensive programming sounds like `-X` should be added to `recreate_gphdfs_external_tables.sh`.

I'm still no quite seeing how exactly the `migration_executor_sql` runs `recreate_gphdfs_external_tables.sh`. But that is ok for now. Maybe offline we can figure it out.

kalensk:
> I'm still no quite seeing how exactly the migration_executor_sql runs recreate_gphdfs_external_tables.sh. But that is ok for now. Maybe offline we can figure it out.

@dkrieger-pvtl helped clarify the workflow for me offline. Thanks!

jchampio:
Let's add a quick comment to explain that `-c` implies `-X` for 8.3, since it's not documented.

jchampio:
Added an explanation.

test/migration_scripts.bats
jchampio: This is far enough away from the touched code that I'll leave it for now.

dkrieger-pvtl (approved):
LGTM.

PR #459, opened 2020-09-29
add fullpath for pg_upgrade error files

dkrieger-pvtl:
Currently, when "pg_upgrade --check" prints a file name for the user,
it only prints the basename.  Now, the full path is added to allow the
user to quickly find the file.

This is done in gpupgrade, to avoid needing to ensure that any new
checks added to pg_upgrade in the future contain the full path.

Co-authored-by: Jacob Champion <pchampion@vmware.com>

### DIFF
--- a/hub/upgrade_master.go
+++ b/hub/upgrade_master.go
@@ -8,6 +8,7 @@ import (
 	"bytes"
 	"fmt"
 	"io"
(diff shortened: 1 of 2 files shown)


kalensk (approved):
LGTM

PR #463, opened 2020-10-05
Format the new trigger DDL statement on gpdb6 to match the old DDL atement for gpdb5

bhuvnesh2703:
Format the new trigger DDL statement on gpdb6 to match the old DDL statement for gpdb5

Also Use a common framework to make replacement for the blocks
identified by a pattern to hide the complexity of the formatting
function rules used by different type of statement from the main loop.
Views, Rule, Trigger, With Clause formatting all follow similar set of
actions to perform the reformatting of the dump, so use a common runner
from the main driver to perform the changes.

### DIFF
--- a/ci/scripts/filters/filter/filter.go
+++ b/ci/scripts/filters/filter/filter.go
@@ -3,7 +3,7 @@
 
 /*
 	The filter command massages the post-upgrade SQL dump by removing known
-	differences. It does this with two sets of rules -- lines and blocks.
+	differences. It does this with the following set of rules
 
 	- Line rules are regular expressions that will cause any matching lines to
 	be removed immediately.
@@ -11,8 +11,8 @@
 	- Block rules are regular expressions that cause any matching lines, and any
 	preceding comments or blank lines, to be removed.
 
-	The main complication here comes from the block rules, which require us to
-	use a lookahead buffer.
+	- Formatting rules are a set of functions that can format the sql statement tokens
+	into a desired format
 
 	filter reads from stdin and writes to stdout. Usage:
 
@@ -35,10 +35,66 @@ import (
 )
 
 type ReplacementFunc func(line string) string
+
 var replacementFuncs []ReplacementFunc
 
-var lineRegexes []*regexp.Regexp
-var blockRegexes []*regexp.Regexp
+// function to identify if the line matches a pattern
+type shouldFormatFunc func(line string) bool
+
+// function to create a formatted string using the tokens
+type formatFunc func(tokens []string) (string, error)
+
+// identifier and corresponding formatting function
+type formatter struct {
+	shouldFormat shouldFormatFunc
+	format       formatFunc
+}
+
+// hold the current tokens for the formatting function
+type formatContext struct {
+	tokens     []string
+	formatFunc formatFunc
+}
+
+var (
+	formatters   []formatter
+	lineRegexes  []*regexp.Regexp
+	blockRegexes []*regexp.Regexp
+)
+
+// is formatting currently in progress
+func (f *formatContext) formatting() bool {
+	return f.formatFunc != nil
+}
+
+func (f *formatContext) addTokens(line string) {
+	f.tokens = append(f.tokens, strings.Fields(line)...)
+}
+
+func endFormatting(line string) bool {
+	return strings.Contains(line, ";")
+}
+
+func (f *formatContext) format() (string, error) {
+	return f.formatFunc(f.tokens)
+}
+
+func newFormattingContext() *formatContext {
+	return &formatContext{}
+}
+
+func (f *formatContext) find(formatters []formatter, line string) {
+	if f.formatFunc != nil {
+		return
+	}
+
+	for _, x := range formatters {
+		if x.shouldFormat(line) {
+			f.formatFunc = x.format
+			break
+		}
+	}
+}
 
 func init() {
 	// linePatterns remove exactly what is matched, on a line-by-line basis.
@@ -60,6 +116,12 @@ func init() {
 
 	replacementFuncs = []ReplacementFunc{
 		filters.FormatWithClause,
+	}
+
+	// patten matching functions and corresponding formatting functions
+	formatters = []formatter{
+		{shouldFormat: filters.IsViewOrRuleDdl, format: filters.FormatViewOrRuleDdl},
+		{shouldFormat: filters.IsTriggerDdl, format: filters.FormatTriggerDdl},
 	}
 
 	for _, pattern := range linePatterns {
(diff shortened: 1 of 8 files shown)


ci/scripts/filters/filter/filter.go
```diff
@@ -34,8 +34,19 @@ import (
 	"github.com/greenplum-db/gpupgrade/ci/scripts/filters"
 )
 
-type ReplacementFunc func(line string) string
-var replacementFuncs []ReplacementFunc
+// function to identify if the buf and line pattern matches a pattern
+type IdentifierFunc func(buf []string, line string) bool
+
+// function to use for formatting a block identified by IdentifierFunc
+type FormatFunc func(line string, allTokens []string) (string, []string, bool)
```
kalensk, line 41: If I am reading the code correctly the returned bool is not used and thus should be removed?

```go
if r.IdentifierFunc(buf, line) {
    completeDDL, resultTokens, _ := r.FormatFunc(line, allTokens)
    return completeDDL, resultTokens, r.FormatFunc
}
```

ci/scripts/filters/filter/filter.go
```diff
@@ -91,6 +104,25 @@ func write(out io.Writer, lines ...string) {
...
+		// identify if any formatting is applicable and call the corresponding formatting function
+		if r.IdentifierFunc(buf, line) {
+			completeDDL, resultTokens, _ := r.FormatFunc(line, allTokens)
+			return completeDDL, resultTokens, r.FormatFunc
+		}
+	}
+
+	// nothing to format
+	return "", nil, nil
+}
```
kalensk, line 124: Could you explain the concept of `allTokens` and `resultTokens` and how they are used? Overall, Im having quite a hard time following the ins & outs of this function. Would a function level comment or diagram or something help?

ci/scripts/filters/filter/filter.go
```diff
@@ -91,6 +104,25 @@ func write(out io.Writer, lines ...string) {
 	}
 }
 
+func formatStmt(allTokens []string, line string, bf FormatFunc, buf []string) (string, []string, FormatFunc) {
```
kalensk, line 107: What does `bf` stand for? What about renaming to just `format`, `formatter`, or `formatFunc` or something more descriptive? I think it should match what is named in the Formatting struct.

ci/scripts/filters/filter/filter.go
```diff
@@ -34,8 +34,19 @@ import (
...
+type IdentifierFunc func(buf []string, line string) bool
+
+// function to use for formatting a block identified by IdentifierFunc
+type FormatFunc func(line string, allTokens []string) (string, []string, bool)
+
+// identifier and corresponding formatting func
+type Formatting struct {
+	IdentifierFunc
+	FormatFunc
+}
```
kalensk, line 47: 1) Do these need to be exported?

2) Im curious if something like `shouldReplace` and `replace` make more sense and is clearer? Then the callers would look something like:

```
if f.shouldReplace() {
    f.Replace()
}
```

ci/scripts/filters/filter/filter.go
bhuvnesh2703: sure did the rename.

ci/scripts/filters/filter/filter.go
```diff
@@ -91,6 +105,31 @@ func write(out io.Writer, lines ...string) {
...
+// and it is processing the next lines read from the dump.
+// if allToken is not empty, it means that a replace function is yet to be identified based on pattern matching
+func formatStmt(allTokens []string, line string, replaceFunc replace, buf []string) (string, []string, replace) {
+	// if allTokens are already populated, continue formatting
+	if len(allTokens) > 0 {
+		completeDDL, allTokens := replaceFunc(line, allTokens)
+		return completeDDL, allTokens, replaceFunc
+	}
+
+	for _, r := range formatting {
```
dkrieger-pvtl, line 121: Consider renaming the variables and types for clarity.  Instead of "formatting", "shouldReplace", "replace", "replaceFunc", etc, maybe use different forms of the same word for them all, like "reformatting", "shouldReformat", "reformat", "reformatFunc" so that would make the intent of the code clearer.  Instead of "allTokens", maybe just "tokens"...it's clear from context that it is a collection of tokens.

dkrieger-pvtl (commented):
Initial comments.

dkrieger-pvtl (commented):
Consider reworking the code to make it more "go-like".  It is currently very "C-like" with a lot of line/buf state mixed in with the reformatting state.  Take a look at my [receiver based implementation](https://github.com/dkrieger-pvtl/gpupgrade/commits/dkrieger_bhuvnesh_fix-filter-diff).  I think it is easier to understand. 
 Note that this makes it easy to move the entire implementation of reformatting to another file.

The only unit test which fails on my branch is this one, but it appears to fail on yours as well:
```
--- FAIL: TestBuildTriggerDdl (0.00s)
    --- FAIL: TestBuildTriggerDdl/still_processing_trigger_ddl (0.00s)
        trigger_test.go:55: got CREATE TRIGGER mytrigger
                AFTER INSERT OR DELETE ON public.foo, want
```

dkrieger-pvtl (commented):
The new implementation is cleaner but there is still state that should be encapsulated into the formatter itself...it still takes extra thought to understand the formatters loop.  `tokens` really belongs to the filter itself, but is global state in the main loop.  And the current code always checks for each formatter matching, even if we are in the middle of formatting a current statement.  But since shouldFormat is based on a single line, it looks like we should choose only one until that statement is formatter.

ci/scripts/filters/filter/filter.go
```diff
@@ -35,10 +35,66 @@ import (
...
+
+func endFormatting(line string) bool {
+	return strings.Contains(line, ";")
+}
+
+func (f *formatContext) format() string {
+	return f.formatFunc(f.tokens)
+}
+
+func newformattingContext() *formatContext {
```
dkrieger-pvtl, line 82: nit: this should be [written as](https://golang.org/doc/effective_go.html#mixed-caps) `newFormattingContext`.

ci/scripts/filters/filter/filter.go
```diff
@@ -99,27 +161,20 @@ func Filter(in io.Reader, out io.Writer) {
...
-			allTokens = resultTokens
-			if finishedFormatting {
-				buf = writeBufAndLine(out, buf, completeDdl)
-				formattingViewOrRuleDdlStmt = false
-				allTokens = nil
+		formattingContext.find(formatters, line)
+		if formattingContext.formatting() {
+			formattingContext.addTokens(line)
+			if endFormatting(line) {
+				stmt := formattingContext.format()
```
dkrieger-pvtl, line 174: I know that we think that `formattingContext.format()` must return a valid SQL statement from the tokens, what happens if it doesn't?  Right now, the output diff will fail but it might take a while to figure out.  So consider some error check here.

ci/scripts/filters/view_and_rule.go
```diff
@@ -47,19 +45,19 @@ func init() {
...
 }
 
-func IsViewOrRuleDdl(buf []string, line string) bool {
-	return len(buf) > 0 && ruleOrViewCommentRegex.MatchString(strings.Join(buf, " ")) && ruleOrViewCreateRegex.MatchString(line)
+func IsViewOrRuleDdl(line string) bool {
+	return ruleOrViewCreateRegex.MatchString(line)
 }
 
-func FormatViewOrRuleDdl(allTokens []string) string {
+func FormatViewOrRuleDdl(tokens []string) string {
```
dkrieger-pvtl, line 52: This code assumes there are at least 5 elements in `allTokens`; if not, it will panic.  Consider having an error check that validates this.  I know that `shouldFormat` must pass to get here, but that only checks the first two tokens.

dkrieger-pvtl (approved):
LGTM as long as the comments about checking for errors in the newly added format functions are addressed.  There are many ways of doing this.  Since the overall script here is meant to "work or error out" perhaps the simplest scheme is to have each formatter panic if its assumptions are not met.

ci/scripts/filters/filter/filter.go
bhuvnesh2703: added error check.

ci/scripts/filters/view_and_rule.go
bhuvnesh2703: added.

PR #464, opened 2020-10-06
data_migration_scripts: correctly find .header file

dkrieger-pvtl:
This PR uses a robust scheme to determine if there is a .header
file associated with the given script.  The existing code only
worked correctly for files ending in ".sql"; if the file ended
in any other suffix, the code treated the entire file as the
.header file.

Note that this PR now has a header file even work on a file with no suffix.  That is, a data_migration_script called "do_this_to_tables" can have a corresponding header file "do_this_to_tables.header".  

Testing was done manually, and a pipeline is running.


kalensk (approved):
LGTM. I tested with the new stats sql which had the original issue.

PR #465, opened 2020-10-07
initialize: PR follow-up: print min version for target cluster

dkrieger-pvtl:
This fixes a bug in commit 08dfdd4817afbbcdf871a2c1bc1301a410630c02.

This is a follow-up to #462.  If a low version(like "5.0.0") is provided as the target cluster, use `6.9.0` instead of `5.28.0` as the required minimum in the error message.

### DIFF
--- a/cli/versions.go
+++ b/cli/versions.go
@@ -61,17 +61,26 @@ func accumulateRanges(a *semver.Range, minVersions map[int]string) {
 	}
 }
 
-func minSourceVersion() string {
-	var min int
-	for major := range minSourceVersions {
-		if min == 0 {
-			min = major
+// returns the version string that is the lowest of the major version of "version"
+// or the lowest version supported in minVersions otherwise
+func getMinVersion(version semver.Version, minVersions map[int]string) string {
+
+	major := version.Major
+	min, ok := minVersions[int(major)]
+	if ok {
+		return min
+	}
+
+	var lowest int
+	for major := range minVersions {
+		if lowest == 0 {
+			lowest = major
 		}
-		if major < min {
-			min = major
+		if major < lowest {
+			lowest = major
 		}
 	}
-	return semver.MustParse(minSourceVersions[min]).String()
+	return semver.MustParse(minVersions[lowest]).String()
 }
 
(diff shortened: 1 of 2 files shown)


cli/versions.go
```diff
@@ -101,14 +101,14 @@ func validateVersion(gpHome string, context string) error {
 		errStr := fmt.Sprintf("%s cluster version %%s is not supported.", context)
 
 		major := version.Major
-		minVersion, ok := minVersions[int(major)]
+		min, ok := minVersions[int(major)]
 		if !ok {
-			minVersion = minSourceVersion()
+			min = minVersion(minVersions)
```
kalensk, line 106: I know this was existing code, but Im curious if there might be any ways of easily differentiating between the `minVersions` map and the `minVersion` function. On first glance maybe a bit confusing?

dkrieger-pvtl: There are many different variations of version...a string "6.9.0", the `semver.Version("6.9.0")` of it, `semver.Range`s of semvers, etc.  I've play around with many combinations, but I didn't find anything very pleasing.

For this one, renaming the function to `getMinVersion()` seemed cleanest...What do you think of the fixup(3cbbc6d6ff3d5ee552bff3e535d3922941426b9e) on this PR?

kalensk: I think pulling in the map access into getMinVersion nicely encapsulates the business logic.

dkrieger-pvtl: What do you think now?

cli/versions.go
```diff
@@ -61,17 +61,26 @@ func accumulateRanges(a *semver.Range, minVersions map[int]string) {
...
-func minSourceVersion() string {
-	var min int
-	for major := range minSourceVersions {
-		if min == 0 {
-			min = major
+// returns the version string that is the lowest of the major version of "version"
+// or the lowest version supported in minVersions otherwise
+func getMinVersion(version semver.Version, minVersions map[int]string) string {
+
+	major := version.Major
```
kalensk, line 68: minor nit: I'd inline `version.Major`

kalensk (approved):

PR #466, opened 2020-10-07
add script to gather basic statistics

dkrieger-pvtl:
We use the data-migration framework to generate and gather basic
statistics about the source cluster.  This can be used to gain insights
into gpupgrade timing performance based on cluster statistics.

These are included and run with the other data-migration scripts
in production.

Co-authored-by: David Krieger <dkrieger@vmware.com>

### DIFF
--- a/data_migration_scripts/stats/gen_gather_stats.sh
+++ b/data_migration_scripts/stats/gen_gather_stats.sh
@@ -1,0 +1,56 @@
+#!/bin/bash
+# Copyright (c) 2017-2020 VMware, Inc. or its affiliates
+# SPDX-License-Identifier: Apache-2.0
+
+cat << 'EOF'
+
+SELECT current_database();
+
+-- Cluster Statistics
+SELECT hostname, COUNT(dbid) AS Primaries FROM pg_catalog.gp_segment_configuration WHERE role='p' GROUP BY hostname;
+SELECT hostname, COUNT(dbid) AS Mirrors FROM pg_catalog.gp_segment_configuration WHERE role='m' GROUP BY hostname;
+
+-- Extensions
+SELECT COUNT(*) AS InstalledExtensions FROM pg_catalog.pg_extension;
+
+-- Database size
+SELECT pg_size_pretty(pg_database_size(current_database())) AS DatabaseSize;
+SELECT COUNT(*) as Databases FROM pg_catalog.pg_database;
+
+-- No. of triggers
+SELECT COUNT(*) AS Triggers FROM pg_catalog.pg_trigger;
+
+-- GUCs
+SELECT COUNT(*) AS NonDefaultParameters FROM pg_catalog.pg_settings WHERE source <> 'default';
+
+-- No. of Tablespaces
+SELECT COUNT(*) AS Tablespaces FROM pg_catalog.pg_tablespace;
+
+-- No. of schemas
+SELECT COUNT(nspname) AS Schemas FROM pg_catalog.pg_namespace;
+
+-- Table Statistics
+SELECT COUNT(*) AS OrdinaryTables FROM pg_catalog.pg_class WHERE RELKIND='r';
+SELECT COUNT(*) AS IndexTables FROM pg_catalog.pg_class WHERE RELKIND='i';
+SELECT COUNT(*) AS Views FROM pg_catalog.pg_class WHERE RELKIND='v';
+SELECT COUNT(*) AS ToastTables FROM pg_catalog.pg_class WHERE RELKIND='t';
+SELECT COUNT(*) AS AOTables FROM pg_catalog.pg_appendonly WHERE columnstore = 'f';
+SELECT COUNT(*) AS AOCOTables FROM pg_catalog.pg_appendonly WHERE columnstore = 't';
+SELECT COUNT(*) AS UserTables FROM pg_catalog.pg_stat_user_tables;
+
+-- No. of Columns in AOCO
+SELECT COUNT(*) AS AOCOColumns FROM information_schema.columns
+ WHERE table_name IN (SELECT relid::regclass::text FROM pg_catalog.pg_appendonly WHERE columnstore = 't');
+
+-- Partition Table Statistics
+SELECT COUNT(DISTINCT parrelid) AS RootPartitions FROM pg_catalog.pg_partition;
+SELECT COUNT(DISTINCT parchildrelid) AS ChildPartitions FROM pg_catalog.pg_partition_rule;
+
+-- No. of views
+SELECT COUNT(*) AS Views FROM pg_catalog.pg_views;
+
+-- No. of indexes
+SELECT COUNT(*) AS Indexes FROM pg_catalog.pg_index;
+
+
+EOF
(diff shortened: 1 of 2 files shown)


data_migration_scripts/stats/gen_gather_stats.sh
```diff
@@ -0,0 +1,54 @@
+#!/bin/bash
+# Copyright (c) 2017-2020 VMware, Inc. or its affiliates
+# SPDX-License-Identifier: Apache-2.0
+
+cat << 'EOF'
+-- Cluster Statistics
+SELECT hostname,count(dbid) AS numPrimaries FROM pg_catalog.gp_segment_configuration WHERE role='p' GROUP BY hostname;
+SELECT hostname,count(dbid) AS numMirrors FROM pg_catalog.gp_segment_configuration WHERE role='m' GROUP BY hostname;
```
kalensk, line 8: nit: space after the comma. ie: ` hostname, COUNT(dbid)`

kalensk, line 8: Please capitalize `COUNT` and other SQL syntax to follow the rest of the code.

kalensk, line 8: Either capitalize `numPrimaries` and `numMirrors` to `NumPrimaries` and `NumMirrors` match the other instances or rename all other variables in a consistent style.

data_migration_scripts/stats/gen_gather_stats.sh
```diff
@@ -0,0 +1,54 @@
...
+-- Table Statistics
+SELECT COUNT(*) AS ordinaryTables FROM pg_catalog.pg_class WHERE RELKIND='r';
+SELECT COUNT(*) AS UserTables FROM pg_catalog.pg_stat_user_tables;
+SELECT COUNT(*) AS indexTables FROM pg_catalog.pg_class WHERE RELKIND='i';
+SELECT COUNT(*) AS views FROM pg_catalog.pg_class WHERE RELKIND='v';
+SELECT COUNT(*) AS toastTables FROM pg_catalog.pg_class WHERE RELKIND='t';
+SELECT COUNT(relid) AS AOTables FROM pg_catalog.pg_appendonly WHERE columnstore = 'f';
+SELECT COUNT(relid) AS AOCOTables FROM pg_catalog.pg_appendonly WHERE columnstore = 't';
+--looks like same as pg_stat_all_tables
+SELECT COUNT(*) AS Tables FROM information_schema.tables;
```
kalensk, line 41: We should settle on what tables to include before committing. That is, it appears `information_schema.tables` is not needed as its essentially covered by `pg_stat_user_tables `

data_migration_scripts/stats/gen_gather_stats.sh
```diff
@@ -0,0 +1,54 @@
...
+-- No. of schemas
+SELECT COUNT(nspname) AS Schemas FROM pg_catalog.pg_namespace;
+
+-- Partition Table Statistics
+SELECT COUNT(DISTINCT parrelid) AS RootPartitions FROM pg_catalog.pg_partition;
+SELECT COUNT(DISTINCT parchildrelid) AS ChildPartitions FROM pg_catalog.pg_partition_rule;
+
+-- Table Statistics
+SELECT COUNT(*) AS ordinaryTables FROM pg_catalog.pg_class WHERE RELKIND='r';
+SELECT COUNT(*) AS UserTables FROM pg_catalog.pg_stat_user_tables;
```
kalensk, line 34: What is the difference between these two? If its not useful to have `pg_stat_user_tables ` given the above let's remove it until proven otherwise.

data_migration_scripts/stats/gen_gather_stats.sh
```diff
@@ -0,0 +1,54 @@
...
+
+-- No. of Tablespaces
+SELECT COUNT(*) AS Tablespaces FROM pg_catalog.pg_tablespace;
+
+-- No. of schemas
+SELECT COUNT(nspname) AS Schemas FROM pg_catalog.pg_namespace;
+
+-- Partition Table Statistics
+SELECT COUNT(DISTINCT parrelid) AS RootPartitions FROM pg_catalog.pg_partition;
+SELECT COUNT(DISTINCT parchildrelid) AS ChildPartitions FROM pg_catalog.pg_partition_rule;
```
kalensk, line 30: Shouldn't these be moved under the Table Statistics section?

kalensk (commented):
I have a fairly strong preference to not commit this until we have finalized what stats we need and remove any extra until proven otherwise such as `information_schema.tables` vs. `pg_stat_user_tables`, etc. This is because we tend not to come back to things and the tech debit piles up and its relatively quick to finalize what queries to not duplicate.

We do not "need" to commit before using it. That is, we have the queries and can use them while waiting for this PR to be commit if necessary, although it's less ideal.

data_migration_scripts/stats/gen_gather_stats.sh
```diff
@@ -0,0 +1,54 @@
...
+SELECT hostname,count(dbid) AS numMirrors FROM pg_catalog.gp_segment_configuration WHERE role='m' GROUP BY hostname;
+
+-- Extensions
+SELECT COUNT(*) AS InstalledExtensions FROM pg_catalog.pg_extension;
+
+-- Database size
+SELECT pg_size_pretty(pg_database_size(current_database())) AS DatabaseSize;
+
+-- No. of triggers
+SELECT COUNT(*) AS Triggers FROM information_schema.triggers;
```
bhuvnesh2703, line 17: can we directly query pg_catalog.pg_trigger to get the count?

dkrieger-pvtl: These two queries give, surprisingly to me, different results...the `information_schema` gives 12 and `pg_trigger` 8.  This is after running our data_migration_scripts/test generation sql.  But `pg_trigger` appears to be "more accurate": it givens constraint triggers on our created tables but the other one does not.  So I'll use `pg_trigger`.

data_migration_scripts/stats/gen_gather_stats.sh
dkrieger-pvtl: The view `pg_stat_user_tables` is all tables from `pg_class` that aren't "system tables": part of pg_catalog, information_schema or toast tables.  This seems useful....to see how many tables the customer created directly.

data_migration_scripts/stats/gen_gather_stats.sh
dkrieger-pvtl: I removed the query with `information_schema.tables`.  The view definition is quite complex and reading it does not give my an intuitive sense of what it means.  The [view description](https://www.postgresql.org/docs/8.3/infoschema-tables.html) is not much help either.

data_migration_scripts/stats/gen_gather_stats.sh
dkrieger-pvtl: I actually removed `Num` as a prefix everywhere since all the queries return counts.

kalensk (approved):
Looks good! Thanks :)

PR #468, opened 2020-10-08
add step.HasStepRun

kalensk:
I am open to discussing ideas on different approaches, etc. This will help very helpful when tackling when user runs "steps out of order" feature which is upcoming.

[Pipeline](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:hasStepRun)

### DIFF
--- a/step/file_store.go
+++ b/step/file_store.go
@@ -13,6 +13,7 @@ import (
 )
 
 type Store interface {
+	HasStepRun(idl.Step) (bool, error)
 	Read(idl.Step, idl.Substep) (idl.Status, error)
 	Write(idl.Step, idl.Substep, idl.Status) error
 }
@@ -82,6 +83,20 @@ func (f *FileStore) Read(step idl.Step, substep idl.Substep) (idl.Status, error)
 	}
 
 	return status.Status, nil
+}
+
+func (f *FileStore) HasStepRun(step idl.Step) (bool, error) {
+	steps, err := f.load()
+	if err != nil {
+		return false, err
+	}
+
+	_, ok := steps[step.String()]
+	if !ok {
+		return false, nil
+	}
+
+	return true, nil
 }
 
 // Write atomically updates the status file.
--- a/step/file_store.go
+++ b/step/file_store.go
@@ -72,12 +72,12 @@ func (f *FileStore) Read(step idl.Step, substep idl.Substep) (idl.Status, error)
 		return idl.Status_UNKNOWN_STATUS, err
 	}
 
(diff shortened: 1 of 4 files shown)


bhuvnesh2703:
For execute to start, we should see if the substep Substep_CHECK_UPGRADE has run? initialize step could have been run, but the ultimate requirement is that Substep_CHECK_UPGRADE has run successfully? If this PR looks for the initialize and not the substep, then help me understand how we can use it.

kalensk:
> For execute to start, we should see if the substep Substep_CHECK_UPGRADE has run? initialize step could have been run, but the ultimate requirement is that Substep_CHECK_UPGRADE has run successfully? If this PR looks for the initialize and not the substep, then help me understand how we can use it.

Before each step I would run [validateStep](https://github.com/kalensk/gpupgrade/commit/bff7220de018a8207cb172c99ec7944731e2a8be#diff-39100c1926bc6fbdf87b909cc4028b8aR142) which based on the current step looks up what previous steps have run or not (hence this PR to introduce `step.HasStepRun`). That commit implements "Cases 1, 2, and 3" for checking when steps are run out of order.

I don't want to use `step.HasRun` as that requires a substep. That is, it requires knowing what the first substep is for each step which will change over development of gpupgrade and there is no easy way to know to update `validateStep` when that occurs. This sounds brittle and like a bug waiting to happen.

I looked if we could pass an arbitrary substep into `step.HasRun` which will return `idl.Status_UNKNOWN_STATUS`  to determine if the step has been run or not. The problem is there is no way to differentiate between an actual error or if the step has been run.

bhuvnesh2703:
To check if execute could be run, we should check if initialize last substep was completed successfully and finalize has not been run. (So, a substep for a previous step and the main step for the following command)..
If initialize was run and half completed and we are running execute, we should be able to catch that but if we say that initialize was run even though half and we can run execute, thats a problem.. let me know what edge cases are you concerned for.

kalensk:
> So, a substep for a previous step and the main step for the following command)..

As explained above, Im pretty concerned about the maintenance costs and potential bugs as explained about.

> If initialize was run and half completed and we are running execute, we should be able to catch that but if we say that initialize was run even though half and we can run execute, thats a problem.. 

If Im following we are able to do that if we say a step has run or not.

Let's syncup offline and discuss.

kalensk:
Closing for now as per offline team discussion. Will implement `step.HasStepCompleted` and attempt to create a `steps.json` to track overall step status.

PR #469, opened 2020-10-09
Add AO, AOCO and partition tables for testing revert.

bhuvnesh2703:
Add AO, AOCO and partition tables for testing revert.


dkrieger-pvtl (approved):
LGTM...this change as is is good.  I am not sure if it completely satisfies the request to "add more data to revert tests" however, so perhaps additional PRs are needed.

PR #471, opened 2020-10-13
Update README

kalensk:
Add crash-only idempotence, supported versions, and link to blog post.

See rendered example [here](https://github.com/kalensk/gpupgrade/blob/updateREADME/README.md)


kalensk:
Updated to include purpose section.

dkrieger-pvtl (approved):
LGTM

PR #472, opened 2020-10-13
Add flag to disable version check

bhuvnesh2703:
Add flag to disable version check


cli/commands/commands.go
```diff
@@ -479,6 +484,7 @@ To return the cluster to its original state, run "gpupgrade revert".
 	subInit.Flags().BoolVarP(&verbose, "verbose", "v", false, "print the output stream from all substeps")
 	subInit.Flags().StringVar(&ports, "temp-port-range", "", "set of ports to use when initializing the target cluster")
 	subInit.Flags().StringVar(&mode, "mode", "copy", "performs upgrade in either copy or link mode. Default is copy.")
+	subInit.Flags().BoolVar(&noVersionCheck, "no-version-check", false, "disable source and target version check")
```
kalensk, line 487: Let's rename this to `skip-version-check` and the associated variable as well.

kalensk (approved):

dkrieger-pvtl (approved):
LGTM.  The added flag does not need to be hidden via the Cobra interface because we override cobra's help system, anyway, so none of the flags are shown explicitly via Cobra.  We show the user explicit flags in our hand-crafted help message.

kalensk:
> LGTM. The added flag does not need to be hidden via the Cobra interface because we override cobra's help system, anyway, so none of the flags are shown explicitly via Cobra. We show the user explicit flags in our hand-crafted help message.

Even though this is true, isn't it still good practice to follow the other hidden flags we have? Basically, whatever we choose to do in terms of hidden flags we should be consistent.

PR #474, opened 2020-10-14
add steps.json to track overall step status

kalensk:
**_Note:_** If viewing all the files changed be sure to click "Load diff" for `step_test.go`.

In order to check if steps are run out of order we first track the overall step status. This includes if the step has run or not or completed. 

We first rename `status.json` to `substeps.json` in order to differentiate it from the new `steps.json` file.

To reduce the code change required and for convenience `steps.json` uses the same data structure and file store as `substeps.json`. An "internal" substep enum STEP_STATUS has been added to track the overall step status. This is _not_ to be used as a normal substep, but rather internally to track the overall step status.

Finally we add `HasStepStarted` and `HasStepCompleted` helpers which will be used in a followup PR to call `validateStep` when constructing a `NewStep` to determine if a step is ran out of order.

[Pipeline](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:stepStatus)


cli/commanders/step.go
```diff
@@ -135,6 +167,14 @@ func (s *CLIStep) SetNextActions(suggestRevert bool) {
 func (s *CLIStep) Complete(completedText string) error {
 	logDuration(s.stepName, s.verbose, s.timer.Stop())
 
+	// After Finalize and Revert have completed the state directory no longer
+	// exists. Thus, the status cannot be updated.
+	if s.step != idl.Step_FINALIZE && s.step != idl.Step_REVERT {
+		if wErr := Write(s.step, idl.Status_COMPLETE); wErr != nil {
+			s.err = errorlist.Append(s.err, wErr)
+		}
```
dkrieger-pvtl, line 175: This function should be modified to write `FAILED` if there is an error anywhere in the call chain for a step...our usage skips all substeps after the first error and then calls `Complete` so we must do this.  Note this goes along with the out-of-band state-dir directory as we are guaranteed to have the config dir (and the steps.json file) if we get here.


```go
+               status := idl.Status_COMPLETE
+               if s.err != nil {
+                       status = idl.Status_FAILED
+               }
+               if wErr := Write(s.step, status); wErr != nil {
                        s.err = errorlist.Append(s.err, wErr)
                }
```

Note this involves delete the three calls during failed substeps in this file from:
```go
			if wErr := Write(s.step, idl.Status_FAILED); wErr != nil {
				s.err = errorlist.Append(s.err, wErr)
			}
```

since the substep call chain always hits `s.Complete()` which is responsible for writing FAILED to the step status file.

dkrieger-pvtl (commented):
Overall, the design looks good.  I think the flow of the creation of the status file directory should be modified as suggested below to reduce the number of special cases in steps.go.

cli/commanders/step.go
kalensk: I might be misunderstanding and we can sync offline for clarity, but what you are describing is already taken care of in the `defer` inside `RunCLISubstep` already handles writing step status `failed` if writing to stdout/stderr fails.

cli/commands/commands.go
```diff
@@ -392,7 +392,10 @@ func initialize() *cobra.Command {
 			// dump on failure.
 			cmd.SilenceUsage = true
 
-			st := commanders.NewStep(idl.Step_INITIALIZE, &step.BufferedStreams{}, verbose)
+			st, err := commanders.NewStep(idl.Step_INITIALIZE, &step.BufferedStreams{}, verbose)
+			if err != nil {
+				return err
+			}
 
 			st.RunInternalSubstep(func() error {
```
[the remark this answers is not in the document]
kalensk: Sure, since `Complete` tries to access the steps.json file its probably easier to attempt the state directory creation outside the step framework. I can update it to:

```go
// Create the state directory outside the step framework to ensure
// we can write to the status file. That is, the step framework
// assumes a valid working state directory.
{
    err := commanders.CreateStateDir()
    if err != nil {
        return cli.NewNextActions(err, strings.ToLower(idl.Step_INITIALIZE.String()), false)
    }
    commanders.Write(idl.Step_INITIALIZE, idl.Status_RUNNING)
}
```

cli/commanders/step.go
dkrieger-pvtl: That `defer` only runs in `RunCLISubstep`...but `st.Compete()` is _still_ called in any case and will write `COMPLETE`.  When I hack the code to have START_HUB reutrn an error `foofoo`, you can see that the steps.json incorrectly gets `COMPLETE` written to it:

```
gpupgrade initialize --source-gphome /usr/local/gpdb5 --target-gphome /usr/local/gpdb6 --source-master-port 15432 --disk-free-ratio 0 --mode copy --skip-version-check

Initialize in progress.

Starting gpupgrade hub process...                                  [FAILED]

Error: substep "START_HUB": foofoo

NEXT ACTIONS
------------
Please address the above issue and run "gpupgrade initialize" again.


If you would like to return the cluster to its original state, please run "gpupgrade revert".

  |2.3.1| dkrieger-a01 in ~/go/src/github.com/greenplum-db/gpupgrade
±  |detached:remotes/kalensk/stepStatus {146} U:1 ?:21 ✗| → cat ~/.gpupgrade/steps.json
{
  "INITIALIZE": {
    "STEP_STATUS": "COMPLETE"
  }
}
```

cli/commanders/step.go
```diff
@@ -84,6 +94,16 @@ func (s *CLIStep) RunCLISubstep(substep idl.Substep, f func(streams step.OutStre
...
+
+			// If deleting the state directory on the master host failed we
+			// cannot write status failed since the state directory may not exist.
+			if substep == idl.Substep_DELETE_MASTER_STATEDIR {
+				return
+			}
+
+			if wErr := Write(s.step, idl.Status_FAILED); wErr != nil {
+				s.err = errorlist.Append(s.err, wErr)
+			}
```
dkrieger-pvtl, line 106: I wonder if the intent of the code would be cleaner with the check inside the write:

```go
			if wErr := Write(s.step, idl.Status_FAILED); wErr != nil {
			    // If deleting the state directory on the master host failed we
			    // cannot write status failed since the state directory may not exist.
			    if substep == idl.Substep_DELETE_MASTER_STATEDIR {
				    return
			    }
				s.err = errorlist.Append(s.err, wErr)
			}

```

cli/commanders/step.go
```diff
@@ -135,6 +155,19 @@ func (s *CLIStep) SetNextActions(suggestRevert bool) {
 func (s *CLIStep) Complete(completedText string) error {
 	logDuration(s.stepName, s.verbose, s.timer.Stop())
 
+	status := idl.Status_COMPLETE
+	if s.Err() != nil {
+		status = idl.Status_FAILED
+	}
+
+	// After Finalize and Revert have completed the state directory no longer
+	// exists. Thus, the status cannot be updated.
```
dkrieger-pvtl, line 164: nit: may this would be better phrased as:
```
// Finalize and Revert delete the state directory.  Thus, if we are running those steps
// we cannot be sure if it exists.  Thus, the status cannot be updated.
```

cli/commanders/step.go
kalensk: Possibly....I find the nested if's a bit hard to read.

cli/commanders/step.go
kalensk: One major consideration is that now that we write the failed status in `Complete` (versus closer to where the error actually happened when running the step) is that if Finalize or Revert fails it won't be written to the `steps.json` due to the "if condition". 

```go
if s.step != idl.Step_FINALIZE && s.step != idl.Step_REVERT {
    if wErr := Write(s.step, status); wErr != nil {
        s.err = errorlist.Append(s.err, wErr)
    }
}
```

High level, this may be "ok"for now as we only are using `HasStepStarted` and `HasStepCompleted`. I think this will be a nasty bug or issue down the line and it does not keep the API steady. 

I suggest moving back writing the failed status back after the substep is run and errors out. Other ideas?

cli/commanders/step.go
```diff
@@ -129,6 +149,19 @@ func (s *CLIStep) RunCLISubstep(substep idl.Substep, f func(streams step.OutStre
...
+	if s.step != idl.Step_FINALIZE && s.step != idl.Step_REVERT {
```
bhuvnesh2703, line 158: i think we shouldn't have any dependence on the step, finalize or revert, if the file is there go ahead and write whether its in finalize or revert.. only in case where the write fails you can return early

kalensk:
Thank you for all the wonderful feedback, ideas, and discussion! I really appreciate it as it makes the code so much better.

Please check out the latest changes:
- Added a StepStore in `step_store.go` to extract and encapsulate all the store related functions.
- The store gets created in `NewStep` and writes to `steps.json`
- The state directory is created outside the step framework
- The store is part of the step framework and writes either error or complete if the store is not disabled (ie: not null)
- The top level (ie: `commands.go`) tells the step framework when the store is no longer valid. That is when the state directory is removed the store is no longer valid. This seemed like the cleanest approach with no "hidden magic" on when things are valid or not. This also does not ignore errors if one can no longer write to the step store for some reason.
- I tried rearranging the step framework and the store, but nothing (so far) seemed like the right approach or clean code.

Please test it out!

cli/commanders/step_store.go
```diff
@@ -0,0 +1,71 @@
...
+// track the overall step status and should not be used as a normal substep.
+type StepStore struct {
+	store *step.FileStore
+}
+
+func NewStepStore() (*StepStore, error) {
+	path, err := utils.GetJSONFile(utils.GetStateDir(), StepsFileName)
+	if err != nil {
+		return &StepStore{}, xerrors.Errorf("getting %q file: %w", StepsFileName, err)
+	}
```
dkrieger-pvtl, line 27: If I run `gpupgrade finalize` first, without having run `initialize`, this will fail in an odd way for the user, since the statedir does not exist yet.  There are a couple of ways to solve this, but perhaps you want this PR to continue with this issue and fix it with the subsequent validateSteps PR....
```
gpupgrade finalize
Error: getting "steps.json" file: open /Users/dkrieger/.gpupgrade/steps.json: no such file or directory
```

kalensk: > If I run gpupgrade finalize first, without having run initialize

Worth nothing, but I don't think this use case applies to this PR. If anything, the upcoming validate step PR perhaps.

dkrieger-pvtl: Sure, we can deal with it in the next PR...

bhuvnesh2703 (approved):

PR #476, opened 2020-10-15
allow substeps to override the default next actions

kalensk:
allow substeps to override the default next actions

[Pipeline](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:nextActions)


dkrieger-pvtl (approved):
LGTM....the override method here is cleaner than passing in args as we can then format nextActions as we want.

PR #477, opened 2020-10-19
Display warning message if mirror and/or standby segment doesn't exist

bhuvnesh2703:
Display warning message if mirror and/or standby segment doesn't exist


dkrieger-pvtl (commented):
LGTM...the approach is clean and uses our existing "map[string]string" method to send (key, value) pairs from the hub back to the cli.  

In this case, the values are strings representing bools("true", "false").  I think it'd be cleaner to convert those values to bools directly somewhere in the chain between `UILoop` where the response from the hub is first seen by the cli and `InitializeWarningMessageIfAny` where we interpret those values.  Since the (key, value) string pair map is really a low-level protocol between the cli and the hub, I think `InitializeCreateCluster` would be the place to do this conversion to bools.  If the conversion fails, it is an error in that low-level interaction.

Ideally, we'd write integration(bats) tests to test this end-to-end but we don't currently test any such clusters, and will need to add it to the list of tests to add at some point in the future.  That said, it'd be great if you manually test a few cases.

cli/commands/commands.go
```diff
@@ -969,3 +982,20 @@ func addHelpToCommand(cmd *cobra.Command, help string) *cobra.Command {
...
+	} else if response.HasStandby == "false" {
+		message = "standby"
+	}
+
+	if message != "" {
+		return fmt.Sprintf(InitializeWarningMessage, message)
+	}
+
+	return message
+}
```
kalensk, line 1001: I think the variety of `if's` is telling us to pause and see if there is a cleaner way perhaps (might not be, but worth considering). What if the response simply sends back the warning text string to interpret. That is, the response has the text something like `[warningText] = "standby and mirror segments"` (or whatever the text needs to say).

Then only print the warning if warningText is not empty.

kalensk:
My personal preference is to simply return a single bool that says there is no standby and/or mirrors and keep the text generic since there is no advantage (as far as I can tell) to special casing the text. The special casing adds a lot of `if statements` and complexity for no real gain. If there was a stronger use case than I could see more of a reason to.

kalensk (approved):

PR #478, opened 2020-10-20
Refactor Response with specific types

bhuvnesh2703:
Earlier, the responses were returned as a map[string]string which
required casting of the values to the desired types once received or
sent. So this refactor introduces response types for each of the
specific phases of gpupgrade and instead of maps relies on simple
structs holding the desired values.


dkrieger-pvtl (commented):
I think we've replaced one problem with another...the original code was very clean provided the cli-side only needed string-type responses from the hub.  Now that we need bools, we've decided we want types.  We could have done this cli-side or hub-side, and we've decided to go hub-side, which is cleaner...after all, the hub knows the types it is sending.

But the PR, as written, has a lot of boilerplate code to translate the Response protobuf struct to a "pure" golang struct.  This is hard to maintain...to add a field in the future, we need to change the protobuf and the struct.

Why not just use the protobuf Response itself?  Our protobuf structs are "first-class" and we intend to use them as such.  See https://github.com/dkrieger-pvtl/gpupgrade/commit/449550bcc7c1c699f3cea442d7d521779b0d7af3 for an idea on this...it doesn't compile(it takes me a while to get all the nested protobuf defs right) but I think you get the idea.  It is simpler, adding a field will just require adding it to protobuf and pulling it out where it is used, AND the Response is nested, since these are in fact related types.

What do you think?

kalensk:
> Why not just use the protobuf Response itself?

Agreed! This would keep the code much simpler and no need to map between the gRPC and golang struct which we don't need anyway.

cli/commanders/steps.go
```diff
@@ -16,23 +16,30 @@ import (
 	"github.com/greenplum-db/gpupgrade/utils/stopwatch"
 )
 
+type UILoopResponse struct {
+	InitializeCreateClusterResponse
+	ExecuteResponse
+	FinalizeResponse
+	RevertResponse
+}
+
```
kalensk, line 25: I think this can be removed with the latest changes.

cli/commands/commands.go
```diff
@@ -555,7 +555,7 @@ If you are satisfied with the state of the cluster, run "gpupgrade finalize"
 to proceed with the upgrade.
 
 To return the cluster to its original state, run "gpupgrade revert".
-`, response.TargetPort, response.TargetMasterDataDir))
+`, strconv.Itoa(int(response.GetTarget().GetPort())), response.GetTarget().GetMasterDataDirectory()))
```
kalensk, line 558: The first interpolation can be simplified to: `response.GetTarget().GetPort()` with using `%d`.

Similar update could be done to the other responses that use an int such as in the finalize text and revert.

kalensk (commented):
Much cleaner with the latest changes! Thank you for doing this refactor as I think its definitely helpful.

I think all the response structs in steps.go can be removed now that we are using the gRPC types? That is, InitializeCreateClusterResponse, ExecuteResponse, FinalizeResponse, RevertResponse in steps.go.

kalensk (approved):

cli/commanders/steps_test.go
```diff
@@ -182,37 +181,108 @@ func TestUILoop(t *testing.T) {
...
+					Contents: &idl.Message_Response{
+						Response: &idl.Response{
+							Contents: &idl.Response_InitializeResponse{
+								InitializeResponse: &idl.InitializeResponse{
+									HasMirrors: true,
+								},
+							},
+						},
+					},
+				}},
```
dkrieger-pvtl, line 202: nit: this and the other similar uses in this PR take up a lot of vertical space.  You could consider writing them like this instead:
```go
 &idl.Message{Contents: &idl.Message_Response{Response: &idl.Response{
					Contents: &idl.Response_InitializeResponse{InitializeResponse: &idl.InitializeResponse{
						HasMirrors: true,
					}}}}},
```

dkrieger-pvtl (approved):
LGTM...please consider the nit to collapse the vertical space of making the Response messages.

PR #479, opened 2020-10-20
validate that steps are run in the correct order

kalensk:
Validate that steps are run in the correct order.

This PR also allows NextActions to use a custom next action string allowing  callers to simply set next actions message directly. This removes the need for suggestRevert. Rename errors.go to next_actions.go

[Pipeline](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:validateStepUsingMap)


cli/commanders/step_store_test.go
```diff
@@ -218,3 +221,210 @@ func TestStepStore(t *testing.T) {
...
+		{
+			"can run revert after execute has started",
+			idl.Step_REVERT,
+			[]stepStatus{
+				{step: idl.Step_INITIALIZE, status: idl.Status_COMPLETE},
+				{step: idl.Step_EXECUTE, status: idl.Status_RUNNING},
+			},
+		},
+	}
+
```
dkrieger-pvtl, line 399: Additional positive test cases: initialize can be re-run if initialize is running, initialize can be re-run if initialize has completed, execute can be re-run if execute is running, execute can be re-run if execute has completed.

cli/commanders/step_store_test.go
```diff
@@ -218,3 +221,210 @@ func TestStepStore(t *testing.T) {
...
+			}
+
+			err = store.ValidateStep(c.currentStep)
+			if err != nil {
+				t.Errorf("unexpected err %#v", err)
+			}
+		})
+	}
+}
+
```
dkrieger-pvtl, line 415: We have 11 negative test cases and 5 positive ones, totalling 16 scenarios.  Looking at it abstracting, we have 4 operations `initialize, execute, finalize, revert` and 4 step states `INVALID(never run), FAILED, RUNNING, COMPLETE`.  This gives us 4 states for each of 4 steps, or 4*4*4*4 or 256 scenarios that are possible.  Now, not all scenarios are reasonable, but I wonder if we can think of it this way to make sure we are covering all the test cases we need to.

We could automatically generate the test cases above and then look at the results(valid or not) and at least pull out more positive test cases and perhaps some negative ones.  Adding in required NextActions complicates things, but we could only do coverage and not combinatoric testing for that.

We could also reduce the combinatorics by logic, that is, there's no need to test if initialize can run depending on the state of finalize as execute/revert are the only direct dependencies for initialize.

Just an idea to make sure we are covering what we need to.

dkrieger-pvtl: After discussion, we've decided the current set of tests is sufficient.

cli/commanders/step_store_test.go
kalensk: Thanks! I added those test cases and a few others. Although I am holding off on execute can be re-run if execute has completed for now. Will re-evaluate based on customer-need/feedback.

bhuvnesh2703 (approved):
LGTM, but the value of RunRevert must be fixed/finalized.

dkrieger-pvtl (approved):
LGTM....I especially like the explicit listing of all other steps on the list of conditions to validate for a step.

PR #483, opened 2020-10-23
Fix RevertResponse to return source cluster information

kalensk:
It was accidentally returning the target cluster information. Also update the style/formatting of the other responses for readability.

[pipeline](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:fixRevertResponse)

### DIFF
--- a/hub/finalize.go
+++ b/hub/finalize.go
@@ -99,13 +99,15 @@ func (s *Server) Finalize(_ *idl.FinalizeRequest, stream idl.CliToHub_FinalizeSe
 		})
 	}
 
-	message := &idl.Message{Contents: &idl.Message_Response{Response: &idl.Response{
-		Contents: &idl.Response_FinalizeResponse{FinalizeResponse: &idl.FinalizeResponse{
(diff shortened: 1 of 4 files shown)


hub/finalize.go
```diff
@@ -99,13 +99,14 @@ func (s *Server) Finalize(_ *idl.FinalizeRequest, stream idl.CliToHub_FinalizeSe
 		})
 	}
 
-	message := &idl.Message{Contents: &idl.Message_Response{Response: &idl.Response{
-		Contents: &idl.Response_FinalizeResponse{FinalizeResponse: &idl.FinalizeResponse{
-			TargetVersion: s.Target.Version.VersionString,
```
dkrieger-pvtl, line 104: This needs to still be included in the output; it gets output to the user at the end of a successful `gpupgrade finalize`.

dkrieger-pvtl (commented):
There is the accidental removal of the target version in finalize.  Other than that, looks fine.

bhuvnesh2703 (approved):
LGTM, just ad the target version in finalize response.

hub/finalize.go
kalensk: doh!

dkrieger-pvtl (approved):
LGTM

PR #485, opened 2020-10-23
Add binary version check

bhuvnesh2703:
add functionality to validate if the hub and agents are of the same version.

### DIFF
--- a/cli/commands/version.go
+++ b/cli/commands/version.go
@@ -11,7 +11,7 @@ var Commit string
 var Release string
 
 func VersionString() string {
-	return fmt.Sprintf("Version: %s\nCommit: %s\nRelease: %s", Version, Commit, Release)
+	return fmt.Sprintf("Version: %s Commit: %s Release: %s", Version, Commit, Release)
 }
 
 func printVersion() {
--- a/hub/version.go
+++ b/hub/version.go
@@ -1,0 +1,102 @@
+// Copyright (c) 2017-2020 VMware, Inc. or its affiliates
+// SPDX-License-Identifier: Apache-2.0
+
+package hub
+
+import (
+	"fmt"
+	"sort"
+	"sync"
+
+	"github.com/greenplum-db/gp-common-go-libs/gplog"
+	"golang.org/x/xerrors"
+
+	"github.com/greenplum-db/gpupgrade/utils/errorlist"
+)
+
+var GetVersionFunc = GetVersion
+
+type agentVersion struct {
+	host    string
+	version string
+	err     error
+}
+
+type agents map[string]string
+
+func (a agents) String() string {
+	hosts := make([]string, 0, len(a))
+	for h := range a {
+		hosts = append(hosts, h)
+	}
+
+	s := ""
+	sort.Strings(hosts)
+	for _, k := range hosts {
+		s += fmt.Sprintf("%s: %s\n", k, a[k])
+	}
+	return s
+}
+
+func ValidateGpupgradeVersion(hubHost string, agentHosts []string) error {
+	path, err := getBinaryPath()
+	if err != nil {
+		return xerrors.Errorf("getting gpupgrade binary path: %w", err)
+	}
+
+	hubVersion, err := GetVersionFunc(hubHost, path)
+	if err != nil {
+		return xerrors.Errorf("getting hub version: %w", err)
+	}
+
+	var wg sync.WaitGroup
+	agentChan := make(chan agentVersion, len(agentHosts))
+
+	for _, host := range agentHosts {
+		wg.Add(1)
+		go func(host string) {
+			defer wg.Done()
+			version, err := GetVersionFunc(host, path)
+			agentChan <- agentVersion{host, version, err}
+		}(host)
+	}
+
+	wg.Wait()
+	close(agentChan)
+
+	var errs error
+	mismatchedAgents := make(agents)
+	for agent := range agentChan {
+		errs = errorlist.Append(errs, agent.err)
+		if hubVersion != agent.version {
+			mismatchedAgents[agent.host] = agent.version
+		}
+	}
+
+	if errs != nil {
+		return errs
+	}
+
+	if len(mismatchedAgents) == 0 {
+		return nil
+	}
+
+	return xerrors.Errorf(`Version mismatch between gpupgrade hub and agent hosts. 
+Hub version: %s
+
+Mismatched Agents:
+%s`, hubVersion, mismatchedAgents.String())
+}
+
+func GetVersion(host, path string) (string, error) {
+	cmd := execCommand("ssh", host, fmt.Sprintf(`bash -c "%s version"`, path))
+	gplog.Debug("running cmd %q", cmd.String())
+	output, err := cmd.CombinedOutput()
+	if err != nil {
+		return "", xerrors.Errorf("%q failed with %q: %w", cmd.String(), string(output), err)
+	}
+
+	gplog.Debug("output: %q", output)
+
+	return string(output), nil
+}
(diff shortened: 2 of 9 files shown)


hub/version.go
```diff
@@ -0,0 +1,87 @@
...
+		return nil
+	}
+
+	sort.Strings(mismatchedHosts)
+	return xerrors.Errorf(`Version mismatch between gpupgrade hub and agent hosts. Found hub version:
+%s
+
+Agents with mismatched version: %s`, hubVersion, strings.Join(mismatchedHosts, ", "))
+}
+
```
dkrieger-pvtl, line 71: I think we should print out the agent version for mismatch agents.  Here's one idea.  There are three parts;
1). separating out the format of the mismatched agents from the function itself
2). making the go func loop "tighter" and deferring processing to outside the loop; this separates the concerns of calling functions on each agent and aggregating the results on the hub
3). returning an error if there is one(missing from the PR; should be added in any case).

Perhaps it'd be easier for us to pair on this....


```go
type agentVersion struct {
	host    string
	version string
	err     error
}

// FIXME: arrange this in the format we want
func (a *agentVersion) String() string {
	return fmt.Sprintf("HOST:VERSION %s:%s", a.host, a.version)
}

type agentVersions []agentVersion

// FIXME: arrange this in the format we want
func (as agentVersions) String() string {
	s := ""
	for _, agent := range as {
		s += agent.String()
	}
	return s
}

func ValidateGpupgradeVersion(hubHost string, agentHosts []string) error {
	path, err := getBinaryPath()
	if err != nil {
		return xerrors.Errorf("getting gpupgrade binary path: %w", err)
	}

	hubVersion, err := GetVersionFunc(hubHost, path)
	if err != nil {
		return xerrors.Errorf("getting hub version: %w", err)
	}

	// collect version information from agents
	var wg sync.WaitGroup
	agentChan := make(chan agentVersion, len(agentHosts))
	for _, host := range agentHosts {
		wg.Add(1)
		go func(host string) {
			defer wg.Done()
			version, err := GetVersionFunc(host, path)
			agentChan <- agentVersion{host, version, err}
		}(host)
	}
	wg.Wait()
	close(agentChan)

	// parse that information
	var errs error
	var mismatched agentVersions
	for r := range agentChan {
		errs = errorlist.Append(errs, r.err)
		mismatched = append(mismatched, r)
	}

	if errs != nil {
		return errs
	}

	if len(mismatched) == 0 {
		return nil
	}

	return xerrors.Errorf(`Version mismatch between gpupgrade hub and agent hosts. Found hub version:
%s

Mismatched Agents: %s`, hubVersion, mismatched.String())
}
```

kalensk: _If_ we end up needing to parse the version this function should help. It can go in `cli/commands/version.go`

```go
func ParseVersionString(text string) (string, error) {
	versionLine := strings.Split(text, "\n")[0]
	if !strings.HasPrefix(versionLine, versionPrefix) {
		return "", xerrors.Errorf("Expected prefix of version to be %q", versionPrefix)
	}
	version, err := semver.Parse(versionLine[len(versionPrefix):])
	if err != nil {
		return "", err
	}
	return version.String(), nil
}
```

hub/version.go
```diff
@@ -0,0 +1,87 @@
...
+		if xerrors.As(err, &exitErr) {
+			return "", xerrors.Errorf("%q failed with %q: %w", cmd.String(), exitErr.Stderr, err)
+		}
+		return "", xerrors.Errorf("%q failed with: %w", cmd.String(), err)
+	}
+
+	gplog.Debug("output: %q", version)
+
+	return string(version), nil
+}
```
kalensk, line 87: I still think using `cmd.CombinedOutput` would simplify the code quite a bit (assuming there are no unforeseen side effects)?

hub/version.go
kalensk: _If_ we end up needing to parse the version this function should help. It can go in `cli/commands/version.go`

```go
func ParseVersionString(text string) (string, error) {
	versionLine := strings.Split(text, "\n")[0]
	if !strings.HasPrefix(versionLine, versionPrefix) {
		return "", xerrors.Errorf("Expected prefix of version to be %q", versionPrefix)
	}
	version, err := semver.Parse(versionLine[len(versionPrefix):])
	if err != nil {
		return "", err
	}
	return version.String(), nil
}
```

dkrieger-pvtl:
This PR uses the version string returned from `gpupgrade version`, whatever it might be, for comparison between the hub and agents.  The version in my working version right now is:
```
Version: 0.4.0
Commit: 04a325df
Release: Dev Build
```

To the user, this would look a bit odd.  Do we just want to compare semvers, that is, `0.4.0`?  This would require us to have confidence that we always change the semver when we release a new version, but we should really be doing that.

kalensk:
> To the user, this would look a bit odd. Do we just want to compare semvers, that is, 0.4.0? 

One idea is to use the iterative approach and compare the whole version output until it becomes an issue. Ideally we are checking that the customer has the same binary installed across their cluster.

kalensk (approved):
LGTM

- Please consider if `cmd.CombinedOutput` would be cleaner.

- If the team decides to go with David's suggested pattern. I have a fairly strong preference for it to be applied across all the code at once as cohesive pattern. Since I find having divergent patterns like that much more confusing and harder to maintain.

hub/version.go
bhuvnesh2703: I did refactoring based on suggestions, but didn't include the producer consume suggestion as of now.
also, we can parse the version, then it begs the question why do we produce the output in a way which we don't want to rely on completely. so will build upon it as we get feedback.

hub/version_test.go
```diff
@@ -0,0 +1,99 @@
...
+		if err == nil {
+			t.Errorf("expected an error")
+		}
+
+		expectedRegex := regexp.MustCompile(`Mismatched Agents:.*\nsdw1.*\nsdw2`)
+		if !expectedRegex.Match([]byte(err.Error())) {
+			t.Errorf("expected sdw1 and sdw2 in mismatched agents, got %s", err)
+		}
+	})
+}
```
dkrieger-pvtl, line 99: I think you should add two new tests: one to validate a success case if all versions match and one to error if there is an error getting the agent's information(most likely due to a temporary ssh error, perhaps).  I also have a few code comments.  See all suggested changes here:

https://github.com/dkrieger-pvtl/gpupgrade/commit/35a3b87590307ec6f8a77880ff49d73abffa13da

You can pull the last commit from https://github.com/dkrieger-pvtl/gpupgrade/commits/review_add-binary-version-check, based on your latest branch, to get the code changes locally.

dkrieger-pvtl (commented):
Overall LGTM but please consider adding a few tests cases as suggested.

cli/commands/version.go
```diff
@@ -11,7 +11,7 @@ var Commit string
 var Release string
 
 func VersionString() string {
-	return fmt.Sprintf("Version: %s\nCommit: %s\nRelease: %s", Version, Commit, Release)
+	return fmt.Sprintf("Version: %s Commit: %s Release: %s", Version, Commit, Release)
```
kalensk, line 14: Let's add a comment that says that `ValidateGpupgradeVersion` expects the version output to be on a single line.

That way if later on someone updates the version output they will know to also update `ValidateGpupgradeVersion` if necessary.

kalensk:
Keep in mind that if it's easier we can keep the version output on multiple lines and simply do a `strings.ReplaceAll("\n", " ")` in `ValidateGpupgradeVersion`

hub/version_test.go
dkrieger-pvtl: Looks good.

dkrieger-pvtl (approved):
LGTM

PR #486, opened 2020-10-23
print run initialize in next actions when any step is run before initialize

kalensk:
When any command is run before initialize print a helpful `Note` about not being able to write to the state directory in case it is a legitimate error. Add the run initialize help text in the next actions. This PR also adds the correct spacing to `stepErr`.

This PR also adds a minor fix to remove an extra trailing newline to the completed text.

Before Fix:
```
Error: getting "steps.json" file: open /home/gpadmin/.gpupgrade/steps.json: no such file or directory

NEXT ACTIONS
------------
gpupgrade commands must be issued in correct order
  1. initialize   runs pre-upgrade checks and prepares the cluster for upgrade
  2. execute      upgrades the master and primary segments to the target
                  Greenplum version
  3. finalize     upgrades the standby master and mirror segments to the target
                  Greenplum version. Revert cannot be run after finalize has started.
Use "gpupgrade --help" for more information
```

After Fix:
```
Error: gpupgrade commands must be issued in correct order

  1. initialize   runs pre-upgrade checks and prepares the cluster for upgrade
  2. execute      upgrades the master and primary segments to the target
                  Greenplum version
  3. finalize     upgrades the standby master and mirror segments to the target
                  Greenplum version. Revert cannot be run after finalize has started.

Use "gpupgrade --help" for more information

Note: If commands were issued in order, ensure gpupgrade can write to /Users/gpadmin/.gpupgrade

NEXT ACTIONS
------------
To begin the upgrade, run "gpupgrade initialize".
```

[Pipeline](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:fixValidateStep)

### DIFF
--- a/cli/commanders/step.go
+++ b/cli/commanders/step.go
@@ -170,7 +170,7 @@ func (s *CLIStep) Complete(completedText string) error {
 	}
 
 	if s.Err() != nil {
-		fmt.Println()
+		fmt.Println() // Separate the step status from the error text
 
 		// allow substpes to override the default next actions
 		var nextActions cli.NextActions
@@ -183,7 +183,7 @@ If you would like to return the cluster to its original state, please run "gpupg
 		return cli.NewNextActions(s.Err(), msg)
 	}
 
-	fmt.Println(completedText)
+	fmt.Print(completedText)
 	return nil
 }
 
--- a/cli/commanders/step.go
+++ b/cli/commanders/step.go
@@ -15,6 +15,7 @@ import (
 	"github.com/greenplum-db/gpupgrade/cli"
 	"github.com/greenplum-db/gpupgrade/idl"
 	"github.com/greenplum-db/gpupgrade/step"
+	"github.com/greenplum-db/gpupgrade/utils"
 	"github.com/greenplum-db/gpupgrade/utils/errorlist"
 	"github.com/greenplum-db/gpupgrade/utils/stopwatch"
 )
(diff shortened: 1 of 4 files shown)


kalensk:
I made some changes. Thus, this PR does the following (I know its a bit of a hodge podge...)

- Instead of bubbling up the error when the state directory can't be written to add a user friendly note instead:
`Note: If commands were issued in order, ensure gpupgrade can write to /Users/gpadmin/.gpupgrade`
- Add the run initialize help text in the next actions.
- Add the correct spacing to `stepErr`.
- Remove an extra trailing newline to the completed text.
- Rename CLIStep to just Step for accuracy and clarity

cli/commanders/step.go
```diff
@@ -32,10 +33,12 @@ type CLIStep struct {
 	err         error
 }
 
-func NewStep(step idl.Step, streams *step.BufferedStreams, verbose bool) (*CLIStep, error) {
+func NewStep(step idl.Step, streams *step.BufferedStreams, verbose bool) (*Step, error) {
 	store, err := NewStepStore()
 	if err != nil {
-		return &CLIStep{}, cli.NewNextActions(err, StepErr.Error())
+		context := fmt.Sprintf("Note: If commands were issued in order, ensure gpupgrade can write to %s", utils.GetStateDir())
```
dkrieger-pvtl, line 39: A couple of comments:

- this assumes that the underlying function `NewStepSore()` creating the error is using `utils.GetStatedir()`; it very well might not be.  That is, we are coupling an explicit assumption about the internals of this function.
- there are other ways that NewStepStore() can fail now(like the permissions on `steps.json` magically changing, for instance) and perhaps other ways in the future.

Since we really want to print a simplified error in the cases where `NewStepStore()` fails, we could address most of these comments by passing back an explicit error type `NewStepStoreError` that contains the path and underlying error.  We could then pull the path out here.

The lack of dealing with other errors, like not being able to open the file or other errors added into `NewStepStore` in the future, appears to be something we are willing to accept as a tradeoff for a simplified user error message.

That said, please log the actual error.

dkrieger-pvtl (commented):
This solves the immediate issue but I think we can make the implementation a little cleaner.  See comment.

cli/commanders/step.go
kalensk: Added a gplog.Error. I still think the current help text is plenty sufficient and logging an error every time a step is run before initialize is not needed.

dkrieger-pvtl: Thanks for adding the error; there are reasons other than being unable to write to the state dir that can cause `NewStepStore` to return an error, so logging will allow us to debug such cases.

dkrieger-pvtl (approved):
LGTM

PR #487, opened 2020-10-28
rename logfile to gpupgrade

kalensk:
rename logfile to gpupgrade

[Pipeline](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:renameLogfile)

### DIFF
--- a/cmd/gpupgrade/main.go
+++ b/cmd/gpupgrade/main.go
@@ -27,7 +27,7 @@ func main() {
 		fmt.Printf("\n%+v\n", err)
 		os.Exit(1)
 	}
-	gplog.InitializeLogging("gpupgrade_cli", logdir)
+	gplog.InitializeLogging("gpupgrade", logdir)
 
 	root := commands.BuildRootCommand()
 	// Silence usage since Cobra prints usage for all errors rather than just


dkrieger-pvtl (commented):
What's the motivation for this?  Internally, we speak of our different processes as the cli, hub and agent.  The log file changed here is for the cli, so that gives us uniform logfile naming: `gpupgrade_{cli,hub,agent}.log`.  Though, of course, it's for different runs of the cli.  For users, we speak of initialize, execute, finalize and revert, and those have log files as well.

So I think the current naming makes sense...I would expect a log called `gpupgrade.log` to somehow contain general logging for the entire upgrade process(data migration, etc), not just the cli.

kalensk:
> What's the motivation for this? Internally, we speak of our different processes as the cli, hub and agent. The log file changed here is for the cli, so that gives us uniform logfile naming: gpupgrade_{cli,hub,agent}.log. Though, of course, it's for different runs of the cli. For users, we have per-step log files for initialize, execute, finalize and revert, and those are called {initialize, execute, finalize, revert}_YYYYMMDD.log. That lends credence to leaving this file called gpupgrade_cli.log.

One might argue that users don't need to know about the CLI. And that the "CLI" should simply be called "gupgrade". This seems much more user-friendly to me.

dkrieger-pvtl:
> One might argue that users don't need to know about the CLI. And that the "CLI" should simply be called "gupgrade". This seems much more user-friendly to me.

This PR expresses a preference; let's discuss as a team.

PR #490, opened 2020-10-28
add confirmation prompt

kalensk:
- Add confirmation text before each step
- When initialize is called with --automatic or -a then no confirmation prompt is presented and gpupgrade proceeds automatically.
- The --automatic flag is not accepted for any other command except initialize.
- Add automatic to other commands, but as a hidden flag. This is used for end to end tests to not hang for user input.
- Fail fast by adding the confirmation prompt "_after_" validating the current step.

[Pipeline](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:confirmationText)


dkrieger-pvtl (commented):
This PR ties the implementation of `automatic` into `NewStep`.  The logic there currently does `ValidateStep` to handle the relatively unlikely out-of-order command requests from a user.  This makes sense, as the creation of the step does not make sense if it is out of order. 

User abort in `NewStep` seems misplaced; it is purely a business rule driven within the UI.  I would prefer that logic all be in the commands.go in each command; if the user chooses to abort, we simply never create the step in the first place.  This does mean that a user could be presented with a Prompt, say yes, only to be told that they cannot run that step.  But that case seems unlikely.

Once we decide where to put the prompt based on UI flow, we can decide how to implement it cleanly.

kalensk:
> Once we decide where to put the prompt based on UI flow, we can decide how to implement it cleanly.

For reference, PM and designer would like to fail fast if the steps are run out of order. That is, I believe only show the confirmation text and prompt after the step has been validate.

dkrieger-pvtl (commented):
There are three cleanups I think can be done here:
-- return UserAbort error and not Skip error upon Prompt "no"
-- place confirmation text in its own file for readability
-- clean up arguments to NewStep.  NewStep is an internal interface, so a [functional options](https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis) approach is too heavy-weight.  A structure with named parameters for callers serves nicely here.


See first three commits in:

https://github.com/dkrieger-pvtl/gpupgrade/tree/dkrieger_kalensk_confirmationText

kalensk:
> There are three cleanups I think can be done here:
-- return UserAbort error and not Skip error upon Prompt "no"

Added. Take a look!

> -- place confirmation text in its own file for readability

Done.

> -- clean up arguments to NewStep. NewStep is an internal interface, so a functional options approach is too heavy-weight. A structure with named parameters for callers serves nicely here.

I agree that that number of parameters to `NewStep` is a lot. For now, I separated the parameters by placing them on separate lines. Depending on how the code changes over time adding a functional options approach would probably make sense. I think its a tad too earlier for now.

dkrieger-pvtl (commented):
Overall, LGTM...see the few comments and consider adding a positive test for the confirmation text.

dkrieger-pvtl (approved):
LGTM

PR #491, opened 2020-10-29
Add default next actions

kalensk:
When there is an error add a default next action based on the step being run.

[Pipeline](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:defaultNextActions)


dkrieger-pvtl (approved):
LGTM.

PR #492, opened 2020-10-29
update config file

kalensk:
- To make it more obvious to customers what the default values are make them live and don't comment them out.
- Eliminate the distinction between optional and required parameters

[Pipeline](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:updateConfig)

### DIFF
--- a/gpupgrade_config
+++ b/gpupgrade_config
@@ -1,52 +1,38 @@
 # ----------------------------
 # gpupgrade configuration file
 # ----------------------------
-#
-# Parameters consist of names and values separated by an equal sign.
-# Comments begin with an "#" and can begin anywhere on the line.
-# Parameters commented out below are default values.
-
 
 # The installation path for the source Greenplum Database.
 # For example, /usr/local/<source-greenplum-version>.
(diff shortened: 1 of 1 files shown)


dkrieger-pvtl (approved):
LGTM for just the `gpupgrade_config` file; it matches exactly the text in the story.

However, there are several points in the story description that are not covered in that story's expected text:
1). listing of default values as a comment in `gpupgrade_config`
2). enforcement of an explicit inclusion of `source_master_port` in `gpupgrade_config`.

There is a comment in that story that might negate this, so if further discussion shows this PR is the only fix required for that story, fine.  If not, modifications will be needed.

kalensk:
Had an offline discussion with PM and designer.

> 1). listing of default values as a comment in gpupgrade_config

We are omitting these to avoid customer confusion and a few other reasons.

> 2). enforcement of an explicit inclusion of source_master_port in gpupgrade_config.

Removed the default value of source_master_port to draw attention to customers which usually are not running on the default.

PR #497, opened 2020-11-04
support --hba-hostnames

kalensk:
This adds support to pass --hba-hostnames to the various utilities.

[Pipeline](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:useHbaHostnames)


hub/upgrade_mirrors.go
```diff
@@ -29,10 +29,16 @@ func writeGpAddmirrorsConfig(mirrors []greenplum.SegConfig, out io.Writer) error
...
 }
 
-func runAddMirrors(r greenplum.Runner, filepath string) error {
+func runAddMirrors(r greenplum.Runner, filepath string, useHbaHostnames bool) error {
+	hbaHostnames := ""
+	if useHbaHostnames {
+		hbaHostnames = "--hba-hostnames"
+	}
+
 	return r.Run("gpaddmirrors",
```
bhuvnesh2703, line 38: why don't we pass a slice of strings? change the parameter for run from ...string to []string (and the other related interfaces/methods) and instead of filtering out the "" we can pass the right thing.. we are adding additional methods which could be avoided?
the call here will look like

```
args := []string{"-a", "-i", filepath}
if useHbaHostnames {
               args = append(args, "--hba-hostnames")
        }
return r.Run("gpaddmirrors", args)

same thing at other callers?

Comment (kalensk) [reply hub/upgrade_mirrors.go:38]: See how easy it is to read all the parameters when they are grouped together like this:

```go
r.Run("gpaddmirrors",
   "-a",
   "-i", filepath,
   hbaHostnames,
```

The cognitive ease is pretty nice. I would like to keep that if possible. The approach you suggested breaks that up. Something to consider.

gpupgrade_config
```diff
@@ -21,6 +21,10 @@ mode = copy
 # Recommended values are 0.6 [60%] for copy mode, and 0.2 [20%] for link mode.
 disk_free_ratio = 0.6
 
+# Whether to use host names or IP addresses in gpinitsystem and other utilities.
+# Choose TRUE to use host names, FALSE to use IP addresses.
```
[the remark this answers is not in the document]
kalensk: Text is from PM. Will have to check-in with them.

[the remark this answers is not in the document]
kalensk: ^^ @Nirali13 some additional text to consider to the `use_hba_hostnames` helper text in the config file.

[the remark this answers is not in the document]
Nirali13: @kalensk @bhuvnesh2703 I am okay with the proposed suggestion.

hub/upgrade_mirrors.go
bhuvnesh2703: args := []string{"-a", "-i", filepath} this as well seems pretty simple to understand what are the arguments are and you can write them in multiple lines as you wrote for Run, normalizing functions which you added adds scope of bugs.. 
```
args := []string{
               "-a", 
               "-i", 
               filepath
             }
if useHbaHostnames {
               args = append(args, "--hba-hostnames")
        }

return r.Run("gpaddmirrors", args)
```
I leave it at your choice, but i don't see the point of adding the normalizing function.

jmcatamney: I don't have strong opinions on the issue; this normalization function is fairly straightforward and being able to handle arguments further ahead of a call is nice, but I wouldn't really mind if we stuck with the more idiomatic Go way of constructing a command string.  Assuming David doesn't weigh in with objections, I'm fine to approve this as-is.

dkrieger-pvtl: I agree...this looks clean to me(note there is no need to change the interface to `Run` using a variadic arg):
```go
	args := []string{"-a",
		"-i", filepath}
	if useHbaHostnames {
		args = append(args, "--hba-hostnames")
	}

	return r.Run("gpaddmirrors", args...)
```

The introduction of `normalize` does not seem worth the extra complexity in the implementation.

gpupgrade_config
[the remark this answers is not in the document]
kalensk: Thanks, updated the text.

hub/upgrade_mirrors.go
kalensk: Based on consensus. Removed normalized function, and updated callers to how @bhuvnesh2703  and @dkrieger-pvtl showed.

kalensk:
Rebased and applied feedback. Letting pipeline run through to verify changes. Take one last look and approve if it looks good. Thanks for the comments and feedback.

dkrieger-pvtl:
@kalensk what about an end-to-end test to show that we actually place hostnames in the pg_hba.conf file?  We can use the existing `finalize.bats` test, say copy mode with IP addresses and link mode with hostnames.

kalensk:
> what about an end-to-end test to show that we actually place hostnames in the pg_hba.conf file? We can use the existing finalize.bats test, say copy mode with IP addresses and link mode with hostnames.

I think we already discussed that as a team, but if not let's do so.

gpupgrade_config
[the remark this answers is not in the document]
dkrieger-pvtl: Users do not care per se what we do `during execution of gpinitsystem and other utilities`.  They are running `gpupgrade`, and need not and should not know what we are doing under the hood.  Perhaps we want to use the [gpinitsystem phrasing](https://gpdb.docs.pivotal.io/6-0/utility_guide/ref/gpinitsystem.html): `Whether to use host names or IP addresses in the pg_hba.conf file when updating the files with addresses that can connect to Greenplum Database`.

hub/restore_source_cluster_test.go
```diff
@@ -198,13 +198,13 @@ func TestRsyncMasterAndPrimaries(t *testing.T) {
...
 
-			expected := []string{"-c", fmt.Sprintf("source /usr/local/greenplum-db/greenplum_path.sh && MASTER_DATA_DIRECTORY=%s PGPORT=%d /usr/local/greenplum-db/bin/gprecoverseg -a", cluster.MasterDataDir(), cluster.MasterPort())}
+			expected := []string{"-c", fmt.Sprintf("source /usr/local/greenplum-db/greenplum_path.sh && MASTER_DATA_DIRECTORY=%s PGPORT=%d /usr/local/greenplum-db/bin/gprecoverseg -a ", cluster.MasterDataDir(), cluster.MasterPort())}
 			if !reflect.DeepEqual(args, expected) {
 				t.Errorf("got %q want %q", args, expected)
 			}
 		})
 
-		err := hub.Recoverseg(&testutils.DevNullWithClose{}, cluster)
+		err := hub.Recoverseg(&testutils.DevNullWithClose{}, cluster, false)
```
dkrieger-pvtl, line 207: Change this to `true` and adjust `expected` to obtain more code coverage.

dkrieger-pvtl (commented):
Overall, looks good with a few minor comments.  Let's discuss, though, the need to add an end-to-end test: there are a lot of intermediate steps that are done to utilize `hba_hostnames` as true.

hub/restore_source_cluster_test.go
kalensk: Updated.

test/finalize.bats
```diff
@@ -164,3 +170,41 @@ validate_data_directories() {
...
+
+        if [[ -n "$HBA_HOSTNAMES" ]] && [[ "$found" = false ]]; then
+            log "expected ${datadir}/pg_hba.conf to contain ${expected_host}. Found" "${actual_hosts[@]}"
+        fi
+
+        if [[ -z "$HBA_HOSTNAMES" ]] && [ "$found" = true ]; then
+            log "expected ${datadir}/pg_hba.conf to 'not' contain ${expected_host}. Found" "${actual_hosts[@]}"
+        fi
+    done
+}
```
dkrieger-pvtl, line 210: I haven't tested, but consider something like this:
```bash
validate_pg_hba_conf() {
    local HBA_HOSTNAMES=$1
    local expected_hosts actual_hosts
    local matched=()
    local unmatched=()

    expected_hosts=($(all_hosts))
    IFS=$'\n' expected_hosts=($(sort <<<"${expected_hosts[*]}"))
    IFS=$'\n' expected_hosts=($(uniq <<<"${expected_hosts[*]}"))
    unset IFS

    for datadir in "${datadirs[@]}"; do
        actual_hosts=($(grep  -v '^#' "${datadir}/pg_hba.conf" | grep -v '^$' | grep -v '^local' | awk '{ print $4 }' | sort | uniq))
        for expected_host in "${expected_hosts[@]}"; do
            for actual_host in "${actual_hosts[@]}"; do
                if [ "$expected_host" == "$actual_host" ]; then
                    matched+=( "$actual_host" )
                    break
                fi
            done
            unmatched+=( "$expected_host" )
        done


        if [[ -n "$HBA_HOSTNAMES" ]] && [[ ${#unmatched[@]} ]]; then
            fail "expected ${datadir}/pg_hba.conf to contain all " "${expected_hosts[@]}" "Found" "${actual_hosts[@]}"
        fi

        if [[ -z "$HBA_HOSTNAMES" ]] && [[ ${#matched[@]} ]]; then
            fail "expected ${datadir}/pg_hba.conf to 'not' contain any " "${expected_hosts[@]}" "Found" "${actual_hosts[@]}"
        fi
    done
}
```

dkrieger-pvtl: Code was modified to something like this.

dkrieger-pvtl:
Please update the `initializeConfirmationText` with the hbaHostnames setting during the confirmation dialog...so we get:
```
gpupgrade initialize will use these values from /Users/dkrieger/go/src/github.com/greenplum-db/gpupgrade/gpupgrade_config
source_gphome:      asdfasdf
target_gphome:      asdfasdfsf
mode:               copy
disk_free_ratio:    0.6
use_hba_hostnames:  true
source_master_port: 12345
temp_port_range:    50432-65535
hub_port:           7527
agent_port:         6416
```

gpupgrade_config
```diff
@@ -21,6 +21,11 @@ mode = copy
 # Recommended values are 0.6 [60%] for copy mode, and 0.2 [20%] for link mode.
 disk_free_ratio = 0.6
 
+# Whether to populate pg_hba.conf with hostnames or IP addresses during
+# execution of gpinitsystem and other utilities.
+# Choose TRUE to use host names, FALSE to use IP addresses.
```
dkrieger-pvtl, line 26: nit: I am not sure we should CAPITALIZE TRUE and FALSE here...maybe `"true"` and `"false"` like what we do for the  comments for `mode`?

dkrieger-pvtl (approved):
LGTM, but please add the `use_hba_hostnames` setting to the initialize confirmation dialog.

PR #498, opened 2020-11-04
update help text

kalensk:
update help text

[Pipeline](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:updateHelp)

### DIFF
--- a/cli/commands/help_text.go
+++ b/cli/commands/help_text.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/greenplum-db/gpupgrade/cli/commanders"
 	"github.com/greenplum-db/gpupgrade/idl"
+	"github.com/greenplum-db/gpupgrade/utils"
 )
 
 var (
@@ -69,10 +70,10 @@ func init() {
 
 const initializeHelp = `
 Runs pre-upgrade checks and prepares the cluster for upgrade.
+This command should be run only during a downtime window.
 
 Initialize will carry out the following steps:
 %s
-
 During or after gpupgrade initialize, you may revert the cluster to its
 original state by running gpupgrade revert.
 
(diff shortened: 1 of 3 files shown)


cli/commands/help_text.go
```diff
@@ -171,22 +186,26 @@ Optional Commands:
...
 	var formattedList string
 	for _, substep := range commandList {
 		formattedList += fmt.Sprintf(" - %s\n", commanders.SubstepDescriptions[substep].HelpText)
 	}
-	return fmt.Sprintf(baseString, formattedList)
 
+	logdir, err := utils.GetLogDir()
+	if err != nil {
+		panic(fmt.Sprintf("failed to get log directory: %v", err))
+	}
```
kalensk, line 206: **_Note_**: I was investigating whether an error here would print a stack trace or not, and whatever I would do it would not even let me compile. That is, it was panic'ing during compilation when trying to force an error during investigation.. 

It appears we only call `log.WritePanics()` in the hub and agent and "_not_" for the CLI.

I think an error is pretty unlikely to occur.

cli/commands/help_text.go
```diff
@@ -104,31 +106,35 @@ Usage: gpupgrade execute
...
 
+NOTE: After running finalize, you must execute data migration scripts. 
+Refer to documentation for instructions.
+
+gpupgrade log files can be found on all hosts in %s
 `
 const revertHelp = `
 Returns the cluster to its original state.
 This command cannot be run after gpupgrade finalize has begun.
+This Command should be run only during a downtime window.
```
jmcatamney, line 137: `s/Command/command/` for consistency.

jmcatamney (approved):

PR #500, opened 2020-11-04
Makefile: add enterprise and oss tarball targets 

kalensk:
The main difference between enterprise and open source tarball is that the version command outputs the associated release type, and the enterprise tarball contains the open_source_licenses.txt file. Committing the license file is the recommended approach as that easily allows us to embed it in the RPM. The only slight gotcha is that we will need to update/commit a new license file before each release.

To allow the RELEASE type to be set the version string variables was enclosed inside the build target.

[Pipeline](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:tarball)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,11 +19,8 @@ VERSION_LD_STR := -X 'github.com/greenplum-db/$(MODULE_NAME)/cli/commands.Versio
 VERSION_LD_STR += -X 'github.com/greenplum-db/$(MODULE_NAME)/cli/commands.Commit=$(COMMIT)'
 VERSION_LD_STR += -X 'github.com/greenplum-db/$(MODULE_NAME)/cli/commands.Release=$(RELEASE)'
 
-BRANCH := $(shell git rev-parse --abbrev-ref HEAD)
 LINUX_ENV := env GOOS=linux GOARCH=amd64
 MAC_ENV := env GOOS=darwin GOARCH=amd64
-LINUX_EXTENSION := .linux.$(BRANCH)
-MAC_EXTENSION := .darwin.$(BRANCH)
 
 # depend-dev will install the necessary Go dependencies for running `go
 # generate`. (This recipe does not have to be run in order to build the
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -99,6 +99,29 @@ build_linux build_mac: build
...
+	rm -r package/data_migration_scripts/test
+	cp cli/bash/bash-completion.sh package
+	cp gpupgrade_config package
+	if [ "$(RELEASE)" = "Enterprise" ]; then \
+		cp open_source_licenses.txt package; \
+	fi
+	# create tarball
+	( cd package; tar czf ../gpupgrade-$(VERSION).tar.gz . )
+	shasum -a 256 gpupgrade-$(VERSION).tar.gz > CHECKSUM
+
```
kalensk, line 124: Should we remove the package directory at the end instead of the beginning and not do a `-f` when removing for safety?

bhuvnesh2703: I think you have done .gitignore and remove it everytime we make a new one, so it's fine not to delete it for checking if someone wants..

bhuvnesh2703 (approved):
Tried creating both the tarballs.
Enterprise version
```
bchaudhary@bchaudhary-a01:/tmp/enterprise$ ./gpupgrade version
Version: 0.4.0 Commit: 42ab517 Release: Enterprise
```
Open Source tarball version
```
bchaudhary@bchaudhary-a01:/tmp/opensource$ ./gpupgrade  version
Version: 0.4.0 Commit: 42ab517 Release: Open Source
```

dkrieger-pvtl (approved):
Consider updating the instructions for tagging.

Makefile
kalensk: Updated. Added `[ ! -d package ] && mkdir package` in the beginning to check if package doesn't exist and bail out if so, and `rm -r package` at the end to cleanup the temp files while leaving the generated tarball.

Makefile
```diff
@@ -84,12 +69,27 @@ sshd_build:
...
 .PHONY: build build_linux build_mac
 
 build:
-	$(BUILD_ENV) go build -o gpupgrade$(EXTENSION) $(BUILD_FLAGS) github.com/greenplum-db/gpupgrade/cmd/gpupgrade
+# TAGGING
+#   YOUR_BRANCH> make all of the changes you want for your tag
+#   follow standard procedures for your PR; commit them(PR is completed)
+#   note git hash for that version(might have been rebased, etc); call it GIT_HASH
+#   YOUR_BRANCH> git tag -a TAGNAME -m "version 0.1.1: add version" GIT_HASH
+#   YOUR_BRANCH> git push origin TAGNAME
```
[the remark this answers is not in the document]
kalensk: Replaced the comment with: `# For tagging a release see the "Upgrade Release Checklist" document.`

And updated the checklist document with:

> Before tagging ensure the HEAD of the branch you are on is the git sha you want to tag, and that the commit is on origin.

PR #501, opened 2020-11-05
update golang to 1.15

dkrieger-pvtl:
Since golang is backward source compatible, we update to the current
latest version of golang for our gpupgrade compilation as well as
testing.

A test that makes sure the bats tests are run with this version of
golang is also included.  We were previously running a golang version
of 1.13.X, even though we were supposed to be on 1.14, so a test to
catch this in the future is in order.

We created a new docker image for this PR, pivotaldata/gpdb6-centos7-test-golang,
which is a GPDB run environment that contains golang 1.15 pre-installed.

After this PR is merged, the prod pipeline will be reflown.

### DIFF
--- a/test/goversion.bats
+++ b/test/goversion.bats
@@ -1,0 +1,13 @@
+#! /usr/bin/env bats
+#
+# Copyright (c) 2017-2020 VMware, Inc. or its affiliates
+# SPDX-License-Identifier: Apache-2.0
+
+load helpers
+
+@test "gpupupgrade is compiled with golang version 1.15.X" {
+    local EXPECTED="gpupgrade: go1.15."
+    run go version gpupgrade
+    [ "$status" -eq 0 ]
+    [[ "$output" =~ $EXPECTED ]] || fail "expected: $EXPECTED got: $output"
+}
--- a/README.md
+++ b/README.md
@@ -117,7 +117,7 @@ pg_upgrade supports two upgrade modes: link and copy.
 
 ### Prerequisites
 
-- Golang. We currently develop against latest stable Golang, which was v1.14 as of April 2020.
+- Golang. We currently develop against latest stable Golang, which was v1.15 as of October 2020.
 - protoc. This is the compiler for the [gRPC protobuf](https://grpc.io/) 
 system which can be installed on macOS with `brew install protobuf`.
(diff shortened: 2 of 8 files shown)


test/goversion.bats
```diff
@@ -0,0 +1,13 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+load helpers
+
+@test "gpupupgrade is compiled with golang version 1.15.X" {
+    local EXPECTED="gpupgrade: go1.15."
+    run go version gpupgrade
+    [ "$status" -eq 0 ]
+    [[ "$output" =~ $EXPECTED ]] || fail "expected: $EXPECTED got: $output"
+}
```
kalensk, line 13: I think this test adds too much rigidity and strictness. That is, there is no problem with compiling with a lower version of golang. It will error if we are using a feature that is unsupported.

dkrieger-pvtl: What is the use case for wanting to support a pre-1.15 golang version?  

Generally, later releases are supposed to be "better"; see the [golang1.15 release notes](https://golang.org/doc/go1.15).  While there is no logical issue with compiling with a previous version of golang, I believe it is bad practice.  New versions are supposed to be "better": performance improvements, bug fixes, etc.   Each new version introduces [behavioral changes[(https://golang.org/doc/go1.15#os) that can change the way tests run.  Locking everyone to the same minor version also eliminates variance between environments, which can make debugging tests easier.

Internally, we have complete control; everyone on the team can update to golang 1.15.  We deliver gpupgrade in its entirety, that is, we do not need to support rolling upgrades.  (So we do not have the potential case of one deployed component built with 1.14 and another built with 1.15.).  Externally, we can expect an open source developer who wants to contribute to gpugrade to use golang 1.15.

In practice, this test really only matters in the CI; anyone running on their local machine likely won't run this test.  If they do, they can see the cause and decide to update to golang 1.15.  That said, everyone should update to golang1.15.

kalensk: > What is the use case for wanting to support a pre-1.15 golang version? 

It's not so much supporting as not being so strict in our requirements (until we have a concrete need/use-case).

kalensk (approved):

test/goversion.bats
dkrieger-pvtl: We decided as a team to keep this test.

PR #502, opened 2020-11-05
Fix name scripts

bhuvnesh2703:
This PR contains couple of commits which does the following
1. Only pulls relkind = 'r' from pg_class else alter table statement was created for views which is incorrect.
2. Excludes relations which has views on it, as alter will fail on those tables
3. Exclude sql statements for altering distribution keys as they also fail when executed.
4. Exclude sql statements for altering partition keys as they will also fail when executed

### DIFF
--- a/data_migration_scripts/start/gen_alter_name_type_columns.header
+++ b/data_migration_scripts/start/gen_alter_name_type_columns.header
@@ -1,8 +1,16 @@
 -- The below SQL alters, where possible, the name data type to varchar(63)
--- in columns other than the first.  For a partition table, SQL is only generated
--- for root partitions as that cascades to the child partitions. Where not possible,
--- a message is logged, and such tables must be manually modified to remove the name
--- data type prior to running gpugprade.
+-- in columns other than the first.
+-- For a partition table, SQL is only generated for root partitions as that
+-- cascades to the child partitions.
+-- SQL statements are not generated for distribution columns with name data type,
+-- one can alter the datatype using the below steps:
+--   1. ALTER TABLE <tablename> SET DISTRIBUTED RANDOMLY;
+--   2. ALTER TABLE <tablename> ALTER COLUMN <columnname> TYPE VARCHAR(63);
+--   3. ALTER TABLE <tablename> SET DISTRIBUTED BY (columnname);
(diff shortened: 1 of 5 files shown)


data_migration_scripts/start/gen_alter_name_type_columns.header
```diff
@@ -1,8 +1,16 @@
...
+-- cascades to the child partitions.
+-- SQL statements are not generated for distribution columns with name data type,
+-- one can alter the datatype using the below steps:
+--   1. ALTER TABLE <tablename> SET DISTRIBUTED RANDOMLY;
+--   2. ALTER TABLE <tablename> ALTER COLUMN <columnname> TYPE VARCHAR(63);
+--   3. ALTER TABLE <tablename> SET DISTRIBUTED BY (columnname);
+-- SQL statements are not generated for partitioning columns with name data type, one
+-- has to manually recreate the table with datatype other than name such as varchar(63).
+-- Where not possible, a message is logged, and such tables must be manually
+modified to remove the name data type prior to running gpugprade.
```
jmcatamney, line 13: The initial `--` is missing on this line.

jmcatamney (approved):
Other than the typo noted in my comment, this looks good to me.

dkrieger-pvtl:
Two of the three main sql scripts are identical, and the other only differs in one line(other than header info).  I am not sure if there's a simple way to have mostly common code for these but it might be worth some thought.

```
 diff ./revert/gen_revert_text_to_name_type_columns.sql ./start/gen_alter_name_type_columns.sql
(header differences)
21c46
<        ' TYPE NAME; EXCEPTION WHEN feature_not_supported THEN PERFORM pg_temp.notsupported(''' || c.oid::pg_catalog.regclass || '''); END $$;'
---
>        ' TYPE VARCHAR(63); EXCEPTION WHEN feature_not_supported THEN PERFORM pg_temp.notsupported(''' || c.oid::pg_catalog.regclass || '''); END $$;'



 diff ./revert/gen_revert_text_to_name_type_columns.sql ./complete/gen_complete_text_to_name_type_columns.sql -s
Files ./revert/gen_revert_text_to_name_type_columns.sql and ./complete/gen_complete_text_to_name_type_columns.sql are identical
```

PR #503, opened 2020-11-09
Finalize cleanup

dkrieger-pvtl:
finalize: cleanup resources after successful run

Like revert, we want finalize to cleanup as many resources as it can
when it finishes.  So, we now:
- archive the log directories
- remove the state directories
- stop the hub and agents

The finalize.bats test has also been updated for a local demo cluster; it has not been updated to run on multi-node.

In addition, the finalize output on successful completion has been updated.  It now reads:

```
Finalize completed successfully.

The target cluster is now ready to use, running Greenplum 6.0.0-beta.1 build dev.
PGPORT: 15432
MASTER_DATA_DIRECTORY: DATADIR/qddir/demoDataDir-1

The source cluster is not running. You may delete the source cluster to recover space.
All source cluster data directores end in "Jda0rddf6xs.old".  For instance, the source master is:
DATADIR/qddir/demoDataDir.Jda0rddf6xs.-1.old

The gpupgrade logs can be found on the master and segment hosts in
HOME/gpAdminLogs/gpupgrade-Jda0rddf6xs-2020-11-09T07:59
```


cli/commands/finalize.go
```diff
@@ -62,18 +63,39 @@ func finalize() *cobra.Command {
...
+
 			return st.Complete(fmt.Sprintf(`
 Finalize completed successfully.
 
 The target cluster is now ready to use, running Greenplum %s.
 PGPORT: %d
 MASTER_DATA_DIRECTORY: %s
 
+The source cluster is not running. You may delete the source cluster to recover space from all hosts. 
+All source cluster data directores end in "%s".
```
kalensk, line 86: We should consider rephrasing to mention the content ID:
`All source cluster data directores end in NkUY7MEe_Kk<contentID>.old`

If so, we would just have finalizeResponse simply pass the upgradeID only with out the `.old` ending.

Since for example after finalize the suffix is actually:
```
./dbfast1/demoDataDir0
./dbfast1/demoDataDir.NkUY7MEe_Kk.0.old
./dbfast_mirror1/demoDataDir0
./dbfast_mirror1/demoDataDir.NkUY7MEe_Kk.0.old

./dbfast2/demoDataDir1
./dbfast2/demoDataDir.NkUY7MEe_Kk.1.old
./dbfast_mirror2/demoDataDir1
./dbfast_mirror2/demoDataDir.NkUY7MEe_Kk.1.old

./dbfast3/demoDataDir2
./dbfast3/demoDataDir.NkUY7MEe_Kk.2.old
./dbfast_mirror3/demoDataDir2
./dbfast_mirror3/demoDataDir.NkUY7MEe_Kk.2.old

./qddir/demoDataDir.NkUY7MEe_Kk.-1.old
./qddir/demoDataDir-1
./standby.NkUY7MEe_Kk.old
./standby/
```

kalensk: Typo in `directories` - missing the last "i".

upgrade/directories.go
```diff
@@ -32,6 +32,10 @@ func GetConfigFile() string {
 	return filepath.Join(utils.GetStateDir(), ConfigFileName)
 }
 
+func GetArchivedSourcePath(path string) string {
+	return path + OldSuffix
+}
+
```
kalensk, line 38: A few thoughts:

1) Depending on the above comment in terms of stating the contentID in the competition text and if so only needing to pass the upgradeID, then this function isn't necessary for only one caller.

2) _If_ we choose to keep it, can we syncup or brainstorm some ideas on the naming and input? That is, this only "really" works when the input is `s.Config.TargetInitializeConfig` and how do we communicate that to the callers who might want to pass `s.Target` or `s.Source`? That is, is there a way to make it more clear what the input should be.

kalensk (commented):
Overall looks good with a few minor comments/thoughts.

cli/commands/finalize.go
```diff
@@ -62,18 +63,39 @@ func finalize() *cobra.Command {
...
+			})
+
+			st.RunCLISubstep(idl.Substep_DELETE_MASTER_STATEDIR, func(streams step.OutStreams) error {
+				// Removing the state directory removes the step status file.
+				// Disable the store so the step framework does not try to write
+				// to a non-existent status file.
+				st.DisableStore()
+				return upgrade.DeleteDirectories([]string{utils.GetStateDir()}, upgrade.StateDirectoryFiles, streams)
+			})
+
```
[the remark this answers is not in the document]
dkrieger-pvtl: I updated `FinalizeHelp`...good catch.  I think this is a sign that eventually we want to somehow automatically generate this list.  

The `RevertHelp` is missing quite a bit and I prefer to place that in a separate PR.

cli/commands/finalize.go
dkrieger-pvtl: Good catch...I've changed the design.  The master source directory is returned in full by `finalize` but the UI constructs the generic contentID string.  I could not think of an elegant way to do it, and it is unlikely that we'll modify the format of the archive directory paths.

upgrade/directories.go
dkrieger-pvtl: This function is now removed in favor of directly inlining the `upgrade.OldSuffix` where needed.  We are unlikely to modify the archive directory path.

cli/commands/finalize.go
```diff
@@ -62,18 +63,40 @@ func finalize() *cobra.Command {
...
+The gpupgrade logs can be found on the master and segment hosts in
+%s
+
 NEXT ACTIONS
 ------------
 Run the “complete” data migration scripts, and recreate any additional tables,
 indexes, and roles that were dropped or altered to resolve migration issues.`,
-				response.GetTargetVersion(), response.GetTarget().GetPort(), response.GetTarget().GetMasterDataDirectory()))
+				response.GetTargetVersion(), response.GetTarget().GetPort(), response.GetTarget().GetMasterDataDirectory(),
+				fmt.Sprintf("%s.<contentID>%s", response.GetUpgradeID(), upgrade.OldSuffix),
```
kalensk, line 97: I still think this is missing the last dot. Can we add that? That is, `fmt.Sprintf("%s.<contentID>.%s"` since the directories are:

```
./dbfast1/demoDataDir.NkUY7MEe_Kk.0.old
./dbfast_mirror1/demoDataDir.NkUY7MEe_Kk.0.old

./dbfast2/demoDataDir.NkUY7MEe_Kk.1.old
./dbfast_mirror2/demoDataDir.NkUY7MEe_Kk.1.old

./qddir/demoDataDir.NkUY7MEe_Kk.-1.old
./standby.NkUY7MEe_Kk.old
```

cli/commands/finalize.go
kalensk: > the UI constructs the generic contentID string. 

To me this is the "correct" way to go.

kalensk (approved):
Looks good with one comment about the last dot in `fmt.Sprintf("%s.<contentID>.%s"`

cli/commands/finalize.go
dkrieger-pvtl: Thanks, but it is correct as is.  I ran it:
```
The source cluster is not running. You may delete the source cluster to recover space from all hosts.
All source cluster data directories end in "Prxl9AYf5Xk.<contentID>.old".
MASTER_DATA_DIRECTORY: /Users/dkrieger/workspace/gpdb/worktree-5X/gpAux/gpdemo/datadirs/qddir/demoDataDir.Prxl9AYf5Xk.-1.old
```

The reason is that the `OldSuffix` is `.old`, e.g. it includes `.`.

PR #504, opened 2020-11-09
data_migration: add help

kalensk:
[Pipeline](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:dataMigrationHelp)

### DIFF
--- a/data_migration_scripts/migration_executor_sql.bash
+++ b/data_migration_scripts/migration_executor_sql.bash
@@ -3,9 +3,42 @@
 # Copyright (c) 2017-2020 VMware, Inc. or its affiliates
 # SPDX-License-Identifier: Apache-2.0
 
+function print_usage() {
+echo '
+Executes scripts addressing catalog inconsistencies between Greenplum versions.
+Before running this command, run migration_generator_sql.bash.
+This command should be run only during the downtime window.
+
+Usage: '$(basename $0)' <GPHOME> <PGPORT> <INPUT_DIR>
+     <GPHOME>    : the path to the source Greenplum installation directory
+     <PGPORT>    : the source Greenplum system port number
+     <INPUT_DIR> : the directory containing the scripts to execute. This is the
+                   <OUTPUT_DIR> you specified earlier in migration_generator_sql.bash,
+                   with a subdirectory appended as in the use cases below. The
+                   subdirectories are pre-initialize, post-finalize, and post-revert.
+
+Use cases:
+- Before "gpupgrade initialize", drop and alter objects:
+migration_executor_sql.bash /path/to/gphome 5432 /path/to/output_dir/pre-initialize
+
+- Following "gpupgrade finalize", restore and recreate objects:
+migration_executor_sql.bash /path/to/gphome 5432 /path/to/output_dir/post-finalize
+
+- Following "gpupgrade revert", restore objects:
+migration_executor_sql.bash /path/to/gphome 5432 /path/to/output_dir/post-revert
+
+Log files can be found in INPUT_DIR/data_migration.log'
+}
+
+if [ "$#" -eq 0 ] || ([ "$#" -eq 1 ] && ([ "$1" = -h ] || [ "$1" = --help ])) ; then
+    print_usage
+    exit 0
+fi
+
 if [ "$#" -ne 3 ]; then
-    echo "Illegal number of parameters"
-    echo "Usage: $(basename $0) <GPHOME> <PGPORT> <INPUT_DIR>"
+    echo ""
+    echo "Error: Incorrect number of arguments"
+    print_usage
     exit 1
 fi
 
--- a/data_migration_scripts/migration_generator_sql.bash
+++ b/data_migration_scripts/migration_generator_sql.bash
@@ -3,9 +3,35 @@
 # Copyright (c) 2017-2020 VMware, Inc. or its affiliates
 # SPDX-License-Identifier: Apache-2.0
 
+function print_usage() {
+echo '
+Identifies catalog inconsistencies between the source and target Greenplum versions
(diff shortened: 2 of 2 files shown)


dkrieger-pvtl (commented):
The current code modifies the interface...it requires a call like `./migration_executor_sql.bash -- $GPHOME $PGPORT indir`.  And it eliminates the check for the correct number of args.  

This is much simpler and I think satisfies the the help request.  Note the request also wants a rename of the directories here.

```
if [[ "$1" =~ -h|--help ]] || [[ -z "$1" ]]; then
  print_usage
  exit 0
fi

if [ "$#" -ne 3 ]; then
    echo "Illegal number of parameters"
    echo "Usage: $(basename $0) <GPHOME> <PGPORT> <INPUT_DIR>"
    exit 1
fi
```

kalensk:
> This is much simpler and I think satisfies the the help request. Note the request also wants a rename of the directories here.

Works for me. Updated.

data_migration_scripts/migration_executor_sql.bash
```diff
@@ -3,9 +3,41 @@
...
+- Following "gpupgrade revert", restore objects:
+migration_executor_sql.bash /path/to/gphome 5432 /path/to/output_dir/post-revert
+
+Log files can be found in INPUT_DIR/data_migration.log'
+}
+
+if [[ "$1" =~ -h|--help ]] || [[ -z "$1" ]]; then
+    print_usage
+    exit 0
+fi
```
dkrieger-pvtl, line 35: This technically should be written as :
```
if [ "$#" -eq 0 ] || ([ "$#" -eq 1 ] && ([ "$1" = -h ] || [ "$1" = --help ])) ; then
    print_usage
    exit 0
fi
```

This then has `script`, `script -h` and `script --help` print the help message, and _anything else_, such as `script -h $PGPORT mydir` or `script -hi` etc not print the help.  One could argue this is overly pedantic, but I think it's still readable.

At a minimum, it should be rewritten as:
```
if [ "$1" = -h ] || [ "$1" = --help ] || [ -z "$1" ]; then
    print_usage
    exit 0
fi
```
to use bash string equality and not pattern matching.  This way, `script -h` prints help but `script -hz` does not.  This still has `script -h $PGPORT DIR` printing help, but in practice a user is very unlikely to set GPHOME to `-h` or `--help`.

data_migration_scripts/migration_executor_sql.bash
```diff
@@ -3,9 +3,41 @@
 # Copyright (c) 2017-2020 VMware, Inc. or its affiliates
 # SPDX-License-Identifier: Apache-2.0
 
+function print_usage {
+echo 'Executes scripts addressing catalog inconsistencies between Greenplum versions.
```
dkrieger-pvtl, line 7: The spec calls for a newline here...
```
./migration_executor_sql.bash

Executes scripts addressing catalog inconsistencies between Greenplum versions.
Before running this command, run migration_generator_sql.bash.
This command should be run only during the downtime window.
```

instead of the current:
```
./migration_executor_sql.bash
Executes scripts addressing catalog inconsistencies between Greenplum versions.
Before running this command, run migration_generator_sql.bash.
This command should be run only during the downtime window.
```

data_migration_scripts/migration_generator_sql.bash
```diff
@@ -3,9 +3,34 @@
...
+fi
+
 if [ "$#" -ne 3 ]; then
-    echo "Illegal number of parameters"
-    echo "Usage: $(basename $0) <GPHOME> <PGPORT> <OUTPUT_DIR>"
+    echo "Error: Incorrect number of parameters"
+    echo ""
+    print_usage
     exit 1
 fi
```
dkrieger-pvtl, line 36: Identical comments apply to this file as well.

dkrieger-pvtl (commented):
See comments.  The current code essentially does what is required, but I make some suggestions to make it more precise.

data_migration_scripts/migration_executor_sql.bash
```diff
@@ -3,9 +3,41 @@
 # Copyright (c) 2017-2020 VMware, Inc. or its affiliates
 # SPDX-License-Identifier: Apache-2.0
 
+function print_usage {
```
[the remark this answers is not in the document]
kalensk: Yeah I was debating that and kind of forgot what our "convention" was. Long term it might be nice to have a cheat sheet of these guidelines perhaps.

data_migration_scripts/migration_executor_sql.bash
```diff
@@ -3,9 +3,41 @@
...
+
+Use cases:
+- Before "gpupgrade initialize", drop and alter objects:
+migration_executor_sql.bash /path/to/gphome 5432 /path/to/output_dir/pre-initialize
+
+- Following "gpupgrade finalize", restore and recreate objects:
+migration_executor_sql.bash /path/to/gphome 5432 /path/to/output_dir/post-finalize
+
+- Following "gpupgrade revert", restore objects:
+migration_executor_sql.bash /path/to/gphome 5432 /path/to/output_dir/post-revert
```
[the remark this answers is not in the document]
Nirali13: @dkrieger-pvtl I agree. This makes it more clear which values needs to be put in the places but the intend here is to show use cases of when this can be used, like examples! 
We also provide "Usage: migration_executor_sql.bash <GPHOME> <PGPORT> <INPUT_DIR>" in the help which specifies the usage. 

I would suggest to keep the test as described in story.

kalensk:
Thanks for the comments @dkrieger-pvtl. I guess BASH is not my strongest. (To me this screams re-write in golang.)

dkrieger-pvtl (approved):
LGTM.

PR #510, opened 2020-11-17
initialize: make init target cluster idempotent

kalensk:
Stop any running target cluster processes and remove any dangling data directories before initializing the target cluster.

[Pipeline](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:initTargetCluster)

### DIFF
--- a/hub/init_target_cluster.go
+++ b/hub/init_target_cluster.go
@@ -61,6 +61,30 @@ func (s *Server) writeConf(sourceDBConn *dbconn.DBConn) error {
 	return WriteInitsystemFile(gpinitsystemConfig, s.initsystemConfPath())
 }
 
+func (s *Server) RemoveTargetCluster(streams step.OutStreams) error {
+	if s.Target == nil {
+		return nil
+	}
+
+	running, err := s.Target.IsMasterRunning(streams)
+	if err != nil {
+		return err
+	}
+
+	if running {
+		if err := s.Target.Stop(streams); err != nil {
+			return xerrors.Errorf("stopping target cluster: %w", err)
+		}
+	}
+
+	err = DeleteMasterAndPrimaryDataDirectories(streams, s.agentConns, s.TargetInitializeConfig)
+	if err != nil {
+		return xerrors.Errorf("deleting target cluster data directories: %w", err)
+	}
+
+	return nil
+}
+
 // CreateTargetCluster runs gpinitsystem using the server's
 // TargetInitializeConfig, then fills in the Target cluster and persists it to
 // disk.
--- a/hub/initialize.go
+++ b/hub/initialize.go
@@ -84,7 +84,12 @@ func (s *Server) InitializeCreateCluster(in *idl.InitializeCreateClusterRequest,
 	})
 
 	st.Run(idl.Substep_INIT_TARGET_CLUSTER, func(stream step.OutStreams) error {
-		err := s.CreateTargetCluster(stream)
+		err := s.RemoveTargetCluster(stream)
+		if err != nil {
+			return err
+		}
+
+		err = s.CreateTargetCluster(stream)
 		if err != nil {
 			return err
 		}
--- a/test/initialize.bats
+++ b/test/initialize.bats
@@ -22,12 +22,15 @@ setup() {
     HELD_PORT_PID=
     AGENT_PORT=
 
+    TARGET_PGPORT=6020
+
     gpupgrade kill-services
     gpupgrade initialize \
         --automatic \
         --source-gphome="${GPHOME_SOURCE}" \
         --target-gphome="${GPHOME_TARGET}" \
-        --source-master-port="${PGPORT}"\
+        --source-master-port="${PGPORT}" \
+        --temp-port-range "$TARGET_PGPORT"-6040 \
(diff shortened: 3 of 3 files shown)


test/initialize.bats
```diff
@@ -264,6 +264,41 @@ wait_for_port_change() {
...
+    cat <<- EOF > "$GPUPGRADE_HOME/substeps.json"
+        {
+          "INITIALIZE": {
+            "GENERATE_TARGET_CONFIG": "COMPLETE",
+            "SAVING_SOURCE_CLUSTER_CONFIG": "COMPLETE",
+            "START_AGENTS": "COMPLETE"
+          }
+        }
+	EOF
+
```
dkrieger-pvtl, line 291: Here you could set-up all kinds of scenarios simulating a partial failure of `gpinitsystem`, like re-starting the target cluster in utility mode, restarting the entire cluster, deleting a few segment data directories, etc.

hub/initialize.go
```diff
@@ -84,7 +84,12 @@ func (s *Server) InitializeCreateCluster(in *idl.InitializeCreateClusterRequest,
 	})
 
 	st.Run(idl.Substep_INIT_TARGET_CLUSTER, func(stream step.OutStreams) error {
-		err := s.CreateTargetCluster(stream)
+		err := s.RemoveTargetCluster(stream)
+		if err != nil {
+			return err
+		}
+
```
dkrieger-pvtl, line 91: I like the idea: crash-only idempotence.  We always run the same logic everytime, on the first try or on a re-try after failure.

hub/init_target_cluster.go
```diff
@@ -61,6 +61,32 @@ func (s *Server) writeConf(sourceDBConn *dbconn.DBConn) error {
...
+	running, err := s.Target.IsMasterRunning(streams)
+	if err != nil {
+		return err
+	}
+
+	if running {
+		if err := s.Target.Stop(streams); err != nil {
+			return xerrors.Errorf("stopping target cluster: %w", err)
+		}
+	}
```
dkrieger-pvtl, line 78: This sequence will only work if `gpinitsystem` went through far enough to either start no segments or all segments.  If the master is running but no other segments are up yet, I think this will fail.  

We might think of other schemes, like stopping all postgres processes on all temporary ports.

hub/init_target_cluster.go
```diff
@@ -61,6 +61,32 @@ func (s *Server) writeConf(sourceDBConn *dbconn.DBConn) error {
...
+
+	if s.TargetInitializeConfig.Primaries != nil && s.TargetInitializeConfig.Master.DataDir != "" {
+		err := DeleteMasterAndPrimaryDataDirectories(streams, s.agentConns, s.TargetInitializeConfig)
+		if err != nil {
+			return xerrors.Errorf("deleting target cluster data directories: %w", err)
+		}
+	}
```
dkrieger-pvtl, line 85: Note the condition on L80 should always be true here...`s.TargetInitializeConfig` is set in `AssignDatadirsAndPorts` which is called when the cluster config is filled above.  Due to our past implementation choices in `DeleteMasterAndPrimaryDataDirectories`, that function will work if a given directory fails to exist entirely.  But it will fail if the directory is there but isn't fully formed yet.  This is probably ok for now.

Note the same logic is used in `revert` but we also delete the tablespace dirs there, since we might have started to run `execute`.

dkrieger-pvtl (commented):
The current design handles some of the possible `gpinitsystem` failures:
- gpinitsystem creates all datadirs and brings the cluster fully up but somehow still errors out
- gpinitsystem fails before creating datadirs or bringing up any processes
- a few others.

However, this does not handle arbitrary failures of `gpinitsystem`, like `gpinisystem` failing after a few segments are brought up(say, the network goes down during initialization).  This comes down to `gpstop` not being robust enough to handle arbitrary cluster states.  

I see a few ways to proceed:
- accept that we are doing better than before and accept that this step is only idempotent under a few possible failure modes
- try a "heavy handed" approach that works in all cases...finding all processes created by `gpinitsystem` and then terminating them
- defer this entirely; if `gpinitsystem` fails, provide manual steps to recover.

Which option we choose depends on how likely we think a `gpinitsystem` failure is.

hub/init_target_cluster.go
kalensk: Yes. start and stop semantics on the cluster need to eventually have idemototence. That work was found to be much too difficult and postponed. I have created a story to track this work.

That work is out of scope for this PR/work.

hub/init_target_cluster.go
kalensk: Removed the if condition.

test/initialize.bats
kalensk: I updated the test to stop a segment and remove its data directory.

bhuvnesh2703 (approved):

dkrieger-pvtl (approved):
LGTM. As we discussed, there are a lot of failure modes of gpinitsystem that will cause this code(in particular, `gpstop`) to fail, but in that case manual cleanup would have been needed anyway.  This PR will work in some of the failure modes of gpinitsystem.

PR #511, opened 2020-11-17
Final diffs

bhuvnesh2703:
This PR does the following:
- Move replacement initialization to a common function
- Add replacement for bit varying pattern to match 5x format
- Use Buffer to ensure what we are formatting

    CREATE VIEW Statement could appear in the body of a function or as a
    standalone DDL in the dump. View formatting should not be done when it's
    in the body of the function, so use the buffer to ensure that only the
    standalone VIEW statement are being formatted. Same applies for triggers
    and rules.
- Exclude precision replacement in lines containing COALESCE, i.e line part of DDL statement
- reindex target database and run filter on the source and target dump before comparision

    During upgrade bitmap indexes are marked invalid, so reindex the target
    database so that pg_dumpall could dump their DDL. The DDL for bitmap
    indexes exists in source dump as they were valid before upgrade.
- remove compare diff variable

### DIFF
--- a/ci/scripts/filters/replacement_test.go
+++ b/ci/scripts/filters/replacement_test.go
@@ -1,0 +1,40 @@
+// Copyright (c) 2017-2020 VMware, Inc. or its affiliates
+// SPDX-License-Identifier: Apache-2.0
+
+package filters
+
+import (
+	"testing"
+)
+
+func TestReplacements(t *testing.T) {
+	tests := []struct {
+		name     string
+		line     string
+		expected string
+	}{
+		{
+			name:     `append B to the pattern 'n'::"bit"`,
+			line:     `a39 bit(1) DEFAULT '0'::"bit" ENCODING`,
+			expected: `a39 bit(1) DEFAULT B'0'::"bit" ENCODING`,
+		},
+		{
+			name:     `append B to the pattern ('n'::"bit")`,
+			line:     `a40 bit varying(5) DEFAULT ('1'::"bit")::bit varying`,
+			expected: `a40 bit varying(5) DEFAULT (B'1'::"bit")::bit varying`,
+		},
+		{
+			name:     `does not append B to the pattern B'n'::"bit"`,
+			line:     `a39 bit(1) DEFAULT B'0'::"bit" ENCODING`,
+			expected: `a39 bit(1) DEFAULT B'0'::"bit" ENCODING`,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			actual := Replacements(tt.line)
+			if actual != tt.expected {
+				t.Errorf("got %v, expected %v", actual, tt.expected)
+			}
+		})
+	}
+}
--- a/ci/scripts/filters/trigger_test.go
+++ b/ci/scripts/filters/trigger_test.go
@@ -9,36 +9,36 @@ import (
 
 func TestFormatTriggerDdl(t *testing.T) {
 	tests := []struct {
-		name    string
-		tokens  []string
-		want    string
-		wantErr bool
+		name        string
+		tokens      []string
+		expected    string
+		expectedErr bool
 	}{
 		{
 			name: "formats create trigger statement and body in to one line",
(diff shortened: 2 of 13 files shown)


ci/scripts/filters/trigger_test.go
```diff
@@ -45,25 +45,35 @@ func TestFormatTriggerDdl(t *testing.T) {
...
 			result: false,
 		},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			if got := IsTriggerDdl(tt.line); got != tt.result {
+			if got := IsTriggerDdl(tt.args.buf, tt.args.line); got != tt.result {
 				t.Errorf("got %t, want %t", got, tt.result)
 			}
 		})
```
dkrieger-pvtl, line 79: nit: I think it is cleaner with [the args struct eliminated](https://gist.github.com/dkrieger-pvtl/8cbda2993c36c00cb616d013f9b8c8ff).

ci/scripts/filters/view_and_rule_test.go
```diff
@@ -49,30 +49,44 @@ func TestFormatViewOrRuleDdl(t *testing.T) {
...
 			result: false,
 		},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			if got := IsViewOrRuleDdl(tt.line); got != tt.result {
+			if got := IsViewOrRuleDdl(tt.args.buf, tt.args.line); got != tt.result {
 				t.Errorf("got %t, want %t", got, tt.result)
 			}
 		})
```
dkrieger-pvtl, line 92: nit: as above, I think this is cleaner with args inlined and result as `expected`.

ci/scripts/upgrade-cluster.bash
```diff
@@ -26,6 +26,17 @@ configure_gpdb_gucs() {
...
+reindex_all_dbs() {
+    local gphome=$1
+    ssh mdw bash <<EOF
+        set -eux -o pipefail
+
+        source ${gphome}/greenplum_path.sh
+        export MASTER_DATA_DIRECTORY=/data/gpdata/master/gpseg-1
+        reindexdb -a
+EOF
+}
```
dkrieger-pvtl, line 38: I see from the pipeline that this takes  less than 1 second.  Is that expected? I would have thought it would take some time...

ci/scripts/filters/replacement_test.go
```diff
@@ -0,0 +1,40 @@
...
+
+import (
+	"testing"
+)
+
+func TestReplacements(t *testing.T) {
+	tests := []struct {
+		name string
+		line string
+		want string
```
dkrieger-pvtl, line 14: nit: s/want/expected(similar in all tests)

ci/scripts/filters/replacement_test.go
```diff
@@ -0,0 +1,40 @@
...
+		},
+		{
+			name: `does not append B to the pattern B'n'::"bit"`,
+			line: `a39 bit(1) DEFAULT B'0'::"bit" ENCODING`,
+			want: `a39 bit(1) DEFAULT B'0'::"bit" ENCODING`,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got := Replacements(tt.line)
```
dkrieger-pvtl, line 34: nit: s/got/actual(similar in all tests)

dkrieger-pvtl (approved):
LGTM...consider some of the nits for renaming variables in the tests.  They are all in unit tests so you do not even have to wait for a full pipeline to run.

ci/scripts/filters/trigger_test.go
kalensk: +1

ci/scripts/upgrade-cluster.bash
bhuvnesh2703: it depends how many tables have the indexes and how much data they have. all these tables in regression test are about functionality so i wouldn't expect much data in tables.

PR #512, opened 2020-11-18
build and verify rpm in CI

dkrieger-pvtl:
We build, verify and upload gpupgrade rpms to GCP

We intend to use rpms to distibute both the open source("oss") and enterprise("enterprise") versions of gpupgrade.

Build the rpms and verify their contents.  Note:

- we upload a version "0.4.0-1.el7.x86_64" right now based on the rpm of greenplum-upgrade-0.4.0-1.el7.x86_64.rpm since that is how Concourse versions a GCP resource
- we need to improve this versioning as that upload will continually be replaced; that is, each build right now is not unique, but only based on the git tag.  We update this infrequently.
- the verification of the rpms just sanity checks the contents; we willbe passing the rpms into the actual tests in the CI in a later PR

Follow up PRs will be done for the following:
- add the verified rpms to the RC GCP bucket for download
- use the verified rpms in the CI tests

### DIFF
--- a/ci/scripts/verify-rpm.bash
+++ b/ci/scripts/verify-rpm.bash
@@ -1,0 +1,65 @@
+#!/bin/bash
+#
+# Copyright (c) 2017-2020 VMware, Inc. or its affiliates
+# SPDX-License-Identifier: Apache-2.0
+
+set -eux -o pipefail
+
+RPM=$1
+RELEASE=$2
+VERSION=$(git describe --tags --abbrev=0)
+
+verify_gpugprade_version_output() {
+  [[ $(/usr/local/greenplum-upgrade/gpupgrade version) == *"Version: ${VERSION}"* ]]
+  [[ $(/usr/local/greenplum-upgrade/gpupgrade version) == *"Release: ${RELEASE}"* ]]
+}
+
+verify_rpm_info() {
+  local info="$1"
+
+  [[ $info == *"Name        : greenplum-upgrade"* ]]
+  [[ $info == *"Architecture: x86_64"* ]]
+  [[ $info == *"Source RPM  : greenplum-upgrade-${VERSION}-1"* ]]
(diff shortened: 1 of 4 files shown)


ci/scripts/verify-rpm.bash
```diff
@@ -0,0 +1,65 @@
...
+set -eux -o pipefail
+
+RPM=$1
+RELEASE=$2
+VERSION=$(git describe --tags --abbrev=0)
+
+verify_gpugprade() {
+  [[ $(/usr/local/greenplum-upgrade/gpupgrade version) == *"Version: ${VERSION}"* ]]
+  [[ $(/usr/local/greenplum-upgrade/gpupgrade version) == *"Release: ${RELEASE}"* ]]
+}
```
kalensk, line 15: nit: I think it would be clearer to call this function `verify_gpugprade_version_output` or similar.

ci/tasks/build.yml
```diff
@@ -25,5 +28,19 @@ run:
     cd gpupgrade_src
     export GOFLAGS="-mod=readonly" # do not update dependencies during build
 
+    # TODO: when we switch to using rpms in tests, this can be removed
     make
     GOBIN=$(realpath ../build_artifacts) make install
+
+    # build the rpms
+    # TODO: remove once pivotaldata/gpdb6-centos7-test-golang is updated to include this rpm
+    yum -y install rpm-build
```
kalensk, line 37: Does this comment still apply as it looks like this is using `pivotaldata/gpdb6-centos7-test-golang` ?

kalensk (commented):
Looks good overall. Will let others take a look as I worked on this.

ci/tasks/build.yml
dkrieger-pvtl: Yes...I still plan to include that package in our docker image so that we do not need to install it on every test pipeline we run.

jmcatamney (approved):

PR #514, opened 2020-11-18
Handle cases where target object is not yet created.

bhuvnesh2703:
This commit handles revert failure when the target cluster is not yet
created.

### DIFF
--- a/hub/revert.go
+++ b/hub/revert.go
@@ -165,7 +165,7 @@ func (s *Server) Revert(_ *idl.RevertRequest, stream idl.CliToHub_RevertServer) 
 			return err
 		}
 
-		return ArchiveSegmentLogDirectories(s.agentConns, s.Config.Target.MasterHostname(), archiveDir)
+		return ArchiveSegmentLogDirectories(s.agentConns, s.Config.Source.MasterHostname(), archiveDir)
 	})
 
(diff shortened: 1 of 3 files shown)


hub/revert.go
```diff
@@ -165,7 +165,7 @@ func (s *Server) Revert(_ *idl.RevertRequest, stream idl.CliToHub_RevertServer)
 			return err
 		}
 
-		return ArchiveSegmentLogDirectories(s.agentConns, s.Config.Target.MasterHostname(), archiveDir)
+		return ArchiveSegmentLogDirectories(s.agentConns, s.Config.Source.MasterHostname(), archiveDir)
```
kalensk, line 168: Can you explain this change? Seems unrelated to this particular commit. Splitting out into a separate commit with explanation in message would be easier to follow.

bhuvnesh2703: i can add a comment if you want, but in this case as well target object is not yet created and instead of using Target Hostname we should rely on source hostname.

kalensk: I see. Thanks for the additional context. (I don't think a comment is needed. The BATs tests and other tests should catch this pretty quick if someone happens to change it back.)

kalensk (approved):

PR #515, opened 2020-11-19
refactor version functions

kalensk:
This is a pre-factor PR with various renames and moving code around for clarity. Namely, to distinguish between gpupgradeVersion and GPDBVersion.

Also, refactor the gpupgrade version consistency error message to print the list of hosts on a single line as this is the more common scenario. That is, it's more common for the same incorrect gpupgrade binary to be on all hosts. Thus, print all mismatched hosts on one line rather than a host per line which could be many lines for when there are numerous hosts.

[Pipeline](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:multiHostValidateVersions)

### DIFF
--- a/hub/version.go
+++ b/hub/version.go
@@ -11,6 +11,7 @@ import (
 	"github.com/greenplum-db/gp-common-go-libs/gplog"
 	"golang.org/x/xerrors"
 
+	"github.com/greenplum-db/gpupgrade/utils"
 	"github.com/greenplum-db/gpupgrade/utils/errorlist"
 )
 
@@ -39,12 +40,12 @@ func (a agents) String() string {
 }
 
 func ValidateGpupgradeVersion(hubHost string, agentHosts []string) error {
-	path, err := getBinaryPath()
+	gpupgradePath, err := utils.GetGpupgradePath()
 	if err != nil {
 		return xerrors.Errorf("getting gpupgrade binary path: %w", err)
 	}
 
-	hubVersion, err := GetVersionFunc(hubHost, path)
+	hubVersion, err := GetVersionFunc(hubHost, gpupgradePath)
 	if err != nil {
 		return xerrors.Errorf("getting hub version: %w", err)
 	}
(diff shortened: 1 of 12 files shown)


kalensk:
I just realized we don't need to move versions.go from the cli to greenplum package if we don't want to (ex: `cli/versions.go → greenplum/allowed_gpdb_versions.go`). Although the file rename still make sense to avoid confusion with the other version related files.

dkrieger-pvtl (commented):
I have some renaming and refactoring suggestions.  See https://github.com/dkrieger-pvtl/gpupgrade/commit/a1f5ae780963d1848666d2c4cc4dc1979dd79773.

kalensk:
> I have some renaming and refactoring suggestions. See dkrieger-pvtl@a1f5ae7.

Thanks. I think I addressed everything. Let me know if I missed anything.

hub/version.go
```diff
@@ -5,91 +5,89 @@ package hub
...
-	version string
-	err     error
+type HostVersionInfo struct {
+	host             string
+	gpupgradeVersion string
+	err              error
 }
 
-type agents map[string]string
+type GpupgradeVersionToHosts map[string][]string
```
dkrieger-pvtl, line 26: nit: This can be a shorter name like `BadVersion`...we know it is for hosts.  Especially since I thought you were going to use this twice: once for gpupgrade versions and once for gpdb versions.

hub/version.go
```diff
@@ -5,91 +5,89 @@ package hub
...
 	}
 
 	wg.Wait()
-	close(agentChan)
+	close(versionChan)
 
 	var errs error
-	mismatchedAgents := make(agents)
-	for agent := range agentChan {
+	gpupgradeVersionToHosts := make(GpupgradeVersionToHosts)
```
dkrieger-pvtl, line 60: nit: this can also be renamed `badGpupgradeVersions`.

dkrieger-pvtl (approved):
LGTM: consider the renames I've listed.  They can be shorter.

PR #517, opened 2020-11-19
CI: use within-pipeline rpms for testing and publish rpms

dkrieger-pvtl:
This PR does two things:
- uses the enterprise rpm created during the `build` job for testing, instead of the previously used `gpupgrade` binary
  - the `no-install` and `install` tests still use their own locally built `gpupgrade` binaries
- publishes both the open source and enterprise rpms if the enterprise rpm passes all tests

Note:
- the intermediate bucket rpms are now of the `gcs versioned_file` type.  Using this has concourse magically use the rpms created during this run of this pipeline in all jobs within the pipeline.  You will note the "generation number" of your rpm(like `1605848954599508`) is the same during the "build" task and the "publish release candidate" task.
- the published release candidates push an explicit version(a semver equivalent of `git describe` on the branch under test).  We assume that all builds of the same git tag are equivalent.  In practice, it is unlikely that two builds would exist, anyway.
- we clean up the version-rc.bash file to use bash instead of perl to parse the output of `git describe`.

We have a single commit since the changes are all coupled...we require all the changes to be commited before any of them will work.  In addition, we are still discussing how to install the rpms locally.  For this PR, we explicitly place `gpupgrade` under `/usr/local/bin`.


ci/template.yml
```diff
@@ -63,33 +63,33 @@ resources:
...
+- name: rpm_gpupgrade_oss
   type: gcs
   source:
     bucket: ((cm-intermediates-bucket))
     json_key: ((cm-gcs-service-account-key))
-    versioned_file: gpupgrade
+    versioned_file: oss/greenplum-upgrade.el7.x86_64.rpm
 
-- name: rpm_gpupgrade_oss
+- name: rpm_gpupgrade_enterprise
```
kalensk, line 73: Just had a thought. Since these names are used everywhere including inside various concourse tasks it might be nice to keep them short. Since know we are gpupgrade what about just simplifying to `enterprise_rpm` and `oss_rpm`. Similarly for the rc names below and everywhere else, etc.

ci/template.yml
```diff
@@ -597,13 +596,19 @@ jobs:
...
-      params:
-        file: gpupgrade_bin_path/gpupgrade-rc-*-linux_x86_64
+          path: gpupgrade_src/ci/scripts/add-semver.sh
+    - in_parallel:
+      - put: rpm_gpupgrade_oss_rc
+        params:
+          file: build_artifacts_rc_oss/greenplum-upgrade*.rpm
+      - put: rpm_gpupgrade_enterprise_rc
+        params:
+          file: build_artifacts_rc_enterprise/greenplum-upgrade*.rpm
```
kalensk, line 612: nit: These can be just `/*.rpm` without the `greenplum-upgrade` part. Perhaps slightly more generic, but not a big deal either way.

ci/template.yml
```diff
@@ -597,13 +596,19 @@ jobs:
...
         outputs:
-          - name: gpupgrade_bin_path
+          - name: build_artifacts_rc_oss
+          - name: build_artifacts_rc_enterprise
         run:
-          path: gpupgrade_src/ci/scripts/version-rc.sh
-    - put: bin_gpupgrade_rc
-      params:
-        file: gpupgrade_bin_path/gpupgrade-rc-*-linux_x86_64
+          path: gpupgrade_src/ci/scripts/add-semver.sh
```
kalensk, line 605: super minor nit: maybe call the file .bash instead of .sh

kalensk (commented):
Overall looks good. A couple of minor thoughts.

ci/template.yml
dkrieger-pvtl: Done...did this renamed.  I also renamed `build_artifacts` to `built` and I used `rename` for the publish step.  It's much cleaner now.  Thanks.

ci/template.yml
dkrieger-pvtl: I prefer `greenplum-upgrade`.  We are very unlikely to change the rpm package name, and this gives us some protection against something weird happening.

ci/template.yml
dkrieger-pvtl: Done.

kalensk (approved):
Looks good. 

Minor nit, it would be nice to quickly cleanup and be consist in referring to the RPM. I see:
- `greenplum-upgrade-*.rpm`
- `greenplum-upgrade*.rpm`
- `*.rpm`
- ...?

dkrieger-pvtl:
> Looks good.
> 
> Minor nit, it would be nice to quickly cleanup and be consist in referring to the RPM. I see:
> 
> * `greenplum-upgrade-*.rpm`
> * `greenplum-upgrade*.rpm`
> * `*.rpm`
> * ...?

Will do in a follow up PR.

PR #519, opened 2020-11-25
specify version output

kalensk:
Users can now specify the output formatting for the version string. Current formats are "oneline" and "json"  For example, `gpupgrade version --format oneline`.

[Pipeline](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:versionOutput)


cli/commands/commands.go
```diff
@@ -78,6 +79,7 @@ func BuildRootCommand() *cobra.Command {
 	}
 
 	root.Flags().BoolVarP(&shouldPrintVersion, "version", "V", false, "prints version")
+	root.Flags().StringVar(&format, "format", "", `specify the output format as either "multiline", oneline", or "json". Default is multiline.`)
```
dkrieger-pvtl, line 82: s/`oneline"`/`"oneline"`/ here and the same below.

cli/commands/version.go
```diff
@@ -3,17 +3,38 @@
...
+		return fmt.Sprintf(`{
+  "Version": %q,
+  "Commit": %q,
+  "Release": %q
+}`, Version, Commit, Release)
+	default:
+		return multiline
+	}
+
 }
```
dkrieger-pvtl, line 36: I think this is cleaner: https://gist.github.com/dkrieger-pvtl/b44ccb1e78027addeff348e4b3fa675c.

dkrieger-pvtl (commented):
We should only allow `--format` when `--version`/`-V` is specified; note `gpupgrade version` is its own subcommand so that case is already covered.  Note this is a very simple check that `format` is `""` unless `shouldPrintVersion` is true.  In the future, we might want to allow the entire output stream to be set this way, but we are not doing that in this PR.

Also consider a cleaner way to code the `version.go` file.

cli/commands/version.go
dkrieger-pvtl: We found a way that seemed good to both of us.

dkrieger-pvtl (approved):
LGTM

PR #520, opened 2020-11-25
Ensure GPDB versions match across all hosts

kalensk:
This refactors and reworks the code to add `EnsureVersionsMatch` which ensures both the gpupgrade and GPDB versions match between the hub and agents.

[Pipeline](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:mismatchVersionsFinal)


dkrieger-pvtl (commented):
I like the approach overall but I think the implementation can be simplified quite a bit.

I think the HubVersion/AgentVersion interface can be simplified.  It really comes down to a way to get the version from localhost or another host.  Right now, there are two layers of indirection.  What about just collapsing that?  See https://github.com/dkrieger-pvtl/gpupgrade/commit/306cefe4f33563d5d73b34ffd8f8276d575ce80b for one idea.

I think naming can be simplified as well:
--what about getting rid of the extra "package" naming in greenplum/version.go and upgrade/gpupgrade_version.go?  That is, drop "GPDB" and "Gpupgrade" for all identifier names so we can do upgrade.Version(host) and greenplum.Version(host), etc.
--ensure_versions_match.go naming can be improved.  Let's pair to discuss.

kalensk:
Definitely cleaner and let's pair. 

One thing that I would strongly advocate for is for the caller "_not_" to pass empty string to get the hub version (`version.Version("")`) since this basically says "empty string means hub" rather than having the name or some other mechanism make that clear and self-documenting to get the hub version vs. agent.

dkrieger-pvtl:
> One thing that I would strongly advocate for is for the end caller to "_not_" to pass empty string to get the hub version (ex: `version.Version("")`) since this basically says "empty string means hub" rather than having the name or some other mechanism make that clear and self-documenting to get the hub vs. agent version.

Yes, perhaps the middle grounds is for the interface to have two methods `LocalVersion` and `RemoteVersion` or some such.  And get rid of the current second level of indirection.  There's nothing "Hub" or "Agent" about getting the versions from the standpoint of the version functions.  From an engineering standpoint, of the options(single method with magic flag, two methods like currently, functional options), likely two methods makes sense here.

kalensk:
> the options(single method with magic flag, two methods like currently, functional options), likely two methods makes sense here.

This is the key. I am advocating for avoiding magic flags and the like and instead use a mechanism that is more self-documenting such as two methods.

hub/initialize.go
```diff
@@ -52,7 +54,11 @@ func (s *Server) Initialize(in *idl.InitializeRequest, stream idl.CliToHub_Initi
 	// we need the cluster information to determine what hosts to check, so we do this check
 	// as early as possible after that information is available
 	st.RunInternalSubstep(func() error {
-		return EnsureGpupgradeAndGPDBVersionsMatch(AgentHosts(s.Source), s.Source.MasterHostname())
+		if err := EnsureVersionsMatch(AgentHosts(s.Source), upgrade.NewVersions()); err != nil {
+			return err
+		}
+
+		return EnsureVersionsMatch(AgentHosts(s.Source), greenplum.NewVersions(s.TargetGPHome))
```
dkrieger-pvtl, line 61: Note we never check that the source version is the same on all hosts.  This _should_ be ok as we assume that they installed their cluster properly.   Furthermore, if we detect an issue, there's nothing we can do at this point...they are already running this cluster in production.   So it is probably better to not check the source cluster versions are the same on all hosts.

dkrieger-pvtl (approved):
LGTM.

PR #521, opened 2020-11-30
Update RPM Packaging

kalensk:
Includes the following changes to the RPM packaging:
- Rename rpm, spec file, and other names to gpupgrade instead of greenplum-upgrade.
- Add executables to /usr/local/bin such they are on the PATH by default.
- Rename migration scripts to include gpupgrade
- Move scripts and config file to utilize the Linux Filesystem Hierarchy.
- Register various "greenplum" and "greenplum/%{name}" directories as files so they are removed when the rpm is erased.
- One can install in an alternate location using: `rpm --prefix=/opt -ivh gpupgrade-*.rpm`
- add openssh and rsync as dependencies

[Pipeline](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:rpm_tweaks)


dkrieger-pvtl (approved):
LGTM.  Though we should have @bmdoil review as well for rpm expertise.

gpupgrade.spec
```diff
@@ -0,0 +1,85 @@
...
+if [ %{gpupgrade_version} = '%%{gpupgrade_version}' ] ; then
+    echo "The macro (variable) gpupgrade_version must be supplied as rpmbuild ... --define='gpupgrade_version [VERSION]'"
+    exit 1
+fi
+
+%setup -q -c -n %{name}
+
+%install
+# executables
+mkdir -p %{buildroot}%{prefix}/usr/local/bin
```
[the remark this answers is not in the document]
kalensk: Note: `%{prefix}` is `/` so I "think" `%{buildroot}/%{prefix}` would be `%{buildroot}//`. I'll let Brent add any notes.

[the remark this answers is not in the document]
dkrieger-pvtl: Yes, I think that's correct.  I initially was concerned that a user could do `--prefix=mydir` but that gives a failure from `rpm` as you see above.

So I think we're good.

gpupgrade.spec
```diff
@@ -0,0 +1,85 @@
...
+mv gpupgrade %{buildroot}%{prefix}/usr/local/bin
+mv data-migration-scripts/migration_generator_sql.bash  %{buildroot}%{prefix}/usr/local/bin/gpupgrade-migration-sql-generator.bash
+mv data-migration-scripts/migration_executor_sql.bash  %{buildroot}%{prefix}/usr/local/bin/gpupgrade-migration-sql-executor.bash
+
+# scripts
+mkdir -p %{buildroot}%{prefix}/usr/local/bin/greenplum/%{name}
+mv data-migration-scripts %{buildroot}%{prefix}/usr/local/bin/greenplum/%{name}
```
bmdoil, line 38: Do we want the SQL and header files to be placed inside /usr/local/bin? It seems that /usr/local/greenplum makes more sense

gpupgrade.spec
```diff
@@ -0,0 +1,85 @@
...
+
+%install
+# executables
+mkdir -p %{buildroot}%{prefix}/usr/local/bin
+mv gpupgrade %{buildroot}%{prefix}/usr/local/bin
+mv data-migration-scripts/migration_generator_sql.bash  %{buildroot}%{prefix}/usr/local/bin/gpupgrade-migration-sql-generator.bash
+mv data-migration-scripts/migration_executor_sql.bash  %{buildroot}%{prefix}/usr/local/bin/gpupgrade-migration-sql-executor.bash
+
+# scripts
+mkdir -p %{buildroot}%{prefix}/usr/local/bin/greenplum/%{name}
```
bmdoil, line 37: Setting up the build environment like this means that if a user specifies `--prefix=/opt/greenplum`, the binary will be placed at `/opt/greenplum/usr/local/bin/gpupgrade`. 

Is this the expected directory tree for relocated installs? 

```
[root@67e6e4a1d4b5 /]# tree -L 4 /opt/greenplum/
/opt/greenplum/
├── etc
│   ├── bash_completion.d
│   │   └── gpupgrade.bash
│   └── greenplum
│       └── gpupgrade
│           └── gpupgrade_config
└── usr
    ├── local
    │   └── bin
    │       ├── gpupgrade
    │       ├── gpupgrade-migration-sql-executor.bash
    │       ├── gpupgrade-migration-sql-generator.bash
    │       └── greenplum
    └── share
        ├── doc
        └── licenses
            └── gpupgrade-0.4.0
```

bmdoil (approved):
LGTM - 2 clarifying comments on the spec file

gpupgrade.spec
kalensk: > Is this the expected directory tree for relocated installs?

I think so, but @kmacoskey would have more insight into that requirement.

bmdoil: Bash completion won't work in this scenario since bash is expecting the files to be in `/etc/bash_completion.d/gpupgrade.bash` 

I'd recommend documenting that if the install is relocated, the user will need to manually move the file to the correct location if they want bash completion.

PR #523, opened 2020-12-08
Change the path of installed artifacts

bhuvnesh2703:
Change the path of installed artifacts

### DIFF
--- a/gpupgrade.spec
+++ b/gpupgrade.spec
@@ -33,53 +33,29 @@ mv gpupgrade %{buildroot}%{prefix}/usr/local/bin
 mv data-migration-scripts/migration_generator_sql.bash  %{buildroot}%{prefix}/usr/local/bin/gpupgrade-migration-sql-generator.bash
 mv data-migration-scripts/migration_executor_sql.bash  %{buildroot}%{prefix}/usr/local/bin/gpupgrade-migration-sql-executor.bash
 
-# scripts
+# additional files
 mkdir -p %{buildroot}%{prefix}/usr/local/bin/greenplum/%{name}
 mv data-migration-scripts %{buildroot}%{prefix}/usr/local/bin/greenplum/%{name}
+mv gpupgrade_config %{buildroot}%{prefix}/usr/local/bin/greenplum/%{name}
+mv gpupgrade.bash %{buildroot}%{prefix}/usr/local/bin/greenplum/%{name}
+mv open_source_licenses.txt %{buildroot}%{prefix}/usr/local/bin/greenplum/%{name}
 
-# config
-mkdir -p %{buildroot}%{prefix}%{_sysconfdir}/greenplum/%{name}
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -195,7 +195,7 @@ To recompile proto files to generate gRPC client and server code run
 
 ## Bash Completion
 
-To enable tab completion of gpupgrade commands source the `cli/bash/gpupgrade.bash`
+To enable tab completion of gpupgrade commands source the `gpupgrade.bash`
```
kalensk, line 198: I'd prefer to leave the local path to the to gpupgrade.bash. That is, `cli/bash` for developers. The README tends to be for developers and not so much for people building or installing the RPM (if so that most likely belongs in a separate README section if needed.)

gpupgrade.spec
```diff
@@ -33,53 +33,27 @@ mv gpupgrade %{buildroot}%{prefix}/usr/local/bin
 mv data-migration-scripts/migration_generator_sql.bash  %{buildroot}%{prefix}/usr/local/bin/gpupgrade-migration-sql-generator.bash
 mv data-migration-scripts/migration_executor_sql.bash  %{buildroot}%{prefix}/usr/local/bin/gpupgrade-migration-sql-executor.bash
 
-# scripts
+# additional files
 mkdir -p %{buildroot}%{prefix}/usr/local/bin/greenplum/%{name}
 mv data-migration-scripts %{buildroot}%{prefix}/usr/local/bin/greenplum/%{name}
+mv gpupgrade_config %{buildroot}%{prefix}/usr/local/bin/greenplum/%{name}
+mv gpupgrade.bash %{buildroot}%{prefix}/usr/local/bin/greenplum/%{name}
+mv open_source_licenses.txt %{buildroot}%{prefix}/usr/local/bin/greenplum/%{name}
```
kalensk, line 41: When installing on centos 7 I am only seeing the migration scripts and not the other files such as config, gpupgrade.bash, and license. The executables are in the correct place. Can you take a look.
```
/usr/local/bin/greenplum/gpupgrade/data-migration-scripts/
```

gpupgrade.spec
```diff
@@ -33,53 +33,27 @@ mv gpupgrade %{buildroot}%{prefix}/usr/local/bin
...
-# license
-mkdir -p %{buildroot}%{prefix}/usr/share/doc
-mkdir -p %{buildroot}%{prefix}/usr/share/licenses
 
 %files
 # executables
-%dir %{prefix}/usr
-%dir %{prefix}/usr/local
 %dir %{prefix}/usr/local/bin
+
```
kalensk, line 47: The following directories need to be registered so that when erasing the rpm from a non-standard location cleans up after itself.
```
%dir %{prefix}/usr/local
%dir  %{prefix}/usr
```

kalensk (approved):

PR #524, opened 2020-12-09
Delete the mirrors and standby tablespace directory in link mode during finalize

bhuvnesh2703:
Delete the mirrors and standby tablespace directory in link mode during finalize


test/finalize.bats
```diff
@@ -171,6 +175,19 @@ validate_data_directories() {
         done
 }
 
+validate_tablespace_dirs() {
+        CHECK_EXISTS=$1
+        shift
+        DATADIRS=("$@")
+        for datadir in "${DATADIRS[@]}"; do
```
kalensk, line 182: update datadir and DATADIRS to `ts` or tablespace.

kalensk (approved):

PR #531, opened 2021-01-28
initialize: do not use checkpoint_segments guc for 7X

dkrieger-pvtl:
The GUC checkpoint_segments is no longer in 7X, so we should not
attempt to pull it from a 7X source cluster in order to populate
the initconfig for the target cluster.

Furthermore, the GUC that supercedes it, max_wal_size, has a
sensible default value.  So we need do not need to replace checkpoint_segments
with anything.

[pipeline](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:7_to_7_remove_checkpoint_segments_guc) has passed.  The code is identical to that in a test branch that successfully did 7 to 7 by hand.

### DIFF
--- a/hub/init_target_cluster.go
+++ b/hub/init_target_cluster.go
@@ -48,7 +48,7 @@ func (s *Server) writeConf(sourceDBConn *dbconn.DBConn) error {
 		return err
 	}
 
-	gpinitsystemConfig, err = GetCheckpointSegmentsAndEncoding(gpinitsystemConfig, sourceDBConn)
+	gpinitsystemConfig, err = GetCheckpointSegmentsAndEncoding(gpinitsystemConfig, s.Source.Version, sourceDBConn)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


hub/init_target_cluster.go
```diff
@@ -119,18 +119,22 @@ func (s *Server) InitTargetCluster(stream step.OutStreams) error {
...
+
+	// The 7X guc max_wal_size supersedes checkpoint_segments and its default value is sufficient.
+	if version.Before("7") {
+		checkpointSegments, err := dbconn.SelectString(dbConnector, "SELECT current_setting('checkpoint_segments') AS string")
+		if err != nil {
+			return gpinitsystemConfig, xerrors.Errorf("retrieve checkpoint segments: %w", err)
+		}
+		gpinitsystemConfig = append(gpinitsystemConfig, fmt.Sprintf("CHECK_POINT_SEGMENTS=%s", checkpointSegments))
+	}
+
```
kalensk, line 137: For reference, assuming [the docs](https://gpdb.docs.pivotal.io/6-13/utility_guide/ref/gpinitsystem.html) are up to date and accurate 6X gpinitsystem requires `CHECK_POINT_SEGMENTS` and `ENCODING`. A more ideal reference would be the source code...I'll see if I can dig that up perhaps.

kalensk: Specifically, it looks like:

For 6X gpintisystem:
- [CHECK_POINT_SEGMENTS](https://github.com/greenplum-db/gpdb/blob/b5b4e16ae6ca2d0f1894357fb00944a23a975900/gpMgmt/bin/gpinitsystem#L404-L408) has a default value.
- [ENCODING](https://github.com/greenplum-db/gpdb/blob/b5b4e16ae6ca2d0f1894357fb00944a23a975900/gpMgmt/bin/gpinitsystem#L409-L414) has a default value.
- Note: That [the docs](https://gpdb.docs.pivotal.io/6-13/utility_guide/ref/gpinitsystem.html) say they are both required.

For 7X gpinitsystem:
- [CHECK_POINT_SEGMENTS](https://github.com/greenplum-db/gpdb/blob/2de82dff48be6a0ac25cd535154c03c897e34b48/gpMgmt/bin/gpinitsystem) is not present.
- [ENCODING](https://github.com/greenplum-db/gpdb/blob/2de82dff48be6a0ac25cd535154c03c897e34b48/gpMgmt/bin/gpinitsystem#L392-L396) has a default value.

dkrieger-pvtl: @kalensk Sure.  Please provide an explicit suggestion for the comment you'd like to see.

dkrieger-pvtl: We decided to decide whether to keep encoding/checkpoint_segments in a follow-up story.

kalensk (approved):
LGTM

PR #532, opened 2021-02-03
7X upgrade: use proper utility mode connection string

dkrieger-pvtl:
NOTE: this PR only does the implementation; it does not modify the bats tests.  That will be done separately, as currently this PR itself enables manual 6->7 gpugrade testing.

[pipeline](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:6_to_7_utility_mode_option_name)

Prior to 7X, a utility mode connection to the cluster uses the string "gp_session_role=utility".  In 7X and later, we use the string "gp_role=utility".  

This PR makes the necessary changes to allow any utility connection to know the correct version string to use.

One subtley in the implementation involves the different types we use for version in gpupgrade.  The Cluster object that is used to store the source and target information uses dbconn.Version, which is in gp-common-go-libs and contains both a string and a semver-v3.  But in checking the compatibility of versions to allow gpugprade, we use just a semver-v4, and that is the only version we want to use in gpupgrade.  But dbconn.Version's semver is NOT just a semver.MustParse(versionString); it is hand-constructed. This and the semver-v3 and semver-v4 incompatibility is why we use semver-v4.MustParse(cluster.Version.semver-v3.String()) method to get the semver for finding the utility connection.

Things to consider:
* is the correct cluster(source or target) being used to determine the version?

### DIFF
--- a/db/connURI/connection.go
+++ b/db/connURI/connection.go
@@ -1,0 +1,101 @@
+//  Copyright (c) 2017-2020 VMware, Inc. or its affiliates
+//  SPDX-License-Identifier: Apache-2.0
+
+package connURI
+
+import (
+	"fmt"
+
+	"github.com/blang/semver/v4"
+)
+
+// NOTE: This package is specific to the supported upgrade matrix of gpupgrade.
+// For instance, URI() will return the wrong result if the request is to
+// AllowSystemTableMods() on a 5X cluster as that parameter differs from 6X and 7X.
+
+// TODO: we should add the source/target ports here too, but they
+//  are known after we need to first call this package.
+type Conn struct {
+	sourceVersion semver.Version
+	targetVersion semver.Version
+}
+
+func Connection(sourceVersion semver.Version, targetVersion semver.Version) *Conn {
+	conn := new(Conn)
+	conn.sourceVersion = sourceVersion
+	conn.targetVersion = targetVersion
+
+	return conn
+}
+
+func (c *Conn) URI(options ...Option) string {
+	opts := newOptionList(options...)
+
+	version := c.sourceVersion
+	if opts.connectToTarget {
+		version = c.targetVersion
+	}
+
+	connURI := fmt.Sprintf("postgresql://localhost:%d/template1?search_path=", opts.port)
+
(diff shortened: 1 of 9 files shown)


kalensk:
We create various connection strings with various options throughout the code base. Personally, I think now is the time to pause and consider creating a more maintainable approach. This is more work in the short term, but worth it in the long term. For example, 

- Determining the version of the cluster on the hub. As the hub is guaranteed to be run on the coordinator. The CLI is better to decouple from business logic and act as a "front end" that may be run anywhere.  The current approach does not separate these concerns.
 - Have we considered prototyping using a functional arguments approach to building the connection string similar to how is done for Rsync and Run.go, etc.?  I have a feeling this would be quite nice. This is just one approach, but other similar ideas are welcomed.

The current approach needs to thread the version throughout the code base, rather than having a single place that callers can build their desired connection string. For example, can the callers simply use something like `conn.ToTargetCluster()` or `conn.ToSourceCluster()` without needing to thread the version throughout?

dkrieger-pvtl:
> * Determining the version of the cluster on the hub. As the hub is guaranteed to be run on the coordinator. The CLI is better to decouple from business logic and act as a "front end" that may eventually be run anywhere. The current approach does not separate these concerns.

Done.  It makes more sense this way seeing it...we grab it just before we use it, and of course the source GPHOME is present on the machine running the hub aka the master host.

dkrieger-pvtl:
> * Have we considered prototyping using a functional arguments approach to building the connection string similar to how is done for Rsync and Run.go, etc.?  I have a feeling this would be quite nice. [See prototype](https://github.com/kalensk/gpupgrade/tree/conn). This is just one approach, but other similar ideas are welcomed.

Done.

kalensk (approved):
LGTM.


Potential future ideas as already noted:
- See if port can be combined into the db/connURI construct.
- See if combing these two lines is helpful such that the db/connURI construct returns a sql connection:
```
connURI := conn.URI(options...)
connection, err := sql.Open("pgx", connURI)
```
- See if we can deprecate `NewDBConn` in the db package.

PR #534, opened 2021-02-09
finalize: for 7X target, do not update non-existent gpperfmon.conf

dkrieger-pvtl:
In a 5X to 6X upgrade, we need to update the contents of the
gpperfmon.conf file located in the master data directory.

In 7X, there is no such conf file.

The bats tests also need to be updated, but will be handled separately.  This PR enables manual 6->7 upgrade testing.

[pipeline](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:7_to_7_no_gpperfmon_conf)

### DIFF
--- a/hub/finalize.go
+++ b/hub/finalize.go
@@ -8,6 +8,7 @@ import (
 	"path/filepath"
 	"time"
 
+	"github.com/blang/semver/v4"
 	"github.com/greenplum-db/gp-common-go-libs/gplog"
 	"golang.org/x/xerrors"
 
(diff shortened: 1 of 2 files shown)


hub/update_conf_files_test.go
```diff
@@ -33,6 +33,8 @@ func TestUpdateConfFiles(t *testing.T) {
 		}
 	}()
 
+	// Note: we only need to UpdateGpperfmonConf() before 7X.  But this test
+	// itself does not have this dependency.
```
jmcatamney, line 37: Nit: Since the change was made to the caller of the function rather than the function itself, I don't think a comment here is necessary.

jmcatamney (approved):

hub/update_conf_files_test.go
dkrieger-pvtl: Fair enough; I'll remove it.

PR #537, opened 2021-03-18
Update README.md

divyeshddv:
gpupgrade is released and can be used in production. Updating README to reflect that.


kalensk (approved):
suggestion: Given this change I think it makes sense to also remove the accompanying phrase "It's still being actively developed and" as it mostly made sense in the context of not using it in production. To me that phrase is not really adding any value or saying anything that is not already obvious.

kalensk:
Looks good. Feel free to merge.

PR #538, opened 2021-03-23
build: update golang to 1.16

dkrieger-pvtl:
Note this requires a corresponding update to the version used in our
test pipelines.

The gpugprade pipeline passes except for the `goversion.bats` that tests that that we built with the correct version.  It's catch-22: the `noinstall` tests use the image-baking image which is currently on `1.15.7` and the `install` tests use the docker image which is changed to `1.16.2` here.  My plan is to, at the end of the day once this is approved, to merge the image-baking pipeline and then in the morning, merge this one.  Then all should be good.

The fact that all bats tests other than the `go version` check pass gives me high confidence that all our other tests will pass on 1.16.2. 

https://github.com/pivotal/gp-image-baking/pull/121 is the pipeline for that.

[pipeline](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:update_golang_to_1.16.2)

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,7 +117,7 @@ pg_upgrade supports two upgrade modes: link and copy.
 
 ### Prerequisites
 
-- Golang. We currently develop against latest stable Golang, which was v1.15 as of October 2020.
+- Golang. We currently develop against latest stable Golang, which was v1.16 as of October 2020.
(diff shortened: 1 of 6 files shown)


kalensk (approved):
LGTM. Thanks for updating. It's good to stay current on dependencies.

PR #540, opened 2021-04-22
Include Schema Name

bhuvnesh2703:
Included schema name


soumyadeep2007 (commented):
Can we reword and add some info into the commit message as to what problem we solved?

test/migration_scripts.bats
```diff
@@ -110,16 +99,16 @@ drop_unfixable_objects() {
...
 
     # Ignore the test tables that break the diff for now.
     # XXX We don't properly handle index constraints after revert, yet.
     # XXX We don't properly handle name type columns after revert, yet.
-    EXCLUSIONS="-T table_with_primary_constraint "
-    EXCLUSIONS+="-T table_with_unique_constraint "
-    EXCLUSIONS+="-T pt_with_index "
-    EXCLUSIONS+="-T sales "
-    EXCLUSIONS+="-T table_with_name_as_second_column "
+    EXCLUSIONS=" -T ${TEST_SCHEMA}.table_with_primary_constraint "
```
soumyadeep2007, line 107: Typo: leading space

soumyadeep2007 (approved):
LGTM!

test/migration_scripts.bats
kalensk: Yeah I did that to align all the table names since the first one is just `=` and the following are `+=`. Makes it easier to read for me.

PR #541, opened 2021-04-29
ci: use cached terraform plugin

dkrieger-pvtl:
Use a cached version of the terraform plugins rather than fetching it remotely from the internet. 

Terraform rotated their gpg key which caused our version of terraform init to fail with the following:
```
terraform init command failed.
Error: exit status 1
Output: 
Initializing provider plugins...
- Checking for available provider plugins on https://releases.hashicorp.com...

Error installing provider "google": openpgp: signature made by unknown entity.

Terraform analyses the configuration and state and automatically downloads
plugins for the providers used. However, when attempting to download this
plugin an unexpected error occured.
```

Specifically, terraform rotated their gpg key and the commit ljfranklin/terraform-resource@1c68991 has not made it to the image ljfranklin/terraform-resource:0.11.14 causing terraform init to fail. As a workaround we cache the terraform plugins on s3 and place them in terraform.d/plugin-cache/linux_amd64.

See also:
https://github.com/greenplum-db/gpdb/pull/11888

---

Note after merging re-fly the prod pipeline with: `make set-pipeline BRANCH=master FLY_TARGET=prod PIPELINE_NAME=gpupgrade`


kalensk:
Thanks! Good idea to track it in code via a PR or branch. Hopefully, releng can merge the root fix soon. (Ugh)

soumyadeep2007 (approved):
Saw a similar [commit](https://github.com/greenplum-db/gpdb/commit/1250c8f0683b12efaa173df001cb4880754e31ed) that might be interesting. It adds and uses a terraform.d cache containing both AWS and Google plugins.

Idk if we want to use something like that - if we have/want to have VMs on AWS for our jobs.

Since I'm not terribly familiar with terraform, I don't have much to add at the moment.

PR #543, opened 2021-05-05
README: add GIT_URI to updaing prod pipeline

kalensk:


soumyadeep2007 (commented):
Curious why we need this when we are already sensing the remote with this line in the Makefile?
```
GIT_URI := $(shell git ls-remote --get-url)
```

kalensk:
> Curious why we need this when we are already sensing the remote with this line in the Makefile?

Maybe you can help.

The prod pipeline has:
```
- name: gpupgrade_src
  source:
    branch: master
    fetch_tags: true
    uri: https://github.com/greenplum-db/gpupgrade
```

However, `git ls-remote --get-url` returns `https://github.com/greenplum-db/gpupgrade.git`. Notice the `.git` extension. This difference causes the state checker to fail unless we update the state checker (which we can). I guess the main question is it appears for some people `git ls-remote --get-url` returns with .git extension and for others it does not?!?

kalensk:
Closing this PR as the default `GIT_URI` in the Makefile has been updated to include the .git extension (see commit https://github.com/greenplum-db/gpupgrade/commit/2eed95f9a8e8a99737c06d6c8733d119727a857b). And the corresponding change has been made to the state checker. Thus, the `git ls-remote --get-url` now matches the default `GIT_URI`  and hopefully everything should just work when a developer fly's the prod pipeline. If not we can tweak things until they do, so that one does not need to manually override `GIT_URI`.

soumyadeep2007:
> I guess the main question is it appears for some people git ls-remote --get-url returns with the .git extension and for others it does not?!?

Maybe its a git version issue. Will have to dig through release notes.

About the ls-remote stuff, I'm not sure why we need it. Why not hard-code the repo url (moving the repo in Github is a rare event) in the Makefile and keep it overrideable?
Then we avoid the ifeq ($(GIT_URI),https://github.com/greenplum-db/gpupgrade) stuff too..(which will fail if ls-remote senses a git style URL?)
If we are to sense the url, we should maybe do git ls-remote --get-url origin (since a repo can have multiple remotes..like I have a "sc" remote for my fork of gpupgrade)

kalensk:
> Why not hard-code the repo url (moving the repo in Github is a rare event) in the Makefile and keep it overrideable?

I agree! Let's go that route. It also simplifies a lot of things such as the Makefile and state checker, etc. We will do a followup shortly.

PR #546, opened 2021-05-07
pass current step name in prompt

kalensk:
Before the prompt was incorrectly hard coded to initialize. Fix it to dynamically display the correct step.

[Pipeline](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:fixPrompt)


soumyadeep2007 (approved):
LGTM!

PR #547, opened 2021-05-17
initialize: validate source and temp target ports

kalensk:
Ensure that the target temp_port_range ports do not overlap with the source cluster ports.

[Pipeline](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:validatePorts)

### DIFF
--- a/hub/fill_cluster_configs.go
+++ b/hub/fill_cluster_configs.go
@@ -5,6 +5,7 @@ package hub
 
 import (
 	"database/sql"
+	"fmt"
 	"path/filepath"
 	"sort"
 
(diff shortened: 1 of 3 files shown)


hub/fill_cluster_configs.go
```diff
@@ -187,3 +192,53 @@ func sanitize(ports []int) []int {
...
+	}
+	sort.Ints(targetPorts)
+
+	for _, sourcePort := range sourcePorts {
+		for _, targetPort := range targetPorts {
+			if sourcePort == targetPort {
+				return newInvalidTempPortRangeError(sourcePort)
+			}
+		}
+	}
```
soumyadeep2007, line 222: Minor optimization opportunity:

The implementation would be slightly more performant (especially if the environment has a large number of segments, in thousands) if we use a set (maybe a `map[string]bool` to capture all of the source cluster ports and then perform a membership check against that set for each target cluster port. This will yield a O(n) check as opposed to the currrent O(n^2) check.

soumyadeep2007 (approved):
LGTM. I have a minor comment that is not a blocker from my perspective, but a nice to have. Thanks for the extensive unit testing.

hub/fill_cluster_configs.go
kalensk: Good idea! I like it. I've updated the code accordingly and unit tests still pass.

(Side rant: I wish golang had better helpers/features around data structures such as set's, iterating, etc.)

soumyadeep2007: Looks good!

PR #548, opened 2021-05-19
config: add target-port

kalensk:
For ease in testing show temporary coordinator port for the target cluster.

Test Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:targetPort

### DIFF
--- a/cli/commands/commands.go
+++ b/cli/commands/commands.go
@@ -157,6 +157,7 @@ func createConfigShowSubcommand() *cobra.Command {
 	subShow.Flags().Bool("source-gphome", false, "show path for the source Greenplum installation")
 	subShow.Flags().Bool("target-gphome", false, "show path for the target Greenplum installation")
 	subShow.Flags().Bool("target-datadir", false, "show temporary data directory for target gpdb cluster")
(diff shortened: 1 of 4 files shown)


cli/commands/commands.go
```diff
@@ -157,6 +157,7 @@ func createConfigShowSubcommand() *cobra.Command {
 	subShow.Flags().Bool("source-gphome", false, "show path for the source Greenplum installation")
 	subShow.Flags().Bool("target-gphome", false, "show path for the target Greenplum installation")
 	subShow.Flags().Bool("target-datadir", false, "show temporary data directory for target gpdb cluster")
+	subShow.Flags().Bool("target-port", false, "show temporary coordinator port for target cluster")
```
soumyadeep2007, line 160: Nit: We don't call it the coordinator in 5 and 6. Should we say `master (coordinator)` or just `master` now?

soumyadeep2007 (approved):

cli/commands/commands.go
kalensk: fixed! I got too excited at using the new name. When needed we should switch to coordinator all at once, rather than certain parts of the code.

PR #549, opened 2021-05-20
initialize: fix overlapping ports test

kalensk:
When checking overlapping ports check target cluster ports in order of master, standby, primaries, and mirrors. This enables the 6->6 initialize.bats to succeed when checking the conflicting port.

[Pipeline](https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:overlapPorts)


kalensk:
@soumyadeep2007 I updated the implementation to check across host/port configurations. This allows users to have the same port across different hosts.

PR #550, opened 2021-05-26
pg_upgrade acceptance tests

soumyadeep2007:
This commit transplants the regression style tests from
pg_upgrade/test/integration (from GPDB 6X_STABLE) into gpupgrade while
making the following improvements:

1. Tests are easy and simple to run locally with minimal setup. The only
dependencies being: a running source cluster, the path to the source
cluster and target cluster binaries and the pg_isolation2 source
directory.
2. Tests are unentangled into two distinct suites: non-upgradeable
(negative) and upgradeable (positive) tests. This improves
understandability drastically and eliminates schedule hell (Where does
this test belong? Will this test run on the source or target?).
3. The test framework runs gpupgrade as opposed to simulating gpupgrade
with pg_upgrade, as it did in the past. This improves understandability
and eliminates the possibility of code divergence between gpupgrade and
its simulation.
4. Significant improvements to the readability, clarity of purpose and
removal of extra fluff from the tests. Some flaky tests were hardened
with regress substitution directives. Some tests were fixed to return a
deterministic sort order in the output.
5. Negative tests exercise the customer provided workarounds from the
gpupgrade documentation. [1]

test/README.md details the internals of the framework and contains
detailed steps to run the tests.

[1] https://gpdb.docs.pivotal.io/upgrade/1-0/gpupgrade_initialize_checks.html#name_user_columns

Test Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:contribTests


kalensk:
Note: need to re-fly the gpupgrade prod pipeline after this is merged to create the new pg-upgrade-tests jobs.

ci/scripts/pg-upgrade-tests.bash
```diff
@@ -0,0 +1,91 @@
...
+# Alternatively, refactor common.bash to use $GPHOME. However, due to unforeseen
+# consequences and stability concerns we cannot do that.
+ln -s "$GPHOME_TARGET" /usr/local/greenplum-db-devel
+source gpdb_src/concourse/scripts/common.bash
+setup_configure_vars
+export LDFLAGS="-L$GPHOME_TARGET/ext/python/lib $LDFLAGS"
+configure
+
+source "${GPHOME_TARGET}"/greenplum_path.sh
+make -j "$(nproc)" -C gpdb_src
```
ashwinstar, line 66: don't need to compile full gpdb source I think, just `gpdb_src/src/test/isolation2` should do the job

ashwinstar (approved):
It would have made it easier for review and also good for long term, if we had commit "acceptance: Targeted test framework" split in two pieces. Real test framework only code commit with maybe a test to showcase and then adding more tests added as separate commit.

ci/scripts/pg-upgrade-tests.bash
kalensk: I tried that and it doesn't work. Unfortunately, I can't remember the exact failure. Let me know if you would like me to remove it, run the pipeline, find the exact failure, and paste it here.

ashwinstar: Nope, its fine. Longer term can optimize this piece out sometime. I tried locally and it worked (might be due to some development setup)

kalensk: Ya, I was getting different results on the CI vs. locally or when manually logging into the box. Note, that we are installing the source and target RPM's and then doing a "make" for isolation2. Its slightly round-about.

shivzone:
General suggestion, would be more relevant if you named the PR as `gpupgrade bats acceptance tests`

PR #555, opened 2021-06-02
Take 2: Sort pg_upgrade --check output independent of platform

soumyadeep2007:
This is a follow-up to commit 62eb4aa. Even with the locale being the
same in CI as our local dev environments, the sort order can be
different with respect to alphanumeric symbols. For instance this can
happen if you have different versions of glibc.
See post: https://stackoverflow.com/questions/27395317/why-does-utf-8-text-sort-in-different-order-between-os-x-and-linux


soumyadeep2007:
Dev pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:lcall

kalensk (approved):

PR #559, opened 2021-06-07
hub: persist archive directory

kalensk:
Persist the archive directory so that during re-runs and restarts the correct directory is sent from the hub to cli and displayed to the user.

Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:archiveDir

### DIFF
--- a/hub/revert.go
+++ b/hub/revert.go
@@ -6,8 +6,6 @@ package hub
 import (
 	"fmt"
 	"os/exec"
-	"path/filepath"
-	"time"
 
 	"github.com/greenplum-db/gp-common-go-libs/gplog"
 	"github.com/pkg/errors"
@@ -15,7 +13,6 @@ import (
 
 	"github.com/greenplum-db/gpupgrade/idl"
 	"github.com/greenplum-db/gpupgrade/step"
-	"github.com/greenplum-db/gpupgrade/upgrade"
 	"github.com/greenplum-db/gpupgrade/utils"
 	"github.com/greenplum-db/gpupgrade/utils/errorlist"
 )
@@ -150,22 +147,24 @@ func (s *Server) Revert(_ *idl.RevertRequest, stream idl.CliToHub_RevertServer) 
 		})
 	}
 
-	// FIXME: archiveDir is not set unless we actually run this substep; it must be persisted.
-	var archiveDir string
+	logArchiveDir, err := s.GetLogArchiveDir()
+	if err != nil {
+		return fmt.Errorf("getting archive directory: %w", err)
+	}
+
 	st.Run(idl.Substep_ARCHIVE_LOG_DIRECTORIES, func(_ step.OutStreams) error {
 		// Archive log directory on master
-		oldDir, err := utils.GetLogDir()
+		logDir, err := utils.GetLogDir()
 		if err != nil {
 			return err
 		}
-		archiveDir = filepath.Join(filepath.Dir(oldDir), upgrade.GetArchiveDirectoryName(s.UpgradeID, time.Now()))
 
-		gplog.Debug("moving directory %q to %q", oldDir, archiveDir)
-		if err = utils.Move(oldDir, archiveDir); err != nil {
+		gplog.Debug("archiving log directory %q to %q", logDir, logArchiveDir)
+		if err = utils.Move(logDir, logArchiveDir); err != nil {
 			return err
 		}
 
-		return ArchiveSegmentLogDirectories(s.agentConns, s.Config.Source.MasterHostname(), archiveDir)
+		return ArchiveSegmentLogDirectories(s.agentConns, s.Config.Source.MasterHostname(), logArchiveDir)
 	})
 
 	st.Run(idl.Substep_DELETE_SEGMENT_STATEDIRS, func(_ step.OutStreams) error {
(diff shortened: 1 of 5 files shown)


dkrieger-pvtl (commented):
As this is a draft PR, I just looked at the design.

What do you mean by "persist" here?  The existing code will keep saving the archive directory to a new location(upon reaching that substep in `finalize` or `revert`)  as a new timestamp is generated each time that substep is run.

I suppose that is fine for now; we do not need to write the archiveDir to disk.  That said, another design would be to generate a timestamp during `initialize`, store it to disk in the config.json file, and then read it here.

But the current solution is fine for now; we do not handle the hub dying and restarting anyway.

kalensk:
> What do you mean by "persist" here? The existing code will keep saving the archive directory to a new location(upon reaching that substep in `finalize` or `revert`) as a new timestamp is generated each time that substep is run. The new code stores the archiveDir in memory via the Server object so there is only one archiveDir per gpupgrade run(provided the hub stays up).

The current PR _does_ in fact persist the `archiveDir` to disk. The function `GetArchiveDir()` checks if `archiveDir` is already set, and if not creates one and persists it to disk by calling `s.SaveConfig()`. This saves it in `config.json`.

Depending on what is _actually_ needed one could theoretically call `LoadConfig()` before checking if `archiveDir` is already set to synchronize the in-memory server object with that on disk. I don't know the full ramifications or side-effects of this. That is, all the interactions and affects of other downstream code.

dkrieger-pvtl:
> Depending on what is _actually_ needed one could theoretically call `LoadConfig()` before checking if `archiveDir` is already set to synchronize the in-memory server object with that on disk. I don't know the full ramifications or side-effects of doing so. That is, all the interactions and affects of other downstream code.

Yeah, we likely don't want to overwrite all config values in Server as I don't think we always Save it after changing any value; but we could just use the archiveDir value here.

hub/server_test.go
```diff
@@ -360,6 +362,54 @@ func TestHubSaveConfig(t *testing.T) {
...
+		expected := fmt.Sprintf("%s-%s-", logDir, id)
+
+		if !strings.HasPrefix(actual, expected) {
+			t.Errorf("got %v want prefix %v", actual, expected)
+		}
+	})
+
+	t.Run("returns persisted archive directory if it is already set", func(t *testing.T) {
+		expected := "/home/gpadmin/gpAdminLogs/gpupgrade-ABC123-2006-01-02T15:04"
+		server.SetArchiveDir(expected)
```
soumyadeep2007, line 400: Instead of having `SetArchiveDir` just for testing, how about we call `GetArchiveDir()` a second time here (1st time being in the previous `t.Run` and check if we get the same output as the previous test?

hub/server_test.go
```diff
@@ -360,6 +362,54 @@ func TestHubSaveConfig(t *testing.T) {
...
+			t.Fatalf("failed to get log dir: %v", err)
+		}
+
+		id := "AAAAAAAAAAA" // all zeroes in base64. 8 bytes decoded -> 11 bytes encoded
+		expected := fmt.Sprintf("%s-%s-", logDir, id)
+
+		if !strings.HasPrefix(actual, expected) {
+			t.Errorf("got %v want prefix %v", actual, expected)
+		}
+	})
```
soumyadeep2007, line 396: Should we also assert that the on-disk config was modified? Read the config file back in and examine the archiveDir?

soumyadeep2007 (approved):
> > Depending on what is _actually_ needed one could theoretically call `LoadConfig()` before checking if `archiveDir` is already set to synchronize the in-memory server object with that on disk. I don't know the full ramifications or side-effects of doing so. That is, all the interactions and affects of other downstream code.
> 
> Yeah, we likely don't want to overwrite all config values in Server as I don't think we always Save it after changing any value; but we could just use the archiveDir value here.

If we do do that, let's add a detailed comment stating why it is okay to do so for `archiveDir` and not the other config params.

hub/server_test.go
soumyadeep2007: In that case, lets have the second test call GetArchiveDir twice. That way it is self-contained and we can remove the function.

hub/server_test.go
kalensk: yes!

hub/server_internal_test.go
```diff
@@ -25,6 +25,7 @@ func TestConfig(t *testing.T) {
 		// forget to add them to this test. Be kind and document those that are
 		// not clear with comments.
 		original := &Config{
+			"archiveLogDir",
```
soumyadeep2007, line 28: s/archiveLogDir/logArchiveDir

hub/revert.go
```diff
@@ -150,22 +147,24 @@ func (s *Server) Revert(_ *idl.RevertRequest, stream idl.CliToHub_RevertServer)
...
-		oldDir, err := utils.GetLogDir()
+		logDir, err := utils.GetLogDir()
 		if err != nil {
 			return err
 		}
-		archiveDir = filepath.Join(filepath.Dir(oldDir), upgrade.GetArchiveDirectoryName(s.UpgradeID, time.Now()))
 
-		gplog.Debug("moving directory %q to %q", oldDir, archiveDir)
-		if err = utils.Move(oldDir, archiveDir); err != nil {
+		gplog.Debug("moving directory %q to %q", logDir, logArchiveDir)
```
soumyadeep2007, line 162: nit: "moving log directory"

kalensk: addressed.

PR #560, opened 2021-06-07
ci: update tasks to use registry-image and GCR

kalensk:
https://github.com/greenplum-db/gpupgrade/pull/554 missed a few references of pivotaldata and docker-image used within tasks.

Use GCR to reduce docker pull count from docker. Use the registry-image resource type which is a newer replacement for the docker-image resource.

Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:updateDocker2

### DIFF
--- a/ci/tasks/build.yml
+++ b/ci/tasks/build.yml
@@ -5,9 +5,9 @@ PLATFORM: linux
 
 # Use a Centos/RHEL image to build the rpms
 image_resource:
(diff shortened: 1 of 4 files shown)


soumyadeep2007 (approved):

kalensk:
Note: Double check by re-flying prod pipeline (but should be a no-op).

PR #562, opened 2021-06-11
Check disk space: include tablespaces and skip filesystems that cannot be accessed

kalensk:
1) Skip filesystems that cannot be accessed. Rather than error'ing out the entire disk space check substep, simply skip filesystems that the gpupgrade user such as gpadmin cannot access.

2) When checking disk size include tablespace in size calculation.

Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:checkDiskSpace


testutils/test_utils.go
```diff
@@ -53,6 +53,71 @@ func CreateMultinodeSampleClusterPair(baseDir string) (*greenplum.Cluster, *gree
 	return sourceCluster, targetCluster
 }
 
+func CreateTablespaces() greenplum.Tablespaces {
+	return greenplum.Tablespaces{
```
soumyadeep2007, line 57: I think we can use this opportunity to correct the examples used here to be closer aligned to reality. We can make the changes together. I'm braindumping here to help our discussion.

Oids 1663 (`pg_default`) and 1664 ('pg_global`) are reserved for the system built-in tablespaces. So we should only use these oids for instances where `UserDefined=0`. Lets use Oids >= 16384 for user-defined tablespaces.

1. Lets change `1664 -> 16384` in these examples.
2. Lets change `Location` of `1663` to `<datadir>/base`
3. Lets change `Location` of `16384` to `/tmp/custom_tablespace`

utils/disk/disk.go
```diff
@@ -68,6 +70,13 @@ func CheckUsage(streams step.OutStreams, d Disk, diskFreeRatio float64, paths ..
 	fsByID := make(map[uint64]string)
 	for _, f := range fs.List {
 		stat, err := d.Stat(f.DirName)
+		if os.IsPermission(err) {
+			// NOTE: Since the user running gpupgrade such as gpadmin might not
+			// have permission to stat all filesystems on the host, ignore the
+			// ones we do not have permission for.
```
soumyadeep2007, line 76: We should transform this comment into a log message of level WARNING.

utils/disk/disk_test.go
```diff
@@ -288,6 +288,63 @@ func TestCheckUsage(t *testing.T) {
 			t.Errorf("returned %#v after all error cases were removed", err)
 		}
 	})
+
+	t.Run("does not error when unable to access a filesystem due to permissions", func(t *testing.T) {
+		d.filesystems = func() (sigar.FileSystemList, error) {
+			return sigar.FileSystemList{List: []sigar.FileSystem{
+				{DirName: "/"},
+				{DirName: "/tmp"},
```
soumyadeep2007, line 296: Can we make the test more minimal by cutting down `/` and `/tmp` ?

hub/check_disk_space_test.go
```diff
@@ -64,14 +68,16 @@ func TestCheckDiskSpace_OnMaster(t *testing.T) {
 
 func TestCheckDiskSpace_OnSegments(t *testing.T) {
```
soumyadeep2007, line 69: Should we add a test case such that the collective space used by segments + tablespaces exceed the free ratio?

soumyadeep2007 (commented):
Looking good. I have some comments we can work through. The disk space refactor really paid off.

utils/disk/disk.go
```diff
@@ -68,6 +70,11 @@ func CheckUsage(streams step.OutStreams, d Disk, diskFreeRatio float64, paths ..
 	fsByID := make(map[uint64]string)
 	for _, f := range fs.List {
 		stat, err := d.Stat(f.DirName)
+		if os.IsPermission(err) {
+			gplog.Warn("Ignoring filesystem %s on host %s. Unable to stat filesystem due to %v.", f.DirName, hostname, err)
```
kalensk, line 74: Consider mentioning "checking disk space". That is, "Not including filesystem %s on host %s in disk space check. ..."

hub/check_disk_space.go
```diff
@@ -77,6 +85,15 @@ func checkDiskSpaceOnStandbyAndSegments(agentConns []*Connection, errs chan<- er
 		var dirs []string
 		for _, seg := range segmentsExcludingMaster {
 			dirs = append(dirs, seg.DataDir)
+
+			segTablespaces := sourceTablespaces[seg.DbID]
+			for _, tsInfo := range segTablespaces {
```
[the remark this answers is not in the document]
kalensk: No longer "applies" after extracting out to using helper function `UserDefinedTablespacesLocations()`

testutils/test_utils.go
kalensk: Updated! Thanks :)

hub/check_disk_space_test.go
```diff
@@ -146,34 +167,106 @@ func TestCheckDiskSpace_OnSegments(t *testing.T) {
...
+			}}
+		mirror := mock_idl.NewMockAgentClient(ctrl)
+		mirror.EXPECT().CheckDiskSpace(
+			gomock.Any(),
+			gomock.Any(),
+		).Return(&idl.CheckDiskSpaceReply{Usage: mirrorUsage}, nil)
+
+		agentConns := []*hub.Connection{
+			{nil, primary, "primary", nil},
+			{nil, mirror, "mirror", nil},
```
soumyadeep2007, line 236: nit: can we use the "key": "val" style initialization of structs that we use in say line 223 above?

hub/initialize.go
```diff
@@ -89,7 +89,12 @@ func (s *Server) Initialize(in *idl.InitializeRequest, stream idl.CliToHub_Initi
 	})
 
 	st.RunConditionally(idl.Substep_CHECK_DISK_SPACE, in.GetDiskFreeRatio() > 0, func(streams step.OutStreams) error {
-		return CheckDiskSpace(streams, s.agentConns, in.GetDiskFreeRatio(), s.Source)
+		conns, err := s.AgentConns()
```
soumyadeep2007, line 92: Nit: should we extract the condition that a refresh of connections is successful into the step's condition (i.e. add an `&&` clause to `in.GetDiskFreeRatio() > 0`)?

Disregard if this is not the prevalent pattern.

soumyadeep2007 (approved):
LGTM

hub/check_disk_space_test.go
kalensk: Updated.

PR #565, opened 2021-06-17
version: use -q when ssh'ing for remote version

kalensk:
If the remote machine uses motd banners or other messages this will get mingled with the output of the version command leading to errors. Suppress motd banner messages by using the ssh -q parameter.

Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:sshVersion

### DIFF
--- a/greenplum/common_test.go
+++ b/greenplum/common_test.go
@@ -21,6 +21,10 @@ func MockExecCommand(ctrl *gomock.Controller) (mock *exectest.MockCommandSpy, cl
 	return mock, ResetExecCommand
 }
 
+func SetExecCommand(cmdFunc exectest.Command) {
+	execCommand = cmdFunc
+}
+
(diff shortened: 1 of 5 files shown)


soumyadeep2007:
Reading the ssh man page on Ubuntu:
```
-q      Quiet mode.  Causes most warning and diagnostic messages to be suppressed.
```

We will lose the ability to see error/warning messages (say the host went down, port not open etc). Idk if we would want that.

kalensk:
> We will lose the ability to see error/warning messages (say the host went down, port not open etc). Idk if we would want that.

The man page talks about warnings, _not_ errors. All the things you listed are errors which would be correctly bubbled up. I think we are fine.

soumyadeep2007 (approved):
Cool, looks good then!

PR #567, opened 2021-07-01
CI: Refactor Pipeline Tasks

kalensk:
Refactor Pipeline Tasks:
- Inline task yaml's to enable faster feedback by making changes to the pipeline on the fly rather than pushing changes to the task yaml and waiting for it flow through.
- Rename CheckJobs to GpupgradeTests to better differentiate between pg_upgrade and gpupgrade jobs, which run gpupgrade and pg_upgrade acceptance tests respectively.

Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:refactorTaskYaml


ci/generated/pipeline.yml
```diff
@@ -550,8 +608,20 @@ jobs:
           args:
             - greenplum-db-6
             - greenplum-db-6
-    - task: load_dump
-      file: gpupgrade_src/ci/tasks/load-dump.yml
+    - task: load_data
```
kalensk, line 611: Im not sure about changing this to load_data from load_dump...

ci/template.yml
```diff
@@ -492,8 +536,20 @@ jobs:
         run:
           path: gpupgrade_src/ci/scripts/load-retail-data.bash
     {{- else }}
-    - task: load_dump
-      file: gpupgrade_src/ci/tasks/load-dump.yml
+    - task: load_data
```
kalensk, line 539: Im not sure about changing this to load_data from load_dump...

soumyadeep2007 (commented):
> ci: rename task from load_dump to load_data
ci: inline install-tests.yml 
ci: inline noninstall-tests 
ci: inline build.yml

1. Inlining:
I don't feel right about these changes, I feel inlining like this is against pipeline best practices. 
Having a clean, short pipeline .yml file is much more important to me than the convenience provided by these patches. I'm feeling the pain right now trying to prune away the monolithic 5X_STABLE pipeline into compile and release jobs for instance.

2. load_dump seems fine to me. Its good to be explicit that we are loading a dump for our testing. I would personally refactor it to be `load_icw_dump`.

> ci: rename CheckJobs to GpupgradeJobs
ci: rename PgUpgradeTests to PgupgradeJobs

These look good!

kalensk:
> 1. Inlining:
>    I don't feel right about these changes, I feel inlining like this is against pipeline best practices.
>    Having a clean, short pipeline .yml file is much more important to me than the convenience provided by these patches. I'm feeling the pain right now trying to prune away the monolithic 5X_STABLE pipeline into compile and release jobs for instance.

The earlier team already had plans and intention of inlining as noted by the following tasks that are already present `on_failure_set_failed`, `lint`, `prepare_source_and_target_installations `, `load_retail_data `,  `upgrade-cluster`, `run_bats `, `add-semver`,  and `pg-upgrade-tests`. This just PR just finishes that intention and tech debt.

The indirection of jumping from task > yml file > to script is frankly too confusing and cumbersome for little to no benefit. I feel like it's a poor pattern that people are simply "just used to". Again this allows one to quickly iterate on pipeline tasks without needing to wait for the changes to flow through.

"If" the pipeline becomes too long one solution that seems better is to split it in logical components and basically do a #import equivalent to reconstruct a master pipeline.

kalensk:
As a prototype I tried breaking up the template into logical sections. On initial glance there does not seem to be an #include type of feature. So simply cat'ing the files works just fine although the order does matter. Probably easiest to checkout the branch and see.

https://github.com/kalensk/gpupgrade/tree/breakApartPipeline

Also, I had some trouble doing the `cat` as a `go:generate` command in `generate.go`, so I put it in the Makefile. My suspicion is there is no "order" when running go generate and thus can't generate the template to feed to the parse_template. I also tried inlining into one command, but that didn't seem to work on initial try.

ci/template.yml
kalensk: Reverted this back to load_dump....

kalensk:
Added a new commit: `ci: rename bats-on-cluster.bash`

soumyadeep2007:
> As a prototype I tried breaking up the template into logical sections. On initial glance there does not seem to be an #include type of feature. So simply cat'ing the files works just fine although the order does matter. Probably easiest to checkout the branch and see.
> 
> https://github.com/kalensk/gpupgrade/tree/breakApartPipeline
> 
> Also, I had some trouble doing the `cat` as a `go:generate` command in `generate.go`, so I put it in the Makefile. My suspicion is there is no "order" when running go generate and thus can't generate the template to feed to the parse_template. I also tried inlining into one command, but that didn't seem to work on initial try.

Yep that looks good. Minor comment for that branch:
1) Let's not check in ci/template.yml as it will be generated by `cat`? Also there is a straggling empty `template.yml` file in that commit.
2) The `rm` before the `cat` is superfluous. '>' should overwrite.

soumyadeep2007 (approved):

PR #569, opened 2021-07-14
README: add intelliJ preferences

kalensk:
Update README with IDE preferences/settings. For now add IntelliJ preferences.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,6 +123,43 @@ system which can be installed on macOS with `brew install protobuf`.
 - Run `make && make depend-dev` to install other developer dependencies. Note 
 make needs to be run first.
 
+### Setting up your IDE
+
+For IntelliJ users set up the following:
+
+Imports:
+- Preferences > Editor > Code Style > Go > select "Imports" tab
+  - uncheck "Use back quotes for imports"
+  - uncheck "Add parentheses for a single import"
+  - uncheck "Remove redundant import aliases"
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -123,6 +123,49 @@ system which can be installed on macOS with `brew install protobuf`.
...
+- Install the "File Watchers" plugin
+- Preferences > Tools > File Watchers
+  - Add a new file watcher choosing "go fmt"
+  - Depending on your personal preference check "Auto-save edited files to trigger the watcher"
+- Now saving the file (ctl+s) will format the file.
+- References: https://blog.jetbrains.com/go/2021/01/12/running-go-fmt-on-save/
+
+Visual Line Guides: 
+- Preferences > Editor > Code Style > select "General" tab
+  - Hard wrap at: 120
```
soumyadeep2007, line 165: I would recommend hard wrapping at 80 lines, which is the default, and we want to be inclusive to developers who have smaller screens. Also, at 80 lines, it is easier to look at split diffs without having to scroll as much.

soumyadeep2007 (approved):
Awesome! Thanks for this!

README.md
kalensk: Im removing that particular section for now. (I'd like to avoid too much controversy and get the other settings in which will be extremely helpful for new members, etc.)

PR #570, opened 2021-07-15
hub: ensure agentConns are ready

kalensk:
Centralize the location of ensuring agentConns are ready inside step.Begin as this ensures the s.agentConns member variable is not nil and safe to use.

However, having the s.agentConns inside the step framework indicates heavy tech debt which needs to be addressed. However, for the time being centrally call it here so that subsequent calls to s.agentConns are populated and updated.

Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:agentConns

### DIFF
--- a/hub/archive_log_directory.go
+++ b/hub/archive_log_directory.go
@@ -9,8 +9,8 @@ import (
 	"github.com/greenplum-db/gpupgrade/idl"
 )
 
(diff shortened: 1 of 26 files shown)


kalensk:
Hmmmm pipeline is showing quite a few gRPC after making this refactor. Will need to take a closer look. There is a possibility we may need to punt on this if its too involved.

PR #573, opened 2021-07-20
ci: refactors

kalensk:
Clean up CI scripts and make them more uniform. Various other refactors and cleanups, etc.

Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:ciRefactors


ci/template.yml
```diff
@@ -433,6 +433,7 @@ jobs:
         - name: rpm_enterprise
         - name: gpupgrade_src
         - name: gpdb_src_source
+        #FIXMEMEMEMEME
```
soumyadeep2007: There's a fixme here?

ci/scripts/install-tests.bash
```diff
@@ -5,64 +5,82 @@
...
+main() {
+    echo "Installing BATS..."
+    ./bats/install.sh /usr/local
 
-echo 'Running data migration scripts to ensure clean cluster...'
-su gpadmin gpupgrade_src/data-migration-scripts/gpupgrade-migration-sql-generator.bash "$GPHOME_SOURCE" "$PGPORT" /tmp/migration gpupgrade_src/data-migration-scripts
-su gpadmin gpupgrade_src/data-migration-scripts/gpupgrade-migration-sql-executor.bash "$GPHOME_SOURCE" "$PGPORT" /tmp/migration/pre-initialize || true
+    echo "Setting up gpadmin user..."
+    mkdir -p gpdb_src
+    ./gpdb_src_source/concourse/scripts/setup_gpadmin_user.bash "centos"
```
soumyadeep2007, line 42: We lost `time` here. Not sure if it's important since the setup should be pretty fast.

soumyadeep2007 (approved):
LGTM! It was super easy to review with the separated commits.

ci/scripts/install-tests.bash
kalensk: Yeah this step is pretty fast. I think `time` is only kinda useful for fairly long and interesting steps such as the upgrade process, and running the tests. It also makes the CI debugging output a bit easier to read when its only used strategically.

ci/template.yml
kalensk: Removed from that commit!

ci/scripts/install-tests.bash
```diff
@@ -8,8 +8,6 @@ set -ex
 # Install BATS
 ./bats/install.sh /usr/local
 
-source gpupgrade_src/test/acceptance/helpers/helpers.bash
-
```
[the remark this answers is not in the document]
kalensk: Updated commit `ci: reword or remove unnecessary comments` with:

> Remove the need to source gpupgrade_src/test/acceptance/helpers/helpers.bash as we now use data migration scripts instead of particular SQL. This allows us to no longer need functions like is_GPDB5 defined in helpers.bash.

PR #576, opened 2021-07-26
ci: use rsync instead of scp to copy BATS to mdw

kalensk:
Use rsync to preserve symlinks when copying BATS to mdw to install for multi-host testing. This was causing the following error on the CI when BATS bumped their version from 1.3.0 to 1.4.1:

> + scp -rpq bats 'centos@mdw:~'
bats/test/fixtures/parallel/suite/helper.bash: No such file or directory
bats/test/fixtures/parallel/setup_file/helper.bash: No such file or directory

Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:fixBATS


orhankislal (approved):
LGTM

PR #578, opened 2021-07-27
ci: use released bucket for PostGIS

kalensk:
Now that postgis has been released pull from the released prod bucket.

Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:updatePostgisBuckets


soumyadeep2007 (approved):
Can we add the fact that Postgis-2.1.5 was re-released to the commit message?

"Now that Postgis-2.1.5 has been re-released pull from the released prod bucket."

kalensk:
> Can we add the fact that Postgis-2.1.5 was re-released to the commit message?
> "Now that Postgis-2.1.5 has been re-released pull from the released prod bucket."

Updated!

PR #580, opened 2021-07-30
 ci: add pxf and non-gpdb extensions to extensions job

kalensk:
This adds pxf to the extensions job as well as a few other non-gpdb extensions.

Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:quotes

### DIFF
--- a/ci/scripts/ci-helpers.bash
+++ b/ci/scripts/ci-helpers.bash
@@ -16,25 +16,25 @@ is_GPDB5() {
 #    so change it to escape to match GPDB 5 representation
 configure_gpdb_gucs() {
     local gphome=$1
-    ssh mdw bash <<EOF
+    ssh -n mdw "
         set -eux -o pipefail
 
         source ${gphome}/greenplum_path.sh
         export MASTER_DATA_DIRECTORY=/data/gpdata/master/gpseg-1
         gpconfig -c bytea_output -v escape
         gpstop -u
-EOF
+"
 }
 
 reindex_all_dbs() {
     local gphome=$1
-    ssh mdw bash <<EOF
+    ssh -n mdw "
         set -eux -o pipefail
 
         source ${gphome}/greenplum_path.sh
         export MASTER_DATA_DIRECTORY=/data/gpdata/master/gpseg-1
         reindexdb -a
-EOF
+"
 }
 
 dump_sql() {
@@ -73,7 +73,7 @@ compare_dumps() {
             target_dump="$target_dump.filtered"
 
             # Run the filter on the source dump
-            ssh mdw "
+            ssh -n mdw "
                 /tmp/filter -version=5 -inputFile='$source_dump' > '$source_dump.filtered'
             "
 
--- a/ci/scripts/ci-helpers.bash
+++ b/ci/scripts/ci-helpers.bash
@@ -11,6 +11,12 @@ is_GPDB5() {
     [[ $version =~ ^"postgres (Greenplum Database) 5." ]]
 }
 
+# Only test pxf when not centos6 since pxf5 for GPDB6 on centos6 is not supported.
+test_pxf() {
+    local os_version=$1
+    [ "$os_version" != "centos6" ]
+}
+
 # set the database gucs
 # 1. bytea_output: by default for bytea the output format is hex on GPDB 6,
 #    so change it to escape to match GPDB 5 representation
--- a/ci/scripts/load-extensions.bash
+++ b/ci/scripts/load-extensions.bash
@@ -52,6 +52,66 @@ SQL_EOF
 SQL_EOF
 "
 
+echo "Installing postgres native extensions and sample data on source cluster..."
+time ssh -n mdw "
+    set -eux -o pipefail
+
+    source /usr/local/greenplum-db-source/greenplum_path.sh
+
+    echo 'Installing amcheck...'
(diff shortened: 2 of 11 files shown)


ci/scripts/load-extensions.bash
```diff
@@ -50,6 +52,69 @@ SQL_EOF
...
+
+
+            CREATE EXTERNAL TABLE pxf_parquet_read (id INTEGER, name TEXT, cdate DATE, amt DOUBLE PRECISION, grade TEXT,
+                                                b BOOLEAN, tm TIMESTAMP WITHOUT TIME ZONE, bg BIGINT, bin BYTEA,
+                                                sml SMALLINT, r REAL, vc1 CHARACTER VARYING(5), c1 CHARACTER(3),
+                                                dec1 NUMERIC, dec2 NUMERIC(5,2), dec3 NUMERIC(13,5), num1 INTEGER)
+                LOCATION ('pxf://gpupgrade-intermediates/extensions/pxf_parquet_types.parquet?PROFILE=gs:parquet&SERVER=google')
+                FORMAT 'CUSTOM' (FORMATTER='pxfwritable_import');
+            CREATE TABLE pxf_parquet_read_materialized AS SELECT * FROM pxf_parquet_read;
+SQL_EOF
```
soumyadeep2007, line 170: This is me being paranoid:

It might be worthwhile but this file lists objects created by pxf:

https://github.com/greenplum-db/pxf/blob/master/external-table/pxf--1.0.sql

Now, `pxf_write`, `pxf_read` and `pxf_validate` are exercised I believe when we
create/read/write to an external table (as they are part of the protocol..see th
 `CREATE PROTOCOL` statement). Double
check `pxf_validate`.

I am not sure how `pxfwritable_import` and `pxfwritable_export` are exercised, if
at all with the DDL and DML commands that we are running in our tests. It might
be worth asking the PXF team. It might be good to exercise them in some way for
more coverage.

Alternatively, one may inspect their function signatures to know if they had
any changes between versions of 5x and 6x (especially the internal backing functions:
`gpdbwritableformatter_import` and `gpdbwritableformatter_export`)

This is not a blocker for this PR but is worth looking into.

ci/scripts/load-extensions.bash
```diff
@@ -52,6 +52,66 @@ SQL_EOF
...
+          FROM pg_index i
+          JOIN pg_opclass op ON i.indclass[0] = op.oid
+          JOIN pg_am am ON op.opcmethod = am.oid
+          JOIN pg_class c ON i.indexrelid = c.oid
+          JOIN pg_namespace n ON c.relnamespace = n.oid
+          WHERE am.amname = 'btree' AND n.nspname = 'pg_catalog'
+            -- Function may throw an error when this is omitted:
+            AND i.indisready AND i.indisvalid
+          ORDER BY c.relpages DESC LIMIT 10;
+SQL_EOF
```
soumyadeep2007, line 76: Nit: I would like to see the code segregate non-gpdb extensions that are shipped with GP vs not shipped. So amcheck should be kept separate?

Also, consider using "Postgres native extension" vs "non-gpdb extension"? I think we should revisit naming - the code here should reflect the terminology in the docs. I don't know if we should call it non-gpdb extension in the docs.

ci/scripts/ci-helpers.bash
```diff
@@ -16,25 +16,25 @@ is_GPDB5() {
 #    so change it to escape to match GPDB 5 representation
 configure_gpdb_gucs() {
     local gphome=$1
-    ssh mdw bash <<EOF
+    ssh -n mdw "
```
soumyadeep2007, line 25: Can we add a comment to the first commit explaining what `-n` gives us? The ssh man page wasn't very clear to me.

soumyadeep2007 (approved):
LGTM. Some food for thought on nomenclature and PXF test coverage.

ci/scripts/load-extensions.bash
kalensk: Yeah, we will definitely look into this with pxf team as a followup.

ci/scripts/load-extensions.bash
kalensk: Will update references to  "Postgres native extension".

ci/scripts/ci-helpers.bash
kalensk: I believe it allows us to do the following psql heredoc construct within an ssh call:

```
        psql -d postgres <<SQL_EOF
            CREATE EXTENSION pxf;
SQL_EOF
```

ci/scripts/load-extensions.bash
bradfordb-vmware: `pxfwritable_import` and `pxfwritable_export` are given as options in the external table DDL, e.g. `FORMAT 'CUSTOM' (FORMATTER='pxfwritable_import')` for the case of a readable external table. So the current tests you have are covering `pxfwritable_import`; if you wanted to cover `pxfwritable_export` you would need to create a writable table.

kalensk: It was decided offline that creating a writeable table is quite a decent amount of additional work for somewhat little test coverage. That is, if readable works, then most likely writeable works as these upgrade tests are mostly "smoke tests" in a sense.

PR #581, opened 2021-07-30
ci: fixups for extensions

kalensk:
Missed these few minor issues:
- When loading pxf data source greenplum_path.sh so that psql can be found.
- After upgrading pxf allow the text_pxf function on the remote machine, so it can be used.

Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:extensionFixups


soumyadeep2007 (approved):
LGTM. The commit message can slightly be reworded to say pxf instead of "extensions".

PR #583, opened 2021-08-16
GPText support

RMTT:
Migrate gptext when doing gpupgrade.

### DIFF
--- a/ci/generated/template.yml
+++ b/ci/generated/template.yml
@@ -201,6 +201,17 @@ resources:
 {{- end }}
 {{end}}
 
+# gptext3.7 supports centos 6/7 and gpdb 5/6
+- name: gptext-3.7.0
+  type: s3
+  source:
+    access_key_id: ((gptext-s3-access-key-id))
+    secret_access_key: ((gptext-s3-secret-access-key))
+    region_name: us-west-2
+    bucket: gptext-artifacts
+    versioned_file: greenplum-text-3.7.0-rhel6_x86_64.tar.gz
+
+
 anchors:
   - &ccp_default_params
     action: create
@@ -621,6 +632,10 @@ jobs:
           resource: madlib1190_gppkg_gpdb{{.Source}}_centos{{.CentosVersion}}
         - get: madlib_gppkg_target
           resource: madlib1190_gppkg_gpdb{{.Target}}_centos{{.CentosVersion}}
+        - get: gptext_pkg_source
+          resource: gptext-3.7.0
+        - get: gptext_pkg_target
+          resource: gptext-3.7.0
         {{- if ne .CentosVersion "6" }}
         - get: pxf_rpm_source
           resource: pxf5_gp{{.Source}}_centos{{.CentosVersion}}_rpm
@@ -707,6 +722,7 @@ jobs:
           - name: gpupgrade_src
           - name: postgis_gppkg_source
           - name: madlib_gppkg_source
+          - name: gptext_pkg_source
           - name: sqldump
           {{- if ne .CentosVersion "6" }}
           - name: pxf_rpm_source
@@ -747,6 +763,7 @@ jobs:
           - name: gpupgrade_src
           - name: postgis_gppkg_target
           - name: madlib_gppkg_target
+          - name: gptext_pkg_target
           {{- if ne .CentosVersion "6" }}
           - name: pxf_rpm_target
           {{- end }}
--- a/ci/generated/template.yml
+++ b/ci/generated/template.yml
@@ -205,8 +205,8 @@ resources:
 - name: gptext-3.7.0
   type: s3
   source:
-    access_key_id: ((gptext-s3-access-key-id))
-    secret_access_key: ((gptext-s3-secret-access-key))
+    access_key_id: ((gptext_s3_access_key_id))
+    secret_access_key: ((gptext_s3_secret_access_key))
     region_name: us-west-2
     bucket: gptext-artifacts
     versioned_file: greenplum-text-3.7.0-rhel6_x86_64.tar.gz
--- a/ci/generated/template.yml
+++ b/ci/generated/template.yml
@@ -205,8 +205,8 @@ resources:
 - name: gptext-3.7.0
   type: s3
   source:
-    access_key_id: ((gptext_s3_access_key_id))
-    secret_access_key: ((gptext_s3_secret_access_key))
+    access_key_id: ((gptext-s3-access-key-id))
+    secret_access_key: ((gptext-s3-secret-access-key))
     region_name: us-west-2
     bucket: gptext-artifacts
     versioned_file: greenplum-text-3.7.0-rhel6_x86_64.tar.gz
--- a/ci/scripts/load-extensions.bash
+++ b/ci/scripts/load-extensions.bash
@@ -113,26 +113,6 @@ SQL_EOF
 SQL_EOF
 "
 
-echo "Installing gptext 3.7.0 on source cluster..."
-time ssh -n mdw "
-export MASTER_DATA_DIRECTORY=$MASTER_DATA_DIRECTORY
-source /usr/local/greenplum-db-source/greenplum_path.sh
-
-tar zxvf /tmp/greenplum-text-3.7.0-rhel6_x86_64.tar.gz -C /tmp/
-chmod +x /tmp/greenplum-text-3.7.0-rhel6_x86_64.bin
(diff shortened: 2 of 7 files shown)


ci/scripts/load-extensions.bash
```diff
@@ -18,6 +18,7 @@ echo "Copying extensions to the source cluster..."
 scp postgis_gppkg_source/postgis*.gppkg gpadmin@mdw:/tmp/postgis_source.gppkg
 scp sqldump/*.sql gpadmin@mdw:/tmp/postgis_dump.sql
 scp madlib_gppkg_source/madlib*.gppkg gpadmin@mdw:/tmp/madlib_source.gppkg
+scp gptext_pkg_source/greenplum-text-3.7.0-rhel6_x86_64.tar.gz gpadmin@mdw:/tmp/greenplum-text-3.7.0-rhel6_x86_64.tar.gz
```
kalensk, line 21: @RMTT 

Can you follow the convention of all the other extensions and abstract out the version number by using source and target. Thus, copy this to `gpadmin@mdw:/tmp/gptext_source.tar.gz`

Full line would look something like:
`scp gptext_source_gppkg/*.tar.gz gpadmin@mdw:/tmp/gptext_source.tar.gz`

Can we also rename the resource to match its type (its not a gppkg). That is, something like `gptext_source_tar`?

ci/scripts/upgrade-extensions.bash
```diff
@@ -21,6 +21,7 @@ export PGPORT=5432
 echo "Copying extensions to the target cluster..."
 scp postgis_gppkg_target/postgis*.gppkg gpadmin@mdw:/tmp/postgis_target.gppkg
 scp madlib_gppkg_target/madlib*.gppkg gpadmin@mdw:/tmp/madlib_target.gppkg
+scp gptext_pkg_target/greenplum-text-3.7.0-rhel6_x86_64.tar.gz gpadmin@mdw:/tmp/greenplum-text-3.7.0-rhel6_x86_64.tar.gz
```
kalensk, line 24: @RMTT  similar comment to one above. Is there a reason we are not following the lines above with *.tar.gz and calling it gptext_target.tar.gz, etc. And naming the resource appropriately based on its actual type?

Something like:
`scp gptext_target_tar/*.tar.gz gpadmin@mdw:/tmp/gptext_target.tar.gz`

ci/scripts/upgrade-extensions.bash
```diff
@@ -143,6 +174,10 @@ ssh -n mdw "
         echo 'Starting pxf...'
         /usr/local/pxf-gp6/bin/pxf cluster start
     fi
+
+    source /usr/local/greenplum-db-text-3.7.0/greenplum-text_path.sh
+    echo 'gptext: start gptext...'
+    gptext-start
```
kalensk, line 180: @RMTT 

Since the version is the same for source and target let's use abstraction and use a `*` instead. That way, the scripts will still work as the version changes.

 So, `/usr/local/greenplum-db-text-*/greenplum-text_path.sh`. Update accordingly through all files.

ci/scripts/upgrade-extensions.bash
```diff
@@ -93,6 +94,36 @@ SQL_EOF
...
     fi
 
+    echo 'Migrating gptext 3.7.0 to target cluster...'
+    source /usr/local/greenplum-db-text-3.7.0/greenplum-text_path.sh
+    echo 'gptext: check gpdb version compatibility...'
+    if [ ! -f \$GPTXTHOME/lib/gptext-gpdb6-* ]
+    then
+        echo 'This version of gptext does not support gpdb6'
+        exit
+    fi
```
kalensk, line 104: @RMTT  This if statement is not needed since we have `set -e`. We should not even be testing with any version for gptext that is not supported for gpdb6... And will fail in other ways if we encounter it. Let's keep the code clean and simple.

ci/scripts/upgrade-extensions.bash
```diff
@@ -93,6 +94,36 @@ SQL_EOF
...
+    then
+        cp $GPHOME_SOURCE/lib/postgresql/gptext-*.so $GPHOME_TARGET/lib/postgresql/
+    fi
+
+    echo 'gptext: copy config files...'
+    cp $MASTER_DATA_DIRECTORY/{gptext.conf,gptxtenvs.conf,zoo_cluster.conf} \$MASTER_DATA_DIRECTORY/
+    if [ -e $MASTER_DATA_DIRECTORY/gptxtauth.conf ]
+    then
+        cp $MASTER_DATA_DIRECTORY/gptxtauth.conf \$MASTER_DATA_DIRECTORY/
+    fi
```
kalensk, line 118: @RMTT 
Let's try and really think through ways of not having all these if statements and  really simplifying this block of code. Again look at the surrounding code and see if we can match it in some sense.

ci/scripts/upgrade-extensions.bash
```diff
@@ -93,6 +94,36 @@ SQL_EOF
...
+
+    echo 'gptext: restart gptext...'
+    gptext-stop
+    gptext-start
+
+    echo 'gptext: stop gptext...'
+    gptext-stop
+
```
kalensk, line 126: I don't think this restart and stopping is needed. Let's really verify that once we get the upgrading of gptext working...See if we can make it simple. If needed, add a comment why restarting gptext is needed. I don't think we need to stop gptext to proceed with the upgrade?

ci/scripts/load-extensions.bash
```diff
@@ -179,6 +180,33 @@ SQL_EOF
...
+chmod +x /tmp/greenplum-text-3.7.0-rhel6_x86_64.bin
+sed -i -r 's/GPTEXT_HOSTS\=\(localhost\)/GPTEXT_HOSTS\=\"ALLSEGHOSTS\"/' /tmp/gptext_install_config
+sed -i -r 's/ZOO_HOSTS.*/ZOO_HOSTS\=\(mdw sdw1 sdw1\)/' /tmp/gptext_install_config
+
+/tmp/greenplum-text-3.7.0-rhel6_x86_64.bin -c /tmp/gptext_install_config -d /usr/local/greenplum-db-text-3.7.0
+createdb demo
+source /usr/local/greenplum-db-text-3.7.0/greenplum-text_path.sh
+gptext-installsql demo
+gptext-start
+"
```
kalensk, line 208: Please move this block inside the `echo "Installing extensions and sample data on source cluster..."` block. And update the echo to specify that they are gpdb extensions. `echo "Installing gpdb extensions and sample data on source cluster..."`

ci/scripts/load-extensions.bash
```diff
@@ -179,6 +180,33 @@ SQL_EOF
...
+        sudo chown gpadmin:gpadmin /usr/local/greenplum-db-text-3.7.0
+        sudo chown gpadmin:gpadmin /usr/local/greenplum-solr
+    "
+done
+time ssh -n mdw "
+export MASTER_DATA_DIRECTORY=$MASTER_DATA_DIRECTORY
+source /usr/local/greenplum-db-source/greenplum_path.sh
+
+tar zxvf /tmp/greenplum-text-3.7.0-rhel6_x86_64.tar.gz -C /tmp/
+chmod +x /tmp/greenplum-text-3.7.0-rhel6_x86_64.bin
```
kalensk, line 199: @RMTT 
Is there any way to have the resource push the .bin directly rather than a tar.gz so this step is not needed? That would make way more sense to me, and greatly simplify multiple downstream steps.

ci/scripts/load-extensions.bash
```diff
@@ -179,6 +180,33 @@ SQL_EOF
 
 test_pxf "$OS_VERSION" && install_pxf || echo "Skipping pxf for centos6 since pxf5 for GPDB6 on centos6 is not supported..."
 
+echo "Installing gptext 3.7.0 on source cluster..."
+mapfile -t hosts < cluster_env_files/hostfile_all
+for host in "${hosts[@]}"; do
+    ssh -n "centos@${host}" "
+        sudo yum install -y java-1.8.0-openjdk
```
kalensk, line 187: Let's pass `-q` as well like we do when installing pxf dependencies in this file.

And let's remove the `. x86_64` from the java line when installing pxf dependencies to match this line for gptext...

ci/generated/template.yml
```diff
@@ -201,6 +201,17 @@ resources:
 {{- end }}
 {{end}}
 
+# gptext3.7 supports centos 6/7 and gpdb 5/6
```
kalensk, line 204: Let's use `&` instead of `/` for clarity. Thus, something like:
`gptext3.7 supports both centos 6 & 7, and gpdb 5 & 6`

PR #586, opened 2021-08-23
rsync: refactor gRPC to be simple pass thru

kalensk:
To keep things clear and consistent and easy simply have the gRPC call for rsync match the actual rsync parameters. This eliminates the need for RsyncPairs at the cost of minor duplication of Options and ExcludedFiles. This overhead is very minor compared to the maintainability and extensibility of the code.

When reviewing please look for the following:
1) That callers correctly use  string(os.PathSeparator) since this is no longer down on the agent. Rather it is the callers responsibility to pass the correct parameters, and downstream code simple calls rsync directly. All business logic should be done on the hub when constructing the rsync request.
2) That the logic for `upgrade.VerifyDataDirectory` refactor which now takes a list is correct.

Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:refactorRsync


upgrade/directories.go
```diff
@@ -157,12 +157,14 @@ func (i *InvalidDataDirectoryError) Is(err error) bool {
...
+func VerifyDataDirectory(path ...string) error {
 	var err error
 
 	for _, f := range PostgresFiles {
-		if !PathExists(filepath.Join(path, f)) {
-			err = errorlist.Append(err, &InvalidDataDirectoryError{path, f})
+		for _, p := range path {
+			if !PathExists(filepath.Join(p, f)) {
+				err = errorlist.Append(err, &InvalidDataDirectoryError{p, f})
+			}
```
kalensk, line 167: Please double check the logic here..

agent/rsync.go
```diff
@@ -54,21 +53,21 @@ func (s *Server) RsyncTablespaceDirectories(ctx context.Context, in *idl.RsyncRe
...
-				rsync.WithSources(pair.GetSource() + string(os.PathSeparator)),
-				rsync.WithDestinationHost(pair.GetDestinationHost()),
-				rsync.WithDestination(pair.GetDestination()),
-				rsync.WithOptions(in.GetOptions()...),
-				rsync.WithExcludedFiles(in.GetExcludes()...),
+				rsync.WithSources(opts.GetSources()...),
+				rsync.WithDestinationHost(opts.GetDestinationHost()),
+				rsync.WithDestination(opts.GetDestination()),
+				rsync.WithOptions(opts.GetOptions()...),
+				rsync.WithExcludedFiles(opts.GetExcludedFiles()...),
```
kalensk, line 70: Please double check the callers to this are now passing in `os.PathSeparator`. That is, all "business logic" should be done on the hub when constructing the gRPC request, and the agent is just a simple pass through.

orhankislal: I checked both of the functions calling this logic (RsyncDataDirectories, RsyncTablespaceDirectories). All of their references seem to have `os.PathSeparator`.

upgrade/directories.go
orhankislal: LGTM

orhankislal (approved):
LGTM. Just a minor observation.

soumyadeep2007 (approved):
LGTM

idl/hub_to_agent.pb.go
```diff
@@ -1020,7 +1020,7 @@ func (m *RsyncReply) Reset()         { *m = RsyncReply{} }
 func (m *RsyncReply) String() string { return proto.CompactTextString(m) }
 func (*RsyncReply) ProtoMessage()    {}
 func (*RsyncReply) Descriptor() ([]byte, []int) {
-	return fileDescriptor_9e73bb06acc917d8, []int{21}
+	return fileDescriptor_9e73bb06acc917d8, []int{20}
```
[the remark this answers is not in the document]
kalensk: `*pb.go` files are auto-generated gRPC files. It is safe to ignore them.

PR #587, opened 2021-08-27
Refactor: unify resource names and remove generated files

orhankislal:
This commit updates resources (and their respective usages in
scripts) to conform to the following format:
productName_1.2.3_gpdb{{.GPVersion}}_centos{{.CentosVersion}}_rpm/pkg/targz

It also removes generated files to prevent committing secret keys
accidentally.

Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:refactor-resource-names


kalensk:
If feasible, I would break this into two commits - 1) refactor resource names, 2) remove generated CI files.

kalensk:
Not for this PR, but it seems like we might be able to remove this confusing indirection in the Makefile when calling `set-pipeline`? That is, replace 
`go generate ./ci`
with
`go run ./parser/parse_template.go generated/template.yml generated/pipeline.yml`

kalensk (approved):

PR #588, opened 2021-08-28
use IntermediateTarget and remove InitializeConfig

kalensk:
Create the following cluster objects: Source, IntermediateTarget, and Target.

This is far more accurate and readable as we use intermediateTarget cluster up until we swap the intermediateTarget with the source when updating the catalog and on disk data directories. This, at any point in time callers can refer to the correct data directory and port. Visually this is represented as something like:

**Source Cluster:**
dataDir:  '/data/dbfast1/demoDataDir0'

**IntermediateTarget:**
dataDir: '/data/dbfast1/demoDataDir.123ABC.0'  
which gets renamed to  '/data/dbfast1/demoDataDir.123ABC.0.old, which we don't need to store as its simply  dataDir + oldSuffix.

**Target:**
dataDir:  '/data/dbfast1/demoDataDir0'

**InitializeConfig:** 
removed in favor of IntermediateTarget using the greenplum.Cluster data type.

Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:sourceTargetIntermediateTarget

### DIFF
--- a/greenplum/cluster_test.go
+++ b/greenplum/cluster_test.go
@@ -278,8 +278,8 @@ func TestClusterFromDB(t *testing.T) {
 		if err == nil {
 			t.Errorf("Expected an error, but got nil")
 		}
-		if actualCluster != nil {
-			t.Errorf("Expected cluster to be nil, but got %#v", actualCluster)
+		if !reflect.DeepEqual(actualCluster, greenplum.Cluster{}) {
+			t.Errorf("got: %#v want empty cluster: %#v", actualCluster, &greenplum.Cluster{})
 		}
 		if !strings.Contains(err.Error(), connErr.Error()) {
 			t.Errorf("Expected error: %+v got: %+v", connErr.Error(), err.Error())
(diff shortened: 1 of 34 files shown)


hub/copy_master.go
```diff
@@ -78,21 +80,21 @@ func Copy(streams step.OutStreams, destinationDir string, sourceDirs, hosts []st
...
 	}
 
 	// include tablespace mapping file which is used as a parameter to pg_upgrade
-	sourcePaths := []string{s.TablespacesMappingFilePath}
-	sourcePaths = append(sourcePaths, s.Tablespaces.GetMasterTablespaces().UserDefinedTablespacesLocations()...)
+	sourcePaths := []string{TablespacesMappingFilePath}
+	sourcePaths = append(sourcePaths, tablespaces.GetMasterTablespaces().UserDefinedTablespacesLocations()...)
 
-	return Copy(streams, destinationDir, sourcePaths, s.IntermediateTarget.PrimaryHostnames())
+	return Copy(streams, destinationDir+string(os.PathSeparator), sourcePaths, hosts)
```
kalensk, line 99: Please double check that the `os.PathSeparator` got correctly used/converted in this refactor for **_both_** `CopyMasterTablespaces` and `CopyMasterDataDir`

greenplum/cluster_test.go
```diff
@@ -278,8 +278,8 @@ func TestClusterFromDB(t *testing.T) {
 		if err == nil {
 			t.Errorf("Expected an error, but got nil")
 		}
-		if actualCluster != nil {
-			t.Errorf("Expected cluster to be nil, but got %#v", actualCluster)
+		if !reflect.DeepEqual(actualCluster, greenplum.Cluster{}) {
+			t.Errorf("got: %#v want empty cluster: %#v", actualCluster, &greenplum.Cluster{})
```
orhankislal, line 282: We might want to keep it in the previous format since the other two errors start with `Expected` as well.

orhankislal (approved):

greenplum/cluster_test.go
kalensk: Just as a note the old code follows the wrong convention as its usually got then want such as `t.Errorf("got: %a want %q", actual, expected)`

PR #589, opened 2021-08-30
Refactor: Remove deprecated PathExists function

orhankislal:
In directories.go, PathExists function has been deprecated and
a PathExist function has been added prior to this commit. This
commit removes the references to that function and replaces
its usage with PathExist in proper format.

Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:refactor-path-exists

### DIFF
--- a/agent/server_test.go
+++ b/agent/server_test.go
@@ -14,6 +14,7 @@ import (
 	"github.com/greenplum-db/gpupgrade/agent"
 	"github.com/greenplum-db/gpupgrade/testutils"
 	"github.com/greenplum-db/gpupgrade/testutils/testlog"
+	"github.com/greenplum-db/gpupgrade/upgrade"
 )
 
 func TestServerStart(t *testing.T) {
@@ -29,14 +30,12 @@ func TestServerStart(t *testing.T) {
 			StateDir: stateDir,
 		})
 
-		if pathExists(stateDir) {
-			t.Fatal("expected stateDir to not exist")
-		}
+		testutils.PathMustNotExist(t, stateDir)
 
 		go server.Start()
 		defer server.Stop()
 
-		exists, err := doesPathEventuallyExist(stateDir)
+		exists, err := doesPathEventuallyExist(t, stateDir)
 		if err != nil {
 			t.Fatalf("unexpected error: %#v", err)
 		}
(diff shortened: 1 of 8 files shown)


agent/server_test.go
```diff
@@ -54,25 +53,26 @@ func TestServerStart(t *testing.T) {
...
 	timeout := 3 * time.Second
 
 	for {
-		exists := pathExists(path)
+		exists, err := upgrade.PathExist(path)
+
+		if err != nil {
+			return false, err
+		}
+
```
kalensk, line 75: This needs to be: 

```suggestion
		exists, err := upgrade.PathExist(path)
		if err != nil {
			t.Fatalf("checking path: %v", err)
		}

```

greenplum/cluster.go
```diff
@@ -351,8 +351,9 @@ func runStartStopCmd(stream step.OutStreams, gphome, command string, env string)
 // IsMasterRunning returns whether the cluster's master process is running.
 func (c *Cluster) IsMasterRunning(stream step.OutStreams) (bool, error) {
 	path := filepath.Join(c.MasterDataDir(), "postmaster.pid")
-	if !upgrade.PathExists(path) {
-		return false, nil
+	exist, err := upgrade.PathExist(path)
+	if !exist {
+		return false, err
```
kalensk, line 356: This needs to check the err from PathExists!

```suggestion
	exist, err := upgrade.PathExist(path)
	if err != nil {
		return false, err	
	}

	if !exist {
		return false, err
```

orhankislal: If there is an error in the `PathExists` call, `exist` will be set to `false`. The first if check is already covered by the second one since they return the exact same pair.

kalensk: I'd much rather be verbose and extra clear in this case as it prevents people from needing to think through things like that.

upgrade/directories.go
```diff
@@ -393,8 +407,9 @@ func Verify5XTablespaceDirectories(tsLocations []string) error {
 			}
 
 			path := filepath.Join(tsLocation, dbOidDir.Name(), PGVersion)
-			if !PathExists(path) {
-				mErr = errorlist.Append(mErr, newTablespaceDirectoryError("5X source cluster", "missing "+path))
+			exist, err := PathExist(path)
+			if !exist {
+				mErr = errorlist.Append(mErr, newTablespaceDirectoryError("5X source cluster", "missing "+path), err)
```
kalensk, line 408: Need to check err...

```suggestion
			exist, err := PathExist(path)
			if err != nil {
							mErr = errorlist.Append(mErr, xerrors.Errorf("checking path %q: %w", path, err))
			}
			
			if !exist {
				mErr = errorlist.Append(mErr, newTablespaceDirectoryError("5X source cluster", "missing "+path), err)
```

upgrade/directories.go
```diff
@@ -171,11 +190,6 @@ func VerifyDataDirectory(path ...string) error {
 	return err
 }
 
-// TODO: Remove alreadyRenamed and use AlreadyRenamed
-func alreadyRenamed(archive, target string) bool {
-	return PathExists(archive) && !PathExists(target)
-}
-
```
kalensk, line 178: This is great! Thank you for removing this tech debt :)

kalensk (approved):
Looks good. 

One really minor suggestion (not really needed).

PR #591, opened 2021-09-01
CI: Remove orafce tests

orhankislal:
The orafce extension has been updated in GPDB6 and the versions
are incompatible with each other. This commit removes
its tests since we suggest the users to uninstall/reinstall.

Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:ci-remove-orafce?group=extensions

### DIFF
--- a/ci/scripts/load-extensions.bash
+++ b/ci/scripts/load-extensions.bash
@@ -87,29 +87,20 @@ SQL_EOF
 SQL_EOF
 
     echo 'Installing hstore...'
-    psql -d postgres <<SQL_EOF
+    psql -v ON_ERROR_STOP=1 -d postgres <<SQL_EOF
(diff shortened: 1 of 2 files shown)


kalensk:
Can you provide more detail for this PR and maybe commit message as to the exact issue? At first glance orafce does seems to work since the pipeline is green. That is, what specific error or issues were being encountered to reveal this incompatibility? Do we need to increase our testing?

orhankislal:
The pipeline is green because the errors in the psql commands that we run via ssh do not propagate. 
If you check the concourse logs, you'll see that orafce install commands are failing. 
I am trying to figure out why since the return code is non 0 in my local tests (even with `time`).

kalensk (approved):
Good catch / investigation on finding the `-v ON_ERROR_STOP=1` construct.

PR #563, opened 2021-06-16
Non-upgradeable test for views referencing deprecated columns

soumyadeep2007:
This tests the functionality added to pg_upgrade --check in order to
flag views referencing deprecated columns as non-upgradeable. This
functionality was added in GPDB PRs:
greenplum-db/gpdb#12168
greenplum-db/gpdb#12234
greenplum-db/gpdb#12169

PR #592, opened 2021-09-12
Improve data migration scripts for deprecated columns

orhankislal:
tsquery and name column types are deprecated in GPDB6 and
have to be changed to VARCHAR temporarily for gpupgrade.
This commit improves the pre- and post- scripts to
handle following cases:

- Index on column
- Clustering on index
- Comment on index
- Inherits on table with column

Co-authored-by: Soumyadeep Chakraborty <soumyadeep2007@gmail.com>

Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:fix-migration-scripts


orhankislal:
Please refrain from reviewing the last two commits on view dependencies since I am working on a more comprehensive approach.

data-migration-scripts/create_find_view_dep_function.sql
```diff
@@ -0,0 +1,120 @@
...
+    for depender, dependee in view2view.iteritems():
+        if dependee in checklist and depender not in checklist:
+            new_checklist[depender] = view_order
+    if len(new_checklist) == 0:
+        break
+    else:
+        checklist.update(new_checklist)
+
+plpy.execute("DROP TABLE IF EXISTS __temp_views_list")
+plpy.execute("CREATE TABLE __temp_views_list (full_view_name TEXT, view_order INTEGER)")
```
soumyadeep2007, line 112: We should create a schema and schema qualify to avoid the outside chance of conflicts.

soumyadeep2007 (commented):
* We should disambiguate the "deprecated columns" terminology. Historically, in gpupgrade, we have referred to deprecated columns as columns that may exist in a catalog table in the source cluster but no longer exists in the target cluster. Refer to PR: #563 for context. So in this PR, lets use "deprecated built-in types" instead. Please update the PR description to say: `Improve data migration scripts for deprecated built-in types`
* Anything I mentioned in the PR for name type columns in the review apply to tsquery type columns as well.
* I noticed that we aren't applying the workaround of altering the column type on the parent table when there is inheritance -> which would lead to corresponding changes in the other scripts involved.
* For the last commit, we need quite a bit of test coverage, covering all the levels of view dependencies.
* In the long run, we might want to think about how we can refactor the last commit so that we can dump and drop all views (and dependent views) that have a specific deprecated property (eg references deprecated function, deprecated table etc). We could call things like `binary_upgrade.view_references_deprecated_table` as an example, as opposed to calling `AND (a.atttypid = 'pg_catalog.tsquery'::pg_catalog.regtype OR a.atttypid = 'pg_catalog.name'::pg_catalog.regtype)`.

data-migration-scripts/post-revert/recreate_indexes_on_deprecated_columns.sql
```diff
@@ -3,6 +3,8 @@
 
 -- generates create index statement to re-create indexes on deprecated types.
 
+SET client_min_messages TO WARNING;
```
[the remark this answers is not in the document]
orhankislal: Not necessary any more thanks to the change in the psql command

data-migration-scripts/test/create_nonupgradable_objects.sql
```diff
@@ -156,5 +159,68 @@ INSERT INTO table_with_tsquery_datatype_columns
...
+    population name,
+    altitude   tsquery
+);
+
+DROP TABLE IF EXISTS name_inherits;
+CREATE TABLE name_inherits (
+    state      char(2)
+) INHERITS (table_with_name_tsquery);
+
+-- view on a view on a name column
```
soumyadeep2007, line 209: Lets add a test case with one more level of recursion.

soumyadeep2007 (approved):
LGTM. Test coverage looks good. Some minor comments still remain about cleanup and commit restructuring, changing the nomenclature of deprecated columns -> deprecated built-in types.

> I noticed that we aren't applying the workaround of altering the column type on the parent table when there is inheritance -> which would lead to corresponding changes in the other scripts involved.

I was wrong about this - the query is filtering out child relations and applying the ALTER on the parent correctly.

PR #594, opened 2021-09-23
Various Refactors

kalensk:
Various refactors. See each commit and their messages for details.

Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:refactors


greenplum/cluster.go
```diff
@@ -182,81 +208,6 @@ func (c Cluster) SelectSegments(selector func(*SegConfig) bool) SegConfigs {
...
-			}
-
-			cluster.Mirrors[content] = seg
-
-		default:
-			return Cluster{}, newInvalidSegmentsError(seg, "unknown role %q", seg.Role)
-		}
-	}
-
-	// Make sure each mirror has a primary.
```
orhankislal, line 225: I don't see a similar check in `WaitForClusterToBeReady`, are we sure we don't need this?

kalensk: Basically, I don't think this is something worth checking. I chatted with @jmcatamney offline and we don't see this being needed, or really how feasible it is to create a cluster where a corresponding primary doesn't have a mirror. Meaning one would really have to "try" to do that.

If needed, we can chat more offline in detail.

As you probably saw, we do call WaitForClusterToBeReady as a sanity check on the source cluster early in initialize when creating the config.

kalensk:
Just added a commit that should help with the test flake. Further analysis will hopefully be done soon with kernel team members.

greenplum/cluster.go
orhankislal: OK, that sounds reasonable.

orhankislal (approved):

PR #595, opened 2021-09-23
cluster: move GreenplumRunner to cluster

kalensk:
See last commit as this is based off https://github.com/greenplum-db/gpupgrade/pull/594

Greenplum runner had its how struct and interface which duplicated much of the information already on the cluster object. Thus, move it to the cluster to remove the need to pass all that information in.

Remove unneeded tests that simple test what we are mocking. That is, the lower level tests already do that.

Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:greenplumRunner


greenplum/cluster_test.go
```diff
@@ -366,6 +369,73 @@ func TestHasAllMirrorsAndStandby(t *testing.T) {
 	}
 }
 
+func TestRunGreenplumCmd(t *testing.T) {
+	testlog.SetupLogger()
+
+	cluster := MustCreateCluster(t, greenplum.SegConfigs{
+		{DbID: 1, ContentID: -1, Hostname: "master", DataDir: "/data/qddir/seg-1", Port: 15432, Role: greenplum.PrimaryRole},
+	})
+	cluster.GPHome = "/usr/local/gpdb6"
```
orhankislal, line 378: Is it possible to use a generic path like `/usr/local/greenplum-db/`?

orhankislal (approved):

greenplum/cluster_test.go
kalensk: Updated.

greenplum/cluster_test.go
```diff
@@ -432,10 +369,77 @@ func TestHasAllMirrorsAndStandby(t *testing.T) {
...
+	cluster.GPHome = "/usr/local/greenplum-db"
+
+	t.Run("executes greenplum utility with greenplum_path.sh set and correct args", func(t *testing.T) {
+		cmd := exectest.NewCommandWithVerifier(Success, func(name string, args ...string) {
+			expected := "bash"
+			if name != expected {
+				t.Errorf("got %q want %q", name, expected)
+			}
+
+			expectedArgs := []string{"-c", "source /usr/local/gpdb6/greenplum_path.sh && /usr/local/gpdb6/bin/gpaddmirrors -a -i mirrors_config --hba-hostnames"}
```
orhankislal, line 387: This line is still using `gpdb6`

Issue #582, opened 2021-08-06
Error during initialize phase


macbethDGPH:
There is some unrecognized error which occurred during init phase of gpupgrade.
All my attempts to find out what does it mean were failed.
Can I get some help (first of all, what does "exit status 127" mean)?  
```
$ gpupgrade initialize -avf gpupgrade_config 

Initialize in progress.

Initialize took 9ms

Error: could not determine source cluster version: exit status 127

NEXT ACTIONS
------------
Please address the above issue and run "gpupgrade initialize" again.
```

kalensk:
There are logs in the following places that will reveal more information:
- /home/gpadmin/gpupgrade/...
- /home/gpadmin/...
- $MASTER_DATA_DIRECTORY/pg_log

Performing a Google search will reveal that exit status 127 means:
> Value 127 is returned by /bin/sh when the given command is not found within your PATH system variable and it is not a built-in shell command. In other words, the system doesn't understand your command, because it doesn't know where to find the binary you're trying to call.

See [here](https://stackoverflow.com/a/1763178) for example.

Those logs will probably reveal that the following command is attempting to be executed:
- Locally on the coordinator host `$GPHOME_SOURCE/bin/postgres --gp-version`
- And remotely on each host with `ssh -q <host> bash -c $GPHOME_SOURCE/bin/postgres --gp-version`

My suspicion is that the `source_gphome` set in the config file is incorrect, or not in the PATH for all hosts such that postgres is not being found. Alternatively, `ssh` might not be found on your system either.

Hope this helps!

macbethDGPH:
Well, thank you for the answer
This is a version of the unitity
```
gpupgrade --version
Version: 1.1.0
Commit: b615d6f
Release: Open Source
```

> My suspicion is that the source_gphome set in the config file is incorrect, or not in the PATH for all hosts such that postgres is 
> not being found. Alternatively, ssh might not be found on your system either.
Your suspicion is wrong, cause if `source_gphome` was incorrect, than `gpupgrade` output would be some like that `could not determine source cluster version: fork/exec /path/to/the/postgres: no such file or directory`

kalensk:
Well without knowing more information its pretty difficult to help. But that error is coming from when gpupgrade checks the minimum versions of GPDB as [indicated here](https://github.com/greenplum-db/gpupgrade/blob/039e97551b4b7057b324be6a574ca04d278caf9b/greenplum/allowed_gpdb_versions.go#L107-L109).

[...]

PR #599, opened 2021-10-11
Remove greenplum version check

orhankislal:
This commit removes the step in which we check if the
greenplum versions are matching between coordinator and
segments. The check was done via ssh which was prone to
failure as described in the following issue:
https://github.com/greenplum-db/gpupgrade/issues/582

We still use the ssh based checks for gpupgrade version.
We cannot use grpc since this check needs to happen before
agents are brought up and a grpc call needs them running.

Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:remove-gpdb-version-check


greenplum/version_test.go
```diff
@@ -114,107 +113,10 @@ func TestVersion_Parsing(t *testing.T) {
...
-		var exitError *exec.ExitError
-		if !errors.As(err, &exitError) {
-			t.Errorf("got %T, want %T", err, exitError)
-		}
-
-		if version != "" {
-			t.Errorf("unexpected version %q", version)
-		}
-	})
-}
```
kalensk, line 168: We should keep these tests as they are still valid since we just removed the remote version and kept local (with the name changed.) That is, just change the caller to use the new local Version() function. And change the test name `TestVersion_Local` to `TestVersion` to reflect that change.

orhankislal: We already have tests for the actual Version function. Without the remote/local layer, these tests are redundant.

PR #603, opened 2021-10-21
Update tested PXF version to 6 in the CI

orhankislal:


ci/scripts/load-extensions.bash
```diff
@@ -114,6 +114,10 @@ install_pxf() {
         scp pxf_rpm_source/*.rpm "gpadmin@${host}":/tmp/pxf_source.rpm
         scp /tmp/key.json "gpadmin@${host}":/tmp/key.json
 
+        ssh -n "${host}" "
+            echo 'export JAVA_HOME=/usr/lib/jvm/jre' >> /home/gpadmin/.bashrc
+        "
+
```
kalensk, line 120: Would it make sense to simplify this and combine it into the other ssh call. (Yes, I know that is done as centos with root access, but there "might" be a way to make it work.) Just an idea.

kalensk (approved):
LGTM. Thanks!

PR #606, opened 2021-10-22
Add gptext to CI testing

orhankislal:
- GPText will officialy support gpupgrade with version 3.8.0.
- GPText is somewhat different compared to the other
extensions, it requires an external solr cluster to test its
functionality.

Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:add-gptext

### DIFF
--- a/ci/scripts/upgrade-extensions.bash
+++ b/ci/scripts/upgrade-extensions.bash
@@ -65,6 +65,9 @@ time ssh -n mdw "
               --temp-port-range 6020-6040
     set -e
 
+    # Remove the expected failure logs such that any legitimate errors can easily be identified.
+    rm -rf /home/gpadmin/gpAdminLogs/gpupgrade/pg_upgrade/*
+
     echo 'Installing extensions on the target cluster...'
     source /usr/local/greenplum-db-target/greenplum_path.sh
     export MASTER_DATA_DIRECTORY=\$(gpupgrade config show --target-datadir)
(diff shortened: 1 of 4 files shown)


ci/scripts/upgrade-extensions.bash
```diff
@@ -92,6 +92,15 @@ SQL_EOF
...
 
+    echo 'Installing gptext on target cluster...'
+    source /usr/local/greenplum-db-text/greenplum-text_path.sh
+    gptext-migrator
+
+    # Since these files are required by the gptext .so file they need to be in the target cluster. Since gpupgrade
+    # initialize is re-run and idempotent place them in the backup of the master data directory.
+    cp /tmp/gptext-source-tmp/* ~/.gpupgrade/master.bak/
+    # cp /tmp/gptext-source-tmp/* \$MASTER_DATA_DIRECTORY/
+
```
kalensk, line 103: Hopefully, this last cp is not needed....

kalensk (approved):
LGTM

(once pipeline runs and lets see about that last `cp`...)

ci/scripts/upgrade-extensions.bash
kalensk: Actually, if its not needed. Then I think it makes more sense to get rid of the `/tmp/gptext-source-tmp` indirection and inline the copy from the source to target MASTER_DATA_DIRECTORY with a comment....

ci/scripts/upgrade-extensions.bash
```diff
@@ -142,6 +151,11 @@ ssh -n mdw "
         echo 'Starting pxf...'
         /usr/local/pxf-gp6/bin/pxf cluster start
     fi
+
+    echo 'Test extensions after upgrade...'
+    psql -v ON_ERROR_STOP=1 -d gptext_db <<SQL_EOF
+        SELECT * FROM gptext.search(table(SELECT 1 SCATTER BY 1), 'gptext_db.public.t1', 'greenplum', NULL);
+SQL_EOF
```
kalensk, line 158: After thinking some more I don't think we need this as the order of operations are: upgrade, compare dumps (which calls pg_dump which indirectly exercises the functions), and then apply any fixups.

Thus, since compare dump happens first which in some sense exercises the functions. I think this is sufficient until we find a specific case where it is needed.

PR #607, opened 2021-10-22
Enable revert operation on initialize for no mirror systems

orhankislal:
Current implementation of revert has a hard stop on systems without mirrors. While this is necessary for production systems, it hinders the ability to quickly test on dev environments for customers. This commit adds a new check to determine if we are at a stage before execute and allows the revert in such cases.

Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:revert-initialize


kalensk (approved):
LGTM

(after pipeline finishes...)

PR #608, opened 2021-10-29
Add scripts for constraint drop/recreate on partitioned tables

orhankislal:
Until GPDB 5.29, dropping constraints on child partition tables was
not allowed, which meant gpupgrade data migration scripts could
not handle them.
Now that there is a GUC to enable this operation, this commit
adds the scripts to detect and drop/recreate them.

Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:mig-scripts-constraints


ci/scripts/cluster-tests.bash
```diff
@@ -27,7 +27,7 @@ function run_migration_scripts_and_tests() {
         data-migration-scripts/gpupgrade-migration-sql-executor.bash "$GPHOME_SOURCE" "$PGPORT" /tmp/migration/pre-initialize || true
 
         make
-        make check --keep-going
+        make check
```
kalensk, line 30: We should avoid this change....(at least within this PR as more discussion is needed). Primarily we had --keep-going so that on the CI its easy to see multiple errors and things for easier debugging and faster feedback.

data-migration-scripts/pre-initialize/gen_drop_constraint_2_primary_unique.sql
```diff
@@ -1,32 +1,9 @@
 -- Copyright (c) 2017-2021 VMware, Inc. or its affiliates
 -- SPDX-License-Identifier: Apache-2.0
 
--- Generate a script to to drop unique/primary key constraints from
--- non-partitioned tables. Exclude all the root and child partitions as we
--- cannot drop unique/primary constraints on the child partition tables
+-- Generate a script to drop unique/primary key constraints.
```
soumyadeep2007, line 4: Currently this script behaves in accordance to a 5X->6X+ upgrade, which is not compatible with a 6X->7X+ upgrade. The old logic needs to be reintroduced for 6X->7X+ upgrades.

The following comment, which would be nice if added reflects this.

"For upgrades from 5X we need to set gp_enable_drop_key_constraint_child_partition so we can drop unique and primary key constraints directly from leaf partitions, in addition to dropping them from the root. Normally, dropping constraints directly from leaves is banned by GPDB - and dropping constraints from the root suffices. However, dropping such constraints from the root doesn't always drop the constraint from the leaves (if an unnamed constraint was added to the partition hierarchy with ALTER TABLE ADD CONSTRAINT at the root level).

For upgrades from 6X onwards, it is sufficient to drop the constraint from the root level."

kalensk (approved):
LGTM! This will be an excellent feature for many customers : )

PR #610, opened 2021-11-04
Update tests to create multiple databases within a tablespace

orhankislal:
This commit expands the existing tests to create multiple databases
inside a user-defined tablespace.

CREATE DATABASE statement indirectly creates a checkpoint and if the
dirty buffers exist at that point the statement will fail. To fix this
issue we added a CHECKPOINT command.

Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:mult-tablespace-test


kalensk (approved):
LGTM

PR #621, opened 2021-12-11
add dynamic_library_path parameter

kalensk:
The dynamic_library_path parameter has been added to allow customers to specify the location of their extensions installed outside of the default location. This GUC is needed by pg_upgrade when performing a pg_dump and pg_restore in order to correctly load the correct extension .so file.

Only set this GUC when it is specified and not the default. This is to keep things simple and there is no need to set this GUC on the customers system if the default locations are used.

Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:dynamicLibraryPath


orhankislal (approved):
LGTM

PR #626, opened 2022-01-04
Add citext and fuzzystrmatch tests to CI

orhankislal:
Add citext and fuzzystrmatch tests to CI
Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:add-ext-ci?group=extensions


ci/6_upgrade_and_functional_jobs.yml
```diff
@@ -180,7 +180,7 @@
       params:
         USE_LINK_MODE: 1
         FILTER_DIFF: 1
-        DIFF_FILE: postgis.diff
+        DIFF_FILE: extensions.diff
```
kalensk, line 183: Let's also rename the actual file `ci/scripts/filters/postgis.diff` to match this new name. That is, rename `ci/scripts/filters/postgis.diff` to `ci/scripts/filters/extensions.diff` (Im a bit concerned that its "working" without renaming the file...)

ci/scripts/load-extensions.bash
```diff
@@ -50,6 +50,26 @@ SQL_EOF
...
+        ) DISTRIBUTED BY (id);
+
+        INSERT INTO citext_test_type VALUES (1,  'larry',  md5(random()::text) );
+SQL_EOF
+
+    echo 'Installing Fuzzy String Match'
+    psql -d ext-test -f /usr/local/greenplum-db-source/share/postgresql/contrib/fuzzystrmatch.sql
+    psql -v ON_ERROR_STOP=1 -d ext-test <<SQL_EOF
+        CREATE VIEW fuzzystrmatch_test_view AS SELECT soundex('a'::text);
+SQL_EOF
```
kalensk, line 72: These are actually postgres native extensions, so lets place them in that appropriate block and have them be installed in the postgres database like all others to keep things simple.

ci/scripts/upgrade-extensions.bash
```diff
@@ -73,13 +73,6 @@ time ssh -n mdw "
 
     gpstart -a
 
-    psql -v ON_ERROR_STOP=1 -d postgres <<SQL_EOF
-        CREATE EXTENSION amcheck;
-        CREATE EXTENSION dblink;
-        CREATE EXTENSION hstore;
-        CREATE EXTENSION pgcrypto;
-SQL_EOF
-
```
kalensk, line 82: Curious why things work after removing this? I think my mental understanding of upgrading with extensions is confused and could use some clarification. (Obviously if this is not needed, then its a better and easier customer experience which is great.)

kalensk (commented):
Thanks for adding this. Always good to have increased test coverage for common postgres extensions. A few comments and tweaks. We can sync offline as well if needed.

ci/6_upgrade_and_functional_jobs.yml
orhankislal: I actually did make this change but it seems it got lost during git rebase etc. steps

ci/scripts/load-extensions.bash
orhankislal: I'll move them to the correct block. The problem with the postgres database is caused by the overloaded operators. Both citext and postgis define operators like '=' or '<' for their custom types (citext, geometry, etc.) During the dump, the order of these operators is different on 5X vs. 6X which means the comparison fails incorrectly. We can embed these changes into the diff file as well but it gets unnecessarily large and I think this is a better solution.

ci/scripts/upgrade-extensions.bash
orhankislal: They were never needed, pg_upgrade docs mention the same: `Do not load the schema definitions, e.g., CREATE EXTENSION pgcrypto, because these will be duplicated from the old cluster`. I realized that while I was recording the video. Placement of the .so files is necessary, enabling the extension on a particular database is not.

PR #629, opened 2022-01-11
gpupgrade_config: expose dynamic_library_path

kalensk:
Also, reorder config parameters to group required parameters together at the top of the file.

Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:exposeDynamicLibPath

### DIFF
--- a/gpupgrade_config
+++ b/gpupgrade_config
@@ -2,6 +2,9 @@
 # gpupgrade configuration file
 # ----------------------------
 
+# The master port for the source Greenplum installation.
+source_master_port =
+
 # The installation path for the source Greenplum Database.
 # For example, /usr/local/<source-greenplum-version>.
 source_gphome =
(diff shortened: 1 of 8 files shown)


gpupgrade_config
```diff
@@ -26,8 +29,10 @@ disk_free_ratio = 0.6
 # Choose "true" to use host names, or "false" to use IP addresses.
 use_hba_hostnames = false
 
-# The master port for the source Greenplum installation.
-source_master_port =
+# For extensions installed outside of target_gphome amend their path to the
+# dynamic_library_path value. For example, if pxf was installed in /opt then
+# set dynamic_library_path to "$libdir:/opt/greenplum/pxf-gp6".
+dynamic_library_path = $libdir
```
kalensk, line 35: Let's ensure this is the best wording for this. Maybe we can sync with @shivzone offline together.

kalensk: We decide to go with:

> # For extensions installed outside of target_gphome include the extension’s 
# path in the dynamic_library_path value. For example, if pxf was installed in /opt then
# set dynamic_library_path to "$libdir:/opt/greenplum/pxf-gp6".

orhankislal (approved):
LGTM

kalensk:
Upon thinking some more we will need to update this PR to filter out $libdir before appending it to dynamic_library_path so that it is not added twice (once as the default value placed from the underlying server code, and again from gpupgrade). Again, we do want to have some "default value" in the config file to not confuse users and have a clear workflow since the vast majority of the time it is not required.

PR #630, opened 2022-01-11
CI: add gptext

kalensk:
- GPText will officialy support gpupgrade with version 3.9.0.
- GPText is somewhat different compared to the other
extensions, it requires an external solr cluster to test its functionality.

Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:add-gptext


orhankislal (approved):
LGTM

PR #633, opened 2022-01-21
Fix dynamic_library_path params in the CI

orhankislal:
1- $GPHOME_TARGET does not resolve to the correct value because of the
quotes. This commit changes it to use the actual path.
2- gptext path is not needed any more once the GUC change is merged into
6X. This commit removes it to keep the code clean.

Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:gpdb-guc-defaults

### DIFF
--- a/ci/scripts/upgrade-extensions.bash
+++ b/ci/scripts/upgrade-extensions.bash
@@ -63,7 +63,7 @@ time ssh -n mdw "
               --source-gphome $GPHOME_SOURCE \
               --source-master-port $PGPORT \
               --temp-port-range 6020-6040 \
-              --dynamic-library-path $GPHOME_TARGET'/madlib/Current/ports/greenplum/6/lib:/usr/local/greenplum-db-text/lib/gpdb6'
+              --dynamic-library-path ${GPHOME_TARGET}/madlib/Current/ports/greenplum/6/lib:/usr/local/greenplum-db-text/lib/gpdb6
     set -e
 
     # Remove the expected failure logs such that any legitimate errors can easily be identified.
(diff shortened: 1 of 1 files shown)


kalensk:
Based on the comments of https://github.com/greenplum-db/gpdb/pull/13022 looks like there are little to no extensions that have a default location that could really be used by dynamic_library_path.

We could probably delete this branch and close out this PR.

orhankislal:
We still need the $GPHOME_TARGET change, I have updated the PR and the commit message accordingly.

kalensk:
> 1- $GPHOME_TARGET does not resolve to the correct value because of the
quotes. 

Have we tried keeping $GPHOME_TARGET and just removing the single quotes around it? I think that would work and be better.

ci/scripts/upgrade-extensions.bash
```diff
@@ -106,7 +106,7 @@ time ssh -n mdw "
               --source-gphome $GPHOME_SOURCE \
               --source-master-port $PGPORT \
               --temp-port-range 6020-6040 \
-              --dynamic-library-path '$GPHOME_TARGET/madlib/Current/ports/greenplum/6/lib:/usr/local/greenplum-db-text/lib/gpdb6'
+              --dynamic-library-path $GPHOME_TARGET'/madlib/Current/ports/greenplum/6/lib:/usr/local/greenplum-db-text/lib/gpdb6'
```
kalensk, line 109: Sorry, I should have been more precise. This is the perfect scenario for curly's! Try the following with no single quotes anywhere:

${GPHOME_TARGET}/madlib/Current/ports/greenplum/6/lib:/usr/local/greenplum-db-text/lib/gpdb6'

kalensk: I made a quick fixup commit to just let the pipeline churn and not have to wait for feedback...

kalensk (approved):

PR #636, opened 2022-01-25
update dependencies and bump to golang 1.17

kalensk:
This does not update gRPC (and some dependencies that rely on gRPC) as that currently breaks kill-services and restart-services.

Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:updateDependencies


orhankislal (approved):
LGTM. We should probably mention that there are completion-related changes in the commit message or just separate it into two commits.

PR #637, opened 2022-01-25
pg_upgrade: partitioned_heap_table_with_dropped_columns

kalensk:
These are the acceptance tests for PR https://github.com/greenplum-db/gpdb/pull/13028 which should be merged first before this PR.

Update tests for readability and maintainability. Test two workarounds 1) doing a CTAS for the individually affected partitions, and 2) doing a CTAS for the entire partition table.

Pipeline (using the above PR branch to test): https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:dropColTests?group=pg_upgrade
Using PR branch RPM: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:dropColTests/resources/gpdb6_centos7_rpm

### DIFF
--- a/test/acceptance/pg_upgrade/non_upgradeable_tests/expected/partitioned_heap_table_with_dropped_columns.out
+++ b/test/acceptance/pg_upgrade/non_upgradeable_tests/expected/partitioned_heap_table_with_dropped_columns.out
@@ -14,56 +14,131 @@
 -- Create and setup non-upgradeable objects
 --------------------------------------------------------------------------------
 
--- 1. Differently aligned dropped column
-CREATE TABLE p_different_aligned_column (a int, b aclitem, c int) DISTRIBUTED BY (a) PARTITION BY range(c) SUBPARTITION BY range(a) (PARTITION p_part_with_different_alignedd_dropped_columns START(0) END(42) (SUBPARTITION subpart_differnt_aligned_column START(0) END(22), SUBPARTITION p_subpart_with_different_alignedd_dropped_columns START(22) END(42)));
+-- 1. Heterogeneous partition table with dropped column
+--    The root and only a subset of children have the dropped column reference.
+CREATE TABLE dropped_column (a int, b int, c char, d varchar(50)) DISTRIBUTED BY (c) PARTITION BY RANGE (a) (PARTITION part_1 START(1) END(5), PARTITION part_2 START(5));
 CREATE
+ALTER TABLE dropped_column DROP COLUMN d;
+ALTER
+
+-- Splitting the subpartition leads to its rewrite, eliminating its dropped column
+-- reference. So, after this, only part_2 and the root partition will have a
+-- dropped column reference.
+ALTER TABLE dropped_column SPLIT PARTITION FOR(1) AT (2) INTO (PARTITION split_part_1, PARTITION split_part_2);
+ALTER
+INSERT INTO dropped_column VALUES(1, 2, 'a');
+INSERT 1
+
+-- 2. Root partitions do not have dropped column references, but some child partitions do
+CREATE TABLE child_has_dropped_column (a int, b int, c char, d varchar(50)) PARTITION BY RANGE (a) (PARTITION part_1 START(1) END(5), PARTITION part_2 START(5));
+CREATE
+
+CREATE TABLE intermediate_table (a int, b int, c char, d varchar(50), to_drop int);
+CREATE
+ALTER TABLE intermediate_table DROP COLUMN to_drop;
+ALTER
+
+ALTER TABLE child_has_dropped_column EXCHANGE PARTITION part_1 WITH TABLE intermediate_table;
+ALTER
+DROP TABLE intermediate_table;
+DROP
+
+-- 3. Root and child partitions have different number of dropped column references
+CREATE TABLE diff_no_dropped_columns (a int, b int, c char, to_drop varchar(50)) PARTITION BY RANGE (a) (PARTITION part_1 START(1) END(5), PARTITION part_2 START(5));
+CREATE
+ALTER TABLE diff_no_dropped_columns DROP COLUMN to_drop;
+ALTER
+
+CREATE TABLE intermediate_table (a int, b int, c char, to_drop varchar(50), to_drop_2 int);
+CREATE
+ALTER TABLE intermediate_table DROP COLUMN to_drop;
+ALTER
+ALTER TABLE intermediate_table DROP COLUMN to_drop_2;
+ALTER
+
+ALTER TABLE diff_no_dropped_columns EXCHANGE PARTITION part_1 WITH TABLE intermediate_table;
+ALTER
+DROP TABLE intermediate_table;
+DROP
+
+-- 4. Differently aligned dropped column
+CREATE TABLE differently_aligned_column (a int, b aclitem, c int) DISTRIBUTED BY (a) PARTITION BY range(c) SUBPARTITION BY range(a) (PARTITION differently_aligned_columns_part START(0) END(42) (SUBPARTITION subpart_1 START(0) END(22), SUBPARTITION subpart_2 START(22) END(42)));
+CREATE
+ALTER TABLE differently_aligned_column DROP COLUMN b;
+ALTER
+INSERT INTO differently_aligned_column VALUES(22, 22), (23, 23);
+INSERT 2
 
 -- 'b' column is intentionally differently aligned - aclitem has 'i' alignment
 -- and timetz has 'd' alignment. If we allow the upgrade then on the new cluster
 -- we will fetch column 'c' at the wrong offset.
-CREATE TABLE subpart_different_aligned_column(a int, b timetz, c int);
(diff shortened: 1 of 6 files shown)


test/acceptance/pg_upgrade/non_upgradeable_tests/sql/partitioned_heap_table_with_dropped_columns.sql
```diff
@@ -14,51 +14,69 @@
 -- Create and setup non-upgradeable objects
 --------------------------------------------------------------------------------
 
--- 1. Differently aligned dropped column
-CREATE TABLE p_different_aligned_column (a int, b aclitem, c int) DISTRIBUTED BY (a)
+-- 1. Partition table with dropped column
```
soumyadeep2007, line 17: We would need to add a case where:
1. We have a partition hierarchy w/ `x` #cols and NO dropped cols.
2. We create a scratch table with the number of cols `x + 1`.
3. We drop the (x + 1)th column from the scratch table, to make it schema compliant w/ the partition hierarchy.
4. Then we exchange it with any leaf partition of the hierarchy.

This corresponds to a sub-case of:
`/* root partition does not have dropped column references */`
See code past the continue:
```
/*
* The child partition must not have any dropped column
* references and the column count must be the same as the
* root partition.
*/
if (childnumattisdropped == 0 && parnatts == childnatts)
continue;
```

soumyadeep2007 (commented):
Some initial feedback. More to come!


We need some upgradeable tests (positive tests) for homogeneous partitions to ensure that they don't get flagged during `initialize` and is properly upgraded w/ `execute`.

For a list of the 3 homogeneous cases we have to add tests for:

```* Valid homogeneous partition tables are defined as:
 *
 * 1. The root partition and its child partitions have no dropped column
 *    references.
 * 2. The root partition and its child partitions have the same dropped column
 *    references.
 * 3. The root partition has a dropped column reference but none of its child
 *    partitions do.
```

kalensk:
> We need some upgradeable tests (positive tests) for homogeneous partitions to ensure that they don't get flagged during initialize and is properly upgraded w/ execute.

There seems to be one already (although the variables and names could be renamed for clarity)...See `p_dropcol` in [partitioned_heap_table.sql]()https://github.com/greenplum-db/gpupgrade/blob/9ceceab6e712520d74bd462c813a0bfdf1c57ce8/test/acceptance/pg_upgrade/upgradeable_tests/source_cluster_regress/sql/partitioned_heap_table.sql#L60-L67.

soumyadeep2007:
> There seems to be one already (although the variables and names could be renamed for clarity)...See p_dropcol in partitioned_heap_table.sql

Yep, we also need to ensure that we handle all 3 cases.

test/acceptance/pg_upgrade/non_upgradeable_tests/sql/partitioned_heap_table_with_dropped_columns.sql
```diff
@@ -14,51 +14,112 @@
...
+ALTER TABLE diff_no_dropped_columns DROP COLUMN to_drop;
+
+CREATE TABLE intermediate_table (a int, b int, c char, to_drop varchar(50), to_drop_2 int);
+ALTER TABLE intermediate_table DROP COLUMN to_drop;
+ALTER TABLE intermediate_table DROP COLUMN to_drop_2;
+
+ALTER TABLE diff_no_dropped_columns EXCHANGE PARTITION part_1 WITH TABLE intermediate_table;
+DROP TABLE intermediate_table;
+
+-- 4. differently named dropped column
```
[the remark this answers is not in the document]
kalensk: Thanks! Removed and renumbered the scenarios.

test/acceptance/pg_upgrade/non_upgradeable_tests/sql/partitioned_heap_table_with_dropped_columns.sql
```diff
@@ -14,51 +14,112 @@
...
-CREATE TABLE subpart_different_size_column(a int, b numeric, c int);
-ALTER TABLE p_different_size_column DROP COLUMN b;
-ALTER TABLE subpart_different_size_column DROP COLUMN b;
-ALTER TABLE p_different_size_column ALTER PARTITION p_part_with_different_sized_dropped_columns EXCHANGE PARTITION subpart_differnt_size_column WITH TABLE subpart_different_size_column;
+CREATE TABLE intermediate_table(a int, b numeric, c int);
+ALTER TABLE intermediate_table DROP COLUMN b;
+
+ALTER TABLE differently_sized_column ALTER PARTITION differently_sized_column_part
+    EXCHANGE PARTITION subpart_1 WITH TABLE intermediate_table;
+DROP TABLE intermediate_table;
```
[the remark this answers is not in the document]
kalensk: Added as the last case. (Note that the scenarios got renumbered after removing the scenario as outlined in your below comment)

test/acceptance/pg_upgrade/non_upgradeable_tests/expected/partitioned_heap_table_with_dropped_columns.out
```diff
@@ -73,33 +148,195 @@ ALTER
...
+  Partitions with misaligned columns:
+    public.child_has_dropped_column_1_prt_part_1
+    public.diff_no_dropped_columns_1_prt_part_1
+    public.diff_no_dropped_columns_1_prt_part_2
+    public.differently_aligned_colu_1_prt_differently_alig_2_prt_subpart_1
+    public.differently_aligned_varl_1_prt_differently_alig_2_prt_subpart_1
+    public.differently_sized_column_1_prt_differently_size_2_prt_subpart_1
+    public.dropped_cols_out_of_order_1_prt_part_1
+    public.dropped_cols_out_of_order_1_prt_part_1
+    public.dropped_column_1_prt_part_2
```
jimmyyih, line 162: The output should not be sorted.

kalensk: I see. Yeah that is masking the detection of the two sub-checks. I added a comment in the tests explaining this.

Ideally it would be great to not need to sort so that the output of the sub-checks can be verified, but it appears things are not deterministic locally and even amongst mac vs. linux without sorting. So I am not sure.

jimmyyih: That's interesting that it would be non-deterministic in line order.  Usually that would happen for dispatched queries because segments return their result set when they individually finish... but for these pg_upgrade queries, there's no dispatch so I would have assumed it would be deterministic in line order.  Wonder why that is. 🤔

soumyadeep2007: See: https://github.com/greenplum-db/gpupgrade/pull/555#issue-909796813

kalensk: @orhankislal helped figure out how to effectively split on the file and sort. That is, we split the file and sort the two sub-checks individually. So I think we are good! Thanks @orhankislal!

test/acceptance/pg_upgrade/non_upgradeable_tests/sql/partitioned_heap_table_with_dropped_columns.sql
```diff
@@ -14,73 +14,254 @@
...
+
+ALTER TABLE dropped_cols_out_of_order EXCHANGE PARTITION part_1 WITH TABLE intermediate_table;
+DROP TABLE intermediate_table;
 
 --------------------------------------------------------------------------------
 -- Assert that pg_upgrade --check correctly detects the non-upgradeable objects
 --------------------------------------------------------------------------------
 !\retcode gpupgrade initialize --source-gphome="${GPHOME_SOURCE}" --target-gphome=${GPHOME_TARGET} --source-master-port=${PGPORT} --disk-free-ratio 0 --automatic;
-! cat ~/gpAdminLogs/gpupgrade/pg_upgrade/p-1/heterogeneous_partitioned_tables.txt | LC_ALL=C sort -b;
+-- NOTE: We sort the output to ensure the test is deterministic. See PR #555. However, this prevents asserting the
```
soumyadeep2007, line 130: Nit: would like to see a commit SHA instead of a PR reference here.

soumyadeep2007 (approved):
LGTM!

test/acceptance/pg_upgrade/non_upgradeable_tests/sql/partitioned_heap_table_with_dropped_columns.sql
jimmyyih: Commit is https://github.com/greenplum-db/gpdb/commit/a21ab9b476f6cb06c44878e07616cd740d6e49ce for reference.

PR #638, opened 2022-01-27
gpupgrade_config: comment out defaults

kalensk:
Also re-order flag definitions in cli/commands/initialize.go to match the order in gpupgrade_config.

Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:commentDefaultConfigValues

### DIFF
--- a/gpupgrade_config
+++ b/gpupgrade_config
@@ -17,32 +17,32 @@ target_gphome =
 # The copy method creates a copy of the primary segments and performs the
 # upgrade on the copies.
 # The link method directly upgrades the primary segments.
-mode = copy
+# mode = copy
 
 # The disk free ratio specifies what fraction of disk space must be free on
 # every host in order for gpupgrade to run. The ratio ranges from 0.0 to 1.0.
 # Recommended values are 0.6 [60%] for copy mode, and 0.2 [20%] for link mode.
-disk_free_ratio = 0.6
+# disk_free_ratio = 0.6
 
 # Whether to populate pg_hba.conf with hostnames or IP addresses during
 # execution of gpinitsystem and other utilities.
 # Choose "true" to use host names, or "false" to use IP addresses.
-use_hba_hostnames = false
+# use_hba_hostnames = false
 
 # For extensions installed outside of target_gphome include the extension’s
 # path in the dynamic_library_path value. For example, if pxf was installed in
 # /opt then set dynamic_library_path to "$libdir:/opt/greenplum/pxf-gp6".
-dynamic_library_path = $libdir
+# dynamic_library_path = $libdir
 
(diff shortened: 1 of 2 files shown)


gpupgrade_config
```diff
@@ -17,32 +17,32 @@ target_gphome =
 # The copy method creates a copy of the primary segments and performs the
 # upgrade on the copies.
 # The link method directly upgrades the primary segments.
-mode = copy
+# mode = copy
```
shivzone, line 20: Very trivial nitpick. For action configs, would be preferable to have every config of the form `#{key} = {$value}` with no space. Eg: `#mode = copy`

shivzone (approved):

gpupgrade_config
kalensk: We could do that, but then the comments don't vertically left align and looks visually funky. I'm "ok" with either way, but just want to point that out.

```
# ----------------------------
# gpupgrade configuration file
# ----------------------------

# The master port for the source Greenplum installation.
source_master_port =

# The installation path for the source Greenplum Database.
# For example, /usr/local/<source-greenplum-version>.
source_gphome =

# The installation path for the target Greenplum Database.
# For example, /usr/local/<target-greenplum-version>.
target_gphome =

# The mode is the upgrade method. The choices are “link” or “copy”.
# The copy method creates a copy of the primary segments and performs the
# upgrade on the copies.
# The link method directly upgrades the primary segments.
#mode = copy

# The disk free ratio specifies what fraction of disk space must be free on
# every host in order for gpupgrade to run. The ratio ranges from 0.0 to 1.0.
# Recommended values are 0.6 [60%] for copy mode, and 0.2 [20%] for link mode.
#disk_free_ratio = 0.6

# Whether to populate pg_hba.conf with hostnames or IP addresses during
# execution of gpinitsystem and other utilities.
# Choose "true" to use host names, or "false" to use IP addresses.
#use_hba_hostnames = false

# For extensions installed outside of target_gphome include the extension’s
# path in the dynamic_library_path value. For example, if pxf was installed in
# /opt then set dynamic_library_path to "$libdir:/opt/greenplum/pxf-gp6".
#dynamic_library_path = $libdir

# The temporary port range for the target Greenplum installation.
# The temporary port range should be reserved prior to initializaton.
# The format is a comma separated list of ports and port ranges, e.g.
# “6000,6002-6005,6012.” The ports will be reconfigured to use the source
# Greenplum installation port range once upgrade is complete.
#temp_port_range = 50432-65535

# The port where the gpupgrade process will be running.
#hub_port = 7527

# The port where the agent process will be running on all hosts.
#agent_port = 6416
```

orhankislal (approved):
LGTM

PR #641, opened 2022-02-17
Add PXF6 to gpupgrade pipeline

bradfordb-vmware:
Authored-by: Bradford D. Boyle <bradfordb@vmware.com>

### DIFF
--- a/ci/1_resources_anchors_groups.yml
+++ b/ci/1_resources_anchors_groups.yml
@@ -203,15 +203,19 @@ resources:
     versioned_file: greenplum-text-release-dev-rhel6_x86_64.tar.gz
 {{- end}}
 
-# NOTE: Skip creating the pxf resources for centos6 since pxf5 gpdb6 is not supported for centos6. Thus, we can only
-# test pxf upgrades on centos7.
+# NOTE: Skip creating the pxf resources for centos6 since pxf6 is not
+# supported for centos6. Thus, we can only test pxf upgrades on centos7.
 {{- if ne .CentosVersion "6" }}
-- name: pxf_5_gpdb{{.GPVersion}}_centos{{.CentosVersion}}_rpm
+- name: pxf_6_gpdb{{.GPVersion}}_centos{{.CentosVersion}}_tgz
   type: gcs
   source:
-    json_key: ((concourse-gcs-resources-service-account-key))
-    bucket: pivotal-gpdb-concourse-resources-prod
-    regexp: pxf/published/gpdb{{.GPVersion}}/pxf-gp{{.GPVersion}}-5.(.*).el{{.CentosVersion}}.x86_64.rpm
+    json_key: ((pxf-storage-service-account-key))
+    bucket: data-gpdb-ud-pxf-build
+    versioned_file: prod/snapshots/pxf-gp{{.GPVersion}}.el{{.CentosVersion}}.tar.gz
+    # after PXF 6.3.0 is released, the released RPMs can be fetched from standard bucket locations
+    # json_key: ((concourse-gcs-resources-service-account-key))
+    # bucket: pivotal-gpdb-concourse-resources-prod
+    # regexp: pxf/published/gpdb{{.GPVersion}}/pxf-gp{{.GPVersion}}-6.(.*).el{{.CentosVersion}}.x86_64.rpm
 {{- end }}
 {{end}}
 
--- a/ci/scripts/load-extensions.bash
+++ b/ci/scripts/load-extensions.bash
@@ -166,14 +166,22 @@ SQL_EOF
 "
 
 install_pxf() {
-    local PXF_CONF=/home/gpadmin/pxf
+    local PXF_BASE=/home/gpadmin/pxf
 
     echo "Installing pxf on all hosts in the source cluster..."
     echo "${GOOGLE_CREDENTIALS}" > /tmp/key.json
 
+    # PXF SNAPSHOT builds are only available as an RPM inside a tar.gz
+    if compgen -G pxf_installer_source/pxf-gp?.el7.tar.gz &>/dev/null; then
+        tar -xf pxf_installer_source/pxf-gp?.el7.tar.gz \
+            --directory pxf_installer_source \
+            --strip-components=1 \
+            --wildcards '*.rpm'
+    fi
+
     mapfile -t hosts < cluster_env_files/hostfile_all
     for host in "${hosts[@]}"; do
-        scp pxf_rpm_source/*.rpm "gpadmin@${host}":/tmp/pxf_source.rpm
+        scp pxf_installer_source/*.rpm "gpadmin@${host}":/tmp/pxf_source.rpm
         scp /tmp/key.json "gpadmin@${host}":/tmp/key.json
 
         ssh -n "centos@${host}" "
@@ -193,14 +201,16 @@ install_pxf() {
 
         echo 'Initialize pxf...'
         export GPHOME=$GPHOME_SOURCE
-        export PXF_CONF=$PXF_CONF
-        export JAVA_HOME=/usr/lib/jvm/jre
+        export PXF_BASE=$PXF_BASE
 
-        mkdir -p ${PXF_CONF}/servers/google
-        /usr/local/pxf-*/bin/pxf cluster init
+        /usr/local/pxf-*/bin/pxf cluster prepare
+        sed -i -e 's|^# export JAVA_HOME=.*|export JAVA_HOME=/usr/lib/jvm/jre|' $PXF_BASE/conf/pxf-env.sh
+        mkdir -p ${PXF_BASE}/servers/google
 
-        cp /home/gpadmin/pxf/templates/gs-site.xml ${PXF_CONF}/servers/google/
-        sed -i 's|YOUR_GOOGLE_STORAGE_KEYFILE|/tmp/key.json|' ${PXF_CONF}/servers/google/gs-site.xml
+        cp /usr/local/pxf-*/templates/gs-site.xml ${PXF_BASE}/servers/google/
+        sed -i 's|YOUR_GOOGLE_STORAGE_KEYFILE|/tmp/key.json|' ${PXF_BASE}/servers/google/gs-site.xml
+
+        /usr/local/pxf-*/bin/pxf cluster register
         /usr/local/pxf-*/bin/pxf cluster sync
         /usr/local/pxf-*/bin/pxf cluster start
 
@@ -230,7 +240,7 @@ SQL_EOF
 "
 }
 
-test_pxf "$OS_VERSION" && install_pxf || echo "Skipping pxf for centos6 since pxf5 for GPDB6 on centos6 is not supported..."
+test_pxf "$OS_VERSION" && install_pxf || echo "Skipping pxf for centos6 since pxf6 for GPDB6 on centos6 is not supported..."
 
 echo "Running the data migration scripts on the source cluster..."
 ssh -n mdw "
--- a/ci/scripts/upgrade-extensions.bash
+++ b/ci/scripts/upgrade-extensions.bash
@@ -23,9 +23,17 @@ scp postgis_gppkg_target/postgis*.gppkg gpadmin@mdw:/tmp/postgis_target.gppkg
 scp madlib_gppkg_target/madlib*.gppkg gpadmin@mdw:/tmp/madlib_target.gppkg
 
 if test_pxf "$OS_VERSION"; then
+    # PXF SNAPSHOT builds are only available as an RPM inside a tar.gz
+    if compgen -G pxf_installer_target/pxf-gp?.el7.tar.gz &>/dev/null; then
+        tar -xf pxf_installer_target/pxf-gp?.el7.tar.gz \
+            --directory pxf_installer_target \
+            --strip-components=1 \
+            --wildcards '*.rpm'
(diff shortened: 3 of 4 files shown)


bradfordb-vmware:
Opening this as a draft/WIP PR since the exact names of the PXF scripts are still being discussed and the Concourse reference currently references a dev build artifact for the PXF RPMs.

ci/scripts/upgrade-extensions.bash
```diff
@@ -34,6 +42,14 @@ if test_pxf "$OS_VERSION"; then
             sudo chown -R gpadmin:gpadmin /usr/local/pxf*
         "
     done
+
+    ssh -n mdw "
+        set -eux -o pipefail
+        export GPHOME=${GPHOME_SOURCE}
+        export PXF_BASE=/home/gpadmin/pxf
+
+        PGDATABASE=postgres /usr/local/pxf-gp5/bin/pxf-before-gpupgrade
```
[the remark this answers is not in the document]
bradfordb-vmware: This is here because the GPDB cluster in this job doesn't have a `gpadmin` database. For customers that are running this against a GPDB cluster with a `gpadmin` user and `gpadmin` database they won't need to specify anything as those are the defaults that the script assumes.

ci/scripts/load-extensions.bash
```diff
@@ -166,11 +166,19 @@ SQL_EOF
...
 
 install_pxf() {
-    local PXF_CONF=/home/gpadmin/pxf
+    local PXF_BASE=/home/gpadmin/pxf
 
     echo "Installing pxf on all hosts in the source cluster..."
     echo "${GOOGLE_CREDENTIALS}" > /tmp/key.json
 
+    # PXF SNAPSHOT builds are only available as an RPM inside a tar.gz
+    if compgen -G pxf_rpm_source/pxf-gp?.el7.tar.gz &>/dev/null; then
```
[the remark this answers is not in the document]
bradfordb-vmware: I was on the fence about hardcoding the GP version here. I think its a little clearer and we can drop the `compgen -G` usage.

ci/scripts/upgrade-extensions.bash
```diff
@@ -63,7 +79,7 @@ time ssh -n mdw "
               --source-gphome $GPHOME_SOURCE \
               --source-master-port $PGPORT \
               --temp-port-range 6020-6040 \
-              --dynamic-library-path ${GPHOME_TARGET}/madlib/Current/ports/greenplum/6/lib:/usr/local/greenplum-db-text/lib/gpdb6
+              --dynamic-library-path ${GPHOME_TARGET}/madlib/Current/ports/greenplum/6/lib:/usr/local/greenplum-db-text/lib/gpdb6:/usr/local/pxf-gp6/gpextable
```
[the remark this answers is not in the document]
bradfordb-vmware: My understanding is that we include it here because the `dynamic_library_path` GUC for the target cluster (the GPDB cluster being upgraded to) hasn't been set yet. This option causes `gpupgrade` to set the GUC when running the new cluster.

My understanding is that this defaults to just `$libdir` but the user can add it either on the command line or in the gpupgrade_configuration file.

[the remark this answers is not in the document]
kalensk: That is correct.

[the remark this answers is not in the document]
denalex: should this value be somehow read from the existing GP5 cluster since we went to the extend of changing that GUC value in our `pre-upgrade` script ? I don't care much what happens in CI, but for the end-users they should not be required to enter this whole thing manually.

denalex (approved):

ci/scripts/load-extensions.bash
```diff
@@ -166,11 +166,19 @@ SQL_EOF
 "
 
 install_pxf() {
-    local PXF_CONF=/home/gpadmin/pxf
+    local PXF_BASE=/home/gpadmin/pxf
 
     echo "Installing pxf on all hosts in the source cluster..."
     echo "${GOOGLE_CREDENTIALS}" > /tmp/key.json
 
+    # PXF SNAPSHOT builds are only available as an RPM inside a tar.gz
```
ashuka24, line 174: this will be removed once 6.3.0 is released as it should no longer be looking as snapshot builds?

ashuka24 (approved):

hpandeycodeit (approved):

ci/scripts/upgrade-extensions.bash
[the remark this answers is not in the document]
orhankislal: We cannot use the values from the GP5 GUC since the GP6 values are different (they have to be, to point to different so files): `/usr/local/pxf-gp5/gpextable` -> `/usr/local/pxf-gp6/gpextable`. We can try to infer but it creates an unnecessary uncertainty since different extensions use different naming conventions.

ci/1_resources_anchors_groups.yml
```diff
@@ -203,15 +203,19 @@ resources:
     versioned_file: greenplum-text-release-dev-rhel6_x86_64.tar.gz
 {{- end}}
 
-# NOTE: Skip creating the pxf resources for centos6 since pxf5 gpdb6 is not supported for centos6. Thus, we can only
-# test pxf upgrades on centos7.
+# NOTE: Skip creating the pxf resources for centos6 since pxf6 is not
+# supported for centos6. Thus, we can only test pxf upgrades on centos7.
 {{- if ne .CentosVersion "6" }}
-- name: pxf_5_gpdb{{.GPVersion}}_centos{{.CentosVersion}}_rpm
+- name: pxf_6_gpdb{{.GPVersion}}_centos{{.CentosVersion}}_rpm
```
orhankislal, line 209: This resource is not an rpm anymore, we should change it to something like `_targz`

ci/scripts/load-extensions.bash
```diff
@@ -166,11 +166,19 @@ SQL_EOF
...
+    if [ -f pxf_rpm_source/pxf-gp5.el7.tar.gz ]; then
+        tar -xf pxf_rpm_source/pxf-gp5.el7.tar.gz \
```
orhankislal, line 176: Can we use wildcards to avoid tying the script to gp5?

ci/scripts/upgrade-extensions.bash
```diff
@@ -23,6 +23,14 @@ scp postgis_gppkg_target/postgis*.gppkg gpadmin@mdw:/tmp/postgis_target.gppkg
 scp madlib_gppkg_target/madlib*.gppkg gpadmin@mdw:/tmp/madlib_target.gppkg
 
 if test_pxf "$OS_VERSION"; then
+    # PXF SNAPSHOT builds are only available as an RPM inside a tar.gz
+    if [ -f pxf_rpm_target/pxf-gp6.el7.tar.gz ]; then
+        tar -xf pxf_rpm_target/pxf-gp6.el7.tar.gz \
```
orhankislal, line 28: Repeat: Can we use wildcards to avoid tying the script to gp6?

ci/scripts/upgrade-extensions.bash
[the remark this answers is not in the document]
shivzone: In ordere to avoid burdening end users was one of the main reasons why Orhan raised the PR to set the defaults in GP6 to include `/usr/local/pxf-gp6/gpextable`, but there were concerns raised about introducing coupling between server and pxf path. 
We do have this documented here https://docs.vmware.com/en/VMware-Tanzu-Greenplum-Upgrade/1.3/greenplum-upgrade/GUID-gpupgrade-config.html, considering major version upgrades require significant planning, i think it is ok to set this expectation for end users.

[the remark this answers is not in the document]
bradfordb-vmware: Can the upgrade config file include these as commented out defaults?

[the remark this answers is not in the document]
shivzone: [gpupgrade_config file ](https://github.com/greenplum-db/gpupgrade/blob/master/gpupgrade_config) is the default config file that we ask uses to use as a template and let them update.
In gpupgrade code per se if nothing is configured, it simply defaults to `$libdir`. We can possibly include this as a reference in both the above config template file and we have also covered this in our docs here https://docs.vmware.com/en/VMware-Tanzu-Greenplum-Upgrade/1.3/greenplum-upgrade/GUID-gpupgrade-config.html#keywords-and-values-2

ci/1_resources_anchors_groups.yml
kalensk: This would be great to fix, but not a blocker (as we can do it later).

ci/1_resources_anchors_groups.yml
```diff
@@ -203,15 +203,19 @@ resources:
...
-    json_key: ((concourse-gcs-resources-service-account-key))
-    bucket: pivotal-gpdb-concourse-resources-prod
-    regexp: pxf/published/gpdb{{.GPVersion}}/pxf-gp{{.GPVersion}}-5.(.*).el{{.CentosVersion}}.x86_64.rpm
+    json_key: ((pxf-storage-service-account-key))
+    bucket: data-gpdb-ud-pxf-build
+    versioned_file: prod/snapshots/pxf-gp{{.GPVersion}}.el{{.CentosVersion}}.tar.gz
+    # after PXF 6.3.0 is released, the RPM should be fetched from standard bucket locations
+    # json_key: ((concourse-gcs-resources-service-account-key))
+    # bucket: pivotal-gpdb-concourse-resources-prod
+    # regexp: pxf/published/gpdb{{.GPVersion}}/pxf-gp{{.GPVersion}}-6.(.*).el{{.CentosVersion}}.x86_64.rpm
```
kalensk, line 218: Thank you so much for adding this here as a comment! This is perfect and exactly what we need.  : )

ci/scripts/load-extensions.bash
kalensk: Oh you mean, `tar -xf pxf_rpm_source/pxf-gp5.el7.tar.gz` would become `tar -xf pxf_rpm_source/pxf*.tar.gz`? Yeah that would be ideal!

ci/scripts/load-extensions.bash
```diff
@@ -230,7 +240,7 @@ SQL_EOF
 "
 }
 
-test_pxf "$OS_VERSION" && install_pxf || echo "Skipping pxf for centos6 since pxf5 for GPDB6 on centos6 is not supported..."
+test_pxf "$OS_VERSION" && install_pxf || echo "Skipping pxf for centos6 since pxf6 for GPDB6 on centos6 is not supported..."
```
kalensk, line 243: Awesome! Thanks for catching this change.

ci/scripts/upgrade-extensions.bash
```diff
@@ -34,6 +42,14 @@ if test_pxf "$OS_VERSION"; then
...
         "
     done
+
+    ssh -n mdw "
+        set -eux -o pipefail
+        export GPHOME=${GPHOME_SOURCE}
+        export PXF_BASE=/home/gpadmin/pxf
+
+        PGDATABASE=postgres /usr/local/pxf-gp5/bin/pxf-pre-gpupgrade
+        "
```
kalensk, line 52: minor nit: outdent ending `"` to align with ssh to match surrounding ssh syntax.

ci/scripts/load-extensions.bash
bradfordb-vmware: I think the answer depends on what test coverage we want to have. If we want to test the newer versions of gpupgrade will work with existing releases of PXF, then we will want to update this after the next PXF release to use release RPMs. However, we then will not have any test coverage for new versions of PXF. I don't anticipate any changes to the upgrade scripts, so that may be OK.

kalensk:
Is there a pipeline of this branch one could take a look at? I couldn't find one in the `cm` or `ud` Concourse instances.

ci/scripts/load-extensions.bash
kalensk: > we then will not have any test coverage for new versions of PXF. 

Thanks for clarifying! If I understand the above (there's a chance I didn't), then I'd strongly prefer to keep it such that we are continually testing the "latest" pxf version and floating on HEAD so to speak. This will enable us to catch any potential regressions or gotchas early on, etc.

bradfordb-vmware:
> Is there a pipeline of this branch one could take a look at? I couldn't find one in the `cm` or `ud` Concourse instances.

Here is the dev pipeline that I've been using for testing this: https://ud.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:add-pxf-6. I've paused most of the other job in order to focus on the extensions group.

ci/scripts/load-extensions.bash
kalensk: Given the above is the commented bucket what we will want to use? That is, does this bucket and comment still align with the above?

```
    # after PXF 6.3.0 is released, the RPM should be fetched from standard bucket locations
    # json_key: ((concourse-gcs-resources-service-account-key))
    # bucket: pivotal-gpdb-concourse-resources-prod
    # regexp: pxf/published/gpdb{{.GPVersion}}/pxf-gp{{.GPVersion}}-6.(.*).el{{.CentosVersion}}.x86_64.rpm

kalensk (approved):
LGTM! (Some minor comments, but no major blockers)

Thanks so much for getting this up and everything! Super helpful for customers.

PR #642, opened 2022-02-18
Add heterogeneous partition table fix script

orhankislal:
Heterogeneous partition tables cannot be upgraded directly.
This commit detects them using the pg_upgrade code and
recreate/exchange the child tables.
Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:fix-hetero-part-tables

### DIFF
--- a/data-migration-scripts/pre-initialize/fix_heterogeneous_partition_tables.sql
+++ b/data-migration-scripts/pre-initialize/fix_heterogeneous_partition_tables.sql
@@ -1,0 +1,115 @@
+-- Copyright (c) 2017-2021 VMware, Inc. or its affiliates
+-- SPDX-License-Identifier: Apache-2.0
+
+-- Detect heterogenous partition tables and CTAS the affected leaf tables
+-- The detection query is based on the GPDB pg_upgrade code at:
+-- contrib/pg_upgrade/greenplum/check_gp.h
+-- check_heterogeneous_partition() in contrib/pg_upgrade/greenplum/check_gp.c
+-- We only handle scenario 1 referenced in check_heterogeneous_partition().
+-- Detection query used: CHECK_PARTITION_TABLE_DROPPED_COLUMN_REFERENCES
+
+SET client_min_messages TO WARNING;
+
+-- Use CREATE SCHEMA IF NOT EXISTS once it is supported
+CREATE SCHEMA __gpupgrade_tmp;
+CREATE OR REPLACE FUNCTION  __gpupgrade_tmp.fix_het()
+RETURNS VARCHAR AS
+$$
+import plpy
+
+swap_sql = ""
+
+res1 = plpy.execute("""
+SELECT cp1.childnamespace, cp1.childrelname, rp.parrelname, p3.schemaname, p3.partitionname, p3.partitionrank, p3.partitionposition, p3.parentpartitiontablename
+    FROM (
+            SELECT p.parrelid, rule.parchildrelid, n.nspname AS childnamespace, c.relname AS childrelname, c.relnatts AS childnatts,
+                   sum(CASE WHEN a.attisdropped THEN 1 ELSE 0 END) AS childnumattisdropped
+            FROM pg_catalog.pg_partition p
+                JOIN pg_catalog.pg_partition_rule rule ON p.oid=rule.paroid AND NOT p.paristemplate
+                JOIN pg_catalog.pg_class c ON rule.parchildrelid = c.oid AND NOT c.relhassubclass
+                JOIN pg_catalog.pg_namespace n ON n.oid = c.relnamespace
+                JOIN pg_catalog.pg_attribute a ON rule.parchildrelid = a.attrelid AND a.attnum > 0
+            GROUP BY p.parrelid, rule.parchildrelid, n.nspname, c.relname, c.relnatts
+        ) cp1
+        JOIN (
+            SELECT p.parrelid, min(c.relnatts) AS minchildnatts, max(c.relnatts) AS maxchildnatts
+            FROM pg_catalog.pg_partition p
+                JOIN pg_catalog.pg_partition_rule rule ON p.oid=rule.paroid AND NOT p.paristemplate
+                JOIN pg_catalog.pg_class c ON rule.parchildrelid = c.oid AND NOT c.relhassubclass
+            GROUP BY p.parrelid
+        ) cp2 ON cp2.parrelid = cp1.parrelid
+        JOIN (
+            SELECT c.oid, n.nspname AS parnamespace, c.relname AS parrelname, c.relnatts AS parnatts,
+                   sum(CASE WHEN a.attisdropped THEN 1 ELSE 0 END) AS parnumattisdropped
+            FROM pg_catalog.pg_partition p
+                JOIN pg_catalog.pg_class c ON p.parrelid = c.oid AND NOT p.paristemplate AND p.parlevel = 0
+                JOIN pg_catalog.pg_namespace n ON n.oid = c.relnamespace
+                JOIN pg_catalog.pg_attribute a ON c.oid = a.attrelid AND a.attnum > 0
+            GROUP BY c.oid, n.nspname, c.relname, c.relnatts
+        ) rp ON rp.oid = cp1.parrelid
+        JOIN pg_partitions p3 ON cp1.childrelname = p3.partitiontablename
+    WHERE NOT (rp.parnumattisdropped = 0 AND rp.parnatts = cp1.childnatts) AND
+          NOT (rp.parnumattisdropped > 0 AND cp2.minchildnatts = cp2.maxchildnatts AND
+               (rp.parnatts = cp1.childnatts OR cp1.childnumattisdropped = 0)) AND
+          NOT (rp.parnumattisdropped > 0 AND cp2.minchildnatts != cp2.maxchildnatts AND
+               cp2.minchildnatts < rp.parnatts AND cp1.childnumattisdropped = 0) AND
+          NOT (rp.parnumattisdropped > 0 AND cp2.minchildnatts != cp2.maxchildnatts AND
+               cp2.minchildnatts >= rp.parnatts)
+    ORDER BY rp.oid, cp1.parchildrelid;
+    """)
+if res1 is not None:
+    for i in res1:
+        schemaname = i["schemaname"]
+        partitionname = i["partitionname"]
+        partitionrank = i["partitionrank"]
+        parrelname = i["parrelname"]
+        childrelname = i["childrelname"]
+        partitionposition = i["partitionposition"]
+        parentpartitiontablename = i["parentpartitiontablename"]
+
+        partition_sql = ""
+        local_parentpartitiontablename = parentpartitiontablename
+        while local_parentpartitiontablename is not None:
+            local_vars = plpy.execute("""
+                SELECT parentpartitiontablename, partitionrank, partitionname
+                FROM pg_partitions
+                WHERE partitiontablename = '{local_parentpartitiontablename}' """.format(**locals()))[0]
+            local_partitionname = local_vars["partitionname"]
+            local_partitionrank = local_vars["partitionrank"]
+            if local_partitionname:
+                partition_sql = " ALTER PARTITION {local_partitionname} ".format(**locals()) + partition_sql
+            elif local_partitionrank:
+                partition_sql = " ALTER PARTITION FOR (RANK({local_partitionrank})) ".format(**locals()) + partition_sql
+            else:
+                plpy.error("Cannot read partition name or rank {local_parentpartitiontablename}".format(**locals()))
+
+            local_parentpartitiontablename = local_vars["parentpartitiontablename"]
+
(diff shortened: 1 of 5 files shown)


kalensk:
Super useful feature for customers and much needed. Thanks for working on this!

Looks like these files got changed accidentally. We should undo those changes:
- go.sum
- test/acceptance/gpupgrade/migration_scripts.bats

data-migration-scripts/pre-initialize/fix_heterogeneous_partition_tables.sql
```diff
@@ -0,0 +1,114 @@
...
+        if partitionname:
+            exchange_sql = " EXCHANGE PARTITION {partitionname} ".format(**locals())
+        elif partitionrank:
+            exchange_sql = " EXCHANGE PARTITION FOR (RANK({partitionrank})) ".format(**locals())
+        else:
+            plpy.error("Cannot read partition name or rank {0}".format(parentpartitiontablename))
+
+        swap_sql = swap_sql + """
+CREATE TABLE __gpupgrade_tmp.scratch_table (LIKE {schemaname}.{childrelname});
+INSERT INTO __gpupgrade_tmp.scratch_table SELECT * FROM {schemaname}.{childrelname};
```
soumyadeep2007, line 98: We can collapse this into a single CTAS:
`CREATE TABLE __gpupgrade_tmp.scratch_table AS SELECT * FROM {schemaname}.{childrelname};`

test/acceptance/gpupgrade/migration_scripts.bats
```diff
@@ -119,13 +119,13 @@ teardown() {
...
+    EXCLUSIONS+="-T ${TEST_SCHEMA}.heterogeneous_ml_partition_table "
 
     MIGRATION_DIR=`mktemp -d /tmp/migration.XXXXXX`
     register_teardown rm -r "$MIGRATION_DIR"
-
+    
     "$GPHOME_SOURCE"/bin/pg_dump --schema-only "$TEST_DBNAME" $EXCLUSIONS -f "$MIGRATION_DIR"/before.sql
 
-
     $SCRIPTS_DIR/gpupgrade-migration-sql-generator.bash $GPHOME_SOURCE $PGPORT $MIGRATION_DIR "$SCRIPTS_DIR"
```
soumyadeep2007, line 129: Nit: some whitespace diff here.

data-migration-scripts/pre-initialize/fix_heterogeneous_partition_tables.sql
```diff
@@ -0,0 +1,114 @@
...
+ALTER TABLE {schemaname}.{parrelname} {partition_sql} {exchange_sql} WITH TABLE __gpupgrade_tmp.scratch_table;
+DROP TABLE __gpupgrade_tmp.scratch_table;
+""".format(**locals())
+
+if swap_sql is not "":
+    swap_sql = """
+SET gp_enable_exchange_default_partition=on;
+CREATE SCHEMA __gpupgrade_tmp;
```
soumyadeep2007, line 106: There is no need to create the schema as we already do that above in line 11.

data-migration-scripts/pre-initialize/fix_heterogeneous_partition_tables.sql
```diff
@@ -0,0 +1,114 @@
...
+-- SPDX-License-Identifier: Apache-2.0
+
+-- Detect heterogenous partition tables and CTAS the affected leaf tables
+-- The detection query is based on the GPDB pg_upgrade code at:
+-- contrib/pg_upgrade/greenplum/check_gp.h
+
+
+SET client_min_messages TO WARNING;
+
+CREATE SCHEMA __gpupgrade_tmp;
```
soumyadeep2007, line 11: Should prefix the `CREATE SCHEMA` with `DROP SCHEMA IF EXISTS __gpupgrade_tmp;` to avoid ERRORs if multiple migration scripts creating the tmp schema are running consecutively.

soumyadeep2007 (approved):
LGTM. Some minor changes suggested.

data-migration-scripts/pre-initialize/fix_heterogeneous_partition_tables.sql
orhankislal: This was done to explicitly set the distribution. CTAS does not inherit the distributed by attribute.

data-migration-scripts/pre-initialize/fix_heterogeneous_partition_tables.sql
orhankislal: This CREATE SCHEMA is run during execution. By that time, we have dropped the temp schema because the generator cleans up after itself.

data-migration-scripts/pre-initialize/fix_heterogeneous_partition_tables.sql
orhankislal: We don't want to drop it because the views-related table might be already in there. The generator does not run the scripts consecutively and getting an error from CREATE SCHEMA does not fail the script. Once we move on from 5X we can use `CREATE SCHEMA IF NOT EXISTS` to avoid this issue.

data-migration-scripts/pre-initialize/fix_heterogeneous_partition_tables.sql
soumyadeep2007: Huh. Maybe we have a bug in CTAS?

I tried this on 5X and 6X both:

```
postgres=# create table foo(i int, j int) distributed by (j);
postgres=# create table fooreplica as select * from foo;
NOTICE:  Table doesn't have 'DISTRIBUTED BY' clause -- Using column(s) named 'j' as the Greenplum Database data distribution key for this table.
postgres=# select * from gp_distribution_policy ;
 localoid | attrnums 
----------+----------
    32768 | {2}
    32771 | {2}
(2 rows)
```

From the notice and gp_distribution_policy table, we are inheriting the distributed by attribute.

orhankislal: Maybe here is my log on 6X:
```
postgres=# create table t1 as select 1::int as a , 2::int as b distributed by (a);
SELECT 1
Time: 88.641 ms
postgres=# create table t2 as select * from t1;
NOTICE:  Table doesn't have 'DISTRIBUTED BY' clause. Creating a NULL policy entry.
SELECT 1
Time: 81.526 ms
postgres=# \d+ t2
                          Table "public.t2"
 Column |  Type   | Modifiers | Storage | Stats target | Description 
--------+---------+-----------+---------+--------------+-------------
 a      | integer |           | plain   |              | 
 b      | integer |           | plain   |              | 
Distributed randomly

postgres=# \d+ t1
                          Table "public.t1"
 Column |  Type   | Modifiers | Storage | Stats target | Description 
--------+---------+-----------+---------+--------------+-------------
 a      | integer |           | plain   |              | 
 b      | integer |           | plain   |              | 
Distributed by: (a)
```

jimmyyih: Copy/paste my comment from Slack about this:
Essentially, the root “issue” is related to ORCA… not on the EXCHANGE PARTITION but in the CTAS.  The CTAS w/ ORCA enabled makes the staging table randomly distributed with a NULL distribution policy.  Then during the EXCHANGE PARTITION, it panics trying to do comparisons on the NULL in an assertion (w/o asserts, it’ll give a proper ERROR).
So for the CTAS, we must provide the correct distribution policy (acquired by running SELECT pg_get_table_distributedby('schema.table'); ) and attach it to the end of the CTAS query.  Or simply disable ORCA with SET optimizer = off in the SQL session… but not sure the repercussions if there are any.

orhankislal: We can rewrite the script to use pg_get_table_distributedby but the LIKE statement sets the distribution as well even with ORCA. Is there a different issue with it?

shivzone: @orhankislal yes funcionally `CREATE TABLE LIKE` followed by `INSERT INTO` will do what you want, but from a performance standpoint would be a lot slower than directly doing a CTAS operation. That is why going with @jimmyyih approach of CTAS + setting the distribution policy explicitly is preferable.

jimmyyih: Yeah, mainly just performance difference.  The CTAS would usually be faster than the INSERT INTO SELECT due to multiple factors (e.g. indexes/constraints existing on the staging table, WAL generation, etc.).

vraghavan78: please set optimizer_enable_ctas = `off`; before the CTAS and reset it after the SQL

orhankislal: OK, I will make the change.

soumyadeep2007: Thanks for the explanation @orhankislal and @jimmyyih . I forgot about the issue w/ ORCA.

> We can rewrite the script to use pg_get_table_distributedby but the LIKE statement sets the distribution as well even with ORCA. Is there a different issue with it?

Great question! In general there are some performance gains w/ using CTAS as opposed to CTL + IIS, specially in the reduction of WAL, meaning it would be more performant if the customer is running in a mirrorless environment w/ `wal_level=minimal`. See https://dba.stackexchange.com/questions/189428/postgresql-why-is-create-table-as-faster-than-create-insert-into

However, for a 5X customer the ^ wouldn't be relevant, as we don't support mirrorless clusters on 5X, and this migration script and associated pg_upgrade check is a 5X->6X only thing. 

So I think its okay to do a CTL + IIS as long as we:
1. Update the docs to say exactly that for scenario 2 that is not covered by this script.
2. Update the pg_upgrade acceptance tests

Side q: what do we do here to ensure that this migration script runs only for upgrades from 5->6?

orhankislal: Thanks for the explanations folks, greatly appreciated. 
@soumyadeep2007 The note regarding scenario 2 is added.

soumyadeep2007: > Yeah, mainly just performance difference. The CTAS would usually be faster than the INSERT INTO SELECT due to multiple factors (e.g. indexes/constraints existing on the staging table, WAL generation, etc.).

Not a whole lot of difference on a mirrorless cluster, but then again I have a very small test case:
```
postgres=# insert into foo select generate_series(1, 1000000000);
INSERT 0 1000000000
Time: 411895.238 ms

postgres=# create table bar as select generate_series(1, 1000000000);
NOTICE:  Table doesn't have 'DISTRIBUTED BY' clause -- Using column(s) named 'generate_series' as the Greenplum Database data distribution key for this table.
HINT:  The 'DISTRIBUTED BY' clause determines the distribution of data. Make sure column(s) chosen are the optimal data distribution key to minimize skew.
SELECT 1000000000
Time: 389729.942 ms
```

But anyway, I don't mind toggling the optimizer setting either. All good!

shivzone: Any performance improvement here is significant considering the some users would likely have to alter many child partitions. Thanks for the inputs !

shivzone (approved):

data-migration-scripts/pre-initialize/fix_heterogeneous_partition_tables.sql
```diff
@@ -0,0 +1,115 @@
...
+-- Detection query used: CHECK_PARTITION_TABLE_DROPPED_COLUMN_REFERENCES
+
+SET client_min_messages TO WARNING;
+
+-- Use CREATE SCHEMA IF NOT EXISTS once it is supported
+CREATE SCHEMA __gpupgrade_tmp;
+CREATE OR REPLACE FUNCTION  __gpupgrade_tmp.fix_het()
+RETURNS VARCHAR AS
+$$
+import plpy
```
bhuvnesh2703, line 18: is plpy a requirement for this query? Will it be available at customer's environment?

bhuvnesh2703: If customer doesn't have PLPYTHONU created?

orhankislal: Yes, in the generator we do create the language if it is not there. gpupgrade-migration-sql-generator.bash lines 107-110

bhuvnesh2703: ok. thank you.

shivzone: The assumption here is also that existing 5X customers in production will have plpython installed, and that's why we `create language`

PR #645, opened 2022-02-24
Add test for pljava

beeender:
Add test for upgrading with for PL/Java.

Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:mc:pljava/jobs/5-to-6-extensions-centos-7

### DIFF
--- a/ci/1_resources_anchors_groups.yml
+++ b/ci/1_resources_anchors_groups.yml
@@ -214,23 +214,12 @@ resources:
     regexp: pxf/published/gpdb{{.GPVersion}}/pxf-gp{{.GPVersion}}-5.(.*).el{{.CentosVersion}}.x86_64.rpm
 {{- end }}
 
-{{- if ne .GPVersion "5" }}
-- name: pljava-gp{{.GPVersion}}-rhel{{.CentosVersion}}_x86_64_gppkg
-  type: s3
-  source:
-    access_key_id: ((bucket-access-key-id))
-    secret_access_key: ((bucket-secret-access-key))
-    region_name: us-west-2
-    bucket: procedural-languages-concourse-gpdb-5x-stable
-    versioned_file: gpdb5/pljava/pljava-1.4.4-gp5-rhel{{.CentosVersion}}-x86_64.gppkg
-{{- else }}
-- name: pljava-gp{{.GPVersion}}-rhel{{.CentosVersion}}_x86_64_gppkg
+- name: pljava_gpdb{{.GPVersion}}_rhel{{.CentosVersion}}_gppkg
   type: gcs
   source:
     json_key: ((concourse-gcs-resources-service-account-key))
     bucket: pivotal-gpdb-concourse-resources-prod
-    regexp: pljava/released/gpdb{{.GPVersion}}/pljava-(.*)-gp{{.GPVersion}}-rhel{{.CentosVersion}}_x86_64.gppkg
-{{- end }}
+    regexp: pljava/released/gpdb{{.GPVersion}}/pljava-(.*)-gp{{.GPVersion}}-rhel{{.CentosVersion}}-x86_64.gppkg
 {{end}}
 
 anchors:
--- a/ci/1_resources_anchors_groups.yml
+++ b/ci/1_resources_anchors_groups.yml
@@ -213,6 +213,24 @@ resources:
     bucket: pivotal-gpdb-concourse-resources-prod
     regexp: pxf/published/gpdb{{.GPVersion}}/pxf-gp{{.GPVersion}}-5.(.*).el{{.CentosVersion}}.x86_64.rpm
 {{- end }}
+
+{{- if ne .GpVersion "5" }}
+- name: pljava-gp{{.GPVersion}}-rhel{{.CentosVersion}}_x86_64_gppkg
+  type: s3
+  source:
+    access_key_id: ((bucket-access-key-id))
+    secret_access_key: ((bucket-secret-access-key))
+    region_name: us-west-2
+    bucket: procedural-languages-concourse-gpdb-5x-stable
+    versioned_file: gpdb5/pljava/pljava-1.4.4-gp5-rhel{{.CentosVersion}}-x86_64.gppkg
+{{- else }}
+- name: pljava-gp{{.GPVersion}}-rhel{{.CentosVersion}}_x86_64_gppkg
+  type: gcs
+  source:
+    json_key: ((concourse-gcs-resources-service-account-key))
+    bucket: pivotal-gpdb-concourse-resources-prod
+    regexp: pljava/released/gpdb{{.GPVersion}}/pljava-(.*)-gp{{.GPVersion}}-rhel{{.CentosVersion}}_x86_64.gppkg
+{{- end }}
 {{end}}
 
 anchors:
--- a/ci/1_resources_anchors_groups.yml
+++ b/ci/1_resources_anchors_groups.yml
@@ -214,7 +214,7 @@ resources:
     regexp: pxf/published/gpdb{{.GPVersion}}/pxf-gp{{.GPVersion}}-5.(.*).el{{.CentosVersion}}.x86_64.rpm
 {{- end }}
 
-{{- if ne .GpVersion "5" }}
+{{- if ne .GPVersion "5" }}
 - name: pljava-gp{{.GPVersion}}-rhel{{.CentosVersion}}_x86_64_gppkg
   type: s3
   source:
--- a/ci/scripts/load-extensions.bash
+++ b/ci/scripts/load-extensions.bash
@@ -19,6 +19,7 @@ scp gptext_targz/greenplum-text*.tar.gz gpadmin@mdw:/tmp/gptext.tar.gz
 scp postgis_gppkg_source/postgis*.gppkg gpadmin@mdw:/tmp/postgis_source.gppkg
 scp sqldump/*.sql gpadmin@mdw:/tmp/postgis_dump.sql
 scp madlib_gppkg_source/madlib*.gppkg gpadmin@mdw:/tmp/madlib_source.gppkg
+scp pljava_gppkg_source/pljava*.gppkg gpadmin@mdw:/tmp/pljava_source.gppkg
 
 echo "Installing extensions and sample data on source cluster..."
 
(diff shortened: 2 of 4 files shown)


ci/1_resources_anchors_groups.yml
```diff
@@ -213,6 +213,24 @@ resources:
     bucket: pivotal-gpdb-concourse-resources-prod
     regexp: pxf/published/gpdb{{.GPVersion}}/pxf-gp{{.GPVersion}}-5.(.*).el{{.CentosVersion}}.x86_64.rpm
 {{- end }}
+
+{{- if ne .GPVersion "5" }}
+- name: pljava-gp{{.GPVersion}}-rhel{{.CentosVersion}}_x86_64_gppkg
```
kalensk, line 218: Update the naem with the following to match the style of all other resource names. 
1) Remove `_x86_64` from the name
2) Use `gpdb` and not just `gp`.
2) Use only underscores with no dashes in the name. (Looks like gptext did it wrong and I'll update that shortly.)

ci/1_resources_anchors_groups.yml
```diff
@@ -213,6 +213,24 @@ resources:
     bucket: pivotal-gpdb-concourse-resources-prod
     regexp: pxf/published/gpdb{{.GPVersion}}/pxf-gp{{.GPVersion}}-5.(.*).el{{.CentosVersion}}.x86_64.rpm
 {{- end }}
+
+{{- if ne .GPVersion "5" }}
```
kalensk, line 217: Add a detailed comment explaining why we need this. See pxf for a possible example.

ci/1_resources_anchors_groups.yml
```diff
@@ -213,6 +213,24 @@ resources:
...
+- name: pljava-gp{{.GPVersion}}-rhel{{.CentosVersion}}_x86_64_gppkg
+  type: s3
+  source:
+    access_key_id: ((bucket-access-key-id))
+    secret_access_key: ((bucket-secret-access-key))
+    region_name: us-west-2
+    bucket: procedural-languages-concourse-gpdb-5x-stable
+    versioned_file: gpdb5/pljava/pljava-1.4.4-gp5-rhel{{.CentosVersion}}-x86_64.gppkg
+{{- else }}
+- name: pljava-gp{{.GPVersion}}-rhel{{.CentosVersion}}_x86_64_gppkg
```
kalensk, line 227: Please refactor this to not need an else. There is a way to refactor the "if clause" to have only one resource type that has both resources centos6 and centos7 for gpdb6. If you need help please reach out on Slack to pair.

ci/scripts/load-extensions.bash
```diff
@@ -232,6 +233,37 @@ SQL_EOF
...
+                RETURNS int
+                AS 'org.postgresql.pljava.example.Parameters.addOne(java.lang.Integer)'
+            IMMUTABLE LANGUAGE java;
+
+            SELECT java_addOne(42);
+SQL_EOF
+"
+}
+
+install_pljava()
```
kalensk, line 265: There is no need to make this a separate function if it runs on both centos6 and centos6. If not, see pxf and how there is an if condition. Please add that condition with proper comments if needed.

ci/scripts/load-extensions.bash
```diff
@@ -232,6 +233,37 @@ SQL_EOF
 
 test_pxf "$OS_VERSION" && install_pxf || echo "Skipping pxf for centos6 since pxf5 for GPDB6 on centos6 is not supported..."
 
+install_pljava() {
+    echo "Installing pljava..."
+    ssh -n mdw "
+        set -eux -o pipefail
+
+        echo "export JAVA_HOME=/usr/lib/jvm/jre" >> /usr/local/greenplum-db-source/greenplum_path.sh
```
kalensk, line 241: This cannot occur. We should never be changing greenplum_path.sh like this as it affects everything!

ci/scripts/load-extensions.bash
```diff
@@ -232,6 +233,37 @@ SQL_EOF
...
+        source /usr/local/greenplum-db-source/greenplum_path.sh
+
+        echo 'Initialize pljava...'
+        export GPHOME=$GPHOME_SOURCE
```
kalensk, line 245: No need to do this I think. Also if it happens to be needed the local variable needs to be called GPHOME_SOURCE.

kalensk (commented):
Some initial feedback for now...

kalensk:
FYI it looks like the extension job for the test pipeline is not running...

https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:mc:pljava/jobs/5-to-6-extensions-centos-7

ci/scripts/load-extensions.bash
beeender: Then we have a problem, `pljava` requires `JAVA_HOME` to find the `libjvm.so` :(
https://github.com/greenplum-db/pljava/blob/8ad2f2732057ed46b2507461dadfd1c6ba8450e4/gpdb/packaging/pljava.spec#L21

ci/1_resources_anchors_groups.yml
beeender: pljava gpdb5 is on s3 for now. If it is really necessary, we can copy it to gcs.

beeender:
> PL/Java releases before 1.5.1 were not aware of pg_upgrade operation. To avoid possible errors during the upgrade involving OID or object clashes, the PL/Java release installed for the new cluster should be 1.5.1 or later.

From https://tada.github.io/pljava/install/upgrade.html

This might be an issue. Currently gpdb pljava release is based on pljava `1.5.0`. Need to check

ci/scripts/upgrade-extensions.bash
```diff
@@ -88,6 +90,14 @@ time ssh -n mdw "
         /usr/local/pxf-gp6/bin/pxf cluster init
     fi
 
+    JAVA_HOME=/usr/lib/jvm gppkg -i /tmp/pljava_target.gppkg
+    # pljava installer will modify LD_LIBRARAY_PATH in the greenplum_path.sh.
+    # And the same modifications needs to be done on all the segments to make sure
+    # they can discover libjvm.so
+    gpscp -f /home/gpadmin/segment_host_list $GPHOME_TARGET/greenplum_path.sh  =:$GPHOME_TARGET/greenplum_path.sh
+    source /usr/local/greenplum-db-target/greenplum_path.sh
+    gpstop -ar
```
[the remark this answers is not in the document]
beeender: Does gpupgrade set `LD_LIBRARY_PATH` somewhere so the `pg_upgrade` didn't see the `LD_LIBRARY_PATH` set in the `greenplum_path.sh`?

beeender:
Possibly related to this commit [e05905560f9b738fec934d19ece960a0e22dc4e3](https://github.com/tada/pljava/commit/e05905560f9b738fec934d19ece960a0e22dc4e3)

ci/scripts/load-extensions.bash
```diff
@@ -232,6 +233,43 @@ SQL_EOF
...
+
+    echo 'Loading pljava data...'
+    psql -v ON_ERROR_STOP=1 -d postgres -f $GPHOME_SOURCE/share/postgresql/pljava/install.sql
+    psql -v ON_ERROR_STOP=1 -d postgres <<SQL_EOF
+    CREATE FUNCTION java_addOne(int)
+    RETURNS int
+    AS 'org.postgresql.pljava.example.Parameters.addOne(java.lang.Integer)'
+    IMMUTABLE LANGUAGE java;
+
+    SELECT java_addOne(42);
```
orhankislal, line 268: We might want to use a view:
`CREATE VIEW pljava_test_view AS SELECT java_addOne(42);`

ci/scripts/load-extensions.bash
```diff
@@ -232,6 +233,43 @@ SQL_EOF
...
+    export JAVA_HOME=/usr/lib/jvm
+
+    echo 'Initializing pljava...'
+
+    gppkg -i /tmp/pljava_source.gppkg
+
+    # pljava installer will modify LD_LIBRARAY_PATH in the greenplum_path.sh.
+    # And the same modifications needs to be done on all the segments to make sure
+    # they can discover libjvm.so
+    gpscp -f /home/gpadmin/segment_host_list $GPHOME_SOURCE/greenplum_path.sh  =:$GPHOME_SOURCE/greenplum_path.sh
```
orhankislal, line 251: I don't see this step in the pljava documentation. Is the docs webpage outdated, or is there a CI-specific issue?

ci/scripts/upgrade-extensions.bash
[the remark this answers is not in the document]
orhankislal: `dynamic_library_path` GUC is used when you load an so file into the database with this command: https://gpdb.docs.pivotal.io/5280/ref_guide/sql_commands/LOAD.html. This is what `create extension` uses to load the `/usr/local/greenplum-db-source/lib/postgresql/pljava.so` file. 

@kalensk In multihost/gpupgrade_cluster_config we set the following:
`LD_LIBRARY_PATH=/usr/local/greenplum-db/lib`
we should try `LD_LIBRARY_PATH=/usr/local/greenplum-db/lib:$LD_LIBRARY_PATH` to see if that fixes this issue.

ci/scripts/load-extensions.bash
```diff
@@ -232,6 +233,43 @@ SQL_EOF
...
+    gpconfig -c pljava_classpath -v 'examples.jar'
+
+    # Source again to get the new LD_LIBRARAY_PATH
+    source /usr/local/greenplum-db-source/greenplum_path.sh
+    # Restart the cluster to reload LD_LIBRARAY_PATH
+    gpstop -ra
+
+
```
orhankislal, line 259: Extra whitespace

ci/scripts/upgrade-extensions.bash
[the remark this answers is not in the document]
kalensk: We no longer rely on `LD_LIBRARY_PATH`. Any references to that in the codebase should be removed as soon as possible, as its no longer used (ie: thats leftover tech debt). It is a critical bug/issue if things rely on `LD_LIBRARY_PATH`. Let me know if I should dig around to pul up the specific commits.

ci/scripts/load-extensions.bash
kalensk: I don't know much on the specifics of pl/java, but for extensions its absolutely critical that we exercise them. As documented in the "Upgrading with Extensions" Confluence page, the following (or similar) is necessary to test the extensions:

> it is recommended to create a table with some sample data utilizing that type. 
If the extensions defines a user defined function, then it is recommended to create a view dependent on that function. It is important to test functions that depend on .so library files in order to test the .so files.

orhankislal: It is not a 100% necessity for pljava (or other language extensions like plr) since we already defined a function directly referencing the extension. 
This is just a low-cost change to add another check and keep the pattern with the other extensions.

kalensk: Sounds good. And yeah the more coverage the better (within reason).

ci/scripts/upgrade-extensions.bash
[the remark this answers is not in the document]
beeender: postgres pljava support `pljava.libjvm_location` to specify the `libjvm.so`. But, that was removed in gpdb pljava https://github.com/greenplum-db/pljava/commit/b638471280d4265e8d0435c7ca627c56feb42ff0 due to lack of custom guc support in gpdb5. maybe it is the time we bring that back.
Is setting a guc before gpupgrade acceptable to solve this issue?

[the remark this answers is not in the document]
beeender: PLJava & gpupgrade

• Customer on GPDB5 with pljava (Recommend to update the pljava to the latest on gpdb5)
• Customer installed latest gpdb6 rpms.
◦ Customer run gpupgrade initialize which creates a blank 6x cluster.
• Customer install the latest (not released) pljava pgpkg on 6x cluster.
• Customer use gpconfig to set the custom guc `pljava.libjvm_location` . (put this into gpupgrade documentation)
• Customer proceed gpupgrade process

ci/scripts/load-extensions.bash
kalensk: Sorry for not being clear. Just to clarify, I was simply concerned with changing `greenplum_path.sh`. If it works you can export JAVA_HOME directly before using it like [pxf5 does](https://github.com/greenplum-db/gpupgrade/blob/1.2.0/ci/scripts/load-extensions.bash#L135). (Note that link is for 1.2.0 tag which is for pxf5 which is different than existing code which is for pxf6. That is, the code changed.)

PR #647, opened 2022-02-28
Add test for plr

beeender:


beeender:
CI can be seen here https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:mc:plr

ci/1_resources_anchors_groups.yml
```diff
@@ -213,8 +213,17 @@ resources:
...
+- name: plr_gpdb{{.GPVersion}}_rhel{{.CentosVersion}}_gppkg
+  type: gcs
+  source:
+    json_key: ((concourse-gcs-resources-service-account-key))
+    bucket: pivotal-gpdb-concourse-resources-prod
+    regexp: plr/released/gpdb{{.GPVersion}}/plr-(.*)-gp{{.GPVersion}}-rhel{{.CentosVersion}}-x86_64.gppkg
+
 {{end}}
 
+
```
orhankislal, line 226: Unnecessary whitespace

ci/scripts/load-extensions.bash
```diff
@@ -232,6 +233,45 @@ SQL_EOF
 
 test_pxf "$OS_VERSION" && install_pxf || echo "Skipping pxf for centos6 since pxf5 for GPDB6 on centos6 is not supported..."
 
+echo 'Installing plr dependencies...'
+mapfile -t hosts < cluster_env_files/hostfile_all
+for host in "${hosts[@]}"; do
+    ssh -n "centos@${host}" "
+        set -eux -o pipefail
+
+        sudo yum install -y R
```
orhankislal, line 242: We should move this line to the dependency loop at line 27-39

ci/scripts/load-extensions.bash
```diff
@@ -232,6 +233,45 @@ SQL_EOF
...
+    "
+done
+
+echo "Installing plr..."
```
orhankislal, line 246: We should add this block to the end of the main gpdb extensions block at line 41-94

ci/scripts/load-extensions.bash
```diff
@@ -232,6 +233,45 @@ SQL_EOF
...
+    CREATE EXTENSION plr;
+    CREATE OR REPLACE FUNCTION r_norm(n integer, mean float8,
+        std_dev float8) RETURNS float8[ ] AS
+    \\\$\\\$
+        x<-rnorm(n,mean,std_dev)
+        return(x)
+    \\\$\\\$
+    LANGUAGE 'plr';
+
+    CREATE TABLE test_norm_var
```
orhankislal, line 268: We should create a view instead of a table like the other extensions. A table uses the function, a view creates a lasting dependency on the function.

ci/scripts/load-extensions.bash
beeender: not sure about this, This block installs the plr like the previous block `test_pxf xxxx` is doing. Should this be moved?

kalensk: Yeah its a bit confusing, but pxf is in its own block because its only supported for centos7 and thus we needed a function `test_pxf "$OS_VERSION" && install_pxf ` to call it when appropriate. 

All other extensions that work as normal fall into two main blocks "GPDB extensions" and then "postgres native extensions". 

I squashed the commits and moved some stuff around. All good!

ci/scripts/load-extensions.bash
```diff
@@ -91,6 +96,23 @@ SQL_EOF
         CREATE VIEW madlib_test_view AS SELECT madlib.normal_quantile(0.5, 0, 1);
         CREATE VIEW madlib_test_agg AS SELECT madlib.mean(value) FROM madlib_test_type;
 SQL_EOF
+
+    echo 'Installing plr...'
+    gppkg -i /tmp/plr_source.gppkg
+
+    # plr added R to the LD_LIBRARY_PATH in greenplum_path.sh
+    # source /usr/local/greenplum-db-source/greenplum_path.sh
+    # gpstop -ra
```
kalensk, line 105: I commented this out for now to see if its "really" needed. I'm hoping its not.

beeender: I think it is needed ... I have been there and met a red job before.

kalensk (approved):

PR #649, opened 2022-03-04
Add test for plcontainer

RMTT:
Install plcontainer for source and target cluster.

I don't add sample data and sql for plcontainer now, because add such things we need install docker and some images for every host, it's cumbersome i think.

Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/rdu:gpupgrade:master?group=extensions

### DIFF
--- a/ci/1_resources_anchors_groups.yml
+++ b/ci/1_resources_anchors_groups.yml
@@ -223,6 +223,27 @@ resources:
     bucket: pivotal-gpdb-concourse-resources-prod
     regexp: plr/released/gpdb{{.GPVersion}}/plr-(.*)-gp{{.GPVersion}}-rhel{{.CentosVersion}}-x86_64.gppkg
 
+# NOTE: plcontainer is only supported for centos7 so skip creating the resource
+# for centos6. plcontainer 1.x is only for GPDB 5X.
+{{- if and (ne .CentosVersion "6") (eq .GPVersion "5") }}
+- name: plcontainer_1.x_gpdb{{.GPVersion}}_rhel{{.CentosVersion}}_gppkg
+  type: gcs
+  source:
+    json_key: ((concourse-gcs-resources-service-account-key))
+    bucket: pivotal-gpdb-concourse-resources-prod
+    regexp: plcontainer/released/gpdb{{.GPVersion}}/plcontainer-1.(.*)-rhel{{.CentosVersion}}-x86_64.gppkg
+{{- end }}
+
+# NOTE: plcontainer is only supported for centos7 so skip creating the resource
+# for centos6. plcontainer 2.x is only for GPDB 6X.
+{{- if and (ne .CentosVersion "6") (ne .GPVersion "5") }}
+- name: plcontainer_2.x_gpdb{{.GPVersion}}_rhel{{.CentosVersion}}_gppkg
+  type: gcs
+  source:
+    json_key: ((concourse-gcs-resources-service-account-key))
+    bucket: pivotal-gpdb-concourse-resources-prod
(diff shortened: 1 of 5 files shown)


ci/scripts/upgrade-extensions.bash
```diff
@@ -21,6 +21,7 @@ export PGPORT=5432
 echo "Copying extensions to the target cluster..."
 scp postgis_gppkg_target/postgis*.gppkg gpadmin@mdw:/tmp/postgis_target.gppkg
 scp madlib_gppkg_target/madlib*.gppkg gpadmin@mdw:/tmp/madlib_target.gppkg
+scp plcontainer_gppkg_target/*.gppkg gpadmin@mdw:/tmp/plcontainer_target.gppkg
```
beeender, line 24: I think at least we can add a UDF which uses plcontainer to make sure the plcontainer related UDF doesn't break the gpupgrade process.

The UDF doesn't have to be runnable if there is no images.

ci/1_resources_anchors_groups.yml
```diff
@@ -213,6 +213,28 @@ resources:
...
+- name: plcontainer_1.6.0_rhel{{.CentosVersion}}_x86_64.gppkg
+  type: gcs
+  source:
+    json_key: ((concourse-gcs-resources-service-account-key))
+    bucket: pivotal-gpdb-concourse-resources-prod
+    regexp: plcontainer/released/gpdb{{.GPVersion}}/plcontainer-(1.6.0)-rhel{{.CentosVersion}}-x86_64.gppkg
+{{- end }}
+{{- end }}
+
+{{- if ne .CentosVersion "6" }}
```
orhankislal, line 228: We can remove this if check and combine them under one `{{- if ne .CentosVersion "6" }}`

ci/scripts/load-extensions.bash
```diff
@@ -91,6 +92,9 @@ SQL_EOF
         CREATE VIEW madlib_test_view AS SELECT madlib.normal_quantile(0.5, 0, 1);
         CREATE VIEW madlib_test_agg AS SELECT madlib.mean(value) FROM madlib_test_type;
 SQL_EOF
+
+    echo 'Installing plcontainer...'
+    gppkg -i /tmp/plcontainer_source.gppkg
```
orhankislal, line 97: I agree with @beeender's comment below regarding the sample usage code.

ci/1_resources_anchors_groups.yml
```diff
@@ -213,6 +213,28 @@ resources:
...
+{{- if ne .GPVersion "5" }}
+- name: plcontainer_2.1.5_gp{{.GPVersion}}_rhel{{.CentosVersion}}_x86_64.gppkg
```
orhankislal, line 230: `gp{{.GPVersion}}` -> `gpdb{{.GPVersion}}`

ci/scripts/upgrade-extensions.bash
beeender: BTW, and create a `VIEW` which uses the UDF to test if there is a similar problem we met with gptext.

RMTT:
I have updated this branch for your comments. @beeender @orhankislal

kalensk:
I made a bunch of changes to allow the extensions-centos-6 job to run since there is no plcontainer for centos6 (just centos7).

In the future creating a branch (ie: not on "master") makes things a lot easier. Happy to help pair on that!

Some personal notes for myself on working with this particular branch since it is named master :
```
git co plcontainer
git push --force-w RMTT HEAD:master
BRANCH=master PIPELINE_NAME=rdu:gpupgrade:master make set-pipeline
```

PR #650, opened 2022-03-10
Update gptext resource

beeender:
To always use the latest release


beeender:
https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:mc:update_gptext_res

kalensk:
Thanks for putting this up! This will be great to catch any potential regressions or errors as new versions of gptext are developed.

I moved the existing comment above the "if" condition and added `_targz` to the resource name to match the style of the other resources. 

I will approve and merge once pipeline is finalized.

PR #662, opened 2022-04-14
CI: unset LD_LIBRARY_PATH

kalensk:
During teardown our BATS test call gpdeletesystem which stops the postgres processes. This was failing with:

> Error: unable to import module: /usr/local/greenplum-db-6.20.2+dev.3.g24b949d/lib/libpq.so.5: symbol gss_acquire_cred_from, version gssapi_krb5_2_MIT not defined in file libgssapi_krb5.so.2 with link time reference

This is most likely because the LD_LIBRARY_PATH is already set in the parent environment and when sourcing the target greenplum_path.sh this also sets the LD_LIBRARY_PATH causing it to be polluted with both the source and target cluster values. Thus, ensure LD_LIBRARY_PATH is unset before sourcing the target greenplum_path.sh.

Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:fixCI_5-to-6-cluster-tests

### DIFF
--- a/test/acceptance/gpupgrade/execute.bats
+++ b/test/acceptance/gpupgrade/execute.bats
@@ -183,7 +183,7 @@ ensure_hardlinks_for_relfilenode_on_master_and_segments() {
     restore_cluster
 
     # Put the source and target clusters back the way they were.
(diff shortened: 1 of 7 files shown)


kyeap-vmware (approved):
Looks good.

PR #663, opened 2022-04-27
CI: pull only pxf6

kalensk:
Previously, the pxf pipelines were pushing both pxf5 and pxf6 snapshot builds to the same bucket with the same name. This did not allow downstream consumers such as gpupgrade the ability to differentiate between pxf versions. In PR https://github.com/greenplum-db/pxf/pull/801 pxf 6 snapshot builds are pushed to a different bucket allowing gpupgrade to pull only pxf 6 builds.

Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:pullSpecificPxfVersion/jobs/5-to-6-extensions-centos-7

**_Note:_** After merging un-pin the pxf resources in our prod CI.

### DIFF
--- a/ci/1_resources_anchors_groups.yml
+++ b/ci/1_resources_anchors_groups.yml
@@ -211,7 +211,7 @@ resources:
   source:
     json_key: ((pxf-storage-service-account-key))
     bucket: data-gpdb-ud-pxf-build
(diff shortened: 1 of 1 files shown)


kalensk:
**_Note:_** After merging un-pin the pxf resources in our prod CI.

bradfordb-vmware (approved):

PR #667, opened 2022-05-07
internally rename master to coordinator

kalensk:
Internally rename master to coordinator.

But keep public facing text to master since gpupgrade currently only supports
GPDB 5 to GPDB 6 upgrades. Rename to coordinator once GPDB 6 to GPDB 7
upgrade is supported.

Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:rename


greenplum/start_or_stop_cluster_test.go
```diff
@@ -298,29 +298,29 @@ func TestIsMasterRunning(t *testing.T) {
 	source.GPHome = "/usr/local/source"
```
kyeap-vmware, line 298: I can't select the exact line because it is unchanged. Line 289 in `greenplum/start_or_stop_cluster_test.go`
TestIsMasterRunning is missing the test function rename. Change from `TestIsMasterRunning` to `TestIsCoordinatorRunning`.

kyeap-vmware (approved):

PR #668, opened 2022-05-12
Remove greenplum.TablespaceInfo in favor of idl.TablespaceInfo

kyeap-vmware:
greenplum.TablespaceInfo and idl.TablespaceInfo is the same struct. Use
the protobuf definition directly.

Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:refactorTSInfo


hub/upgrade_mirrors_using_rsync.go
```diff
@@ -125,7 +125,7 @@ func RsyncMirrorTablespacesOnSegments(agentConns []*idl.Connection, source *gree
 			intermediateMirror := intermediate.Mirrors[sourcePrimary.ContentID]
 
 			for tsOid, sourcePrimaryTsInfo := range source.Tablespaces[sourcePrimary.DbID] {
-				if !sourcePrimaryTsInfo.IsUserDefined() {
+				if !sourcePrimaryTsInfo.UserDefined {
```
kalensk, line 128: All calls throughout this PR should be `GetUserDefined` instead of `UserDefined` to better protect against nulls. Its just a small additional level of safety that is easy to do.

hub/upgrade_primaries.go
```diff
@@ -39,7 +39,7 @@ func UpgradePrimaries(agentConns []*idl.Connection, source *greenplum.Cluster, i
 				NewDataDir:    intermediatePrimary.DataDir,
 				NewPort:       strconv.Itoa(intermediatePrimary.Port),
 				NewDBID:       strconv.Itoa(intermediatePrimary.DbID),
-				Tablespaces:   getProtoBufSegmentTablespaces(source.Tablespaces, intermediatePrimary.DbID),
+				Tablespaces:   map[int32]*idl.TablespaceInfo(source.Tablespaces[intermediatePrimary.DbID]),
```
kalensk, line 42: No need for casting to the map type (intelliJ can show you this). Thus, simplify the line to:

`Tablespaces:   source.Tablespaces[intermediatePrimary.DbID],`

greenplum/tablespace.go
```diff
@@ -134,10 +125,10 @@ func (t TablespaceTuples) Write(w io.Writer) error {
 	for _, tablespace := range t {
 		line := []string{
 			strconv.Itoa(tablespace.DbId),
-			strconv.Itoa(tablespace.Oid),
+			strconv.Itoa(int(tablespace.Oid)),
 			tablespace.Name,
 			tablespace.Info.Location,
-			strconv.Itoa(tablespace.Info.UserDefined)}
+			strconv.Itoa(Btoi(tablespace.Info.UserDefined))}
```
kalensk, line 131: We can reduce this to a single function call if `Btoi` return a string version of 0 or 1. (We would probably rename the function to illustrate that as well). This makes more sense to me since this is the only place in the code that needs that. This would be a different situation if it was needed in three or more places. 

Thus, something like `boolToString(tablespace.Info.UserDefined)` (function name not finalized). I'd also not export the function.

greenplum/tablespace.go
```diff
@@ -35,24 +35,19 @@ const tablespacesQuery = `
 		) upgrade_tablespace`
 
 // map<tablespaceOid, tablespaceInfo>
-type SegmentTablespaces map[int]TablespaceInfo
+type SegmentTablespaces map[int32]*idl.TablespaceInfo
 
 // map<DbID, map<tablespaceOid, tablespaceInfo>>
 type Tablespaces map[int]SegmentTablespaces
```
kalensk, line 41: For consistency lets make this use an int32 (`map[int32]SegmentTablespaces`) and also update DbId below to be int32.

kalensk (approved):
LGTM!

Just remember to squash the two commits before merging, and ensure pipeline is green.

PR #670, opened 2022-05-18
Additional setup for test that verifies default port usage

kyeap-vmware:
Sets stateDir to isolate test and verifies the config file does not
exist.

Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:clean-test-env

### DIFF
--- a/cli/commands/commands_test.go
+++ b/cli/commands/commands_test.go
@@ -46,6 +46,14 @@ func TestGetHubPort(t *testing.T) {
 	})
 
 	t.Run("uses default port if the config file does not exist", func(t *testing.T) {
+		stateDir := testutils.GetTempDir(t, "")
+		defer testutils.MustRemoveAll(t, stateDir)
+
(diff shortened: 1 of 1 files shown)


cli/commands/commands_test.go
```diff
@@ -46,6 +48,21 @@ func TestGetHubPort(t *testing.T) {
...
+
+		// confirm config file does not exist
+		configFile := upgrade.GetConfigFile()
+		if _, err := os.Stat(configFile); err == nil {
+			t.Errorf("found unexpected config file %q", configFile)
+		} else if errors.Is(err, os.ErrNotExist) {
+		} else {
+			t.Errorf("got unexpected error %#v", err)
+		}
+
```
kalensk, line 65: Let's use `testutils.PathMustNotExist()`. You can see the other callers to that function for examples.

kalensk: We don't really need to test that the config file doesn't exist since its guaranteed to not exist with a state directory set as a tempdir.

PR #671, opened 2022-05-19
check active connections before execute and finalize

kalensk:
pg_upgrade and other steps will fail if there are active connections to the database. Thus, check them upfront before proceeding.

Pipeline https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:activeConns


kyeap-vmware (approved):
Looks good.

PR #679, opened 2022-06-06
Add child partition ownership tests for pg_upgrade.

bmdoil:
https://github.com/greenplum-db/gpdb/pull/13622 resolves an issue where `ALTER TABLE ... OWNER TO` 
on a partition parent table was not propagating to child partitions during pg_upgrade. 

This PR adds tests to exercise the fix.

### DIFF
--- a/test/acceptance/pg_upgrade/upgradeable_tests/source_cluster_regress/sql/partitioned_heap_table.sql
+++ b/test/acceptance/pg_upgrade/upgradeable_tests/source_cluster_regress/sql/partitioned_heap_table.sql
@@ -101,3 +101,11 @@ INSERT INTO dropped_and_added_column SELECT i, i, i, i FROM generate_series(1, 1
 ALTER TABLE dropped_and_added_column DROP COLUMN b;
 ALTER TABLE dropped_and_added_column ADD COLUMN e int;
 INSERT INTO dropped_and_added_column SELECT i, i, i, i FROM generate_series(10, 20) i;
+
+---
+--- partitioned table with alter owner
+---
+
+CREATE ROLE testrole;
+CREATE TABLE p_alter_owner (id INTEGER, name TEXT) DISTRIBUTED BY (id) PARTITION BY RANGE(id) (START(1) END(3) EVERY(1));
+ALTER TABLE p_alter_owner OWNER TO testrole;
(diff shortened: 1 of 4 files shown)


test/acceptance/pg_upgrade/upgradeable_tests/source_cluster_regress/sql/partitioned_heap_table.sql
```diff
@@ -101,3 +101,14 @@ INSERT INTO dropped_and_added_column SELECT i, i, i, i FROM generate_series(1, 1
...
+---
+--- partitioned table with alter owner
+---
+
+CREATE ROLE testrole;
+CREATE TABLE p_basic_alterowner (id INTEGER, name TEXT) DISTRIBUTED BY (id) PARTITION BY RANGE(id) (START(1) END(3) EVERY(1));
+ALTER TABLE p_basic_alterowner OWNER TO testrole;
+SELECT c.relname, pg_catalog.pg_get_userbyid(c.relowner) as owner
+FROM pg_class c
+WHERE relname like '%p_basic_alterowner%';
```
kalensk, line 114: super minor suggestions.

- Consider renaming to `p_alter_owner` or similar as I think "basic" isn't adding much value and adds to the name clouding the actual purpose which is altering the owner.
- Consider removing the SELECT statement from the source cluster tests, and only keeping it in the target_cluster tests. Asserting that the target cluster has the expected owner is the focus of this pg_upgrade test. Whereas testing that alter succeeds on the source cluster is a different type of test which should already be there in the main server tests (and not necessarily here). In this case, it's more clear to me to have a test focused on a single aspect that we are concerned with - upgrading the owner.

Again not a big deal, just some possible considerations.

kalensk (approved):
General note, merge only after the related pg_upgrade PR has been merged **and** gone through the pipeline to build a release candidate for gpupgrade pipeline to consume.

test/acceptance/pg_upgrade/upgradeable_tests/target_cluster_regress/sql/partitioned_heap_table.sql
```diff
@@ -22,3 +22,7 @@ SELECT b, c FROM dropped_column WHERE a=10;
 SELECT b, c FROM root_has_dropped_column WHERE a=10;
 
 SELECT c, d FROM dropped_and_added_column WHERE a=10;
+
+SELECT c.relname, pg_catalog.pg_get_userbyid(c.relowner) as owner
+FROM pg_class c
+WHERE relname like '%p_basic_alterowner%';
```
soumyadeep2007, line 28: Nit: the leading % is unnecessary.

soumyadeep2007 (approved):

PR #691, opened 2022-07-22
data-migration-script to drop parent partitions with seg entries

kalensk:
data-migration-script to drop parent partitions with seg entries

Turns out the ICW regression database has more AO and AOCO parent partitions with seg entries which is currently causing the 5 to 6 upgrade jobs to fail. Thus, the easiest way to address all these tables seems to add a pre-initialize data migration script. There is no need for a post-finalize or post-revert script since the seg entries are not needed.

Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:fixParentPartitionsWithSegEntries

### DIFF
--- a/data-migration-scripts/pre-initialize/gen_drop_parent_partitions_with_seg_entries.sql
+++ b/data-migration-scripts/pre-initialize/gen_drop_parent_partitions_with_seg_entries.sql
@@ -1,0 +1,19 @@
+-- Copyright (c) 2017-2022 VMware, Inc. or its affiliates
+-- SPDX-License-Identifier: Apache-2.0
+
+-- Generates a script to truncate non-empty segrels for AO and AOCO parent
+-- partitions.
+
+SELECT 'SET allow_system_table_mods TO DML;';
+
+SELECT 'DELETE FROM ' || segrelid::regclass || ';'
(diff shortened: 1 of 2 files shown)


data-migration-scripts/pre-initialize/gen_drop_parent_partitions_with_seg_entries.sql
```diff
@@ -0,0 +1,16 @@
+-- Copyright (c) 2017-2022 VMware, Inc. or its affiliates
+-- SPDX-License-Identifier: Apache-2.0
+
```
soumyadeep2007, line 3: Add a header like the other migration scripts do?

```
-- generates a script to truncate non-empty segrels for AO and AOCO parent partitions.
```

soumyadeep2007 (approved):
LGTM!

bmdoil (approved):
+1

data-migration-scripts/pre-initialize/gen_drop_parent_partitions_with_seg_entries.sql
kalensk: Thanks! I knew I forgot something, but didn't realize it...

PR #692, opened 2022-07-22
Enable option ON_ERROR_STOP on psql calls

kyeap-vmware:
If a sql file is executed by psql without this option turned on, any
errors in sql execution will be ignored and psql will keep executing sql
statements. The resulting error code will be the status of the last SQL
statement executed. By forcing psql to stop if an error is encountered
we can stop SQL execution and return with a non zero error code.

Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:ON_ERROR_STOP2

### DIFF
--- a/ci/scripts/load-retail-data.bash
+++ b/ci/scripts/load-retail-data.bash
@@ -59,8 +59,8 @@ ssh mdw <<EOF
 
     source ${GPHOME_SOURCE}/greenplum_path.sh
     cd /home/gpadmin/industry_demo
-    psql -d template1 -e -f data_generation/prep_database.sql
-    psql -d gpdb_demo -e -f data_generation/prep_external_tables.sql
+    psql -v ON_ERROR_STOP=1 -d template1 -e -f data_generation/prep_database.sql
+    psql -v ON_ERROR_STOP=1 -d gpdb_demo -e -f data_generation/prep_external_tables.sql
 EOF
 
 # copy extracted demo_data to segments and start gpfdist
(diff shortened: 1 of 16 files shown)


data-migration-scripts/test/create_nonupgradable_objects.sql
```diff
@@ -49,7 +49,8 @@ CREATE INDEX sales_idx_bitmap on sales using bitmap(office_id);
 CREATE INDEX sales_1_prt_2_idx on sales_1_prt_2(office_id, region);
 CREATE INDEX sales_1_prt_3_2_prt_asia_idx on sales_1_prt_3_2_prt_asia(region);
 CREATE INDEX sales_1_prt_outlying_dates_idx on sales_1_prt_outlying_dates(trans_id);
-CREATE UNIQUE INDEX sales_unique_idx on sales(office_id);
+INSERT INTO sales VALUES (1, 2, 'usa');
+CREATE UNIQUE INDEX sales_unique_idx on sales(trans_id);
```
kalensk, line 53: Nice catch! Great commit message clearly explaining the issue.

ci/scripts/load-retail-data.bash
```diff
@@ -88,24 +88,26 @@ time ssh mdw <<EOF
...
     # generate data
-    psql -d gpdb_demo -e -f data_generation/gen_order_base.sql
-    psql -d gpdb_demo -e -f data_generation/gen_facts.sql
-    psql -d gpdb_demo -e -f data_generation/gen_load_files.sql
-    psql -d gpdb_demo -e -f data_generation/load_RFMT_Scores.sql
+    psql -v ON_ERROR_STOP=1 -d gpdb_demo -e -f data_generation/gen_order_base.sql
+    psql -v ON_ERROR_STOP=1 -d gpdb_demo -e -f data_generation/gen_facts.sql
+    # gen_load_files.sql is trying to query table that doesn't exist. Disabled
+    # ON_ERROR_STOP until we can look into why this is the case.
+    psql -v ON_ERROR_STOP=0 -d gpdb_demo -e -f data_generation/gen_load_files.sql
```
kalensk, line 105: Thanks for adding the various commit messages like these that give an indication that its disabled and a little bit about why.

kalensk (approved):
LGTM!

Hopefully, in the future as a team we can figure out ways to enable it everywhere.

PR #694, opened 2022-07-27
add --pg-upgrade-verbose flag to initialize & execute

kalensk:
When the user calls initialize or execute with --pg-upgrade-verbose than pg_upgrade is executed with the --verbose flag enabled. This is useful when pg_upgrade fails and more detailed logging is needed.

Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:pgUpgradeVerbose


kyeap-vmware (approved):
If --pg-upgrade-verbose requires the --verbose flag. We might want to add a check of some sort.

kalensk:
> If --pg-upgrade-verbose requires the --verbose flag. We might want to add a check of some sort.

Thanks. Updated to add this check including bats tests. Also updated the help text with this.

PR #696, opened 2022-08-01
Hardcode unchanging ci non-secrets

kyeap-vmware:
pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:vault-non-secret-hardcode


kalensk (approved):
If it can, I think the `cm-artifacts-bucket` bucket should be hard coded as well?

PR #699, opened 2022-08-03
Use secrets stored in Vault

kyeap-vmware:
To get our build processes in compliance with the VMware Supply Chain
Security Requirements, all of our secrets have been migrated to Vault.
Our CI secret repo will be depreciated and all secrets going forward
will be stored in Vault.

Go YAML templating is used to swap between dev and prod secrets.

Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:vault-secret

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -195,10 +195,14 @@ endif
 WORKSPACE ?= ~/workspace
 PIPELINE_NAME ?= gpupgrade:$(shell git rev-parse --abbrev-ref HEAD | tr '/' ':')
 FLY_TARGET ?= cm
+
+# YAML templating is used to switch between prod and dev pipelines. The
+# environment variable JOB_TYPE is used to determine whether a dev or prod
+# pipeline is generated. It is used when go generate runs our yaml parser.
 ifeq ($(FLY_TARGET),prod)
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -195,10 +195,14 @@ endif
...
+
+# YAML templating is used to switch between prod and dev pipelines. The
+# environment variable TARGET is used to determine whether a dev or prod
+# pipeline is generated. It is used when go generate runs our yaml parser.
 ifeq ($(FLY_TARGET),prod)
-	TARGET := prod
+set-pipeline: export TARGET=prod
 else
-	TARGET := dev
+set-pipeline: export TARGET=dev
```
kalensk, line 205: Hopefully, we can we do a followup PR right after this one to see if we can come up with a more clear now instead of TARGET to minimize confusion with FLY_TARGET. Both closely related concepts, but distinct.

kalensk (approved):
Looks great! Thanks for taking on this work. A few very minor notes.

kalensk:
Just a reminder to re-fly the prod pipeline after merging.

PR #701, opened 2022-08-04
Transpose Active Connections Output

kalensk:
Reformat output of active connections to better handle multi-line queries.

Before:
```
Application:  User:    Database:  Query:
etl_job       gpadmin  postgres   SELECT
  clock_timestamp(),
  pg_sleep(1),
  clock_timestamp(),
  pg_sleep(30),
  clock_timestamp(),
  pg_sleep(60),
  clock_timestamp();
status_checker        gpcc  stats_db  SELECT
  clock_timestamp(),
  pg_sleep(1),
  clock_timestamp(),
  pg_sleep(30),
  clock_timestamp(),
  pg_sleep(60),
  clock_timestamp();
```

After:
```
Application:  etl_job
User:         gpadmin
Database:     postgres
Query:        SELECT
  clock_timestamp(),
  pg_sleep(1),
  clock_timestamp(),
  pg_sleep(30),
  clock_timestamp(),
  pg_sleep(60),
  clock_timestamp();

Application:  status_checker
User:         gpcc
Database:     stats_db
Query:        SELECT
  clock_timestamp(),
  pg_sleep(1),
  clock_timestamp(),
  pg_sleep(30),
  clock_timestamp(),
  pg_sleep(60),
  clock_timestamp();
```

Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:activeConns

### DIFF
--- a/greenplum/query_pg_stat_activity.go
+++ b/greenplum/query_pg_stat_activity.go
@@ -16,7 +16,7 @@ import (
 )
 
 type StatActivity struct {
-	Usename          string
+	User             string
 	Application_name string
 	Datname          string
 	Query            string
@@ -43,7 +43,7 @@ func (s StatActivities) Error() string {
 func (s StatActivities) table() [][]string {
 	var rows [][]string
 	for _, activity := range s {
-		rows = append(rows, []string{activity.Application_name, activity.Usename, activity.Datname, activity.Query})
+		rows = append(rows, []string{activity.Application_name, activity.User, activity.Datname, activity.Query})
 	}
 
 	sort.Sort(utils.TableRows(rows))
@@ -67,7 +67,7 @@ func QueryPgStatActivity(db *sql.DB, cluster *Cluster) error {
 	var activities StatActivities
 	for rows.Next() {
 		var activity StatActivity
-		err := rows.Scan(&activity.Datname, &activity.Usename, &activity.Application_name, &activity.Query)
+		err := rows.Scan(&activity.Datname, &activity.User, &activity.Application_name, &activity.Query)
 		if err != nil {
 			return xerrors.Errorf("pg_stat_activity: %w", err)
 		}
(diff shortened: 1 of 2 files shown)


greenplum/query_pg_stat_activity.go
```diff
@@ -24,32 +24,38 @@ type StatActivity struct {
...
 }
 
-func (s StatActivities) table() [][]string {
-	var rows [][]string
+func (s StatActivities) Error() string {
+	var sb strings.Builder
+	var tw tabwriter.Writer
+	tw.Init(&sb, 0, 0, 1, ' ', 0)
+
+	sort.Sort(s)
```
kyeap-vmware, line 48: We can remove the sorting interface if we sort the results of the query using SQL.

kalensk: Good idea! I updated to order by within the query and remove the sorting to simplify the code.

greenplum/query_pg_stat_activity_test.go
```diff
@@ -51,8 +51,8 @@ func TestQueryPgStatActivity(t *testing.T) {
 			target.Version = semver.MustParse("6.0.0")
 		}()
 
-		mock.ExpectQuery(`SELECT datname, usename, application_name, current_query FROM pg_stat_activity WHERE procpid <> pg_backend_pid\(\);`).
-			WillReturnRows(sqlmock.NewRows([]string{"datname", "usename", "application_name", "query"}))
+		mock.ExpectQuery(`SELECT application_name, usename, datname, current_query FROM pg_stat_activity WHERE procpid <> pg_backend_pid\(\) ORDER BY application_name, usename, datname;`).
```
kyeap-vmware, line 54: is usename supposed to be username?

kyeap-vmware (approved):

greenplum/query_pg_stat_activity_test.go
kalensk: In the database they call it it [usename](https://www.postgresql.org/docs/9.2/monitoring-stats.html#PG-STAT-ACTIVITY-VIEW) so here in the query we need to reference it by that name. But in golang for the struct field I called it user, since I found usename weird. So outside of the query we can reference it by user.

PR #706, opened 2022-08-09
revert: check active connections before stopping cluster

kalensk:
Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:checkActiveConnsOnRevert


kyeap-vmware (approved):
Does the same need to be done in RemoveIntermediateCluster?

PR #711, opened 2022-09-13
Rename Data Migration Script Phases

kalensk:
1) Rename data migration script phases
Since the generator and executor will be folded into the gpupgrade commands/steps have the data migration script phase match those.

2) Create data migration script seed directory.
This makes it more clear which scripts are the seed scripts and which are the generated scripts. Also, make it clear which scripts are for 5-to-6 upgrades compared to 6-to-7 upgrades. (Although, ideally these seed scripts would be outputted as part of the pg_upgrade --check and not here.)

3) Add data migration script README

Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:dataMigrationRenamePhases


data-migration-scripts/README.md
```diff
@@ -0,0 +1,18 @@
...
+Important Reminders:
+- The reason the **generator** and **executor** is split into two stages allows customers to inspect the generated SQL 
+to be run on their cluster which affects their data. That is, some customers need to see exactly what will be run on 
+their data.
+- The **generator** takes a "snapshot" of the current source cluster to create generated SQL migration scripts. If new 
+problematic objects are added **after** the generator was first run, then the previously generated scripts are outdated.
+The generator will need to be re-run to capture the newly added objects.
+- All **seed scripts** used to generate the data migration scripts are executed on the **source cluster**.
+- The **generated scripts** for stats, initialize, and revert are executed on the **source cluster**.
+- The **generated scripts** for finalize and revert are executed on the **target cluster**.
```
kyeap-vmware, line 18: Is this right? finalize and revert are executed on the target cluster?

kyeap-vmware (approved):

data-migration-scripts/README.md
kalensk: Thanks! Fixed.

PR #714, opened 2022-09-20
Fix some bug during gp6 upgrade to gp7

wangxuqi:
1.  Fix check_active_connections_on_target_cluster during finalize phase during gp6 upgrade to gp7.
2.  Only add '--old-tablespaces-file' option for target major version less than 7. The option is not supported in gp7 for now.

### DIFF
--- a/greenplum/query_pg_stat_activity.go
+++ b/greenplum/query_pg_stat_activity.go
@@ -41,7 +41,7 @@ func (s StatActivities) Error() string {
 }
 
 func QueryPgStatActivity(db *sql.DB, cluster *Cluster) error {
-	query := `SELECT application_name, usename, datname, query FROM pg_stat_activity WHERE pid <> pg_backend_pid() ORDER BY application_name, usename, datname;`
+	query := `SELECT application_name, usename, datname, query FROM pg_stat_activity WHERE pid <> pg_backend_pid() AND datname <>'' ORDER BY application_name, usename, datname;`
 	if cluster.Version.Major < 6 {
 		query = `SELECT application_name, usename, datname, current_query FROM pg_stat_activity WHERE procpid <> pg_backend_pid() ORDER BY application_name, usename, datname;`
 	}
--- a/upgrade/run.go
+++ b/upgrade/run.go
@@ -62,7 +62,7 @@ func Run(stdout, stderr io.Writer, opts *idl.PgOptions) error {
 		args = append(args, "--old-options", opts.GetOldOptions())
 	}
 
-	if opts.Action != idl.PgOptions_check {
+	if opts.Action != idl.PgOptions_check && semver.MustParse(opts.TargetVersion).Major < 7 {
 		args = append(args, "--old-tablespaces-file", utils.GetTablespaceMappingFile())
 	}
 


wangxuqi:
Also find a bug during initialize phase. In my test, A `Substep_shutdown_source_cluster ` is needed after `Substep_backup_target_master`. Or the initialize phase will be failed as pg_upgrade need to start the old master. But the old master is still on. Port  already in use will make upgrade  initialize fail. Im not sure if this is only my problem, do not fix this problem in this commit for now. If needed, I will do it.

greenplum/query_pg_stat_activity.go
```diff
@@ -41,7 +41,7 @@ func (s StatActivities) Error() string {
 }
 
 func QueryPgStatActivity(db *sql.DB, cluster *Cluster) error {
-	query := `SELECT application_name, usename, datname, query FROM pg_stat_activity WHERE pid <> pg_backend_pid() ORDER BY application_name, usename, datname;`
+	query := `SELECT application_name, usename, datname, query FROM pg_stat_activity WHERE pid <> pg_backend_pid() AND datname <>'' ORDER BY application_name, usename, datname;`
```
kalensk, line 44: I'm concerned that this is omitting pg_stat_activity entries that don't have a datname entirely, rather than coalescing them to empty string and still displaying them. Omitting entries rather than simply displaying them with empty string or "no database found" or something.

kalensk:
> Also find a bug during initialize phase. In my test, A Substep_shutdown_source_cluster is needed after Substep_backup_target_master. Or the initialize phase will be failed as pg_upgrade need to start the old master. But the old master is still on. Port already in use will make upgrade initialize fail. Im not sure if this is only my problem, do not fix this problem in this commit for now. If needed, I will do it.

We do not yet support 6->7 upgrades. That is the next thing we are working on in about a month from now. During our initial preliminary tests this was not an issue and seems to be only something you are encountering. 

We are very actively making changes to pg_upgrade master (7X) and pg_upgrade (6X) at the moment. If one does not have the latest there may be potential issues. I find it odd that pg_upgrade behavior changed like that between 6x and 7x.

What you are suggesting is a very major change and one that requires very thorough testing and understanding, as well as ensuring the code is clear and clean.

wangxuqi:
> @wangxuqi **1)** Can you please rename your branch from "master" to something else such as "gpupgradeFixes"? This will allow us to more easily run all the tests which is required before merging.
> 
> Before merging we will need to "fly a pipeline" of all our tests. However, your branch is named "master" which will causes lots of issues.
> 
> **2)** At the minimum unit tests are needed before merging. Can you try adding some yourself? You can use the existing tests as a guide. Or if you need help maybe add me as GitHub collaborator on your fork, and I can push some to your branch? Or I can try adding some as a comment in this PR.

OK. I will close this PR and reopen a new one to fix the problem you mentioned. Thx

greenplum/query_pg_stat_activity.go
wangxuqi: > I'm concerned that this is omitting pg_stat_activity entries that don't have a datname _entirely_, rather than coalescing them to an empty string and _still_ displaying them. That is, we do not want to omit entries, rather we want to still display them, but with dataname being an empty string.

So, what you want fix is to add datname _entirely_ in pg_stat_activity and do not do some work around in gpupgrade? Or the better way is deal with them both?

wangxuqi: What I want to known is does this change on 'pg_stat_activity' is still need or not?

kalensk: Your change will not show active connections when datname is '' which is bad since we still want to display these active connections to the user.

My original suggestion was to still show those active connections when datname is NULL, but display it as empty string to the user using `COALESCE(datname, '')`. However, the more I looked into this the better fix is to change the Datname type from `String` to `sql.NullString` on line 20 in the struct.

```go
type StatActivity struct {
	User             string
	Application_name string
	Datname          sql.NullString
	Query            string
}
```

And update the existing test in `greenplum/query_pg_stat_activity_test.go` on line 64 to:
```go

	t.Run("errors when pg_stat_activity shows active connections and database is NULL", func(t *testing.T) {
		expectPgStatActivityToReturn(mock).WillReturnRows(sqlmock.NewRows([]string{"application_name", "usename", "datname", "query"}).
			AddRow("etl_job", "gpadmin", nil, "SELECT * FROM my_table;").
			AddRow("status_checker", "gpcc", "stats_db", "SELECT * FROM stats;"))

		expected := greenplum.StatActivities{
			{Application_name: "etl_job", User: "gpadmin", Datname: sql.NullString{String: "", Valid: false}, Query: "SELECT * FROM my_table;"},
			{Application_name: "status_checker", User: "gpcc", Datname: sql.NullString{String: "stats_db", Valid: true}, Query: "SELECT * FROM stats;"},
		}

		err = greenplum.QueryPgStatActivity(db, target)
		var nextActionsErr utils.NextActionErr
		if !errors.As(err, &nextActionsErr) {
			t.Errorf("got type %T want %T", err, nextActionsErr)
		}

		if !strings.Contains(nextActionsErr.Err.Error(), expected.Error()) {
			t.Errorf("got %#v, want %#v", err, expected)
		}

		if !strings.Contains(nextActionsErr.NextAction, "close") {
			t.Errorf("got %q, want 'close'", nextActionsErr.NextAction)
		}
	})
```

wangxuqi: OK, It seems a better choice.  And we may filter internal background worker process after we get active process. Close this pr and create a new one,  only fix gp7 upgrade skip '--old-tablespaces-file' option.

kalensk: > And we may filter internal background worker process after we get active process.

Why would we filter background worker processes rather than displaying them as active connections to the user? I don't understand. Doesn't pg_upgrade fail if there are background worker processes connected to the database?

PR #723, opened 2022-10-13
gpupgrade generate & apply fixups

kalensk:
See individual commits.

- tweak output based on user feedback
- colorize text to send stats output
- bold directory paths
- gpupgrade apply append output to file rather than truncating
- output gpupgrade apply logs to gpupgrade log directory
- rename functions to match their command name

Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:dataMigrationFixups

### DIFF
--- a/testutils/test_utils.go
+++ b/testutils/test_utils.go
@@ -138,7 +138,7 @@ func MustGetPort(t *testing.T) int {
 func GetTempDir(t *testing.T, prefix string) string {
 	t.Helper()
 
-	dir, err := os.MkdirTemp("", prefix+"-")
+	dir, err := os.MkdirTemp("", prefix)
 	if err != nil {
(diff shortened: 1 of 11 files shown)


testutils/test_utils.go
```diff
@@ -138,7 +138,7 @@ func MustGetPort(t *testing.T) int {
 func GetTempDir(t *testing.T, prefix string) string {
 	t.Helper()
 
-	dir, err := os.MkdirTemp("", prefix+"-")
+	dir, err := os.MkdirTemp("", prefix+"")
```
kyeap-vmware, line 141: We can also remove adding the empty string. prefix+""

kyeap-vmware (approved):

PR #725, opened 2022-10-25
pass --old-tablespaces-file only for 6X

kalensk:
Only pass --old-tablespaces-file to pg_upgrade when in check mode and target version is less than 7X. The pg_upgrade parameter --old-tablespaces-file is not supported or needed in 7X or higher.

Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:pgUpgradeRun

### DIFF
--- a/upgrade/run_test.go
+++ b/upgrade/run_test.go
@@ -484,7 +484,7 @@ func TestRun(t *testing.T) {
 			},
 		},
 		{
-			name:        "sets --old-tablespaces-file when upgrading and not calling --check",
+			name:        "sets --old-tablespaces-file when upgrading and not calling --check when target version is 6x",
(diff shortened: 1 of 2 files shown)


upgrade/run_test.go
```diff
@@ -463,7 +463,7 @@ func TestRun(t *testing.T) {
 			},
 		},
 		{
-			name:        "does not set --old-tablespaces-file when --check is passed",
+			name:        "does not set --old-tablespaces-file when --check is passed and version is less than 7x",
```
kyeap-vmware, line 466: nit: and target version is less than 7x.

kyeap-vmware (approved):

PR #736, opened 2022-12-01
Add gp_fastsequence test

kyeap-vmware:
Test case for related  6X PR: https://github.com/greenplum-db/gpdb/pull/14576

gpupgrade: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:fastsequence-test
6X: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/upgrade-gp-fastsequence


test/acceptance/pg_upgrade/5-to-6/upgradeable_tests/source_cluster_regress/sql/gp_fastsequence.sql
```diff
@@ -0,0 +1,15 @@
...
+--------------------------------------------------------------------------------
+-- Create and setup upgradeable objects
+--------------------------------------------------------------------------------
+
+CREATE TABLE aotable1_fastsequence (i int) WITH (appendonly=true);
+INSERT INTO aotable1_fastsequence SELECT * FROM generate_series(1,10);
+
+CREATE TABLE aotable2_fastsequence (i int) WITH (appendonly=true);
+INSERT INTO aotable2_fastsequence SELECT * FROM generate_series(1,10);
+INSERT INTO aotable2_fastsequence SELECT * FROM generate_series(11,20);
```
jimmyyih, line 15: The source cluster SQL doesn't look thorough enough.  I suggest doing something like this:
```
CREATE TABLE aotable_fastsequence (i int) WITH (appendonly=true);
INSERT INTO aotable_fastsequence SELECT generate_series(1,101);
CREATE INDEX aotable_fastsequence_idx ON aotable_fastsequence USING btree(i);

CREATE TABLE aocotable_fastsequence (i int) WITH (appendonly=true, orientation=column);
INSERT INTO aocotable_fastsequence SELECT generate_series(1,101);
CREATE INDEX aocotable_fastsequence_idx ON aocotable_fastsequence USING btree(i);
```

And then on the target cluster, do these verifications for both tables:
1. Does SELECT * on the tables work.  (it's okay if the expected output file will be a bit big)
2. INSERT some rows with generate_series(1001, 1010) and use the query from the JIRA to validate unique ctids:
```
SELECT gp_segment_id, ctid, count(ctid) FROM <tablename> GROUP BY gp_segment_id, ctid HAVING count(ctid) > 1;
```
3. Set enable_seqscan off and try to do a filtered query (e.g. i < 10 would be good).  It should fail.  Do a REINDEX and try the query again.  You should only see 10 rows and not ~20.

test/acceptance/pg_upgrade/5-to-6/upgradeable_tests/source_cluster_regress/expected/gp_fastsequence.out
```diff
@@ -0,0 +1,34 @@
+-- Copyright (c) 2017-2022 VMware, Inc. or its affiliates
```
kalensk, line 1: Copyright needs to be updated to 2023.

test/acceptance/pg_upgrade/5-to-6/upgradeable_tests/source_cluster_regress/upgradeable_source_schedule
```diff
@@ -1,3 +1,4 @@
 test: alter_statistic_on_column ao_table_multi_segfile ao_table_without_base_relfilenode aoco_table_multi_segfile check_constraints external_table indexes partitioned_ao_aoco_table partitioned_heap_table partitions_using_keywords pl_functions resource_group_queue user_defined_aggregates user_defined_types views_referencing_deprecated_columns views_with_lag_lead_functions
 test: clog_preservation
 test: vacuum_freeze_tables
+test: gp_fastsequence
```
kalensk, line 4: Can this test be run in parallel? That is, move it above to after "views_with_lag_lead_functions"?

test/acceptance/pg_upgrade/5-to-6/upgradeable_tests/target_cluster_regress/sql/gp_fastsequence.sql
```diff
@@ -0,0 +1,56 @@
...
+ORDER BY 1, 2, 3;
+
+-- Verify table data is not corrupt
+SELECT * FROM aotable_fastsequence ORDER BY i;
+
+-- Verify INSERTs produce no duplicate ctids
+INSERT INTO aotable_fastsequence SELECT generate_series(1001, 1010);
+SELECT gp_segment_id, ctid, count(ctid) FROM aotable_fastsequence GROUP BY gp_segment_id, ctid HAVING count(ctid) > 1;
+
+-- Indexes are invalid and should not work after upgrade
```
kalensk, line 23: Can we add one other sentence explaining why with maybe a reference to the GPDB code base where this is the case? Similarly for line 49 where indexes are invalid is also mentioned.

test/acceptance/pg_upgrade/5-to-6/upgradeable_tests/target_cluster_regress/expected/gp_fastsequence.out
```diff
@@ -0,0 +1,212 @@
...
+ 1             | 0      | 0             
+ 1             | 1      | 301           
+ 2             | 0      | 0             
+ 2             | 1      | 301           
+(6 rows)
+
+-- Verify table data is not corrupt
+SELECT * FROM aocotable_fastsequence ORDER BY i;
+ERROR:  Unexpected internal error (appendonlytid.h:66)
+DETAIL:  FailedAssertion("!(h->bytes_4_5 <= 0x8000)", File: "../../../../src/include/access/appendonlytid.h", Line: 66)
```
kalensk, line 178: Im assuming this will need to be updated once https://github.com/greenplum-db/gpdb/pull/14576 is merged? That is, it says verify the data is not corrupt, but this is showing an error....

kyeap-vmware: Oh. this is a good catch. I missed this. No error supposed to be here. The previous table verify disabled seqscan which caused this failure.

test/acceptance/pg_upgrade/5-to-6/upgradeable_tests/source_cluster_regress/upgradeable_source_schedule
kyeap-vmware: It doesn't look like it can be. I see pipeline failures when attempting parallelization.

jimmyyih: I don't see why it wouldn't be able to.  What were the pipeline failures?

kyeap-vmware: Resolved. Parallelization failures were due to another test running a VACUUM FREEZE and expected test FATALs causing session disconnects.

kalensk: What exactly did you have to do to fix it? Can you elaborate in detail? (Looks like maybe use the isolation2 framework to run commands in separate sessions...?)

jimmyyih (approved):

PR #737, opened 2022-12-07
Do path lookup for sed in update_conf_files unit test

jimmyyih:
The update_conf_files unit test assumes that the sed tool is located in /usr/bin/sed. However, this assumption isn't guaranteed since developers could have their sed somewhere else
(e.g. /usr/local/bin/sed from MacOS brew or compiling from scratch on linux). It's safer to do a path lookup for sed and use that for the expected output string to avoid developer environment assumptions.

### DIFF
--- a/hub/update_conf_files_test.go
+++ b/hub/update_conf_files_test.go
@@ -6,6 +6,7 @@ package hub_test
 import (
 	"errors"
 	"fmt"
+	"os/exec"
 	"path/filepath"
 	"strings"
 	"testing"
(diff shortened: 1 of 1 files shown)


hub/update_conf_files_test.go
```diff
@@ -526,8 +527,13 @@ primary_slot_name = 'internal_wal_replication_slot'
 			t.Fatalf("got error count %d, want %d", len(errs), len(opts))
 		}
 
+		sedPath, lookupErr := exec.LookPath("sed")
+		if lookupErr != nil {
+			t.Fatalf("tests require sed (%v)", lookupErr)
```
kalensk, line 532: Minor comments to follow surrounding code conventions and consistency:
- No need for the parenthesis around the %v. 
- Prefer `err` over `lookupErr` unless explicitly needed for shadowing, etc.

Something like:
```
		sedPath, err := exec.LookPath("sed")
		if err != nil {
			t.Fatalf("expected sed: %v", err)
		}
```

kalensk (approved):

PR #743, opened 2023-01-10
use an enum for the mode

kalensk:
Using an enum makes the code much more readable and explicit compared to booleans. This leads to less makes, errors, and code clarity. It also allows flexibility for introducing pg_upgrade's --clone mode.

https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:modeEnum


idl/common.pb.go
```diff
@@ -0,0 +1,101 @@
...
+// A compilation error at this line likely means your copy of the
+// proto package needs to be updated.
+const _ = proto.ProtoPackageIsVersion3 // please upgrade the proto package
+
+type Mode int32
+
+const (
+	Mode_unknown_mode Mode = 0
+	Mode_copy         Mode = 1
+	Mode_link         Mode = 2
```
jimmyyih, line 28: Nit: I think MODE_UNKNOWN, MODE_COPY, and MODE_LINK would look better as constant enum names but I guess this is the standard for this gpupgrade repository after https://github.com/greenplum-db/gpupgrade/commit/32b37a6c504bbc9a33df18d96827c4a07964c8aa.  All good with me.

jimmyyih (approved):

idl/common.pb.go
kalensk: We switched away from that since when the enum's get printed which is often they are lower case which is needed for the messaging. It prevents needing to have strings.Tolower litter throughout which gets rather annoying.

kyeap-vmware (approved):

PR #744, opened 2023-01-10
pg_upgrade: add upgradeable sequence test

kalensk:
Upgrading sequences from 5->6 use the script `adjust_sequences.sql`. Ensure sequecnes are properly upgraded by adding a pg_upgrade acceptance test.

https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:sequenceTest


test/acceptance/pg_upgrade/5-to-6/upgradeable_tests/source_cluster_regress/sql/sequences.sql
```diff
@@ -0,0 +1,17 @@
...
+-- SPDX-License-Identifier: Apache-2.0
+
+-- Test to ensure that tables with simple check constraints can be upgraded.
+
+--------------------------------------------------------------------------------
+-- Create and setup upgradeable objects
+--------------------------------------------------------------------------------
+
+CREATE SEQUENCE serial_seq START 100;
+SELECT nextval('serial_seq');
```
jimmyyih, line 11: A `SELECT * FROM serial_seq;` in the source and target SQL files to get an expected before and after would be useful.

test/acceptance/pg_upgrade/5-to-6/upgradeable_tests/target_cluster_regress/sql/sequences.sql
```diff
@@ -0,0 +1,10 @@
+-- Copyright (c) 2017-2023 VMware, Inc. or its affiliates
+-- SPDX-License-Identifier: Apache-2.0
+
+--------------------------------------------------------------------------------
+-- Validate that the upgradeable objects are functional post-upgrade
+--------------------------------------------------------------------------------
+
+SELECT last_value FROM serial_seq;
+
+SELECT * FROM tbl_with_sequence;
```
jimmyyih, line 10: We should add more test coverage around post-upgrade scenarios as well and not only validate that upgrade was successful.  Maybe something like this would give more test confidence:
```
SELECT * FROM serial_seq;

SELECT * FROM tbl_with_sequence;

SELECT nextval('serial_seq');
INSERT INTO tbl_with_sequence(t) VALUES ('test3');
SELECT * FROM tbl_with_sequence;
SELECT * FROM serial_seq;
```

kalensk: Updated.


Note, that on the target cluster I purposely made the order of operations:
```
SELECT nextval('serial_seq');
INSERT INTO tbl_with_sequence(t) VALUES ('test3');
```

To see if it properly increments on the insert even if the serial_seq was already advanced.

test/acceptance/pg_upgrade/5-to-6/upgradeable_tests/source_cluster_regress/sql/sequences.sql
```diff
@@ -0,0 +1,19 @@
+-- Copyright (c) 2017-2023 VMware, Inc. or its affiliates
+-- SPDX-License-Identifier: Apache-2.0
+
+-- Test to ensure that tables with simple check constraints can be upgraded.
```
jimmyyih, line 4: Should change to something like: Test to ensure that sequences can be upgraded.

jimmyyih (approved):

test/acceptance/pg_upgrade/5-to-6/upgradeable_tests/target_cluster_regress/sql/sequences.sql
jimmyyih: Great, looks good.  The sequence metadata looks good to before/after upgrade (log_cnt is changed to 0 but that's okay and may actually be a good thing; it's a decrementing counter to trigger a WAL write).

PR #746, opened 2023-01-14
Use cdw instead of mdw in CI scripts and yaml templates

jimmyyih:
Our CI uses a custom tool called CCP to provision GPDB test clusters on the cloud for our automated tests. There was a recent change that changed the hostnames mdw and smdw (master and standby master) to cdw and scdw (coordinator and standby coordinator) as part of the inclusive terminology changes. The master data directory was also changed to use coordinator instead of master. Technically we could pin our CCP to the v2.6.0 tag but updating the hostnames here as well is better so that we keep ourselves aligned with the latest CCP code. The CI scripts and yaml templates have been updated accordingly.

Green developer pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:fix_ci_mdw_to_cdw

### DIFF
--- a/ci/5_multi_host_gpupgrade_jobs.yml
+++ b/ci/5_multi_host_gpupgrade_jobs.yml
@@ -24,7 +24,7 @@
       params:
         <<: *ccp_default_params
         vars:
-          standby_master: true
+          standby_coordinator: true
           instance_type: n1-standard-2
           number_of_nodes: 4
           PLATFORM: {{.OSVersion}}
--- a/ci/6_upgrade_and_functional_jobs.yml
+++ b/ci/6_upgrade_and_functional_jobs.yml
@@ -72,7 +72,7 @@
           {{- if .PrimariesOnly}}
           mirrors: false
           {{- else if not .NoStandby}}
-          standby_master: true
+          standby_coordinator: true
           {{- end}}
           instance_type: n1-standard-2
           number_of_nodes: 4
@@ -249,7 +249,7 @@
               source gpupgrade_src/test/acceptance/helpers/finalize_checks.bash
 
               echo 'Doing failover tests of mirrors and standby...'
-              validate_mirrors_and_standby /usr/local/greenplum-db-target mdw 5432
+              validate_mirrors_and_standby /usr/local/greenplum-db-target cdw 5432
     {{- end -}}
     {{- end }}
   ensure:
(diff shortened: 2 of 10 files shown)


kyeap-vmware (approved):
I was gonna ask how this isn't breaking some other parts of gpupgrade, but it looks like we pull hostname from source cluster. 

Maybe this is probably a future concern for 6 > 7, but there might be additional work involved in upgrading a mdw cluster to a cdw cluster.

jimmyyih:
> I was gonna ask how this isn't breaking some other parts of gpupgrade, but it looks like we pull hostname from source cluster.
> 
> Probably a future concern for 6 > 7, but there might be additional work involved in upgrading a mdw cluster to a cdw cluster.

Most likely no extra work.  It was a blanket change from mdw to cdw so there wouldn't be any mentions of mdw/master on the CCP clusters anymore.

kyeap-vmware:
> Most likely no extra work. It was a blanket change from mdw to cdw so there wouldn't be any mentions of mdw/master on the CCP clusters anymore.

The scenario I was thinking of was in a prod setting if that changes anything. An existing cluster running real workloads is setup with mdw references or folders. But then the new cluster to upgrade to is assumes cdw.

kalensk:
I believe our pipeline didn't pull the latest CCP changes until just recently. For example, you can see our jobs start to fail with a specific CCP commit: https://prod.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade/jobs/5-to-6-centos6-e2e-copy-mode/builds/2

jimmyyih:
> The scenario I was thinking of was in a prod setting if that changes anything. An existing cluster running real workloads is setup with mdw references or folders. But then the new cluster to upgrade to assumes cdw.

There shouldn't be anything in the gpupgrade product that makes master/coordinator assumptions besides maybe trivial stdout/log output if applicable.  This PR was strictly to fix the pipeline CCP usage and some test references.  On a user production system, it doesn't matter what hostnames or folder names they use (e.g. master, mdw, coordinator, sf-node1, xyz-1, thisismyhostname-23).

If you're talking about a copy-mode upgrade that involves a target cluster on separate machines with different hostnames (not in-place upgrade), that's not currently supported I think (nothing about hostnames in pg_upgrade nor gpupgrade documentation).

> I believe our pipeline didn't pull the latest CCP changes until just recently. For example, you can see our jobs start to fail with a specific CCP commit: https://prod.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade/jobs/5-to-6-centos6-e2e-copy-mode/builds/2

No that was my bad.  I forgot to set the pipeline before/after merging and only remembered like 15 minutes afterwards.  I only retriggered one of the failing jobs afterwards so thanks for retriggering the rest (if that was you).

PR #747, opened 2023-01-26
Add acceptance test for external partitions

jimmyyih:
External partitions were originally banned from upgrade but that has changed after changing some GPDB logic and fixing pg_dump. To test the change, a new acceptance test is added that covers most of the external partition scenarios.

GPDB commit reference:
https://github.com/greenplum-db/gpdb/commit/3cf14766073b97848115b6e1db5fbbd8523a755e
https://github.com/greenplum-db/gpdb/commit/6d7bfd099f59cd0e5d90a8d77e736f4626637b99
https://github.com/greenplum-db/gpdb/commit/8ecf11fa6eb936afa6875da76fd82332dfb793d2
https://github.com/greenplum-db/gpdb/commit/ca97e829b1403a0e72914a6aa5afc788f9de4ddc

Dev pipeline:
https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:exchange_external_partition_tests


test/acceptance/pg_upgrade/5-to-6/upgradeable_tests/target_cluster_regress/expected/exchange_external_partition.out
```diff
@@ -0,0 +1,109 @@
...
+ 3 | 2 
+ 5 | 1 
+ 5 | 2 
+ 7 | 2 
+ 8 | 2 
+ 9 | 2 
+(9 rows)
+-- this should fail
+INSERT INTO one_level_partition_table VALUES (1,1);
+ERROR:  insert into external partitions not supported
```
kalensk, line 35: Naive question perhaps, but why is inserting into external partitions is not supported? Do we need a high level comment or breadcrumb stating why?

kalensk (approved):

test/acceptance/pg_upgrade/5-to-6/upgradeable_tests/target_cluster_regress/expected/exchange_external_partition.out
jimmyyih: Writable external tables cannot be exchanged into a partition table.  Only readable external tables are allowed so the INSERTs into external partitions would never work.  I'll add a comment at the top of the test for future reference.

PR #748, opened 2023-01-27
Fix some AOCO test queries in the pg_upgrade acceptance tests

jimmyyih:
The original queries were querying the wrong table (the append-only row-oriented one). Fix the queries to query against the correct table (the append-only column-oriented one).


kalensk (approved):
Nice! good find.

PR #752, opened 2023-02-09
ci: update icw.diff filter file

jimmyyih:
After a recent GPDB 6X change, the icw.diff filter file needed to be updated to ignore the new expected diffs. In fact, the previous icw.diff contents were not relevant anymore and were skipped by the patch command. The contents of this newly updated icw.diff filter file was created by running the following command on one of the CI test systems:

```
diff -U3 --speed-large-files --ignore-space-change --ignore-blank-lines /tmp/source.sql.filtered /tmp/target.sql.filtered > /tmp/icw.diff
```

The patch command in our CI script was updated as well to use the --ignore-whitespace flag since `diff --ignore-space-change` can cause some hunks to have missing whitespace diffs which throws off the patch command... but it's fine and is actually a good thing so that we can lower the patch size.

GPDB commit reference:
\<insert commit later>

----

The GPDB PR this change is for:
https://github.com/greenplum-db/gpdb/pull/14818


kalensk:
Is there a way for the patch command to error out and bubble up to fail on our CI when the a hunk is ignored or skipped? For example, when this output occurs fail the job:
```
Skipping patch.
6 out of 6 hunks ignored -- saving rejects to file 
```

jimmyyih:
> Is there a way for the patch command to error out and bubble up to fail on our CI when the a hunk is ignored or skipped? For example, when this output occurs fail the job:
> 
> ```
> Skipping patch.
> 6 out of 6 hunks ignored -- saving rejects to file 
> ```

Testing it out:
```
[gpadmin@cdw tmp]$ patch -R /tmp/target.sql.filtered < icw.diff
patching file /tmp/target.sql.filtered
Unreversed patch detected!  Ignore -R? [n]
Apply anyway? [n]
Skipping patch.
1 out of 1 hunk ignored -- saving rejects to file /tmp/target.sql.filtered.rej
[gpadmin@cdw tmp]$ echo $?
1
```
Seems like we get a nonzero return code so yeah.  We could also detect the reject file but checking the RC should be good enough.  I'll do it in a separate PR later.

kyeap-vmware (approved):
LGTM

PR #756, opened 2023-02-17
ci: fail early if patching the target dump file failed or was skipped

jimmyyih:
When the ci filter patch is out-of-date, the patching command may skip when it finds a diff in the patch that it cannot apply. We were not checking the return code for the patch command so the job would continue and run the long diff against the source and target dumps which may eventually fail depending on if there were any GPDB changes to the dump output. When looking at the dump diffs as the result of the patch apply skip, the unexpected diffs could be deceptive when the ci filter patch is what needs to be updated. Check the return code and fail early if the patching fails or is skipped.

As a result of checking the return code, it was also found that the retail_demo.diff filter patch was no longer needed because the `START WITH ` filter was integrated into scripts/filters/rules6x.go a while back. Just in case we need it later, keep the empty file around for now.

This is a follow-up PR to resolve a comment in a different PR:
https://github.com/greenplum-db/gpupgrade/pull/752#issuecomment-1428627862


ci/scripts/ci-helpers.bash
```diff
@@ -73,6 +73,11 @@ compare_dumps() {
                 patch --ignore-whitespace -R '$target_dump.filtered'
             " < ./ci/scripts/filters/${DIFF_FILE}
 
+            if [ $? -ne 0 ]; then
+                echo "error: patching failed"
+                exit 1
+            fi
+
```
kalensk, line 80: Looks good! BASH is weird. Im a little surprised that this is even needed and that if patch returns a non-zero exit status that it doesn't just bubble up to `ssh` command and up to the controlling script and to concourse to fail normally.

kalensk (approved):
> As a result of checking the return code, it was also found that the retail_demo.diff filter patch was no longer needed because the START WITH  filter was integrated into scripts/filters/rules6x.go a while back. Just in case we need it later, keep the empty file around for now.

My vote is to remove the `retail_demo.diff` file as it's easy to add it back if we ever need it again. Not a huge deal either way.

kyeap-vmware (approved):

jimmyyih:
> My vote is to remove the retail_demo.diff file as it's easy to add it back if we ever need it again. Not a huge deal either way.

I'm just gonna leave it for now just in case (also to not spend any more cycles on this) and we can remove it later when approaching 6 to 7 upgrade testing.

PR #760, opened 2023-02-25
Add acceptance test for list partitioned tables with custom type

jimmyyih:
A partition table can be list partitioned by a custom type but this previously had issues when running through upgrade because the CREATE TABLE assumes the custom type's equality operator is in the pg_catalog schema or in search_path. Since the SQL dump from pg_dump explicitly sets the search_path to blank, only pg_catalog was searched which would result in the pg_restore section of upgrade to fail. This has been fixed in a recent GPDB commit and we should be able to support this scenario now.

GPDB commit reference:
https://github.com/greenplum-db/gpdb/commit/d18208cf0bb82e2ba78121547f44f9615196744e

### DIFF
--- a/test/acceptance/pg_upgrade/5-to-6/upgradeable_tests/source_cluster_regress/sql/partitioned_heap_table.sql
+++ b/test/acceptance/pg_upgrade/5-to-6/upgradeable_tests/source_cluster_regress/sql/partitioned_heap_table.sql
@@ -109,3 +109,41 @@ INSERT INTO dropped_and_added_column SELECT i, i, i, i FROM generate_series(10, 
 CREATE ROLE testrole;
 CREATE TABLE p_alter_owner (id INTEGER, name TEXT) DISTRIBUTED BY (id) PARTITION BY RANGE(id) (START(1) END(3) EVERY(1));
 ALTER TABLE p_alter_owner OWNER TO testrole;
+
+--
+-- list partitioned by custom type where equality operator is in different schema
+-- Note: On 5X, inserts into the table won't work due to a bug where it assumes
+--       the equality operator is in pg_catalog. This is fixed in 6X so we'll at
+--       least test the table creation here.
+--
+
+CREATE TYPE equal_operator_not_in_search_path_type AS (a int, b int);
+CREATE FUNCTION equal_operator_not_in_search_path_func (equal_operator_not_in_search_path_type, equal_operator_not_in_search_path_type) RETURNS boolean
+  AS 'SELECT $1.a = $2.a;'
+  LANGUAGE SQL IMMUTABLE
+  RETURNS NULL ON NULL INPUT;
+
+CREATE SCHEMA equal_operator_not_in_search_path_schema;
+CREATE OPERATOR equal_operator_not_in_search_path_schema.= (
+        LEFTARG = equal_operator_not_in_search_path_type,
+        RIGHTARG = equal_operator_not_in_search_path_type,
+        PROCEDURE = equal_operator_not_in_search_path_func
+);
+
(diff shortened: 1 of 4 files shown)


jimmyyih:
Pending https://github.com/greenplum-db/gpdb/pull/14997 to be merged.

test/acceptance/pg_upgrade/5-to-6/upgradeable_tests/target_cluster_regress/sql/partitioned_heap_table.sql
```diff
@@ -26,3 +26,9 @@ SELECT c, d FROM dropped_and_added_column WHERE a=10;
 SELECT c.relname, pg_catalog.pg_get_userbyid(c.relowner) as owner
 FROM pg_class c
 WHERE relname like 'p_alter_owner%';
+
+INSERT INTO equal_operator_not_in_search_path_table VALUES (1, '(1,1)');
+SELECT * FROM equal_operator_not_in_search_path_table;
+
+INSERT INTO equal_operator_not_in_search_path_table_multi_key VALUES (1, '(1,1)', 1);
+SELECT * FROM equal_operator_not_in_search_path_table_multi_key;
```
kalensk, line 34: Naive question perhaps, but shouldn't we exercise the = operator with something like `WHERE a = 1`. Meaning actually use the equal operator that we defined in another schema.

jimmyyih: Not naive at all.  The equality operator is exercised by the INSERT where it's saying the value `'(1,1)'` belongs to a particular partition that holds all rows with value `'(1,1)'` (or for the multi key one, rows with value `('(1,1)', 1)`).  While a filter expression using the equality operator is valid, that's not what the test is validating.  We're validating the partitioning aspect.

kalensk: Ah!!! Thanks for that clarification. (Not super obvious to me on first pass.) That totally makes sense though. Thanks.

test/acceptance/pg_upgrade/5-to-6/upgradeable_tests/source_cluster_regress/sql/partitioned_heap_table.sql
```diff
@@ -109,3 +109,41 @@ INSERT INTO dropped_and_added_column SELECT i, i, i, i FROM generate_series(10,
...
+DISTRIBUTED BY (a) PARTITION BY LIST(b)
+(
+  PARTITION part1 VALUES('(1,1)')
+);
+
+CREATE TABLE public.equal_operator_not_in_search_path_table_multi_key (a int, b equal_operator_not_in_search_path_type, c int)
+DISTRIBUTED BY (a) PARTITION BY LIST(b, c)
+(
+  PARTITION part1 VALUES(('(1,1)', 1))
+);
```
kalensk, line 149: minor: Does the multi_key table cover the use case for the single_key table? That is, do we need the single_key table given the multi_table key?

kalensk (approved):

test/acceptance/pg_upgrade/5-to-6/upgradeable_tests/source_cluster_regress/sql/partitioned_heap_table.sql
jimmyyih: AFAIK, some but not all.  There are areas where the logic diverges IIRC and it's best to cover both just in case there's a future GPDB change that diverges the logic further.

PR #762, opened 2023-02-28
Add post-upgrade test to verify that AO/CO indexes are handled properly

jimmyyih:
For upgrading AO/CO indexes, we need to verify that the provided REINDEX workaround fixes the indexes and that they'll be usable afterwards.


test/acceptance/pg_upgrade/5-to-6/upgradeable_tests/target_cluster_regress/sql/ao_indexes.sql
```diff
@@ -0,0 +1,98 @@
...
+SELECT * FROM aoco_with_bitmap WHERE a = 1;
+SELECT * FROM aoco_with_gist WHERE a @@ to_tsquery('footext1');
+
+-- Provided REINDEX workaround should fix all the AO indexes
+REINDEX INDEX ao_with_btree_idx;
+REINDEX INDEX ao_with_bitmap_idx;
+REINDEX INDEX ao_with_gist_idx;
+REINDEX INDEX aoco_with_btree_idx;
+REINDEX INDEX aoco_with_bitmap_idx;
+REINDEX INDEX aoco_with_gist_idx;
```
kalensk, line 43: We currently have data-migration scripts for [partitioned_tables_indexes](https://github.com/greenplum-db/gpupgrade/tree/ea2b2464717736487659d48bf5b61394cce99e34/data-migration-scripts/5-to-6-seed-scripts/initialize/partitioned_tables_indexes). Curious if those still relevant or can be simplified to just a REINDEX in the finalize scripts?

jimmyyih: I'm not aware of why partitioned table indexes need to be dropped but I can see it being a bit different from regular tables.  I can do some digging and testing to verify later.

kalensk (approved):

PR #764, opened 2023-03-02
pg_upgrade tests: expand external table tests

kalensk:
Tests for GPDB PR: [[6X] support dumping external tables with serial column types #15109](https://github.com/greenplum-db/gpdb/pull/15109)

Add external table tests for tables with serial column types to ensure attaching the sequence to the table succeeds. That is, Serial column types are dumped in components and added together. The CREATE TABLE is dumped, followed by the CREATE SEQUENCE, and then attached to the table column with ALTER TABLE ONLY public.foobar ALTER COLUMN b SET DEFAULT nextval('public.foobar_b_seq'::regclass);

Until GPDB commit XXX the restore failed with “unsupported ALTER command for external table” causing pg_upgrade to fail. Specifically, the pg_dump SQL contains the following ALTER which is not supported for external tables: `ALTER TABLE ONLY public.foobar ALTER COLUMN b SET DEFAULT nextval('public.foobar_b_seq'::regclass);`

Add additional tests for external tables with with user defined type containing default value, and external tables with dropped columns. See GPDB commit XXX.

Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:externalTablesWithSerialCols


test/acceptance/pg_upgrade/5-to-6/upgradeable_tests/target_cluster_regress/input/external_table.source
```diff
@@ -0,0 +1,20 @@
...
+SELECT * FROM readable_external_table;
+
+INSERT INTO writable_external_table_with_serial_column_type VALUES (1);
+SELECT * FROM readable_external_table_with_serial_column_type;
+
+INSERT INTO writable_external_table_with_user_type_and_default_value VALUES (1);
+SELECT * FROM readable_external_table_with_user_type_and_default_value;
+
+INSERT INTO writable_external_table_with_dropped_column VALUES (1);
+SELECT * FROM readable_external_table_with_dropped_column;
```
kalensk, line 18: Doing some DELETEs or UPDATEs in addition to the INSERTs might be useful.

test/acceptance/pg_upgrade/5-to-6/upgradeable_tests/source_cluster_regress/output/external_table.source
```diff
@@ -6,5 +6,82 @@
...
+(exited with code 0)
+!\retcode > @abs_srcdir@/data/external_table_with_user_type_and_default_value.csv;
+-- start_ignore
+
+-- end_ignore
+(exited with code 0)
+!\retcode > @abs_srcdir@/data/external_table_with_dropped_column.csv;
+-- start_ignore
+
+-- end_ignore
```
jimmyyih, line 23: These start_ignore and end_ignore lines don't seem to be needed.

test/acceptance/pg_upgrade/5-to-6/upgradeable_tests/source_cluster_regress/output/external_table.source
```diff
@@ -6,5 +6,82 @@
...
+(exited with code 0)
+INSERT INTO writable_external_table_with_dropped_column VALUES (1);
+INSERT 1
+
+!\retcode pkill gpfdist;
+-- start_ignore
+
+-- end_ignore
```
jimmyyih, line 86: Same with these start_ignore and end_ignore lines.

test/acceptance/pg_upgrade/5-to-6/upgradeable_tests/target_cluster_regress/input/external_table.source
jimmyyih: I don't think you can do DELETEs or UPDATEs on a writable external table.  I think it's only INSERTs and maybe COPY?

jimmyyih (approved):

test/acceptance/pg_upgrade/5-to-6/upgradeable_tests/target_cluster_regress/input/external_table.source
kalensk: Yeah, you're right!
 `ERROR:  cannot update or delete from external relation "writable_external_table_with_user_type_and_default_value"`
 
 COPY TO/FROM seems to work...I can add that.
 ```
COPY (SELECT * FROM readable_external_table_with_user_type_and_default_value ) to '/tmp/out';
COPY 1

COPY writable_external_table_with_user_type_and_default_value (a,b) FROM PROGRAM 'for i in `seq 1 3`; do echo $i,$i; done' DELIMITER ',';
COPY 3
 ```

PR #765, opened 2023-03-02
Add acceptance test for tables with name columns

jimmyyih:
Tables with name columns were previously banned from pg_upgrade. However, we recently removed the restriction in a recent GPDB commit. Add an acceptance test for test coverage.

GPDB commit reference:
https://github.com/greenplum-db/gpdb/commit/5850858ddb77e430d20c619a5a44092390afad48


test/acceptance/pg_upgrade/5-to-6/upgradeable_tests/source_cluster_regress/sql/name_data_type.sql
```diff
@@ -0,0 +1,22 @@
...
+CREATE TABLE table_with_name_columns (a int, b name, c text, d name);
+
+-- Do a simple insert.
+INSERT INTO table_with_name_columns SELECT i, 'bbb' || i, 'ccc' || i, 'ddd' || i FROM generate_series(1,10)i;
+
+-- Do an insert that maxes out the name data type. Name data is
+-- truncated to 63 chars so only the text value will store the 'z'
+-- at the end.
+INSERT INTO table_with_name_columns VALUES (88, 'aaaaaaaaaabbbbbbbbbbccccccccccddddddddddeeeeeeeeeeffffffffffgggz', 'aaaaaaaaaabbbbbbbbbbccccccccccddddddddddeeeeeeeeeeffffffffffgggz', 'aaaaaaaaaabbbbbbbbbbccccccccccddddddddddeeeeeeeeeeffffffffffgggz');
+
```
kalensk, line 20: Can we tweak the wording of the second sentence to be more clear. Maybe something like:

> The name data type is limited to 63 chars which will truncate the `z`. However, the text data type does not have this limitation and will show the ending `z`.

test/acceptance/pg_upgrade/5-to-6/upgradeable_tests/source_cluster_regress/expected/name_data_type.out
```diff
@@ -0,0 +1,39 @@
...
+ 4  | bbb4                                                            | ccc4                                                             | ddd4                                                            
+ 5  | bbb5                                                            | ccc5                                                             | ddd5                                                            
+ 6  | bbb6                                                            | ccc6                                                             | ddd6                                                            
+ 7  | bbb7                                                            | ccc7                                                             | ddd7                                                            
+ 88 | aaaaaaaaaabbbbbbbbbbccccccccccddddddddddeeeeeeeeeeffffffffffggg | aaaaaaaaaabbbbbbbbbbccccccccccddddddddddeeeeeeeeeeffffffffffgggz | aaaaaaaaaabbbbbbbbbbccccccccccddddddddddeeeeeeeeeeffffffffffggg
```
kalensk, line 35: Minor Side note: My eye got initially confused at seeing these...I wonder if putting the two name types side-by-side is more clear than having the text type in-between them.

jimmyyih: The text column in-between the two name columns was mainly to blindly test on-disk alignment (padding) before/after upgrade.  The int and text column are 4-byte aligned whereas the name column used to be 4-byte aligned but was changed to have no alignment needed.  So having one name column being in the middle of two 4-byte aligned columns and having the other name column be at the end of the tuple is what this is implicitly testing.

test/acceptance/pg_upgrade/5-to-6/upgradeable_tests/source_cluster_regress/sql/name_data_type.sql
jimmyyih: Updated although I used "store" instead of "show" since that's more correct for this INSERT statement.

test/acceptance/pg_upgrade/5-to-6/upgradeable_tests/source_cluster_regress/expected/name_data_type.out
kalensk: Makes sense and is definitely good to have! I like it and think having what you basically just said as a code comment would be awesome (at least for me).

test/acceptance/pg_upgrade/5-to-6/upgradeable_tests/source_cluster_regress/sql/name_data_type.sql
```diff
@@ -0,0 +1,22 @@
...
+-- Test to ensure that a table with name columns can be upgraded. This
+-- was previously banned but found to be upgradeable.
+
+--------------------------------------------------------------------------------
+-- Create and setup upgradeable objects
+--------------------------------------------------------------------------------
+
+CREATE TABLE table_with_name_columns (a int, b name, c text, d name);
+
+-- Do a simple insert.
```
kalensk, line 13: minor nit: 

I'd remove all these comments and just leave the one below about 'z' being truncated. A couple of reasons:
- It forces one to write self-documenting clear code. This to me is the real benefit and power.
- Obvious comments tend to not add any real value and get stale over time. They can often times be misleading or distracting to what the actual code is doing.

Brief yet detailed comments for "why" something non-obvious is being done is super helpful and is encouraged.

kalensk (approved):

kalensk:
Do we want to expand to add tests around the following. See [following tests](https://github.com/greenplum-db/gpupgrade/blob/8d758a09a9aef4b23fd11275fc87588830fd0f53/data-migration-scripts/5-to-6-seed-scripts/test/create_nonupgradable_objects.sql#L112-L150).
- create partition table with name datatype attribute as the not-first column as the partition key
- create table with name datatype attribute as the not-first column as the distribution key
- create table / views with name datatype
- multilevel partition table with partitioning keys using name datatype
- multilevel partition table with partitioning keys using not using name datatype

Im about to put up a PR today to remove the name data type migration scripts.

jimmyyih:
> Do we want to expand to add tests around the following. See [following tests](https://github.com/greenplum-db/gpupgrade/blob/8d758a09a9aef4b23fd11275fc87588830fd0f53/data-migration-scripts/5-to-6-seed-scripts/test/create_nonupgradable_objects.sql#L112-L150). The one with a view is pretty important since that's an edge case we ran into in various ways.
>
>    create partition table with name datatype attribute as the not-first column as the partition key
>    create table with name datatype attribute as the not-first column as the distribution key
>    create table / views with name datatype.
>    multilevel partition table with partitioning keys using name datatype
>    multilevel partition table with partitioning keys using not using name datatype

Interesting, thanks for catching that stuff.  For those proposed test cases, none of them really stand out to me as needed.

>    create partition table with name datatype attribute as the not-first column as the partition key

This is covered by this PR but more thoroughly.

>    create table with name datatype attribute as the not-first column as the distribution key

This doesn't seem too interesting given all supported types should work as distribution key.  If anything, we would create a table for every supported type in `distribution_key_and_data.sql` instead.  Might not be a bad idea actually to give extra coverage.

>    create table / views with name datatype.

Not sure why views would be relative here.  Views are dependent on their underlying tables so just checking if tables work is all we need.  Even if the view is on catalog tables with name columns, those catalog tables would've been recreated anyways with correctly aligned data.  What is the edge case that was hit?

>    multilevel partition table with partitioning keys using name datatype

Same thing with the distribution key.  The assumed issue with name datatypes that resulted in the pg_upgrade check was strictly for on-disk format/alignment.  This shouldn't matter for partitioning keys.  If anything, we would create a table for every supported type in a different SQL file instead.  Might not be a bad idea actually to give extra coverage.

>    multilevel partition table with partitioning keys using not using name datatype

I'm not sure what value this one provided since coverage relating to name datatypes would basically be the same as a regular heap table for just table data.

jimmyyih:
> Not sure why views would be relative here. Views are dependent on their underlying tables so just checking if tables work is all we need. Even if the view is on catalog tables with name columns, those catalog tables would've been recreated anyways with correctly aligned data. What is the edge case that was hit?

Okay, I think the edge case you might be talking about is that the pg_upgrade check also tagged views with name columns and that some postgis views (from https://github.com/greenplum-db/gpupgrade/pull/766, I see views `geography_columns`, `raster_columns`, and `raster_overviews` being dropped) that were just relaying information from catalog tables with name columns (e.g. `geography_columns` view outputs `pg_namespace.nspname`, `pg_class.relname`, and `pg_attribute.attname`).  I would guess other views would get tagged by this as well since those are pretty common columns to output in database administration views (e.g. gp_toolkit but the pg_upgrade check filtered that schema out).  IMO, that should have been a bug in the pg_upgrade check where the query should have filtered out views.

jimmyyih:
Updated to add an index on the name column as well since indexes have the name data on-disk as well.

PR #768, opened 2023-03-06
Remove developer stdout debug message

jimmyyih:
When running initialize, we see this output:
```
Proceeding with upgrade
Starting gpupgrade hub process...                                  [IN PROGRESS]Started hub
Starting gpupgrade hub process...                                  [COMPLETE]
Saving source cluster configuration...                             [COMPLETE]
```

The "Started hub" part doesn't seem like it should be there and messes with the IN PROGRESS to COMPLETE line replacement. With this patch, it'll now look like this:
```
Proceeding with upgrade
Starting gpupgrade hub process...                                  [COMPLETE]
Saving source cluster configuration...                             [COMPLETE]
```

### DIFF
--- a/cli/commanders/initialize.go
+++ b/cli/commanders/initialize.go
@@ -80,7 +80,6 @@ func StartHub() (err error) {
 		return xerrors.Errorf("%q failed with %q: %w", cmd.String(), string(output), err)
 	}
 
-	fmt.Println("Started hub")
 	log.Printf("%s", output)
 	return nil
 }
--- a/cli/commands/commands.go
+++ b/cli/commands/commands.go
@@ -194,6 +194,10 @@ var restartServices = &cobra.Command{
 			return err
 		}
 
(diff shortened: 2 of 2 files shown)


kalensk (changes requested):
This also removes "Started hub" when the user runs "gpupgrade kill-services". The required change is more complicated.

jimmyyih:
> This also removes "Started hub" when the user runs "gpupgrade restart-services". We want to preserve that message for restart-services. The required change is a bit more complicated.

Good catch.  Made the suggested change.

kalensk (approved):
Looks good! Thanks for making the change : )

(Note: For a future PR or similar. Just making an observation. During my limited testing I noticed `gpupgrade restart-services` print `Restarted agents on:` even if there are no agents to restart. That should probably be conditionalized and maybe add another message saying nothing to restart or all good or something for the user.)

kalensk:
Can probably close GPUP-62

PR #770, opened 2023-03-14
Initialize: UI Tweaks

kalensk:
1) Don't prompt to proceed with the upgrade if the user chose to quit from executing the data migration scripts.

2) Remove --automatic from initialize

3) Update initialize confirmation text since data migration scripts are now folded into the initialize command.

Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:UItweaks


cli/commands/initialize.go
```diff
@@ -303,7 +303,6 @@ To return the cluster to its original state, run "gpupgrade revert".`,
 	subInit.Flags().BoolVarP(&verbose, "verbose", "v", false, "print the output stream from all substeps")
 	subInit.Flags().BoolVar(&pgUpgradeVerbose, "pg-upgrade-verbose", false, "execute pg_upgrade with --verbose")
 	subInit.Flags().StringVarP(&file, "file", "f", "", "the configuration file to use")
-	subInit.Flags().BoolVarP(&nonInteractive, "automatic", "a", false, "do not prompt for confirmation to proceed")
 	subInit.Flags().BoolVar(&nonInteractive, "non-interactive", false, "do not prompt for confirmation to proceed")
 	subInit.Flags().MarkHidden("non-interactive") //nolint
```
jimmyyih, line 307: Interesting.  So this `non-interactive` hidden flag was always there.  Curious, I found that it was introduced in https://github.com/greenplum-db/gpupgrade/commit/6ec133fa0572c48509864677c74ecb31fea6fff2 but as part of a very large PR.  What was it going to be for?

cli/commands/help_text.go
```diff
@@ -109,7 +109,7 @@ Required Flags:
 
 Optional Flags:
 
-  -a, --automatic            suppress summary & confirmation dialog
+  --non-interactive          suppress summary & confirmation dialog
```
kyeap-vmware, line 112: It looks like we override commanders default generated help text. So marking the flags as hidden won't remove it from help text.

We'll need to remove line 112 and 165 to hide from users.

jimmyyih (approved):

cli/commands/help_text.go
kalensk: Ah! Good catch. Thank you.

cli/commands/initialize.go
kalensk: Yeah, `--non-interactive` is present across all commands for CI and internal testing. It is not exposed to end users.

Originally, initialize also had the `automatic` flag which we essentially aliased to `non-interactive`. This was done to be consistent with all the other commands. Now that automatic is being removed for initialize we are still keeping the "hidden" non-interactive flag for all commands for CI and internal testing.

kyeap-vmware (approved):

PR #771, opened 2023-03-15
Add test to verify AO tables with unused aoblkdirs can be upgraded

kyeap-vmware:
When an index is created for an AO table an aoblkdir relation gets created for the index. But when all indexes are dropped for the table, the aoblkdir does not get removed. This scenario needs to be handled gracefully by gpupgrade. This test ensures The aoblkdirs for AO tables with no indexes are filtered out and not upgraded to the target cluster.

With the changes to upgrading AO aux tables using relfilenodes, this is now the expected behavior.

Upgrade AO aux table using relfilenodes reference commit: https://github.com/greenplum-db/gpdb/commit/xxxxxx

Must be merged after AO aux table optimizations are merged.


test/acceptance/pg_upgrade/5-to-6/upgradeable_tests/source_cluster_regress/sql/ao_indexes.sql
```diff
@@ -40,3 +40,29 @@ SELECT * FROM ao_with_gist WHERE a @@ to_tsquery('footext1');
...
+
+
+-- When the last index on an AO table is deleted, the table's corresponding
+-- pg_aoblkdir is not deleted. The aoblkdir is first created when an index is
+-- created for the AO table. If an index is created for an AO table then
+-- deleted, an aodblkdir relation is left on the source cluster. During
+-- gpupgrade, pg_dump will not dump an index which means the aoblkdir does not
+-- get created on the target cluster. If this is not accounted for, gpupgrade
+-- will error out due to relation mismatch. To resolve this edge case,
+-- gpupgrade filters out aoblkdirs for AO tables with no indexes.
```
jimmyyih, line 53: s/gpupgrade/pg_upgrade/

Suggestion is made to better distinguish where the filtering is happening (e.g. search in pg_upgrade code instead of gpupgrade code).

jimmyyih (approved):

PR #775, opened 2023-03-21
Update pg_upgrade acceptance tests upgradeable source schedule

jimmyyih:
The first parallel test group in the test schedule may open up too many database connections (compared to the max_connections GUC setting which is set to 25 in our CI) and can cause one of the tests to hit this error:
```
ERROR:  failed to acquire resources on one or more segments
DETAIL:  FATAL:  sorry, too many clients already
```

Simply move some of the newer tests over to the next parallel group to prevent the error.


kyeap-vmware (approved):

kalensk (approved):
LGTM. 

Wouldn't it make more sense overall to increase max_connections for the pg_upgrade tests?  Run more tests in parallel to have a faster feedback cycle, etc. Seems win-win.

Seems like an easy thing to do in `ci/scripts/pg-upgrade-tests.bash` with a detailed comment explaining why such as:

```BASH
    echo "Creating the source demo cluster..."
    create_source_cluster
    
    # Increase max_connections since running many tests in parallel saturates 
    # the default connections of 25 resulting in 
    # "ERROR:  failed to acquire resources on one or more segments"
    ${GPHOME_SOURCE}/bin/gpconfig -c max_connections -v 100
    gpstop -u
    
    echo "Running tests..."
    run_pg_upgrade_tests
```

May need to source greenplum_path.sh so the following instead:
```
(source ${GPHOME_SOURCE}/greenplum_path.sh && ${GPHOME_SOURCE}/bin/gpconfig -c max_connections -v 100) && ${GPHOME_SOURCE}/bin/gpstop -u`
```

jimmyyih:
> Wouldn't it make more sense overall to increase max_connections for the pg_upgrade tests? Run more tests in parallel to have a faster feedback cycle, etc. Seems win-win.

Not necessarily.  The max_connections default is 25 for gpdemo cluster so increasing it and putting everything into one parallel test group would make local dev testing more annoying if we ourselves do not use the default gpdemo cluster settings that every other team uses.  Also, keeping it at 25 helps with keeping a low testing memory footprint since max_connections is used to calculate the overall shared memory size for each segment (great for testing with containers and locally).  Also, the speed up for putting everything into one parallel group wouldn't really be much with these particular tests; if anything it would only be 1 or 2 seconds saved.

kalensk:
> Not necessarily. The max_connections default is 25 for gpdemo cluster so increasing it and putting everything into one parallel test group would make local dev testing more annoying if we ourselves do not use the default gpdemo cluster settings that every other team uses. Also, keeping it at 25 helps with keeping a low testing memory footprint since max_connections is used to calculate the overall shared memory size for each segment (great for testing with containers and locally). Also, the speed up for putting everything into one parallel group wouldn't really be much with these particular tests; if anything it would only be 1 or 2 seconds saved.

Sounds good.

> would make local dev testing more annoying if we ourselves do not use the default gpdemo cluster settings that every other team uses. 

FWIW I pretty much never run all the tests locally. Just the one Im working on and then fly a dev pipeline to let the cloud do the annoying work for me while I do something else. It also ensures that the dev pipeline "passes".

> Also, keeping it at 25 helps with keeping a low testing memory footprint since max_connections is used to calculate the overall shared memory size for each segment (great for testing with containers and locally). 

Ah, interesting. Good to know. Presumably the memory would still be manageable even for containers. And it doesn't just statically allocate a bunch of memory all at once.

> Also, the speed up for putting everything into one parallel group wouldn't really be much with these particular tests; if anything it would only be 1 or 2 seconds saved.

True.

PR #776, opened 2023-03-22
Fix spelling of quitting

jimmyyih:
Online dictionaries show that it should be quitting, not quiting. Do a quick search and replace to fix the trivial spelling mistake.


kalensk (approved):
LGTM. Ha! At least I was consistent.

jimmyyih:
> LGTM. Ha! At least I was consistent.

Consistent indeed. 😄

PR #777, opened 2023-03-22
pass logDir to finalize & revert data migration scripts

kalensk:
**Note:** Let me know if there is a better interface for `ApplyDataMigrationScripts(nonInteractive bool, gphome string, port int, logDir string, currentScriptDirFS fs.FS, currentScriptDir string, phase idl.Step)` in terms of passing in logDir and deriving currentScriptDir from logDir.

Previously, applying the finalize and revert data migration scripts would fail trying to open
/home/gpadmin/gpAdminLogs/gpupgrade/apply_revert.log

This is because the gpupgrade log directory has already been archived. And internally the apply data migration script code is calling utils.GetLogDir() rather than the archived directory. Thus, pass either the logDir during initialize, or the logArchiveDirectory during finalize & revert.

By passing in the log directory this removes the need for input-dir and output-dir as those can be derived from the log directory.

Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:passLogDir

### DIFF
--- a/cli/commanders/data_migration_apply.go
+++ b/cli/commanders/data_migration_apply.go
@@ -25,7 +25,7 @@ import (
 	"github.com/greenplum-db/gpupgrade/utils/errorlist"
 )
 
-func ApplyDataMigrationScripts(nonInteractive bool, gphome string, port int, currentScriptDirFS fs.FS, currentScriptDir string, phase idl.Step) error {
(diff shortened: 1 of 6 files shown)


kalensk:
>  Shouldn't one of the gpupgrade acceptance tests have caught this? It was pretty easy to reproduce the issue manually.

Unfortunately, I don't believe so since this bug only appears when running revert or finalize _interactively_. The gpugprade accpetance tests run the data migration scripts manually with `gpupgrade generate` and `gpupgrade apply` since we run initialize and the other commands _non-interactively_ for the CI.

PR #778, opened 2023-03-23
Consolidate substep text in help text and confirmation prompts

kyeap-vmware:
Help text and confirmation prompt both show the substeps for the
command gpupgrade intends to run. They should be the same.

pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:consolidate-substep-text

### DIFF
--- a/cli/commands/help_text.go
+++ b/cli/commands/help_text.go
@@ -8,6 +8,8 @@ import (
 	"os"
 
 	"github.com/spf13/cobra"
+	"golang.org/x/text/cases"
+	"golang.org/x/text/language"
 
 	"github.com/greenplum-db/gpupgrade/cli/commanders"
 	"github.com/greenplum-db/gpupgrade/idl"
@@ -15,15 +17,24 @@ import (
 )
 
 var (
-	Help           map[string]string
-	InitializeHelp string
-	ExecuteHelp    string
-	FinalizeHelp   string
-	RevertHelp     string
+	Help               map[idl.Step]string
+	initializeSubsteps commanders.Substeps
+	executeSubsteps    commanders.Substeps
+	finalizeSubsteps   commanders.Substeps
+	revertSubsteps     commanders.Substeps
+	InitializeHelp     string
+	ExecuteHelp        string
+	FinalizeHelp       string
+	RevertHelp         string
 )
 
 func init() {
-	InitializeHelp = GenerateHelpString(initializeHelp, []idl.Substep{
+	logDir, err := utils.GetLogDir()
+	if err != nil {
+		panic(fmt.Sprintf("failed to get log directory: %v", err))
+	}
+
+	initializeSubsteps = commanders.Substeps{
 		idl.Substep_generate_data_migration_scripts,
 		idl.Substep_execute_stats_data_migration_scripts,
 		idl.Substep_execute_initialize_data_migration_scripts,
@@ -39,16 +50,18 @@ func init() {
 		idl.Substep_shutdown_target_cluster,
 		idl.Substep_backup_target_master,
 		idl.Substep_check_upgrade,
-	})
-	ExecuteHelp = GenerateHelpString(executeHelp, []idl.Substep{
+	}
+
+	executeSubsteps = commanders.Substeps{
 		idl.Substep_check_active_connections_on_source_cluster,
 		idl.Substep_shutdown_source_cluster,
 		idl.Substep_upgrade_master,
 		idl.Substep_copy_master,
 		idl.Substep_upgrade_primaries,
 		idl.Substep_start_target_cluster,
-	})
-	FinalizeHelp = GenerateHelpString(finalizeHelp, []idl.Substep{
+	}
+
+	finalizeSubsteps = commanders.Substeps{
 		idl.Substep_check_active_connections_on_target_cluster,
 		idl.Substep_upgrade_mirrors,
 		idl.Substep_upgrade_standby,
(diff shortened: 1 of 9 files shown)


cli/commands/help_text.go
```diff
@@ -15,75 +15,94 @@ import (
...
-		idl.Substep_stop_hub_and_agents,
-		idl.Substep_delete_master_statedir,
-		idl.Substep_execute_revert_data_migration_scripts,
-	})
+	logdir, err := utils.GetLogDir()
+	if err != nil {
+		panic(fmt.Sprintf("failed to get log directory: %v", err))
+	}
+
+	InitializeSubsteps = "Initialize will carry out the following steps:\n" +
```
kalensk, line 35: Is there any way we can remove the text `"Initialize will carry out the following steps:\n" +` from these variables so its simply a list of substeps? Have it solely be a list of substeps. To me that is much clearer and follows "single responsibility principle". Maybe move the phrase up a level?

Side Note: If it happens to be useful one can use the step variable instead of hard coding to "Initialize will carry out..." with `"%s will carry out ...", cases.Title(language.English).String(idl.Step_initialize.String()))`

cli/commands/help_text.go
```diff
@@ -294,18 +309,13 @@ gpupgrade log files can be found on all hosts in %s
...
-	}
-
-	logdir, err := utils.GetLogDir()
-	if err != nil {
-		panic(fmt.Sprintf("failed to get log directory: %v", err))
+		substepStr += fmt.Sprintf(" - %s\n", commanders.SubstepDescriptions[substep].HelpText)
 	}
 
-	return fmt.Sprintf(baseString, formattedList, logdir)
+	return substepStr
```
kalensk, line 318: very minor suggestion: I realize this was already there, but for clarity renaming `commandList` to `substeps` and renaming `substepStr` to `output` perhaps.

Another neat idea is to create a []idl.Substep type and rename this function to String() as a method on that type. Then when you call %s on that type it will "automatically: print it out. For example,

```go
type Substeps []idl.Substep

func (s Substeps) String() string {
	var output string
	for _, substep := range s {
		output += fmt.Sprintf(" - %s\n", commanders.SubstepDescriptions[substep].HelpText)
	}

	return output
}
```

All callers would be updated from using type `[]idl.Substep` to type `Substeps`.

cli/commands/help_text.go
```diff
@@ -15,75 +15,94 @@ import (
...
+			idl.Substep_delete_segment_statedirs,
+			idl.Substep_stop_hub_and_agents,
+			idl.Substep_delete_master_statedir,
+			idl.Substep_execute_revert_data_migration_scripts,
+		})
+
+	InitializeHelp = fmt.Sprintf(initializeHelp, InitializeSubsteps, logdir)
+	ExecuteHelp = fmt.Sprintf(executeHelp, ExecuteSubsteps, logdir)
+	FinalizeHelp = fmt.Sprintf(finalizeHelp, FinalizeSubsteps, logdir)
+	RevertHelp = fmt.Sprintf(revertHelp, RevertSubsteps, logdir)
```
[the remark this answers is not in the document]
kyeap-vmware: These variables are being used in a few other places. Making the changes above would result in a more diffs with little benefit.

cli/commands/help_text.go
kyeap-vmware: Creating a new variable type for just one minor function seems excessive. We can re-evaluate if this expands further. The variable renames suggestions can be taken though.

cli/commands/help_text.go
```diff
@@ -83,7 +94,18 @@ func init() {
 		idl.Substep_stop_hub_and_agents,
 		idl.Substep_delete_master_statedir,
 		idl.Substep_execute_revert_data_migration_scripts,
-	})
+	}
+
+	initializeSubstepHelpText = fmt.Sprintf("%s will carry out the following steps:\n%s", cases.Title(language.English).String(idl.Step_initialize.String()), GenerateSubstepString(initializeSubsteps))
+	executeSubstepHelpText = fmt.Sprintf("%s will carry out the following steps:\n%s", cases.Title(language.English).String(idl.Step_execute.String()), GenerateSubstepString(executeSubsteps))
+	finalizeSubstepHelpText = fmt.Sprintf("%s will carry out the following steps:\n%s", cases.Title(language.English).String(idl.Step_finalize.String()), GenerateSubstepString(finalizeSubsteps))
+	revertSubstepHelpText = fmt.Sprintf("%s will carry out the following steps:\n%s", cases.Title(language.English).String(idl.Step_revert.String()), GenerateSubstepString(revertSubsteps))
```
kalensk, line 102: Let's inline these can get rid of them! Would make things a lot more simple.

cli/commands/help_text.go
```diff
@@ -294,18 +312,13 @@ gpupgrade log files can be found on all hosts in %s
...
-	if err != nil {
-		panic(fmt.Sprintf("failed to get log directory: %v", err))
+func GenerateSubstepString(substepList []idl.Substep) string {
+	var output string
+	for _, substep := range substepList {
+		output += fmt.Sprintf(" - %s\n", commanders.SubstepDescriptions[substep].HelpText)
 	}
 
-	return fmt.Sprintf(baseString, formattedList, logdir)
+	return output
```
kalensk, line 321: This will make so much more sense as a function on a []Substeps struct inside substeps.go. Can we do that?

```go
type Substeps []idl.Substep

func (s Substeps) String() string {
	var output string
	for _, substep := range s {
		output += fmt.Sprintf(" - %s\n", SubstepDescriptions[substep].HelpText)
	}

	return output
}
```

cli/commands/initialize.go
```diff
@@ -164,7 +164,7 @@ func initialize() *cobra.Command {
 				return err
 			}
 
-			confirmationText := fmt.Sprintf(initializeConfirmationText, logdir, configPath,
+			confirmationText := fmt.Sprintf(initializeConfirmationText, initializeSubstepHelpText, logdir, configPath,
```
kalensk, line 167: This is a code smell!! Seeing a variable called `initializeSubstepHelpText` being passed to the `initializeConfirmationText`. Rather we should fix this to be passing `initializeSubsteps`. 

This can occur if we inline the variables `initializeSubstepHelpText`.

kalensk (approved):

PR #779, opened 2023-03-25
Fix revert after early initialize exit

jimmyyih:
With the migration scripts being generated and applied at the beginning of the initialize command now, starting up the hub and creating the config file is done much later. Both are required for revert to succeed but we allow users to quit early during the migration scripts part of Initialize. If the user quits Initialize during that early phase, running revert will fail. To fix the issue, we create and prepopulate the config file and start the hub prior to the migration script logic in Initialize. New step logic has been introduced as well to allow us to skip over unnecessary revert substeps.

Dev pipeline:
https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:revert_after_migration_scripts

### DIFF
--- a/cli/commanders/initialize.go
+++ b/cli/commanders/initialize.go
@@ -4,13 +4,14 @@
 package commanders
 
 import (
-	"fmt"
 	"log"
 	"os"
 	"os/exec"
 
 	"golang.org/x/xerrors"
 
+	"github.com/greenplum-db/gpupgrade/greenplum"
+	"github.com/greenplum-db/gpupgrade/hub"
 	"github.com/greenplum-db/gpupgrade/step"
 	"github.com/greenplum-db/gpupgrade/upgrade"
 	"github.com/greenplum-db/gpupgrade/utils"
@@ -37,7 +38,7 @@ func CreateStateDir() (err error) {
 	return nil
 }
 
-func CreateConfigFile(hubPort int) error {
+func CreateConfigFile(hubPort int, sourcePort int, sourceGPHome string) error {
 	path := upgrade.GetConfigFile()
 
 	exist, err := upgrade.PathExist(path)
@@ -53,13 +54,17 @@ func CreateConfigFile(hubPort int) error {
 	// Bootstrap with the port to enable the CLI helper function connectToHub to
 	// work with both initialize and all other CLI commands. This overloads the
 	// hub's persisted configuration with that of the CLI when ideally these
-	// would be separate.
-	err = os.WriteFile(path, []byte(fmt.Sprintf(`{"Port": %d}`, hubPort)), 0644)
-	if err != nil {
-		return err
-	}
+	// would be separate. Also bootstrap with the source $GPHOME and coordinator
+	// port to handle the initialize scenario where the user quits before the
+	// InitializeRequest is sent to the hub and wants to revert.
+	config := hub.Config{}
+	config.Port = hubPort
+	config.Source = &greenplum.Cluster{}
+	config.Source.Primaries = make(greenplum.ContentToSegConfig)
+	config.Source.Primaries[-1] = greenplum.SegConfig{Port: sourcePort}
+	config.Source.GPHome = sourceGPHome
 
-	return nil
+	return config.SaveConfig()
 }
 
 func StartHub() (err error) {
--- a/hub/fill_cluster_configs.go
+++ b/hub/fill_cluster_configs.go
@@ -18,27 +18,55 @@ import (
 	"github.com/greenplum-db/gpupgrade/utils/errorlist"
 )
 
-// FillConfiguration populates the Config saves it to disk.
-func FillConfiguration(config *Config, request *idl.InitializeRequest, saveConfig func() error) error {
+var InitializeConnectionFunc = initializeConnection
+
+func initializeConnection(gphome string, port int) (*sql.DB, error) {
 	tempSource, err := greenplum.NewCluster([]greenplum.SegConfig{})
 	if err != nil {
-		return err
+		return nil, err
 	}
 
 	// Need greenplum version to use correct utility mode parameter when making the database connection URI.
-	sourceVersion, err := greenplum.Version(request.GetSourceGPHome())
+	tempSource.Version, err = greenplum.Version(gphome)
 	if err != nil {
-		return err
+		return nil, err
 	}
 
 	tempSource.Destination = idl.ClusterDestination_source
-	tempSource.Version = sourceVersion
-	conn := tempSource.Connection([]greenplum.Option{greenplum.Port(int(request.GetSourcePort())), greenplum.UtilityMode()}...)
-
+	conn := tempSource.Connection([]greenplum.Option{greenplum.Port(port), greenplum.UtilityMode()}...)
 	db, err := sql.Open("pgx", conn)
 	if err != nil {
-		return err
+		return nil, err
 	}
+
+	return db, nil
+}
+
+// We only need specific config values to be set for the hub RevertResponse
+// to handle reverting an early Initialize exit.
+func GetEarlyInitializeConfiguration(hubPort int, coordinatorPort int, gphome string) (*Config, error) {
+	db, err := InitializeConnectionFunc(gphome, coordinatorPort)
+	if err != nil {
+		return nil, xerrors.Errorf("create connection: %w", err)
+	}
+	defer db.Close()
+
+	source, err := greenplum.ClusterFromDB(db, gphome, idl.ClusterDestination_source)
+	if err != nil {
+		return nil, xerrors.Errorf("retrieve source configuration: %w", err)
+	}
+
+	conf := &Config{}
+	conf.Source = &source
+	conf.UpgradeID = upgrade.NewID()
+	conf.Port = hubPort
+
+	return conf, nil
+}
+
+// FillConfiguration populates the Config saves it to disk.
+func FillConfiguration(config *Config, request *idl.InitializeRequest, saveConfig func() error) error {
+	db, err := InitializeConnectionFunc(request.GetSourceGPHome(), int(request.GetSourcePort()))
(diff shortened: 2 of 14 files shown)


hub/fill_cluster_configs.go
```diff
@@ -18,27 +18,55 @@ import (
...
+        if err != nil {
+                return nil, err
+        }
+
+        return db, nil
+}
+
+// We only need specific Server values to be set for the hub RevertResponse
+// to handle reverting an early Initialize exit.
+func LoadEarlyInitializeConfiguration(server *Server) error {
```
kalensk, line 48: I strongly encourage just passing in what is needed rather than the full Server object. In this case looks like just the source cluster seems to be needed, so `source greenplum.Cluster` as the arg should suffice?

If you are trying to assign values to the server config, its best to have the function return what is needed rather than passing in the object C-style. It keeps things so much easier to understand where mutation is occurring, and is one of the powers of a higher level language (in a sense). Most bugs or misunderstandings come from mutation. Examples down below in `FillConfiguration()` with lines like `config.UpgradeID = upgrade.NewID()` or  `source, err := greenplum.ClusterFromDB(...)`

cli/commanders/initialize.go
```diff
@@ -37,7 +39,27 @@ func CreateStateDir() (err error) {
...
+	enc := json.NewEncoder(&buffer)
+	enc.SetIndent("", "  ")
+	if err := enc.Encode(c); err != nil {
+		return xerrors.Errorf("save config: %w", err)
+	}
+
+	return utils.AtomicallyWrite(file, buffer.Bytes())
+}
+
+func CreateConfigFile(hubPort int, sourcePort int, sourceGPHome string) error {
```
kalensk, line 62: This looks like a great opportunity to extract out the hub config struct and associated functions - Save, Load, SaveConfig, etc. into its own package which can be used by both the CLI and hub. So much repeated code and extra code that could be collapsed and simplified. Maybe its just me, but Im not seeing the benefit yet of EarlyInitializeConfig rather than re-using the config struct.

greenplum/segconfig.go
```diff
@@ -60,6 +60,31 @@ func (s *SegConfig) IsOnHost(hostname string) bool {
...
+	}
+
+	var datadir string
+	if err := db.QueryRow(query).Scan(&datadir); err != nil {
+		return "", xerrors.Errorf("scanning gp_segment_configuration: %w", err)
+	}
+
+	return datadir, nil
+}
+
```
kalensk, line 87: Couldn't we just use `greenplum.ClusterFromDB()` instead of creating a new function? Just re-use the same code/pattern that FillConfiguration() is doing to initialize the source cluster config object which is exactly what we need. No need for extra code or special-one off code. Keeps things simple.

jimmyyih: The `greenplum.ClusterFromDB()` function grabs too much unnecessary information which will be unnecessarily outputted to the early config file (e.g. all primaries and mirrors segment info) and would make it more annoying to observe the early config file for dev debugging and/or production debugging (I wouldn't want to look at 99% unrelated info in the middle of looking at a production database).  But I think I can filter out the unneeded segment info by adding a `func (c ContentToSegConfig) OnlyCoordinator() ContentToSegConfig` function.  I'll do that since it's enhancing something existing rather than adding something new as you're saying.

hub/fill_cluster_configs.go
jimmyyih: > I strongly encourage just passing in what is needed rather than the full Server object. In this case looks like just the source cluster seems to be needed, so source greenplum.Cluster as the arg should suffice?
>
>  If you are trying to assign values to the server config, its best to have the function return what is needed rather than passing in the object C-style. It keeps things so much easier to understand where mutation is occurring, and is one of the powers of a higher level language (in a sense). Most bugs or misunderstandings come from mutation. Examples down below in FillConfiguration() with lines like config.UpgradeID = upgrade.NewID() or source, err := greenplum.ClusterFromDB(...)

The function name references loading so doing pointer logic to define the struct makes more sense and is pretty common.  Although I could change the argument from a Server object pointer to the underlying Config object pointer (to match what `FillConfiguration()` is doing), I'll change it to something like this instead which makes more sense (we're technically not loading anything anyways; more so instead getting stuff) and better matches your suggestion:
```
func GetEarlyInitializeConfiguration(gphome string) *Config, error
```

jimmyyih: > But, it's actually not common (if used at all) in gpupgrade code base as we purposely avoided that for the reasons I listed. I think it's useful to follow conventions of the code base and surrounding code.

One quick example is the hub Initialize which calls the `FillConfiguration()` function that takes in a Config object pointer and fills it out instead of returning a Config object.  But other than that, yes it does seem to be the code base's convention as you mention.  I'll update `FillConfiguration()` in a different PR to meet the convention since I was incorrectly using that function's logic as the basis for mine.

cli/commanders/initialize.go
jimmyyih: Yes, it didn't look good to import the hub package into the commanders package (the two packages looked like they shouldn't mix) so I went with this quick implementation.  Doesn't seem too big of a refactor to move the logic into its own package so I'll take care of that now instead of a separate PR.  Does the `upgrade` package seem good for this?  It already references the config with `upgrade.ConfigFileName` and `upgrade.GetConfigFile()` so kinda makes sense to me.

greenplum/segconfig.go
jimmyyih: Sure, updated.

cli/commanders/initialize.go
kalensk: Creating a new package called `config` or `state` something along those lines is the way to go. Thinking of good name is key. We should move `upgrade.ConfigFileName` and `upgrade.GetConfigFile()` and anything else like that there.

We try to keep packages somewhat small (except for like the CLI, hub, and agent packages) such that they have a "single responsibility and purpose". Similar to how a struct or class should do one thing and do it well. This keeps things manageable, clear, and prevents circular dependencies (a problem this code base used to have). 

The upgrade package's responsibility is to run pg_upgrade and anything related to pg_upgrade. Its morphed _slightly_ from that over time, but its not useful to contribute to that morphing. Thus, a new package is the way to go.

jimmyyih: So I created a refactor patch to create a `config` package under `upgrade/config/config.go` but the refactor patch became too big at the end for this PR (as you say, there's a lot touching this config logic).  Also, I found that the `commands` package imports the `hub` package in some files so my main nit of not mixing the packages didn't feel as correct anymore... and also the `Config` object being in the `hub` package kept making more sense as I worked on the refactor since it's mainly the hub that uses it.

With that said, I simply removed all the new `EarlyInitializeConfig` stuff and just imported the `hub` package to satisfy the PR.  When the PR is merged, I'll open up a separate PR for the Config refactor patch as I feel it deserves its own separate attention and greatly escapes the scope of this PR of fixing a minor bug.

hub/revert.go
```diff
@@ -41,6 +41,20 @@ func (s *Server) Revert(_ *idl.RevertRequest, stream idl.CliToHub_RevertServer)
...
+		if err != nil {
+			return err
+		}
+
+		st.JumpToAndOnlyRun(
+			idl.Substep_archive_log_directories,
+			idl.Substep_delete_segment_statedirs,
+		)
+	}
+
```
kalensk, line 57: Another maybe more clearer idea is to change the condition to use the following:

```go
	configFilled, err := step.HasCompleted(idl.Step_initialize, idl.Substep_saving_source_cluster_config)
	if err != nil {
		return err
	}


	if !configFilled {
          // ...
}	
```

jimmyyih: Updated.

cli/commanders/initialize.go
```diff
@@ -53,13 +54,17 @@ func CreateConfigFile(hubPort int) error {
...
+	// port to handle the initialize scenario where the user quits before the
+	// InitializeRequest is sent to the hub and wants to revert.
+	earlyInitializeConfig := hub.Config{}
+	earlyInitializeConfig.Port = hubPort
+	earlyInitializeConfig.Source = &greenplum.Cluster{}
+	earlyInitializeConfig.Source.Primaries = make(greenplum.ContentToSegConfig)
+	earlyInitializeConfig.Source.Primaries[-1] = greenplum.SegConfig{Port: sourcePort}
+	earlyInitializeConfig.Source.GPHome = sourceGPHome
+
+	return earlyInitializeConfig.SaveConfig()
```
kalensk, line 67: minor: For easy of reading let's just name this `config` instead of `earlyInitializeConfig`

hub/fill_cluster_configs.go
```diff
@@ -18,27 +18,55 @@ import (
...
+        tempSource.Destination = idl.ClusterDestination_source
+	conn := tempSource.Connection([]greenplum.Option{greenplum.Port(port), greenplum.UtilityMode()}...)
+        db, err := sql.Open("pgx", conn)
+        if err != nil {
+                return nil, err
+        }
 
-	db, err := sql.Open("pgx", conn)
+        return db, nil
+}
```
kalensk, line 43: Might be the github UI, but double check the indentation around here. I use intelliJ, but `gofmt` and a few other similar tools should automatically take care of that and organizing the imports.

kalensk (approved):
LGTM. Looking forward to the followup on extracting config to its own package and cleaning up and organizing some of its uses now that its used in both the CLI and Hub.

(Minor suggestion on variable naming and to double checking indentation.)

cli/commanders/initialize.go
jimmyyih: Updated.  Thanks for the reminder.

hub/fill_cluster_configs.go
jimmyyih: Updated.

PR #781, opened 2023-03-28
data migration scripts: refactor global scripts

kalensk:
Global scripts are ones that are only applied once for the postgres database, rather than multiple times for each database.

Refactor to check if a seed script is global rather than if the seed script directory is global. The impetus for this change is to combine the cluster_stats and database_stats into a single directory such that customers are forced to run "all" stats scripts rather than "some".

Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:refactorGlobalScripts


cli/commanders/data_migration_generate.go
```diff
@@ -282,9 +282,9 @@ func GenerateScriptsPerDatabase(database DatabaseName, gphome string, port int,
 	return nil
 }
 
-// Generate one global script for the postgres database rather than all databases.
 func isGlobalScript(scriptDir string, database string) bool {
-	return database != "postgres" && (scriptDir == "gphdfs_user_roles" || scriptDir == "cluster_stats")
+	// Generate one global script for the postgres database rather than all databases.
+	return database != "postgres" && (scriptDir == "gen_alter_gphdfs_roles.sql" || scriptDir == "generate_cluster_stats.sh")
 }
```
jimmyyih, line 288: Should we change `scriptDir` to just `script` now that we're giving it `script.Name()` instead of `scriptDir.Name()`?

jimmyyih (approved):

cli/commanders/data_migration_generate.go
kalensk: Ah, yeah! Updated.

PR #782, opened 2023-03-28
bump minimum supported GPDB versions

kalensk:
Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:bumpMinVersions

### DIFF
--- a/greenplum/allowed_gpdb_versions.go
+++ b/greenplum/allowed_gpdb_versions.go
@@ -10,9 +10,9 @@ import (
 )
 
 // Change these values to bump the minimum supported versions and associated tests.
-var min5xVersion = "5.29.7"
-var min6xVersion = "6.21.1"
-var min7xVersion = "7.0.0"
+const min5xVersion = "5.29.10"
+const min6xVersion = "6.24.0"
+const min7xVersion = "7.0.0"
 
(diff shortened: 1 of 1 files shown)


jimmyyih (approved):

kyeap-vmware (approved):

li-bupt:
Hi, I would like to know why the minimum version was updated to 6.24? Are there any issues with using gpupgrade in previous versions？(⊙o⊙)

kalensk:
> Does this mean that I can port the latest pg_upgrade to an older version of gp and upgrade without version checking?

I don't think I understand your question. Can you be more specific in what you are asking and trying to do? 

With this update gpupgrade requires GPDB server 5.29.10 and 6.24.0. This is there for the safety and benefit of the user as explained above since there are pg_upgrade fixes and features some of which are required.

li-bupt:
I found that gp6 to gp7 upgrade is supported in the code, so I want to try to upgrade 6.21.1 gp version to 7.0. Can I port 6.24 pg_upgrade to 6.21.1 gp and then use gpupgrade? :-)

li-bupt:
Thank you for such a detailed reply!

PR #785, opened 2023-03-29
Refactor hub config into its own top-level package

jimmyyih:
The hub config logic was becoming too cumbersome as more usage was added. To make things easier, we extract the hub config into its own package that can be used by the cli and hub. Along with that, the functionality has been updated to better reflect current gpupgrade coding best practices. The backupdir stuff was also put into its own package under the config directory.

dev pipeline:
https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:refactor_config


cli/commands/initialize.go
```diff
@@ -191,7 +192,7 @@ func initialize() *cobra.Command {
 			})
 
 			st.RunCLISubstep(idl.Substep_saving_source_cluster_config, func(streams step.OutStreams) error {
-				return commanders.CreateConfigFile(hubPort, sourcePort, sourceGPHome)
+				return config.CreateConfigFile(hubPort, sourcePort, sourceGPHome)
```
kalensk, line 195: Not necessarily part of this PR. Given this change and where we are at I think it make more sense to rename this from `config.CreateConfigFile()` to possible `config.Create()`.

config/fill_cluster_configs.go
```diff
@@ -42,58 +43,43 @@ func initializeConnection(gphome string, port int) (*sql.DB, error) {
...
-	conf.Source = &source
-	conf.UpgradeID = upgrade.NewID()
-	conf.Port = hubPort
-
-	return conf, nil
-}
-
-// FillConfiguration populates the Config saves it to disk.
-func FillConfiguration(config *Config, request *idl.InitializeRequest, saveConfig func() error) error {
+func GetInitializeConfiguration(hubPort int, request *idl.InitializeRequest, wasEarlyExit bool) (*Config, error) {
```
kalensk, line 46: Let's not add this parameter `wasEarlyExit bool` for this PR.

It's more more clearer and simpler to simply re-use the _entire_ function for both initialize and revert. For example, we can use generics to change the `InitializeRequest` to be just a generic request. Revert would create the entire config - no need for a weird bool, or a condition, or anything. Keep it simple.

config/fill_cluster_configs_test.go
```diff
@@ -344,19 +346,24 @@ func TestEnsureTempPortRangeDoesNotOverlapWithSourceClusterPorts(t *testing.T) {
...
 			}
 		})
 	}
 }
 
+// Enable exectest.NewCommand mocking.
+func TestMain(m *testing.M) {
+	os.Exit(exectest.Run(m))
+}
+
```
kalensk, line 366: Let's create a new file called like `config_suite_test.go` and then basically do the same idea as `greenplum/init_test.go` where we move the following int there `TestMain`, `PostgresGPVersion_5_29_10`, init(), etc.

kalensk (approved):
Great change! Thanks : )

cli/commands/initialize.go
jimmyyih: Sure, that makes sense.  Updated.

config/fill_cluster_configs_test.go
jimmyyih: Done.

config/fill_cluster_configs.go
jimmyyih: As discussed offline, we'll keep the `wasEarlyExit` variable for now and another PR will come in to do as you mentioned in your comment.

PR #786, opened 2023-03-30
Remove unused substeps from protobuf

kyeap-vmware:
Remove unused substeps from protobuf
Add missing mockgen version in Makefile
Sync go.mod and go.sum with codebase using `go mod tidy`

pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:remove-unused-substeps


jimmyyih (approved):

Makefile
```diff
@@ -25,7 +25,7 @@ depend-dev: export GOFLAGS := -mod=readonly # do not update dependencies during
 depend-dev:
 	mkdir -p $(GOBIN)
 	go install github.com/golang/protobuf/protoc-gen-go@v1.3.2
-	go install github.com/golang/mock/mockgen
+	go install github.com/golang/mock/mockgen@v1.6.0
```
kalensk, line 28: Why are we pinning the version of mockgen if 1.6.0 is the latest version? Meaning if latest works, then no need to pin the version. By not pinning we roll with the latest which is what we want unless there is a specific issue. If there is an issue then we should add a comment on why we need to pin to a specific version (which is missing from the protoc-gen-go...oops).

jimmyyih: Yeah, I was wondering about this too when I was reviewing but I ran the commands just to verify and got this:
```
$ go install github.com/golang/mock/mockgen
../../go/pkg/mod/github.com/golang/mock@v1.6.0/mockgen/mockgen.go:41:2: missing go.sum entry for module providing package golang.org/x/mod/modfile (imported by github.com/golang/mock/mockgen); to add:
	go get github.com/golang/mock/mockgen@v1.6.0
../../go/pkg/mod/github.com/golang/mock@v1.6.0/mockgen/mockgen.go:42:2: missing go.sum entry for module providing package golang.org/x/tools/imports (imported by github.com/golang/mock/mockgen); to add:
	go get github.com/golang/mock/mockgen@v1.6.0

$ go install github.com/golang/mock/mockgen@v1.6.0
go: downloading golang.org/x/mod v0.4.2
go: downloading golang.org/x/tools v0.1.1
go: downloading golang.org/x/sys v0.0.0-20210510120138-977fb7262007
```

So seems like the pin is just to fix that.

kyeap-vmware: I ran into the same mockgen error posted above. It looks it was caused by missing some dependencies in our go.mod and go.sum file. `go mod tidy` fixes the mockgen install errors. We can keep the mockgen version rolling.

kalensk (approved):

PR #787, opened 2023-03-30
Refactor config.Create and GetInitializeConfiguration

kalensk:
1) rename config.Port to config.HubPort

2) bootstrap config.Create with all known values
- This allows revert to no longer needs to call Config.GetInitializeConfiguration since the upgradeID and other known values are bootstraped in the config. Having good knonw values in the config allow the rest of revert to succeed.
- GetInitializeConfiguration no longer needs to save the same parameters that have already been bootstrapped in the config.

3) add GetInitializeConfiguration to config struct

4) refactor GetInitializeConfiguration
- Since we now bootstrap the config with all known values, there is no need to send them in the initialize request.
- Rename to AddClusters.

Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:refactorRevertFillConfig3


integration/hub_test.go
```diff
@@ -30,12 +31,6 @@ func TestHub(t *testing.T) {
 		args           []string
 		configContents string
 	}{
-		{
-			name:           "starts on default port when flag is not set and config file is empty",
-			expected:       upgrade.DefaultHubPort,
-			args:           []string{"hub", "--daemonize"},
-			configContents: `{}`,
-		},
```
kalensk, line 38: This test doesn't make much sense since this scenario won't happen given this section of the cli/commands/hub.go code:

```go
			finfo, err := os.Stat(stateDir)
			if os.IsNotExist(err) {
				return fmt.Errorf("gpupgrade state dir (%s) does not exist. Did you run gpupgrade initialize?", stateDir)
			} else if err != nil {
				return err
			} else if !finfo.IsDir() {
				return fmt.Errorf("gpupgrade state dir (%s) does not exist as a directory.", stateDir)
			}
```

Since the config.Read() refactor we now clobber rather than merge conf settings. I think this makes much more sense from an API perspective and gpupgrade life cycle. Thus, the conf defaults went away when reading and we bubble up errors right away when reading. See the change from `cli/commands/hub.go`. I think this is actually better, rather than trying to be "too clever" especially for "edge cases" we don't need to support. We can definitely revisit if the need arises.

hub/revert.go
```diff
@@ -50,12 +49,6 @@ Cannot revert and restore the source cluster. Please contact support.`)
 	// If CLI Initialize exited before the InitializeRequest was sent
 	// to the hub, we will only need to do a couple revert substeps.
 	if !hasInitializeStarted {
-		request := &idl.InitializeRequest{SourceGPHome: s.Source.GPHome, SourcePort: int32(s.Source.CoordinatorPort())}
-		s.Config, err = config.GetInitializeConfiguration(s.Port, request, true)
-		if err != nil {
-			return err
-		}
-
```
jimmyyih, line 58: This will cause a regression in early Initialize revert where the revert output will give incorrect output:
```
The source cluster is now running version 0.0.0.
source /usr/local/gpdb5/greenplum_path.sh
export MASTER_DATA_DIRECTORY=
export PGPORT=15432
```

This was why the `config.GetInitializeConfiguration()` with the `wasEarlyExit` logic was done so that the revert response would have the version and master data directory set for the CLI revert to properly output to stdout.

hub/revert.go
```diff
@@ -71,13 +64,13 @@ Cannot revert and restore the source cluster. Please contact support.`)
...
 	st.RunConditionally(idl.Substep_delete_target_cluster_datadirs,
-		s.Intermediate != nil && s.Intermediate.Primaries != nil && s.Intermediate.CoordinatorDataDir() != "",
+		s.Intermediate.Primaries != nil && s.Intermediate.CoordinatorDataDir() != "",
 		func(streams step.OutStreams) error {
 			return DeleteCoordinatorAndPrimaryDataDirectories(streams, s.agentConns, s.Intermediate)
 		})
 
 	st.RunConditionally(idl.Substep_delete_tablespaces,
-		s.Intermediate != nil && s.Intermediate.Primaries != nil && s.Intermediate.CoordinatorDataDir() != "",
+		s.Intermediate.Primaries != nil && s.Intermediate.CoordinatorDataDir() != "",
```
jimmyyih, line 73: This will cause a regression in early Initialize revert where the hub will panic because `s.Intermediate.Primaries` is not accessible due to `s.Intermediate` being `nil`.

hub/revert.go
kalensk: I have a followup PR in the works that addresses this by merging  config.AddClusters into config.Create. (Its much too big of a change for this PR.) This will enable the config to be essentially fully initialized and fully usable the minute it is created. This allows us to not need any nil check conditionals in revert! It also allows all downstream users of config to be confident in using it which is critical in preventing nil checks and access bugs. Its a very similar concept to [RAII](https://en.wikipedia.org/wiki/Resource_acquisition_is_initialization).

As I dived deeper into this code, I think its good to move the logic of AddClusters to the CLI since it doesn't actually do much heavy lifting in that it only makes a single database call to the source cluster to get gp_segment_configuration, and does a postgres --version. To me the benefits outweigh any potential downsides.

hub/revert.go
kalensk: Added back.

hub/server_test.go
```diff
@@ -148,10 +148,10 @@ func mustListen(t *testing.T) (int, func()) {
...
 }
 
-// TODO: These tests would be faster and more testable if we pass in a gRPC
-//  dialer to AgentConns similar to how we test RestartAgents. Thus, we would be
-//  able to use bufconn.Listen when creating a gRPC dialer. But since there
-//  are many callers to AgentConns that is not an easy change.
+//	 TODO: These tests would be faster and more testable if we pass in a gRPC
+//		dialer to AgentConns similar to how we test RestartAgents. Thus, we would be
+//		able to use bufconn.Listen when creating a gRPC dialer. But since there
+//		are many callers to AgentConns that is not an easy change.
```
jimmyyih, line 154: This diff looks a bit weird to shift the comment over with tabs.  Was this intentional?  If so, I'm fine with it.

jimmyyih (approved):

hub/server_test.go
kalensk: Yeah kinda.

For the past year my IDE auto formats those two TODO's in a super weird way. Im not quite sure how to fix it and editing it by hand to undo the changes got super annoying. Finally I got fed up and this felt like the current formatting in this diff is a good "compromise" that satisfies my IDE and doesn't look "too" weird when looking at the code overall. Sigh.

PR #789, opened 2023-04-02
record substep status for substeps run on the CLI

kalensk:
We record the substep status in sbusteps.json in the hub. We have yet to do the similar in the CLI. This adds support for recording the substep status for substeps run on the CLI.

Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:recordCLIsubsteps


cli/commands/initialize.go
```diff
@@ -195,7 +195,7 @@ func initialize() *cobra.Command {
 				return nil
 			})
 
-			st.RunCLISubstep(idl.Substep_saving_source_cluster_config, func(streams step.OutStreams) error {
+			st.RunCLISubstep(idl.Substep_initialize_saving_source_cluster_config, func(streams step.OutStreams) error {
```
kalensk, line 198: This is needed since we currently also have the same substep name on the hub (soon to be removed). These names collide and overwrite one another causing all kinds of issues. Use different substep names until https://github.com/greenplum-db/gpupgrade/pull/790 is merged which removes this.

jimmyyih (approved):

PR #790, opened 2023-04-02
merge AddClusters into config.Create

kalensk:
Enable the config to be fully initialized and usable the minute it is created. A very similar concept to RAII.It also allows all downstream users of config to be confident in using it which is critical in preventing nil checks and access bugs.

This allows us to also set the fields LogArchiveDir and backupDir when creating the config.

Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:refactorConfigCreate


cli/commanders/data_migration_generate.go
```diff
@@ -122,6 +121,18 @@ Logs:
...
+// XXX: for internal testing only
+func SetBootstrapConnectionFunction(connectionFunc func(destination idl.ClusterDestination, gphome string, port int) (*sql.DB, error)) {
+	bootstrapConnectionFunc = connectionFunc
+}
+
+// XXX: for internal testing only
+func ResetBootstrapConnectionFunction() {
+	bootstrapConnectionFunc = connection.Bootstrap
+}
+
```
jimmyyih, line 135: Seems like this should go in the connection package so that any unit test that runs into `connection.Bootstrap()` can use it but I'm fine with this as it can be changed later if needed.

config/config.go
```diff
@@ -94,12 +94,18 @@ func Create(db *sql.DB, hubPort int, agentPort int, sourceGPHome string, targetG
...
+		return Config{}, err
+	}
+
 	config := Config{}
 	config.HubPort = hubPort
 	config.AgentPort = agentPort
 	config.Mode = mode
 	config.UseHbaHostnames = useHbaHostnames
 	config.UpgradeID = upgrade.NewID()
+	config.LogArchiveDir = filepath.Join(filepath.Dir(logDir), upgrade.GetArchiveDirectoryName(config.UpgradeID, time.Now()))
```
jimmyyih, line 120: Was mentioned in the other PR but probably good to reference here too:
`LogArchiveDir` probably shouldn't be recorded in the Config file since it can cause multiple Revert calls (unlikely but possible) to fail and the `time.Now()` affixed to the end is the Initialize timestamp instead of Finalize or Revert.  IMO, the logic should probably be similar to calling a simpler version of `GetLogArchiveDir()` on each Finalize/Revert call so that things are archived properly and idempotently (log dir and upgrade id will always be the same; just the timestamp needs to be more unique for each call).  Can discuss this later for a separate PR when/if needed.

jimmyyih (approved):

config/config.go
kalensk: Yeah, separate PR with more discussion is needed.

Since we support crash-only idempotentence, ideally each substep should be idempotent including archiving the log directory.

One important consideration is that we use the upgradeID is various other places such as intermediate data directory names. It's important (unless we think of a whole different approach) to correlate that upgradeID in the archive log directory with those other places. This can be very useful when correlating different gpupgrade runs.

cli/commanders/data_migration_generate.go
kalensk: Possibly! Will leave that for another PR since the existing standard is how it is here. Not saying we can't consider changing it, just that if we do we should do it for all SetFunctions and ResetFunctions.

PR #795, opened 2023-04-05
refactor ArchiveLogDirectories

kalensk:
The current implementation does not work if a substep after archive log directory fails and revert or finalize are re-run then the logs won’t be properly archived.

Derive the archive log directory name each time it is called. This is useful for substep idempotence if revert or finalize need to be re-run then the archive log directory will succeed with the same upgradeID, but different timestamp.

Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:refactorLogArchiveDir


hub/archive_log_directory.go
```diff
@@ -28,17 +28,23 @@ func ArchiveLogDirectories(logArchiveDir string, agentConns []*idl.Connection, t
...
 		return err
 	}
 
 	return ExecuteRPC(agentConns, request)
 }
+
+// GetLogArchiveDir returns the name of the file to be used to store logs
+// from this run of gpupgrade during a revert.
+func GetLogArchiveDir(logDir string, upgradeID upgrade.ID, t time.Time) string {
+	archiveName := fmt.Sprintf("gpupgrade-%s-%s", upgradeID.String(), t.Format("20060102T1504"))
```
jimmyyih, line 48: Probably best to add seconds in the timestamp too just in case.  (e.g. `"20060102T150405"`)

config/config.go
```diff
@@ -106,18 +106,12 @@ func Create(db *sql.DB, hubPort int, agentPort int, sourceGPHome string, targetG
...
-		return Config{}, err
-	}
-
 	config := Config{}
 	config.HubPort = hubPort
 	config.AgentPort = agentPort
 	config.Mode = mode
 	config.UseHbaHostnames = useHbaHostnames
 	config.UpgradeID = upgrade.NewID()
-	config.LogArchiveDir = filepath.Join(filepath.Dir(logDir), upgrade.GetArchiveDirectoryName(config.UpgradeID, time.Now()))
```
jimmyyih, line 120: Shouldn't we remove `LogArchiveDir` from the Config struct as well now that we will not be storing it?

hub/finalize.go
```diff
@@ -85,7 +87,12 @@ func (s *Server) Finalize(req *idl.FinalizeRequest, stream idl.CliToHub_Finalize
 	})
 
 	st.Run(idl.Substep_archive_log_directories, func(_ step.OutStreams) error {
-		return ArchiveLogDirectories(s.LogArchiveDir, s.agentConns, s.Config.Target.CoordinatorHostname())
+		logDir, err := utils.GetLogDir()
+		if err != nil {
+			return err
+		}
+
+		return ArchiveLogDirectories(logDir, GetLogArchiveDir(logDir, s.UpgradeID, time.Now()), s.agentConns, s.Config.Target.CoordinatorHostname())
```
jimmyyih, line 95: For both hub Finalize and Revert, it seems like we need to store the output of `GetLogArchiveDir()` (or at least `time.Now()`) for the FinalizeResponse and RevertResponse.  Currently, the responses' `LogArchiveDirectory` will be blank and the UX output at the end of Finalize and Revert will not show an archive directory path.

config/config.go
kalensk: Ah! I knew I forgot something. Thanks.

hub/finalize.go
kalensk: Yup, fixed. 

Side note, I wonder if it makes sense to set ArchiveLogDirectory substep to AlwaysRun in Finalize as well...

hub/archive_log_directory.go
kalensk: Yeah I was a bit on the fence about that. Added.

hub/finalize.go
jimmyyih: > Side note, I wonder if it makes sense to set ArchiveLogDirectory substep to AlwaysRun in Finalize as well...

Yeah, I think we should.  A Finalize failure after archiving the log directory will later cause a rerun of Finalize to create a new `~/gpAdminLogs/gpupgrade/` directory (that won't get archived) and not output an archive directory location in the Finalize success output.  I think it's safe to make that change (whether you wanna do it in this PR or open another; should be fine).

jimmyyih (approved):

hub/finalize.go
kalensk: I'll open another PR just to keep things a bit separate and clear and not hold up anything.

PR #796, opened 2023-04-06
Rename userCanceled to userQuit and Add Quit Status

kalensk:
1) Rename userCanceled to userQuit

2) Add Quit status and don't show the default next action text when user quits.

3) Minor data migration UI tweaks to reduce unneeded text output to make it easier to see what script is being generated or applied.

Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:userQuit


jimmyyih (approved):
Looks good. 👍

PR #801, opened 2023-04-15
Preserve the 5X address when upgrading

kalensk:
That is, populate the gpinitsystem config file with both hostname and address when creating the target cluster.

Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:initTargetClusterWithAddress


kyeap-vmware (approved):

jimmyyih (approved):
General example of address would be a NIC name (e.g. hostname sdw1 and address sdw1-1) instead of IPs but it's fine.

kalensk:
> General example of address would be a NIC name (e.g. hostname sdw1 and address sdw1-1) instead of IPs but it's fine.

Ah. Updated. `Address` is definitely not the name I would have chosen ;)

PR #802, opened 2023-04-20
Topological sort dependant views using deprecated tsquery

kyeap-vmware:
To drop views using deprecated types in the correct order, they
 must be sorted based on their dependencies. The previous sort
 implementation was coded under the assumption that a view could only
 depend on 1 other relation. This resulted in a few unaccounted edge
 cases like the ones listed below.
    
  1. A view depending on 1 view and 1 table
  2. A view depending on 2+ tables
  3. A view depending on 2+ views
    
 Trying to fix these edge cases for the old custom sorting algorithm
 uncovered more issues and the code logic quickly became unreadable and
 very difficult to follow. Redoing the sorting using a standard
 topological sort is the right thing to do as it would guarantee the
 correct view order and would be much easier to maintain moving forward.
 
 pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:topological-sort-depricated-views

### DIFF
--- a/data-migration-scripts/5-to-6-seed-scripts/create_find_view_dep_function.sql
+++ b/data-migration-scripts/5-to-6-seed-scripts/create_find_view_dep_function.sql
@@ -10,23 +10,24 @@
 SET client_min_messages TO WARNING;
 
 CREATE OR REPLACE FUNCTION  __gpupgrade_tmp_generator.find_view_dependencies()
-RETURNS VOID AS
+RETURNS void AS
 $$
 import plpy
 
-
-# First find views that do not depend on other views (and directly on the table)
-
-leaf_view = plpy.execute("""
+# Deprecated views are views whose tables contain a deprecated object.
+# The sql joins on pg_rewrite to get view names because the pg_depend entries for
+# views are against their internal definitions in pg_rewrite rather than their
+# relational definition in pg_class.
+deprecated_views_result = plpy.execute("""
 SELECT
-    schema,
-    view,
-    owner
+    view_schema,
+    view_name,
+    view_owner
 FROM (
     SELECT DISTINCT
-        nv.nspname AS schema,
-        v.relname AS view,
-        pg_catalog.pg_get_userbyid(v.relowner) AS owner
+        nv.nspname AS view_schema,
+        v.relname AS view_name,
+        pg_catalog.pg_get_userbyid(v.relowner) AS view_owner
     FROM
         pg_depend d
         JOIN pg_rewrite r ON r.oid = d.objid
@@ -52,30 +53,26 @@ FROM (
     ) subq;
 """)
 
-checklist = {}
-view_order = 1
-for row in leaf_view:
-    checklist[(row['schema'], row['view'], row['owner'])] = view_order
-
-rows = plpy.execute("""
+# Get views that depend on other views
+dependent_views_result = plpy.execute("""
 SELECT
-    nsp1.nspname AS depender_schema,
-    depender,
-    depender_owner,
-    nsp2.nspname AS dependee_schema,
-    dependee,
-    dependee_owner
+    nsp1.nspname AS view_schema,
+    view_name,
+    view_owner,
+    nsp2.nspname AS dependent_view_schema,
+    dependent_view_name,
+    dependent_view_owner
 FROM
     pg_namespace AS nsp1,
     pg_namespace AS nsp2,
     (
         SELECT
-            c.relname depender,
-            c.relnamespace AS depender_nsp,
-            pg_catalog.pg_get_userbyid(c.relowner) as depender_owner,
-            c1.relname AS dependee,
-            c1.relnamespace AS dependee_nsp,
-            pg_catalog.pg_get_userbyid(c1.relowner) as dependee_owner
+            c.relname view_name,
+            c.relnamespace AS view_nsp,
+            pg_catalog.pg_get_userbyid(c.relowner) AS view_owner,
+            c1.relname AS dependent_view_name,
+            c1.relnamespace AS dependent_view_nsp,
+            pg_catalog.pg_get_userbyid(c1.relowner) AS dependent_view_owner
         FROM
             pg_rewrite AS rw,
             pg_depend AS d,
@@ -90,11 +87,11 @@ FROM
             AND c1.relkind = 'v'
             AND c.relname <> c1.relname
         GROUP BY
-            depender, depender_nsp, depender_owner, dependee, dependee_nsp, dependee_owner
+            view_name, view_nsp, view_owner, dependent_view_name, dependent_view_nsp, dependent_view_owner
     ) t1
 WHERE
-    t1.depender_nsp = nsp1.oid
-    AND t1.dependee_nsp = nsp2.oid
+    t1.view_nsp = nsp1.oid
+    AND t1.dependent_view_nsp = nsp2.oid
     AND nsp1.nspname NOT LIKE 'pg_temp_%'
     AND nsp1.nspname NOT LIKE 'pg_toast_temp_%'
     AND nsp1.nspname NOT IN ('pg_catalog', 'information_schema', 'gp_toolkit')
@@ -103,27 +100,89 @@ WHERE
     AND nsp2.nspname NOT IN ('pg_catalog', 'information_schema', 'gp_toolkit')
 """)
 
-view2view = {}
-for row in rows:
-    key = (row['depender_schema'], row['depender'], row['depender_owner'])
-    val = (row['dependee_schema'], row['dependee'], row['dependee_owner'])
-    view2view[key]=val
+class View:
+    def __init__(self, schema, name, owner):
+        self.schema = schema
+        self.name = name
+        self.owner = owner
+        self.visited = False
 
-while True:
-    view_order += 1
-    new_checklist = {}
-    for depender, dependee in view2view.iteritems():
-        if dependee in checklist and depender not in checklist:
-            new_checklist[depender] = view_order
-    if len(new_checklist) == 0:
-        break
-    else:
-        checklist.update(new_checklist)
+    def __eq__(self, other):
+        if isinstance(other, View):
+            return self.schema == other.schema and self.name == other.name and self.owner == other.owner
+        return False
+
+    def __hash__(self):
+        return hash((self.schema, self.name, self.owner))
+
+deprecated_views = []
+dependent_view_to_views = {}  # a dependent view is one that references one or more view
+deprecated_view_to_dependent_views = {}
+for row in deprecated_views_result:
+    deprecated_view = View(row['view_schema'], row['view_name'], row['view_owner'])
+    deprecated_views.append(deprecated_view)
+    dependent_view_to_views[deprecated_view] = []
+    deprecated_view_to_dependent_views[deprecated_view] = []
+
+# Build reversed dependency map. This is done because it will be easier to
+# discover deprecated views starting from known deprecated views.
+for row in dependent_views_result:
+    view = View(row['view_schema'], row['view_name'], row['view_owner'])
+    dependent_view = View(row['dependent_view_schema'], row['dependent_view_name'], row['dependent_view_owner'])
+    dependent_view_to_views.setdefault(view, []) # Add view to the dependency map, but no value is needed since there are no other views that depend on it.
+    dependent_view_to_views.setdefault(dependent_view, []).append(view)
+
+def create_dependency_graph(graph, node, dependency_graph):
+    if node.visited:
+        return
+
+    node.visited = True
+    for neighbor in graph[node]:
+        dependency_graph.setdefault(neighbor, []).append(node)
+        create_dependency_graph(graph, neighbor, dependency_graph)
+
+deprecated_view_to_dependencies = {}
+for view in deprecated_views:
+    create_dependency_graph(dependent_view_to_views, view, deprecated_view_to_dependencies)
+
+# Add deprecated views and their dependencies to deprecated_view_to_dependent_views
+for deprecated_view, dependencies in deprecated_view_to_dependencies.items():
+    deprecated_view_to_dependent_views.setdefault(deprecated_view, []).extend(dependencies)
+
+def sort_nodes(graph, node):
+    if node.visited:
+        return []
+
+    node.visited = True
+    undiscovered_neighbors = []
+    for neighbor in graph[node]:
+        undiscovered_neighbors += sort_nodes(graph, neighbor)
+
+    undiscovered_neighbors.append(node)
+    return undiscovered_neighbors
+
+def topological_sort(graph):
+    # reset graph
+    for node, nodes in graph.items():
+        node.visited = False
+        for neighbor in nodes:
+            neighbor.visited = False
+
+    sorted_nodes = []
+    for node in graph:
+        sorted_nodes += sort_nodes(graph, node)
+
+    sorted_nodes.reverse()
+    return sorted_nodes
+
+# Sort deprecated views such that DROP VIEW statements will be generated in the
+# correct order since some views dependent on others.
+sorted_deprecated_views = topological_sort(deprecated_view_to_dependent_views)
 
 plpy.execute("DROP TABLE IF EXISTS  __gpupgrade_tmp_generator.__temp_views_list")
 plpy.execute("CREATE TABLE  __gpupgrade_tmp_generator.__temp_views_list (full_view_name TEXT, view_owner TEXT, view_order INTEGER)")
-for v, view_order in checklist.items():
-    sql = "INSERT INTO  __gpupgrade_tmp_generator.__temp_views_list VALUES('{0}.{1}', '{2}', {3})".format(v[0],v[1],v[2],view_order)
+for index, view in enumerate(sorted_deprecated_views):
+    sql = "INSERT INTO  __gpupgrade_tmp_generator.__temp_views_list VALUES('{0}.{1}', '{2}', {3})".format(view.schema, view.name, view.owner, index)
     plpy.execute(sql)
 $$ LANGUAGE plpythonu;
 
--- a/data-migration-scripts/5-to-6-seed-scripts/create_find_view_dep_function.sql
+++ b/data-migration-scripts/5-to-6-seed-scripts/create_find_view_dep_function.sql
@@ -40,7 +40,7 @@ FROM (
         AND d.classid = 'pg_rewrite'::regclass
         AND d.refclassid = 'pg_class'::regclass
         AND d.deptype = 'n'
-        AND (a.atttypid = 'pg_catalog.tsquery'::pg_catalog.regtype OR a.atttypid = 'pg_catalog.name'::pg_catalog.regtype)
+        AND a.atttypid = 'pg_catalog.tsquery'::pg_catalog.regtype
         AND c.relkind = 'r'
         AND NOT a.attisdropped
         AND nv.nspname NOT LIKE 'pg_temp_%'
--- a/data-migration-scripts/5-to-6-seed-scripts/finalize/tables_using_tsquery_type/recreate_views_on_deprecated_built_in_types.sql
+++ b/data-migration-scripts/5-to-6-seed-scripts/finalize/tables_using_tsquery_type/recreate_views_on_deprecated_built_in_types.sql
@@ -4,4 +4,4 @@
 SELECT
     $$CREATE VIEW $$ || full_view_name || $$ AS $$ || pg_catalog.pg_get_viewdef(full_view_name::regclass::oid, false) || $$;$$ || E'\n'||
     $$ALTER VIEW $$ || full_view_name || $$ OWNER TO $$ || view_owner || $$;$$
-FROM __gpupgrade_tmp_generator.__temp_views_list ORDER BY view_order;
+FROM __gpupgrade_tmp_generator.__temp_views_list ORDER BY view_order DESC;
--- a/data-migration-scripts/5-to-6-seed-scripts/initialize/tables_using_tsquery_type/gen_drop_depr_built_in_type_dependent_views.sql
+++ b/data-migration-scripts/5-to-6-seed-scripts/initialize/tables_using_tsquery_type/gen_drop_depr_built_in_type_dependent_views.sql
@@ -1,5 +1,5 @@
 -- Copyright (c) 2017-2023 VMware, Inc. or its affiliates
 -- SPDX-License-Identifier: Apache-2.0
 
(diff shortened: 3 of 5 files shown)


data-migration-scripts/5-to-6-seed-scripts/create_find_view_dep_function.sql
```diff
@@ -103,27 +100,73 @@ WHERE
...
+    if dependee in visited:
+        continue
+
+    visited.add(dependee)
+    stack.extend(reverse_dep_map[dependee])
+    for depender in reverse_dep_map[dependee]:
+        if depender not in dep_map:
+            dep_map[depender] = []
+        dep_map[depender].append(dependee)
+
```
kalensk, line 140: Why do we need `starting_nodes` if `views_of_tables_using_deprecated` already has the same thing essentialy?

data-migration-scripts/5-to-6-seed-scripts/create_find_view_dep_function.sql
```diff
@@ -103,27 +100,73 @@ WHERE
...
-        checklist.update(new_checklist)
+# Create start node list and add them to dependency maps
+reverse_dep_map = {}
+dep_map = {}
+starting_nodes = []
+for row in views_of_tables_using_deprecated:
+    node = (row['schema'], row['view'], row['owner'])
+    starting_nodes.append(node)
+    reverse_dep_map[node] = []
+    dep_map[node] = []
```
kalensk, line 111: I know the existing code already had it, but espcially with all this new code is there any possible way we can consider renaming depender and dependee to something more clear. Im having a really hard time following all this code and I think the naming of all these new objects and maps could be much more detailed/descriptive. Some initial ideas `dependent_child`, `dependent_parent`, instead of node create a class called DepreciatedVeiows, rename reverse_dep_map to dependent_child_to_depreciatedView, and dependent_parent_to_depreciatedView instead of dep_map, etc. Again those are initial ideas and maybe require a bit more thought.

Maybe we can pair on that.

data-migration-scripts/5-to-6-seed-scripts/create_find_view_dep_function.sql
kyeap-vmware: views_of_tables_using_deprecated is the raw result of our sql query. `starting_nodes` is the list of views after formatting the results of the query.

data-migration-scripts/5-to-6-seed-scripts/create_find_view_dep_function.sql
```diff
@@ -103,27 +100,73 @@ WHERE
...
+    if dependee not in reverse_dep_map:
+        reverse_dep_map[dependee] = []
+    if depender not in reverse_dep_map:
+        reverse_dep_map[depender] = []
+    reverse_dep_map[dependee].append(depender)
+
+# Discover views that depend on views using deprecated types
+# and build dependency map that contains only relevant views.
+visited = set()
+stack = starting_nodes
```
kalensk, line 128: This line is kinda weird to me in that its creating yet another variable which doesn't really seem to be needed in a sense. That and it's named too generically to easily follow. It almost seems like most of this below logic could be done when first iterating the result of the returned SQL rows and simplified....

data-migration-scripts/5-to-6-seed-scripts/create_find_view_dep_function.sql
kalensk: Maybe I am missing something, but I still don't see it. I mean yeah, I definitely get that `views_of_tables_using_deprecated` is the raw sql, but that's already an iterable, and we are basically copying it into another iterable (starting_nodes). Is that even really necessary? Is there a way to simplify things? Something's not adding up for me. I'll need to keep playing with it.

data-migration-scripts/5-to-6-seed-scripts/create_find_view_dep_function.sql
```diff
@@ -103,27 +100,73 @@ WHERE
...
-    if len(new_checklist) == 0:
-        break
-    else:
-        checklist.update(new_checklist)
+# Create start node list and add them to dependency maps
+reverse_dep_map = {}
+dep_map = {}
+starting_nodes = []
+for row in views_of_tables_using_deprecated:
+    node = (row['schema'], row['view'], row['owner'])
```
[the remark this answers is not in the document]
jimmyyih: It's possible to create a class like this:
```
class View:
    def __init__(self, schema, name, owner):
        self.schema = schema
        self.name = name
        self.owner = owner

    def __eq__(self, other):
        isView = isinstance(other, self.__class__)
        if not isView:
            return False

        if self.schema == other.schema and self.name == other.name and self.owner == other.owner:
            return True
        else:
            return False

    def __hash__(self):
        # return some calculated hash value somehow
```

It's a bit extra for Python scripting but I'm fine with it.  Here, I believe `node` refers to the map points that will be sorted and so is aptly named for map sorting logic.  If anything, it could be renamed to `view_node` to be a bit more detailed (or simply just `view`).

data-migration-scripts/5-to-6-seed-scripts/create_find_view_dep_function.sql
kalensk: Since python is not strongly typed I really appreciate when maps are named something like `key_to_value`, so something like `view_dependency_to_depreciated_view`. Its verbose, but makes it so clear when trying to follow the code. Or whatever other name makes it easy to follow.

data-migration-scripts/5-to-6-seed-scripts/create_find_view_dep_function.sql
kyeap-vmware: The results needs to be more than just iterable. We are putting the data in a common format.
query1 returns a list of views on tables that use tsquery (1 view per entry)
query2 returns a list of view dependencies (2 views per entry). 

The views we get from both queries need to be in the same format in order to properly compare them against each other. Later in the code we are popping from our starting_nodes stack and looking them up in our graph.

data-migration-scripts/5-to-6-seed-scripts/create_find_view_dep_function.sql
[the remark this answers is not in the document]
kyeap-vmware: renamed the node in this area to reprecated_view

data-migration-scripts/5-to-6-seed-scripts/create_find_view_dep_function.sql
kyeap-vmware: Renamed maps
reverse_dep_map => view_dependency_to_views
dep_map => deprecated_view_to_view_dependencies

In the reverse map it is not yet known which views are deprecated. In the input map that gets topologically sorted it is known all views inside are deprecated.

data-migration-scripts/5-to-6-seed-scripts/create_find_view_dep_function.sql
kyeap-vmware: It seems to be all needed. The logic below is a while loop implementation of DFS to discover deprecated views. While that is happening we simultaneously build our deprecated view dependency map

data-migration-scripts/5-to-6-seed-scripts/create_find_view_dep_function.sql
```diff
@@ -103,27 +100,72 @@ WHERE
...
+# Build reversed dependency map. This is done because it will be easier to
+# discover deprecated views starting from known deprecated views.
+for row in views_dependent_on_views:
+    view = (row['view_schema'], row['view_name'], row['view_owner'])
+    view_dependency = (row['view_dependency_schema'], row['view_dependency_name'], row['view_dependency_owner'])
+    if view_dependency not in view_dependency_to_views:
+        view_dependency_to_views[view_dependency] = []
+    if view not in view_dependency_to_views:
+        view_dependency_to_views[view] = []
+    view_dependency_to_views[view_dependency].append(view)
```
kalensk, line 122: I am pretty sure (double check) that this block can be simplified to something like the following using setdefault()

```
for row in views_dependent_on_views:
    view = (row['view_schema'], row['view_name'], row['view_owner'])
    view_dependency = (row['view_dependency_schema'], row['view_dependency_name'], row['view_dependency_owner'])    
    view_dependency_to_views.setdeafult(view_dependency, []).append(view)
```

Another potential possibility is to use `defaultdict(list)`
```
views_dependent_on_views = collections.defaultdict(list)

for row in views_dependent_on_views:
    views_dependent_on_views[view_dependency].append(view)
```

data-migration-scripts/5-to-6-seed-scripts/create_find_view_dep_function.sql
```diff
@@ -103,27 +100,72 @@ WHERE
...
+    view_dependency = stack.pop()
+    if view_dependency in visited:
+        continue
+
+    visited.add(view_dependency)
+    stack.extend(view_dependency_to_views[view_dependency])
+    for view in view_dependency_to_views[view_dependency]:
+        if view not in deprecated_view_to_view_dependencies:
+            deprecated_view_to_view_dependencies[view] = []
+        deprecated_view_to_view_dependencies[view].append(view_dependency)
```
kalensk, line 138: Same concept as the above comment in that I think this block can be simplified with setdefault():

```
for view in view_dependency_to_views[view_dependency]:
    deprecated_view_to_view_dependencies.setdefault(view, []).append(view_dependency)
```

data-migration-scripts/5-to-6-seed-scripts/create_find_view_dep_function.sql
```diff
@@ -103,27 +100,72 @@ WHERE
...
+
+# Build reversed dependency map. This is done because it will be easier to
+# discover deprecated views starting from known deprecated views.
+for row in views_dependent_on_views:
+    view = (row['view_schema'], row['view_name'], row['view_owner'])
+    view_dependency = (row['view_dependency_schema'], row['view_dependency_name'], row['view_dependency_owner'])
+    if view_dependency not in view_dependency_to_views:
+        view_dependency_to_views[view_dependency] = []
+    if view not in view_dependency_to_views:
+        view_dependency_to_views[view] = []
```
kalensk, line 121: Is this needed? I don't see `view_dependency_to_views[view]` being used anywhere?

data-migration-scripts/5-to-6-seed-scripts/create_find_view_dep_function.sql
```diff
@@ -103,27 +100,72 @@ WHERE
...
-        if dependee in checklist and depender not in checklist:
-            new_checklist[depender] = view_order
-    if len(new_checklist) == 0:
-        break
-    else:
-        checklist.update(new_checklist)
+# Create formatted list of known deprecated views and add to dependency maps
+view_dependency_to_views = {}
+deprecated_view_to_view_dependencies = {}
+known_deprecated_views = []
```
kalensk, line 106: Can we rename `known_deprecated_views` to just `deprecated_views`?

data-migration-scripts/5-to-6-seed-scripts/create_find_view_dep_function.sql
```diff
@@ -103,27 +100,72 @@ WHERE
...
+        stack.append(node)
+
+    for node in graph:
+        if node not in visited:
+            DFS(node)
+
+    while stack:
+        result.append(stack.pop())
+
+    return result
```
kalensk, line 161: Dumb question perhaps, but is this just adding the reverse stack to a new variable called result? If so, why not just do
`return stack.reverse()` and rename stack to result or similar? Keeps things simple and removes an intermediate variable.

data-migration-scripts/5-to-6-seed-scripts/create_find_view_dep_function.sql
```diff
@@ -103,27 +100,72 @@ WHERE
...
-        checklist.update(new_checklist)
+# Create formatted list of known deprecated views and add to dependency maps
+view_dependency_to_views = {}
+deprecated_view_to_view_dependencies = {}
+known_deprecated_views = []
+for row in views_of_tables_using_deprecated:
+    deprecated_view = (row['view_schema'], row['view_name'], row['view_owner'])
+    known_deprecated_views.append(deprecated_view)
+    view_dependency_to_views[deprecated_view] = []
+    deprecated_view_to_view_dependencies[deprecated_view] = []
```
kalensk, line 111: Can maybe remove these lines if we use `defaultdict(list)` perhaps? (Please, double check though as my python is a bit rusty.) That is, initialize the maps with the following and your good:
```
views_dependent_on_views = collections.defaultdict(list)
deprecated_view_to_view_dependencies = collections.defaultdict(list)
```

data-migration-scripts/5-to-6-seed-scripts/create_find_view_dep_function.sql
```diff
@@ -103,27 +100,72 @@ WHERE
...
-    for depender, dependee in view2view.iteritems():
-        if dependee in checklist and depender not in checklist:
-            new_checklist[depender] = view_order
-    if len(new_checklist) == 0:
-        break
-    else:
-        checklist.update(new_checklist)
+# Create formatted list of known deprecated views and add to dependency maps
+view_dependency_to_views = {}
+deprecated_view_to_view_dependencies = {}
```
kalensk, line 105: Can we rename this to `deprecated_view_to_dependencies`? No need for the second view in the name as its implied. Makes it quite a bit easier to follow the below block found further down in the code:

```
    visited.add(view_dependency)
    stack.extend(view_dependency_to_views[view_dependency])
    for view in view_dependency_to_views[view_dependency]:
        if view not in deprecated_view_to_view_dependencies:
            deprecated_view_to_view_dependencies[view] = []
        deprecated_view_to_view_dependencies[view].append(view_dependency)
```

data-migration-scripts/5-to-6-seed-scripts/create_find_view_dep_function.sql
```diff
@@ -103,27 +100,72 @@ WHERE
...
+    if view_dependency not in view_dependency_to_views:
+        view_dependency_to_views[view_dependency] = []
+    if view not in view_dependency_to_views:
+        view_dependency_to_views[view] = []
+    view_dependency_to_views[view_dependency].append(view)
+
+# Discover views that depend on deprecated views using DFS and build dependency
+# map that contains only deprecated views.
+visited = set()
+stack = known_deprecated_views
```
kalensk, line 127: Ohhh.... `stack` should be called `deprecated_views` and keep `known_deprecated_views` as is.

data-migration-scripts/5-to-6-seed-scripts/create_find_view_dep_function.sql
```diff
@@ -10,23 +10,24 @@
...
 
-
-# First find views that do not depend on other views (and directly on the table)
-
-leaf_view = plpy.execute("""
+# Get views that depend on at least one table that use deprecated tsquery type.
+# The sql joins on pg_rewrite to get view names because the pg_depend entries for
+# views are against their internal definitions in pg_rewrite rather than their
+# relational definition in pg_class.
+views_of_tables_using_deprecated = plpy.execute("""
```
kalensk, line 21: nit: Can we rename to `views_of_tables_using_deprecated_objects`? I feel like its missing a noun at the end.

data-migration-scripts/5-to-6-seed-scripts/create_find_view_dep_function.sql
```diff
@@ -103,27 +100,79 @@ WHERE
...
+        self.owner = owner
+        self.visited = False
+
+    def __eq__(self, other):
+        if isinstance(other, View):
+            return self.schema == other.schema and self.name == other.name and self.owner == other.owner
+        return False
+
+    def __hash__(self):
+        return hash((self.schema, self.name, self.owner))
```
kalensk, line 116: Ugh. I know I was the one that recommended this originally, but after seeing it I don't think its getting us much unfortunately. We can probably remove it, unless there is benefit.

data-migration-scripts/5-to-6-seed-scripts/create_find_view_dep_function.sql
```diff
@@ -52,30 +53,26 @@ FROM (
...
-    depender_owner,
-    nsp2.nspname AS dependee_schema,
-    dependee,
-    dependee_owner
+    nsp1.nspname AS view_schema,
+    view_name,
+    view_owner,
+    nsp2.nspname AS view_dependency_schema,
+    view_dependency_name,
+    view_dependency_owner
```
kalensk, line 64: Food for thought: Instead of `view_dependency` throughout the code what about `dependent_view`?

data-migration-scripts/5-to-6-seed-scripts/create_find_view_dep_function.sql
```diff
@@ -103,27 +100,79 @@ WHERE
...
+# map that contains only deprecated views.
+while known_deprecated_views:
+    view_dependency = known_deprecated_views.pop()
+    if view_dependency.visited:
+        continue
+
+    view_dependency.visited = True
+    known_deprecated_views.extend(view_dependency_to_views[view_dependency])
+    for view in view_dependency_to_views[view_dependency]:
+        deprecated_view_to_dependencies.setdefault(view, []).append(view_dependency)
```
kalensk, line 146: I still don't understand this block and hope that it could be greatly simplified and more clear. Some possible ideas perhaps.

1\. Instead of a while loop couldn't this block simply be:
```py
for depreciated_view in known_deprecated_views:
   ...
```  

This eliminates the need for both mutating `known_deprecated_views` with the `.pop()` as well as needing a `.visited` field. Seems much easier to follow.

2\. For example, why couldn't this line:
`known_deprecated_views.extend(view_dependency_to_views[view_dependency])`
just be moved to line 134 as something like: `known_deprecated_views.append(view)`. That seems much more simple to me.

3\. This section:
```py
for view in view_dependency_to_views[view_dependency]:
    deprecated_view_to_dependencies.setdefault(view, []).append(view_dependency)
```  

Would be much clearer as something like:
```py
for depreciated_view in known_deprecated_views:
    for view in view_dependency_to_views[depreciated_view]:
        deprecated_view_to_dependencies.setdefault(view, []).append(depreciated_view)
```  

Also the name of the map `deprecated_view_to_dependencies` doesn't seem to match what the actual variables are. That is, we are adding a `view_dependency` as the key and a `depreciated_view` as the value. Thus, the map should be called something like `view_dependency_to_depreciated_view` or even better `dependent_view_to_depreciated_view`.

data-migration-scripts/5-to-6-seed-scripts/create_find_view_dep_function.sql
```diff
@@ -103,27 +100,79 @@ WHERE
...
+
+        undiscovered_neighbors.append(node)
+        return undiscovered_neighbors
+
+    visited = []
+    for node in graph:
+        visited += DFS(node)
+
+    visited.reverse()
+    return visited
```
kalensk, line 168: minor nit: I'd move this block to be at the top of the `topological_sort` function and then have the DFS function under that. Easier to follow that way.

data-migration-scripts/5-to-6-seed-scripts/create_find_view_dep_function.sql
```diff
@@ -103,27 +100,95 @@ WHERE
...
+# discover deprecated views starting from known deprecated views.
+for row in dependent_views_result:
+    view = View(row['view_schema'], row['view_name'], row['view_owner'])
+    dependent_view = View(row['dependent_view_schema'], row['dependent_view_name'], row['dependent_view_owner'])
+    dependent_view_to_views.setdefault(view, []) # Add view to the dependency map, but no value is needed since there are no other views that depend on it.
+    dependent_view_to_views.setdefault(dependent_view, []).append(view)
+
+# Discover views that depend on deprecated views using DFS and build dependency
+# map that contains only deprecated views.
+def create_graph_with_dependencies(dependency_graph, graph, node):
```
kalensk, line 138: consider renaming to `create_dependency_graph` and move return parameter as the last argument

data-migration-scripts/5-to-6-seed-scripts/create_find_view_dep_function.sql
```diff
@@ -103,27 +100,95 @@ WHERE
...
+    for neighbor in graph[node]:
+        dependency_graph.setdefault(neighbor, []).append(node)
+        create_graph_with_dependencies(dependency_graph, graph, neighbor)
+
+def find_deprecated_views(dependent_view_to_views, deprecated_views):
+    dependency_graph = {}
+    for view in deprecated_views:
+        create_graph_with_dependencies(dependency_graph, dependent_view_to_views, view)
+
+    return dependency_graph
```
kalensk, line 152: Consider removing this function and inlining things, but keeping the similar naming style. I'd have to see it, but I don't think this function is getting us much now given the names are so much clearer.

data-migration-scripts/5-to-6-seed-scripts/create_find_view_dep_function.sql
```diff
@@ -103,27 +100,95 @@ WHERE
...
+            neighbor.visited = False
+
+    sorted_nodes = []
+    for node in graph:
+        sorted_nodes += DFS(graph, node)
+
+    sorted_nodes.reverse()
+    return sorted_nodes
+
+ordered_view_list = topological_sort(deprecated_view_to_dependent_views)
```
kalensk, line 186: Consider renaming `ordered_view_list` to `sorted_depreciated_views` or similar. This makes it clear what kind of views they are.

data-migration-scripts/5-to-6-seed-scripts/create_find_view_dep_function.sql
```diff
@@ -103,27 +100,95 @@ WHERE
...
+
+    return dependency_graph
+
+deprecated_views_graph = find_deprecated_views(dependent_view_to_views, deprecated_views)
+for node, nodes in deprecated_views_graph.items():
+    deprecated_view_to_dependent_views.setdefault(node, []).extend(nodes)
+
+# Topological sort assumes that every node has an entry in the graph to
+# it's list of neighbors even if the list ends up being empty.
+def DFS(graph, node):
```
kalensk, line 160: Consider renaming DFS to `sort_nodes` and if useful can have a comment explaining that its a DFS search.

Consider removing the comment as the code shows that. If anything we might have a comment above `topological_sort` explaining _why_ we need to sort the depreciated views. Meaning we need to generate the DELETE VIEW statements in order based on their dependencies otherwise it won't work and some error occurs, etc.

kalensk (approved):
Thanks for working with me! Great work.

Hopefully, useful in learning the importance of good variable naming and things for clarity and maintainability.

PR #804, opened 2023-04-21
Add tests for tables names with special characters

kyeap-vmware:
Recently it was found that tables ending with $ would fail upgrade
because we were using blank dollar syntax tags $$ to escape special
characters in table names.

pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:special-char-tests.

7x PR that fixes this issue: https://github.com/greenplum-db/gpdb/pull/15408

### DIFF
--- a/test/acceptance/pg_upgrade/5-to-6/upgradeable_tests/source_cluster_regress/sql/special_characters.sql
+++ b/test/acceptance/pg_upgrade/5-to-6/upgradeable_tests/source_cluster_regress/sql/special_characters.sql
@@ -1,0 +1,50 @@
+-- Copyright (c) 2017-2023 VMware, Inc. or its affiliates
+-- SPDX-License-Identifier: Apache-2.0
+
+-- Test to ensure that tables with special characters can be upgraded.
+
+--------------------------------------------------------------------------------
+-- Create and setup upgradeable objects
+--------------------------------------------------------------------------------
+
+CREATE TABLE "foo$"(i int);
+CREATE TABLE "$$foo$"(i int);
+CREATE TABLE "my_table_@pple"(i int);
+CREATE TABLE "users!@#$%^&*()_+`-={}|[];':\""<>,.?/"(i int);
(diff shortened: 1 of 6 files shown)


test/acceptance/pg_upgrade/5-to-6/upgradeable_tests/source_cluster_regress/sql/special_characters.sql
```diff
@@ -0,0 +1,50 @@
...
+CREATE TABLE "users!@#$%^&*()_+`-={}|[];':\""<>,.?/"(i int);
+CREATE TABLE "Café_Latté"(i int);
+CREATE TABLE "data_2021-09-25"(i int);
+CREATE TABLE "Sales@2023"(i int);
+CREATE TABLE "table_(parenthesis)"(i int);
+CREATE TABLE "Ελληνικά_τραπέζια"(i int);
+CREATE TABLE "table_with_underscores and spaces"(i int);
+CREATE TABLE "table_with_ö_umlaut"(i int);
+CREATE TABLE "table_with_की_hindi_characters"(i int);
+CREATE TABLE "学生表"(i text);
```
kalensk, line 22: Should we have a table name that ends in two $$'s?

kyeap-vmware: It doesn't really matter how many `$` a table end in. If there are multple `$`'s the same error would occur just earlier. We should be covered.

kalensk (approved):
LGTM. 

I really wish we didn't need this test here and that upgrading the ICW database in our e2e tests would be enough....sigh.

(General Comment: I really want to keep the pg_upgrade acceptance tests "targeted" for what is _actually_ needed rather than just testing anything and everything. I'm concerned it's a slippery slope that can quickly snowball and get out of hand.)

PR #805, opened 2023-04-24
finalize: run analyze_new_cluster.sh

kalensk:
On the target cluster run analyze_new_cluster.sh to generate statistics on the upgraded cluster.

**Question:** Do we want to run this before, or after the finalize data migration scripts? Depending on what is needed these scripts do the following: 1) add partition table indexes,  2) change data type back to tsquery from text, 3) recreate primary and foreign key constraints. Where I believe adding partition table indexes can take quite a long time (several hours or more).

Currently, we are running the analyze before the data migration scripts since the code was a bit easier, but totally happy to close this PR and swap it.

Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:refactorFinalize


hub/finalize.go
```diff
@@ -86,6 +87,15 @@ func (s *Server) Finalize(req *idl.FinalizeRequest, stream idl.CliToHub_Finalize
...
 	})
 
+	st.Run(idl.Substep_analyze_target_cluster, func(streams step.OutStreams) error {
+		upgradeDir, err := utils.GetPgUpgradeDir(s.Target.Coordinator().Role, int32(s.Target.Coordinator().ContentID))
+		if err != nil {
+			return err
+		}
+
+		return s.Target.RunCmd(streams, filepath.Join(upgradeDir, "analyze_new_cluster.sh"))
+	})
```
jimmyyih, line 97: Wondering if this should be in CLI finalize instead of hub finalize and be interactive (run/skip)... but this patch should be fine for now.

jimmyyih (approved):

hub/finalize.go
kalensk: 1) Yeah, once we do a bit more team discussion we can see if adding a prompt is desired before running analyze. I "may" hold off a bit on merging this PR until that meeting.

2) Just one note that if its in the CLI we can't call `s.Target.RunCmd()` since there is no "target" cluster object at that point. (Although maybe there is a way of doing that...we just don't have precinct for reading the Config object in the CLI as that tends to be business logic reserved for the Hub.) 

If we move this to the CLI we would essentially need to write a helper function of sorts that does what RunCmd does with additional code. I wanted to leverage the cluster cmd functions with `s.Target.RunCmd()` and hence why its in the hub rather than CLI. Relatively easy to move it to the CLI, just different implementation is needed.

PR #806, opened 2023-04-25
Add non-upgradable test for tables distributed on duplicate columns

kyeap-vmware:
Because of a GPDB5 bug in it's parser, it is possible to create tables distributed on duplicate columns. Ensure pg_upgrade does not proceed with upgrade and generates a txt file with the problem tables. Since this is patched on GPDB6 allowing pg_upgrade to continue would result in pg_restore failure during `gpupgrade execute`.

### DIFF
--- a/test/acceptance/pg_upgrade/5-to-6/non_upgradeable_tests/expected/distribution_on_duplicated_columns.out
+++ b/test/acceptance/pg_upgrade/5-to-6/non_upgradeable_tests/expected/distribution_on_duplicated_columns.out
@@ -1,0 +1,37 @@
+-- Copyright (c) 2017-2023 VMware, Inc. or its affiliates
+-- SPDX-License-Identifier: Apache-2.0
+
+-- There is a bug in the GPDB5 parser that allows distribution on duplicated
+-- columns. This has been patched in GPDB6+ so users need to fix any affected
+-- table's distribution before upgrading. A check has been put in GPDB6 to
+-- ensure upgrade will not continue if there are tables distributed on
+-- duplicate columns.
+
+--------------------------------------------------------------------------------
+-- Create and setup non-upgradeable objects
+--------------------------------------------------------------------------------
+CREATE TABLE distributed_on_duplicated_columns1 (a int, b int) DISTRIBUTED BY (a, a, b);
+CREATE
+CREATE TABLE distributed_on_duplicated_columns2 (a int, b int, c int) DISTRIBUTED BY (a, a, b, b, a, b, c);
+CREATE
+
+--------------------------------------------------------------------------------
+-- Assert that pg_upgrade --check correctly detects the non-upgradeable objects
+--------------------------------------------------------------------------------
+!\retcode gpupgrade initialize --source-gphome="${GPHOME_SOURCE}" --target-gphome=${GPHOME_TARGET} --source-master-port=${PGPORT} --disk-free-ratio 0 --non-interactive;
+-- start_ignore
+-- end_ignore
+(exited with code 1)
+! cat ~/gpAdminLogs/gpupgrade/pg_upgrade/p-1/duplicate_column_distribution.txt;
+Database: isolation2test
+  public.distributed_on_duplicated_columns1: 1
+  public.distributed_on_duplicated_columns2: 1, 2
+
+
+--------------------------------------------------------------------------------
+-- Workaround to unblock upgrade
+--------------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


test/acceptance/pg_upgrade/5-to-6/non_upgradeable_tests/sql/distribution_on_duplicated_columns.sql
```diff
@@ -0,0 +1,26 @@
+-- Copyright (c) 2017-2023 VMware, Inc. or its affiliates
+-- SPDX-License-Identifier: Apache-2.0
+
+-- There is a bug in the GPDB5 parser that allows distribution on duplicated
+-- columns. This has been patched in GPDB6+ so users need to fix any affected
+-- table's distribution before upgrading. A check has been put in GPDB6 to
+-- ensure upgrade will not continue if there are tables distributed on
+-- duplicate columns.
```
kalensk, line 8: Thanks for this comment! Great explanation and bread crumb for later when we might need to figure out why this is needed and what not.

test/acceptance/pg_upgrade/5-to-6/non_upgradeable_tests/sql/distribution_on_duplicated_columns.sql
```diff
@@ -0,0 +1,26 @@
...
+
+--------------------------------------------------------------------------------
+-- Create and setup non-upgradeable objects
+--------------------------------------------------------------------------------
+CREATE TABLE distributed_on_duplicated_columns1 (a int, b int) DISTRIBUTED BY (a, a, b);
+CREATE TABLE distributed_on_duplicated_columns2 (a int, b int, c int) DISTRIBUTED BY (a, a, b, b, a, b, c);
```
kalensk, line 14: Maybe a naive question, but what happens when the last value is repeated like `DISTRIBUTED BY (a, a, b, b)` or `DISTRIBUTED BY (a, a, b, b, a, b, c, c);`? Do we need a table for that?

test/acceptance/pg_upgrade/5-to-6/non_upgradeable_tests/expected/distribution_on_duplicated_columns.out
```diff
@@ -0,0 +1,39 @@
...
+!\retcode gpupgrade initialize --source-gphome="${GPHOME_SOURCE}" --target-gphome=${GPHOME_TARGET} --source-master-port=${PGPORT} --disk-free-ratio 0 --non-interactive;
+-- start_ignore
+-- end_ignore
+(exited with code 1)
+! cat ~/gpAdminLogs/gpupgrade/pg_upgrade/p-1/duplicate_column_distribution.txt;
+Database: isolation2test
+  tablename : duplicated distributed column
+  public.distributed_on_duplicated_columns1 : 1
+  public.distributed_on_duplicated_columns2 : 1
+  public.distributed_on_duplicated_columns2 : 2
```
kalensk, line 30: Some thoughts on the output formatting in the associated [GPDB PR](https://github.com/greenplum-db/gpdb/pull/15421/files#r1178540350).

test/acceptance/pg_upgrade/5-to-6/non_upgradeable_tests/sql/distribution_on_duplicated_columns.sql
kyeap-vmware: My understanding of the 5X parsing logic is that it is just taking the last column listed and checking against all the previous ones as if they were a set.

in test cases
`DISTRIBUTED BY (a, a, b);`
parser checks if b is in (a, a). b is not in the set so table is created

`DISTRIBUTED BY (a, a, b, b, a, b, c);`
parser checks if c is in (a, a, b, b, a, b). c is not in this set so table is created.

`DISTRIBUTED BY (a, a, b, b)`
parser checks b is in (a, a, b). b is in this set so this fail with duplicate column error

`DISTRIBUTED BY (a, a, b, b, a, b, c, c);`
parser checks c is in (a, a, b, b, a, b, c). c is in this set so this fails with duplicate column error

In 6X now the logic looks recursive and checks every column against all previously seen . 
`DISTRIBUTED BY (a, b, c, e);`
a checks against ()
b checks against (a)
c checks against (a, b)
e checks against (a, b, c)

kalensk: Thanks! 

(Im pretty surprised that 5X logic was shipped since its ummm so wrong.)

kalensk (approved):

PR #810, opened 2023-05-04
alwaysRun certain substeps

kalensk:
When recently debugging a live customer issue when upgrading the coordinator failed resulted in needing to always run certain substeps rather than skipping them. The source cluster was started for inspection and then gpupgrade execute was re-run. However, stopping the source cluster substep was skipped since it was previously run causing pg_upgrade to fail since it expects the cluster to be stopped. Thus, always run some substeps which are needed if a failure occurs downstream.

Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:alwaysRun

### DIFF
--- a/hub/finalize.go
+++ b/hub/finalize.go
@@ -46,7 +46,7 @@ func (s *Server) Finalize(req *idl.FinalizeRequest, stream idl.CliToHub_Finalize
 		return nil
 	})
 
-	st.Run(idl.Substep_check_active_connections_on_target_cluster, func(streams step.OutStreams) error {
+	st.AlwaysRun(idl.Substep_check_active_connections_on_target_cluster, func(streams step.OutStreams) error {
 		return s.Intermediate.CheckActiveConnections(streams)
 	})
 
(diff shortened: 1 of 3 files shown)


kyeap-vmware (approved):

hub/finalize.go
```diff
@@ -51,7 +51,7 @@ func (s *Server) Finalize(req *idl.FinalizeRequest, stream idl.CliToHub_Finalize
 		return s.Intermediate.WaitForClusterToBeReady()
 	})
 
-	st.Run(idl.Substep_shutdown_target_cluster, func(streams step.OutStreams) error {
+	st.AlwaysRun(idl.Substep_shutdown_target_cluster, func(streams step.OutStreams) error {
```
jimmyyih, line 54: I feel like this needs a `!HasCompleted(idl.Substep_start_target_cluster)` conditional wrapping it to make sure it doesn't unnecessarily stop the target cluster before just starting it right back up (this costs time; especially on big production systems).  But I'm fine without the additional conditional check.

jimmyyih (approved):

hub/finalize.go
kalensk: We already have that built into `.start()` and `.stop()` internally as they basically check (although not super robustly) if the cluster is already running or not and does nothing if there is no work to do.

jimmyyih: Yes, but the scenario I outlined is not covered by those internal checks.  If the target cluster is already up and the Finalize call expects to skip over from the beginning to `idl.Substep_analyze_target_cluster`, then the target cluster will be unnecessarily shut down and started back up before the `idl.Substep_analyze_target_cluster` step because shutdown is always run/checked and start is always run/checked (in that order) with these changes.  However, if we add in the suggested conditional, then we know that the target cluster does not need to be shut down for any extra Finalize calls (e.g. to do that deferred ANALYZE) and will only call the start step to start the target cluster if the user had shut it down.

PR #812, opened 2023-05-04
Refactor and Update CLI Step Framework

kalensk:
1\) Refactor to more closely match the hub step framework. Ideally, we could consider consolidating the CLI and Hub step frameworks or organizing things better between them.

2\) Most importantly skip completed substeps.

By not skipping substeps like creating the config can result in a serious bug where execute fails upgrading primaries. Basically, after the target cluster is created and initialize is re-run a new config file will be generated with a new upgradeID. Later on when rsyc'ing the upgrading coordinator to the primaries to bootstrap their upgrade will simply create a new directory with the coordinator containing no postgresql.conf as that is ignored when copying resulting in pg_upgrade failing.

Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:updateCLIstepFramework


cli/commands/initialize.go
```diff
@@ -245,22 +232,23 @@ func initialize() *cobra.Command {
...
 
-			st.RunInternalSubstep(func() error {
 				if nonInteractive {
 					return nil
 				}
 
 				fmt.Println()
-				return commanders.Prompt(bufio.NewReader(os.Stdin), idl.Step_initialize)
+				return clistep.Prompt(bufio.NewReader(os.Stdin), idl.Step_initialize)
 			})
```
jimmyyih, line 252: This change looks a bit weird joining the two steps... but it seems fine logically for now (`idl.Substep_execute_initialize_data_migration_scripts` seems to be forced to always be the last step before running the hub step now since that's where the "move forward" prompt is stored now).  Although the inner `nonInteractive` check will never trigger so it can be removed.

jimmyyih (approved):

PR #816, opened 2023-05-18
add --pg-upgrade-skip-checks

kalensk:
Since pg_upgrade checks can take a long time for large clusters add --pg-upgrade-skip-checks to speed up internal testing feedback cycles.

This is supported for both initialize and execute as hidden flags. The command is "not" persisted in config.json. Thus, when specifying in initialize it is not persisted to execute and will need to be manually specified on the command line each time.

Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:skipPgUpgradeChecks


kyeap-vmware (approved):

kyeap-vmware:
We should be good to merge even though the flag doesn't exist  in pg_upgrade yet as long as we do not call it.

PR #818, opened 2023-05-19
Check more often that the source cluster is synchronized before upgrade

jimmyyih:
```
commit bb583a16cfe181c963dfbb53b0042e33f74d8dd8
Author: Jimmy Yih <jyih@vmware.com>
Date:   Thu May 18 12:15:50 2023 -0700

    Check that master and standby master are synchronized on GPDB 5X as well

    It's very important to make sure the master and standby master are
    synchronized before running Initialize and Execute. This is because
    Revert will rsync the standby master data directory onto the master
    data directory which can cause severe master corruption if the standby
    master not up-to-date. We have this check for GPDB 6X+ but it was
    seemingly missed for 5X because the catalog view gp_stat_replication
    does not exist on 5X. However, we're only checking the master and
    standby master replication status so there's no need to use the
    cluster-wide gp_stat_replication view; we can simply use
    pg_stat_replication which will only report the replication status of
    the individual segment (in this case, just the master segment).

commit b27b14311806bd1e34e34273cd701247085adfc4
Author: Jimmy Yih <jyih@vmware.com>
Date:   Thu May 18 12:17:34 2023 -0700

    Check more often that the source cluster is synchronized before upgrade

    We already check that the cluster is synchronized when creating the
    source cluster config file at the beginning of Initialize. However,
    the state of the source cluster could have changed any time afterwards
    between that early Initialize step and upgrading the master segment in
    Execute. If the cluster was out-of-sync and we run Revert after
    Execute, there is potential for severe catalog corruption. In
    particular, this is more likely to happen to the master and standby
    master since there's a GPDB-specific mechanism to bypass the
    synchronous WAL replication when the standby master is not
    responsive. To prevent this, we add a quick sanity check for cluster
    synchronization before the Initialize pg_upgrade checks and more
    importantly before doing the actual upgrade operation in Execute.
```

Dev pipeline:
https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:fix_wait_for_standby_5x
^ One red job but it's not relevant to these two patches


hub/initialize.go
```diff
@@ -146,6 +146,13 @@ func (s *Server) InitializeCreateCluster(req *idl.InitializeCreateClusterRequest
 		return RsyncCoordinatorDataDir(stream, sourceDir, targetDir)
 	})
 
+	st.AlwaysRun(idl.Substep_wait_for_cluster_to_be_ready_before_check_upgrade, func(streams step.OutStreams) error {
+		return s.Source.WaitForClusterToBeReady()
+	})
+
+	// Doing the upgrade check should always be last since it's the
+	// longest running Initialize step and rerunning it due to a
+	// following step failing would be a bad user experience.
```
kalensk, line 155: I'd remove this comment. I don't think its really needed at this moment in time as there are no relevant substeps that would even go after this.

hub/execute.go
```diff
@@ -49,6 +49,21 @@ func (s *Server) Execute(req *idl.ExecuteRequest, stream idl.CliToHub_ExecuteSer
...
 	})
 
+	// We do not always run this cluster synchronization check
+	// because checking requires the source cluster to be available
+	// and Substep_upgrade_master makes the source cluster
+	// unavailable.
+	st.Run(idl.Substep_wait_for_cluster_to_be_ready_before_upgrade_master, func(streams step.OutStreams) error {
+		// We can only check if the cluster is synchronized if
+		// the cluster is up and running. It's assumed that most
+		// of the time, this will be a no-op.
```
kalensk, line 59: I'd remove this particular comment. I don't think its really needed to be honest.

hub/initialize.go
```diff
@@ -146,6 +146,13 @@ func (s *Server) InitializeCreateCluster(req *idl.InitializeCreateClusterRequest
 		return RsyncCoordinatorDataDir(stream, sourceDir, targetDir)
 	})
 
+	st.AlwaysRun(idl.Substep_wait_for_cluster_to_be_ready_before_check_upgrade, func(streams step.OutStreams) error {
+		return s.Source.WaitForClusterToBeReady()
+	})
```
kalensk, line 151: Curious if this is _actually_ needed for the pg_upgrade checks as it doesn't seem so on first glance? Also, given that we do a sanity check early-on when creating the config file of checking the segments.

hub/initialize.go
jimmyyih: Sure, I'll remove it.  I mainly added it as a note/reminder to new devs since I had moved the new substep to after the upgrade checks and realized a couple minutes later that it was a bad idea.

hub/initialize.go
jimmyyih: > Curious if this is actually needed for the pg_upgrade checks as it doesn't seem so on first glance?

It's not needed for the upgrade checks; the naming is mainly to fit in with how other substeps were being named (presumably it's better than what I had originally with `Substep_wait_for_cluster_to_be_ready_initialize`).  The `before_check_upgrade` part is just positional rather than any dependency.  Any suggestion on naming?  Maybe `Substep_wait_for_cluster_to_be_ready_initialize_sanity_check`?

> Also, given that we do a sanity check early-on when creating the config file of checking the segments.

As mentioned in the commit message, that early sanity check could have happened days or weeks before the actual Execute upgrade step is run.  The standby master could have gone out-of-sync or a primary could have failed over to its mirror.  While the newly added check step in Execute will catch the main issue, it's somewhat cheap to add this sanity check here in Initialize as well to catch the issue mid-way (e.g. if their Initialize upgrade check fails and they make changes to the cluster with their standby master down).  This mid-way check can be considered optional as a courtesy check but it's cheap and effective for preventing a really big disaster IMO.  So essentially, there would be the early check (run once beginning of Initialize), this mid-way check (run always near end of Initialize), and a final check (run once before Execute upgrade step).

kalensk: I see. The thought being its a cheap check with large benefits, so best to essentially scatter it throughout the upgrade process. I am on board. Below was just a tangential thought/concern I had about that which I think is negligible.

My only thought around that which I don't think is a big deal is that theoretically initialize is designed to be run on the prod cluster weeks before where data could be ingesting such that the clusters could be out of sync. And this check preventing them from fully running initialize. However, in actual practice I think customers run initialize during maintenance windows or low-work periods (ie: late at night) such that the chance of this check (out of sync) being prohibitive is negligible.

kalensk:
Probably not for this PR, but wondering if we should call `s.Source.WaitForClusterToBeReady()` towards the end of revert. I believe I tried this before, but due to needing to run gprecoverseg in certain cases it cased caused `WaitForClusterToBeReady()` to take awhile. This would be annoying during development and testing. You might have a way to force that synchronization or speed that up in cases like that.

hub/initialize.go
kalensk: In terms of naming, maybe just move the initialize/execute part to the beginning, such that `Substep_initialize_wait_for_cluster_to_be_ready`. What you had or this is good.

kalensk (approved):
LGTM!

greenplum/wait_for_segments.go
```diff
@@ -77,14 +77,14 @@ WHERE content > -1 AND status = 'u' AND (role = preferred_role) ` + whereClause)
 		whereClause = "sent_lsn = flush_lsn;"
 	}
 
-	row = db.QueryRow("SELECT COUNT(*) FROM gp_stat_replication WHERE gp_segment_id = -1 AND state = 'streaming' AND " + whereClause)
+	row = db.QueryRow("SELECT COUNT(*) FROM pg_stat_replication WHERE state = 'streaming' AND " + whereClause)
```
kalensk, line 80: For my own benefit can you give a brief explanation here in the github comments why `gp_stat_replication` wasn't working for us compared to `pg_stat_replication`? Or is it linked in the JIRA.

hub/initialize.go
jimmyyih: Renamed to `Substep_initialize_wait_for_cluster_to_be_ready` but kept it before the upgrade checks.  It feels better to couple it with the always running upgrade checks.

greenplum/wait_for_segments.go
jimmyyih: The query previously was fine for 6X but `gp_stat_replication` is not a catalog view that exists in 5X.  This is why the previous logic was to skip master/standby synchronization checking on 5X which we've seen to be dangerous.  Since `gp_stat_replication` is basically just a more informational and distributed version of `pg_stat_replication`, the majority of its info is not needed (the query was literally filtering on only the master) and we can just use the basic `pg_stat_replication` view.  The `pg_stat_replication` view when run against the master will give the replication status between the master and its standby.  Best part about this is that the same query against `pg_stat_replication` can be used on all versions of GPDB that supports master/standby.

PR #821, opened 2023-05-23
Fixups

kalensk:
See individual commits.

Please take a close look to ensure the logic around generating an upgrade ID is still the same.

Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:fixups


kyeap-vmware (approved):
NewID changes look fine to me.

config/config_test.go
```diff
@@ -28,7 +28,7 @@ func TestConfig(t *testing.T) {
 		HubPort:      12345,
 		AgentPort:    54321,
 		Mode:         idl.Mode_copy,
-		UpgradeID:    0,
+		UpgradeID:    "0",
```
jimmyyih, line 31: Shouldn't the default uninitialized upgrade id string be blank `""` for this and all other mentions of `"0"`?

kalensk: It's testing that the configuration can be saved and read back. So we don't want to use defaults like `""`, but rather a value. For example, HubPort is `12345`, and in this case upgradeID is `"0"`, but probably having it be something like "fByNj2ldsTE" would be more clear. I can update that.

kalensk: 1) Updated this line to be `"ABC123"` to make it clear that it has a non-default value. The early tests in the file are asserting the conf and be written to and read. 
2) The following line on 212 I updated to check that its not default value of empty string (ie: `if conf.UpgradeID == ""`) rather than "0".
3) And then removed a few unnecessary initializes of `"0"` in `hub/server_test.go` which weren't technically needed. It helps keep things clean and easy to follow.

jimmyyih (approved):

PR #823, opened 2023-05-25
Functional pipeline fixes

kalensk:
See individual commits.

Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:functionalPipelineFixes

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -195,6 +195,7 @@ pipeline functional-pipeline: export JOB_TYPE=dev
 endif
 
 .PHONY: pipeline functional-pipeline expose-pipeline
+pipeline functional-pipeline: export DUMP_PATH=${DUMP_PATH:-}
 pipeline functional-pipeline: export 5X_GIT_USER=${5X_GIT_USER:-}
 pipeline functional-pipeline: export 5X_GIT_BRANCH=${5X_GIT_BRANCH:-}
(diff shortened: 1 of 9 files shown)


jimmyyih (approved):

kalensk:
Unfortunately, due to unforeseen issues I had to push a few fix-ups after approval. Please re-review the last two commits.

**1)** `max_statement_mem` needs to be greater than `statement_mem`. Thus, we now set `max_statement_mem` as well. 
These GUCs are dynamically set based on the total memory size. Set `max_statement_mem` to be half of the total memory, and set `statement_mem` to be 1/16th of the `max_statement_mem`.

For an n2-standard-16 instance with 64GB of memory this sets `max_statement_mem` to 32GB and `statement_mem` to 2GB.


**2)** Change naming implementation of `saved_cluster_env_files` and `dummy_resource` resources.
Ensure resource filenames are unique with git branch name to avoid collisions when multiple pipelines are run. Specify `BranchName` and to set unique filenames during pipeline creation since the Concourse GCS resource does not guarantee pulling the latest resource. See issue frodenas/gcs-resource#47

kyeap-vmware (approved):

jimmyyih (approved):

PR #824, opened 2023-05-26
Add --pg-upgrade-jobs flag

kyeap-vmware:
This value propagates to pg_upgrade's --jobs flag.  The pg_upgrade --jobs flag parallelizes the number of databases that it dumps and restores at a time. The default value for --pg-upgrade-jobs is set to 4. It was an arbitrary number that was chosen because it seemed like a safe default.

pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:jobs


idl/cli_to_hub.proto
```diff
@@ -33,6 +33,7 @@ message ExecuteRequest {
   bool pgUpgradeVerbose = 1;
   bool skipPgUpgradeChecks = 2;
   string parentBackupDirs = 3;
+  int32 pgUpgradeJobs = 4;
```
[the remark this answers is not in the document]
kyeap-vmware: files are regenerated, and there are two empty comments line left behind.

cli/commands/initialize.go
```diff
@@ -323,6 +325,7 @@ To return the cluster to its original state, run "gpupgrade revert".`,
 	subInit.Flags().BoolVar(&pgUpgradeVerbose, "pg-upgrade-verbose", false, "execute pg_upgrade with --verbose")
 	subInit.Flags().BoolVar(&skipPgUpgradeChecks, "skip-pg-upgrade-checks", false, "skips pg_upgrade checks")
 	subInit.Flags().MarkHidden("skip-pg-upgrade-checks") //nolint
+	subInit.Flags().UintVar(&pgUpgradeJobs, "pg-upgrade-jobs", 4, "databases to upgrade in parallel based on the number of specified threads. Defaults to 4.")
```
jimmyyih, line 328: Is enabling parallel upgrade by default what we want?  Shouldn't this be off by default and the user chooses?

kalensk: While generally true, I think in this case it makes sense to enable it by default (assuming we have a sane default for most scenarios with no major negative side-effects). Most customers will benefit from it and want it enabled (at least to some extent). 

This flag is most handy during execute. However, currently it can only set it during initialize. Thus, I think we should _consider_ whether adding the flag to execute (as well), and having it take precedence over what was passed during initialize. Basically what we do for --parent-backup-dirs

I'm happy to keep discussing all the options and see what makes the most sense.

cli/commands/initialize.go
```diff
@@ -50,6 +50,7 @@ func initialize() *cobra.Command {
 	var useHbaHostnames bool
 	var dynamicLibraryPath string
 	var dataMigrationSeedDir string
+	var pgUpgradeJobs uint
```
kalensk, line 53: minor nit: I'd move this to above `useHbaHostnames` to match the order of the parameters to the confiramtionText, or move it after `skipPgUpgradeChecks`.

gpupgrade_config
```diff
@@ -29,6 +29,9 @@ target_gphome =
 # link mode.
 # disk_free_ratio = 0.6
 
+# Databases to upgrade in parallel based on the number of specified threads.
+# pg_upgrade_jobs = 4
+
```
kalensk, line 34: I don't think we do, but slightly wondering if we need a second line that says:
```
# Needs to be at least 1.
```

kalensk (approved):
LGTM. 

- Some minor nit's to possibly consider.
- Probably a bit more team discussion around what the default value should be.

hub/upgrade_coordinator.go
```diff
@@ -31,6 +31,7 @@ func UpgradeCoordinator(streams step.OutStreams, backupDir string, pgUpgradeVerb
 		BackupDir:           backupDir,
 		PgUpgradeVerbose:    pgUpgradeVerbose,
 		SkipPgUpgradeChecks: skipPgUpgradeChecks,
+		PgUpgradeJobs:       strconv.Itoa(int(pgUpgradeJobs)),
```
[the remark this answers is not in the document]
kyeap-vmware: this value might break either way. protocol buffer is a unit32.

gpupgrade_config
kyeap-vmware: pg_upgrade checks for `user_opts.jobs <= 1`. It will work either way.

hub/upgrade_coordinator.go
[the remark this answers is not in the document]
kalensk: > this value might break either way. protocol buffer is a unit32.

The protocol buffer is actually a string (`string pgUpgradeJobs = 4;`), so it won't break there. I think the main place it will error (need to actually test it) is when parsing the config file since the flag is set to `UintVar`.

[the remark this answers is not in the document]
kyeap-vmware: > this value might break either way. protocol buffer is a unit32.

I have no idea what I was thinking. You're right. the code even implies it's a string because of strconv.

gpupgrade_config
kalensk: Yeah no need for the additional sentence. Original phrasing is good.

kyeap-vmware:
Force push to squash commits that address PR comments and rebase merge conflicts.

PR #825, opened 2023-06-01
fix progress bar

kalensk:
Need to wait for "both" the work being done and for the progress bars being incremented (ie: rendered).

Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:fixProgressBar

### DIFF
--- a/cli/commanders/data_migration_apply.go
+++ b/cli/commanders/data_migration_apply.go
@@ -13,6 +13,7 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+	"sync"
 
 	"github.com/fatih/color"
 	"github.com/vbauerster/mpb/v8"
@@ -60,11 +61,14 @@ func ApplyDataMigrationScripts(nonInteractive bool, gphome string, port int, log
 	}()
 
 	progressBar := mpb.New()
+	var wg sync.WaitGroup
 	errChan := make(chan error, len(scriptDirsToRun))
 	outputChan := make(chan []byte, len(scriptDirsToRun))
 
 	fmt.Printf("\nApplying data migration scripts...\n")
 	for _, scriptDir := range scriptDirsToRun {
+		wg.Add(1)
+
 		scriptDirEntries, err := utils.System.ReadDirFS(utils.System.DirFS(scriptDir), ".")
 		if err != nil {
 			return err
@@ -77,6 +81,8 @@ func ApplyDataMigrationScripts(nonInteractive bool, gphome string, port int, log
 				decor.CountersNoUnit("  %d/%d scripts applied")))
 
 		go func(gphome string, port int, scriptDir string, bar *mpb.Bar) {
+			defer wg.Done()
+
 			output, err := ApplyDataMigrationScriptSubDir(gphome, port, utils.System.DirFS(scriptDir), scriptDir, bar)
 			if err != nil {
 				errChan <- err
@@ -89,6 +95,7 @@ func ApplyDataMigrationScripts(nonInteractive bool, gphome string, port int, log
 	}
 
 	progressBar.Wait()
+	wg.Wait()
 	close(errChan)
 	close(outputChan)
 
--- a/cli/commanders/data_migration_generate.go
+++ b/cli/commanders/data_migration_generate.go
@@ -79,15 +79,19 @@ func GenerateDataMigrationScripts(nonInteractive bool, gphome string, port int, 
 
 	fmt.Printf("\nGenerating data migration scripts for %d databases...\n", len(databases))
 	progressBar := mpb.New()
+	var wg sync.WaitGroup
 	errChan := make(chan error, len(databases))
 
 	for _, database := range databases {
+		wg.Add(1)
 		bar := progressBar.New(int64(database.NumSeedScripts),
 			mpb.NopStyle(),
 			mpb.PrependDecorators(decor.Name("  "+database.Datname, decor.WCSyncSpaceR)),
 			mpb.AppendDecorators(decor.NewPercentage("%d")))
 
 		go func(database DatabaseInfo, gphome string, port int, seedDir string, outputDir string, bar *mpb.Bar) {
+			defer wg.Done()
+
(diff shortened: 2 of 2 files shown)


jimmyyih:
If the issue is with the progress bar and concurrency, the issue is more likely to be the progress bar(s) being completed too early.  Once those progress bars reach 100%, the main thread will be allowed to continue and possibly exit early (e.g. running the generate command instead of going through Initialize interactive mode) before the last data migration script is written to disk.  This makes sense when you look at `cli/commanders/data_migration_generate.go:GenerateScriptsPerPhase()`:
```
bar.Increment() // Increment for seed scripts run rather than actual scripts generated

if len(scriptOutput) == 0 {
        continue
}

var contents bytes.Buffer
contents.WriteString(`\c ` + database.QuotedDatname + "\n")

headerOutput, err := utils.System.ReadFileFS(seedDirFS, filepath.Join(phase.String(), scriptDir.Name(), strings.TrimSuffix(script.Name(), path.Ext(script.Name()))+"\

if err != nil && !errors.Is(err, fs.ErrNotExist) {
        return err
}

contents.Write(headerOutput)
contents.Write(scriptOutput)

outputPath := filepath.Join(outputDir, "current", phase.String(), scriptDir.Name())
err = utils.System.MkdirAll(outputPath, 0700)
if err != nil {
        return err
}

outputFile := "migration_" + database.QuotedDatname + "_" + strings.TrimSuffix(script.Name(), filepath.Ext(script.Name())) + ".sql"
err = utils.System.WriteFile(filepath.Join(outputPath, outputFile), contents.Bytes(), 0644)
if err != nil {
        return err
}
```
The bar is being incremented before the migration script is being written to disk which releases the main thread from its wait and allows it to exit quick since it's at the end of the generate command.  Since the main thread has exited, all the Go routines will be immediately stopped and killed resulting in the last migration script to not be written to disk.  Presumably, the main fix should be to move the `bar.Increment()` to after the migration script is written.  The wait group stuff you're adding in this PR should also help assist concurrency coordination so I think it'll still be useful.

kalensk:
> Presumably, the main fix should be to move the bar.Increment() to after the migration script is written. 

Sure we can do that.

> The wait group stuff you're adding in this PR should also help assist concurrency coordination so I think it'll still be useful.

Right, the additional wait group in this PR is still needed based on [this comment](https://github.com/vbauerster/mpb/issues/130#issuecomment-1509752800) from the issue I opened regarding using this progress bar library. The relevant text being:
> In other words if range databases loop has completed it doesn't mean that bars rendering loop completed as well and vise versa. There may be surprising side effects if forgetting to wait either, for example if you forget to wait for progress and you program ends right after wg.Wait() then bars may end up with incomplete state like showing 98%.

jimmyyih:
> > Presumably, the main fix should be to move the bar.Increment() to after the migration script is written.
> 
> I am not sure about that. For the generation code see the comment where we increment the bar: `bar.Increment() // Increment for seed scripts run rather than actual scripts generated`
> 
> Since the progress bar needs to know up front how many units of work to do which in this case how many seed scripts to exercise rather than how many scripts are outputted from those. That is, not all seed scripts result in an outputted script as that depends on the cluster and its data. Thus, we have the progress bar for seed scripts exercised rather than actual scripts outputted since we can't know up front how many scripts will be outputted.

Seems like it's just an issue with the code structure then and it should look like this instead:
```
if strings.HasSuffix(script.Name(), ".sh") || strings.HasSuffix(script.Name(), ".bash") {
        scriptOutput, err = executeBashFile(gphome, port, filepath.Join(seedDir, phase.String(), scriptDir.Name(), script.Name()), database.Datname)
        if err != nil {
                return err
        }
}

if len(scriptOutput) != 0 {
        var contents bytes.Buffer
        contents.WriteString(`\c ` + database.QuotedDatname + "\n")

        headerOutput, err := utils.System.ReadFileFS(seedDirFS, filepath.Join(phase.String(), scriptDir.Name(), strings.TrimSuffix(script.Name(), path.Ext(script.Name()))+".header"))

        if err != nil && !errors.Is(err, fs.ErrNotExist) {
                return err
        }

        contents.Write(headerOutput)
        contents.Write(scriptOutput)

        outputPath := filepath.Join(outputDir, "current", phase.String(), scriptDir.Name())
        err = utils.System.MkdirAll(outputPath, 0700)
        if err != nil {
                return err
        }

        outputFile := "migration_" + database.QuotedDatname + "_" + strings.TrimSuffix(script.Name(), filepath.Ext(script.Name())) + ".sql"
        err = utils.System.WriteFile(filepath.Join(outputPath, outputFile), contents.Bytes(), 0644)
        if err != nil {
                return err
        }
}

bar.Increment()
```

kalensk:
> Seems like it's just an issue with the code structure then and it should look like this instead:

We should sync off line since I'm not following. To me the current location of bar.increment is fine since its for when when the SQL or BASH seed scripts are run, and _not_ when the generated output scripts are created. 

The fix is to denote the difference between having wait groups for the word being done versus the rendering of the progress bar. These two are different for us since there is a difference between executing the seed scripts and writing any generated files if necessary.

kalensk:
The progress bar is tied to the number of seed scripts executed rather than the number of actual generated scripts written. This is because the progress bar needs to know up front how many units of work to do. It makes sense to increment the bar right after the seed scripts are executed since this is what the actual bar is tied to. The previous implementation had that.

However, there is a small section of code in which the bar is incremented before the generated script is fully written to. This may show the bar being completed even though there is still a little work to do in writing the file. In reality this is negligible since for even extremely large scripts the file is written very fast on current systems.

Given all of the above some people find it easier to increment the bar after the script was written even though this is technically not correct. Given the way the code is structured this results in the bar being incremented in two places: 1) once after the seed scripts are run and there is no generated script output to write, and 2) after the seed scripts are run and after the generated script is written.

So it is a trade off of having the increment scattered in two places versus a negligible scenario where the progress bar could show being completed while the file is being written. After discussing offline the majority of others wanted to go with ensuring the bar is incremented after the script is written. I would have preferred leaving the increment where it was.

jimmyyih (approved):

PR #832, opened 2023-06-10
Port two easy bats test to golang

kyeap-vmware:
```
Author: Kevin Yeap <kyeap@vmware.com>
Date:   Fri Jun 9 17:14:26 2023 -0700

    port golang compile version bats test to golang
    
    This test verifies that the version of go used to compiled gpupgrade is
    greater than the minimum version of go required. The minimum version of
    go required to compiled gpupgrade is specified in the go.mod file.

```
```
Author: Kalen Krempely <kalensk@gmail.com>
Date:   Fri Jun 9 15:43:32 2023 -0700

    port version bats test to golang
```

pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:portBATsTests


kalensk:
Technically I think these two tests should _eventually_ be moved to the integration directory. So far, its looking like acceptance tests are those that need a running demo cluster, and integration tests do not. Both integration and acceptance tests use a compiled gpupgrade binary.

test/acceptance/gpupgrade/goversion_test.go
```diff
@@ -0,0 +1,71 @@
...
+	t.Run("gpupgrade is compiled with the expected golang version from go.mod", func(t *testing.T) {
+		// Since the go mod version contains the minimum required go version it
+		// must be less than or equal to the compiled version.
+		if !goModVersion(t).LTE(compiledVersion(t)) {
+			t.Errorf("got %q want %q", goModVersion(t), compiledVersion(t))
+		}
+	})
+}
+
+func compiledVersion(t *testing.T) semver.Version {
```
kalensk, line 30: This needs a `	t.Helper()`

jimmyyih (approved):

jimmyyih:
One weird thing that I found is that the test failed because my gpupgrade binary from `make install` was outputting this:
```
$ gpupgrade version
Version:
Commit:
Release:
```

This was because `make install` does its own compilation but without baking in the versioning stuff.  The binary from `make build` worked fine with the test.  Should `make install` be changed to not use `go install` but instead to copy the compiled binary from `go build` to `$GOBIN`?

kalensk:
> One weird thing that I found is that the test failed because my gpupgrade binary from `make install` was outputting this: ...

Yeah, there is a long running tech debt story to look into this and fix it which has not yet been addressed. The story is: GPUP-9 - Makefile: Fix Makefile and why version output is sometimes blank

PR #835, opened 2023-06-15
Finalize: Refactor Analyze Target Cluster

kalensk:
See commit messages for details, but basically:

- Prompt user if they want to run analyze or not. This user input requires analyze target cluster to run on the CLI instead of the hub.
- Instead of running analyze_new_cluster.sh we can just run the single command: vacuumdb --all --analyze-only

Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:analyzeTargetCluster2


kalensk:
New implementation here https://github.com/greenplum-db/gpupgrade/pull/837

There were too many gotchas (ie: chicken and egg issues) when reading from the config.json file in the CLI.

PR #837, opened 2023-06-17
Finalize: Refactor Analyze Target Cluster

kalensk:
See commit messages for details, but basically:

- Prompt user if they want to run analyze or not. This user input requires analyze target cluster to run on the CLI instead of the hub.
- Instead of running analyze_new_cluster.sh we can just run the single command: vacuumdb --all --analyze-only

Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:gob


cli/commands/finalize.go
```diff
@@ -88,6 +90,28 @@ func finalize() *cobra.Command {
 					response.GetLogArchiveDirectory(), utils.System.DirFS(currentDir), currentDir, idl.Step_finalize)
 			})
 
+			st.Run(idl.Substep_analyze_target_cluster, func(streams step.OutStreams) error {
+				if !nonInteractive {
+					fmt.Println()
+					fmt.Println(`
+It is strongly recommended to create optimizer statistics to ensure performant operations. 
+However, this could take quite awhile and you may need your cluster now.
+If you choose to postpone after the upgrade run "vacuumdb --all --analyze-in-stages".`)
```
jimmyyih, line 99: This sentence reads a bit weird to me.  Maybe it needs a comma after "postpone" and a bit of shifting to something like: If you choose to postpone, run "vacuumdb --all --analyze-in-stages" after the upgrade.

jimmyyih (approved):
For the gob and --verbose commits, could you add thoughts/details to the commit message.  Specifically on why we're using gob and why we're suggesting users to use --verbose for all commands.

cli/commands/finalize.go
kalensk: Updated to `If you postpone creating statistics then after the upgrade run "vacuumdb --all --analyze-in-stages".` Ideally I'm trying to keep the command at the end of the sentence since its easier to parse/read than having it elsewhere. Totally happy to change it to whatever the majority of people prefer.

kalensk:
> Specifically on why we're using gob and why we're suggesting users to use --verbose for all commands.

Updated. Let me know if you have any other suggestions.

jimmyyih:
> > Specifically on why we're using gob and why we're suggesting users to use --verbose for all commands.
> 
> Updated. Let me know if you have any other suggestions.

Updated commit messages look good.  Thanks!

kalensk:
Holding off on merging as some unforeseen consequences were found on the CI that complicates things further. More investigation and likely help from the team on best approach forward are needed.

All the other e2e jobs are passing fine except two, which is a bit strange at first.

The 5-to-6 upgrade extensions job fails when analyzing:
```
vacuumdb: vacuuming of database "postgres" failed: ERROR:  function raster_hash(public.raster) does not exist
LINE 1:  SELECT raster_hash($1) = raster_hash($2) 
                ^
HINT:  No function matches the given name and argument types. You might need to add explicit type casts.
QUERY:   SELECT raster_hash($1) = raster_hash($2) 
CONTEXT:  SQL function "raster_eq" during startup
```

The 6-to-6 copy mode job fails when analyzing:
```
vacuumdb: vacuuming of database "regression" failed: ERROR:  read beyond eof in table "m_aocs" file "base/60402/26105", read position 0 (small offset 0), actual read length 0 (large read length 48) (cdbbufferedread.c:211)  (seg1 10.0.154.25:20001 pid=18253) (cdbbufferedread.c:211)
```

jimmyyih:
In commit 4b7f72e6241312b40298970d649e0a4e202a26b2, there's a couple occurrences of `steams` that need to be changed to `streams`.

ci/main/scripts/load-dump.bash
```diff
@@ -112,38 +63,10 @@ echo "${databases}" | while read -r database; do
...
-                psql -v ON_ERROR_STOP=1 -d regression -c 'DROP MATERIALIZED VIEW IF EXISTS ${view}';
-            " || echo "Dropping materialized views failed. Continuing..."
-        fi
-    done
+if is_GPDB5 ${GPHOME_SOURCE}; then
+    echo "Applying 5-to-6 workarounds..."
+    bash gpupgrade_src/ci/main/scripts/5-to-6-workarounds.bash
+elif is_GPDB6 ${GPHOME_SOURCE}; then
+    echo "Applying 6-to-7 workarounds..."
+    bash gpupgrade_src/ci/main/scripts/6-to-7-workarounds.bash
```
jimmyyih, line 71: Nice, there definitely will be workaround differences so this will help in the near future.

jimmyyih (approved):

PR #839, opened 2023-06-22
force copy when make installing

kalensk:
Sometimes when running make install the following error will show up:
```
cp: cannot create regular file '/home/gpadmin/go/bin/gpupgrade': Text file busy
make: *** [Makefile:141: install] Error 1
```

Where cp -f will remove the existing destination file if it cannot be opened and try again.

https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:fixMakeInstall

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -138,7 +138,7 @@ rpm:
 
 install:
 	@test $${GOPATH?Error GOPATH not set}
-	cp gpupgrade $(GOPATH)/bin/
+	cp -f gpupgrade $(GOPATH)/bin/
 
 # To lint, you must install golangci-lint via one of the supported methods
 # listed at


jimmyyih:
Doesn't that error message mean that you're trying to modify an executable while it's running?  Most likely you're OS is somehow modifying gpupgrade while your hub and agent are doing something?  Example:
```
$ lsof | grep gpupgrade | grep txt
gpupgrade 39432 jyih  txt       REG               1,11     22099640           259638593 /Users/jyih/go/bin/gpupgrade
gpupgrade 39441 jyih  txt       REG               1,11     22099640           259638593 /Users/jyih/go/bin/gpupgrade

$ ps -ef | grep [3]9432
  501 39432     1   0  3:05PM ??         0:00.05 gpupgrade hub --daemon
  
$ ps -ef | grep [3]9441
  501 39441     1   0  3:05PM ??         0:00.04 /Users/jyih/go/bin/gpupgrade agent --daemon --port 6416 --state-directory /var/folders/pd/gn5d8jn96mz14v8v6tncjjtm0000gn/T/2938115789
```

However, `cp` shouldn't be doing any modifications... just replacing the file so that the original binary's inode would still be intact for the existing processes to still run.  That's a weird one but `cp -f` should be fine if that fixes the issue for you.

jimmyyih (approved):

PR #840, opened 2023-06-23
Fix Concourse CI lint job

jimmyyih:
The golangci/golangci-lint image was recently updated from v1.52 to v1.53 which caused our Concourse CI lint job to start failing. It seems like this new image does not play well with our prod Concourse workers since the new image seems to work on our developer Concourse instances that are on newer versions of Concourse. To work around the issue for now, pin the image to v1.52.

The image is already pinned on our gpupgrade prod pipeline which is now green.  This PR is to save the change.

### DIFF
--- a/ci/main/pipeline/2_build_lint.yml
+++ b/ci/main/pipeline/2_build_lint.yml
@@ -40,6 +40,8 @@ jobs:
           type: registry-image
           source:
             repository: golangci/golangci-lint
+            # Pin the image to v1.52 because v1.53 has issues with current prod Concourse
+            tag: v1.52
         inputs:
           - name: gpupgrade_src
         run:


kalensk:
> It seems like this new image does not play well with our prod Concourse workers

Are there any specifics around this?

If the timeline to upgrading the prod concourse is somewhat reasonable it might be best to just let the job fail for a bit. Alternatively, can we investigate running lint in the build job? I'm pretty against pinning unless there is "absolutely" no alternatives.   (Im more okay with temporarily pining via just flying the prod pipeline compared to committing the change in .yml). It would be nice to brainstorm some alternatives before doing so.

jimmyyih:
> Are there any specifics around this?

As you've pointed out, it's due to the Debian OS version bump which bumped the Linux kernel version to something incompatible with the prod Concourse workers.  You can see more info here:
https://www.debian.org/releases/stable/amd64/release-notes/ch-whats-new.en.html#idm120

> If the timeline to upgrading the prod concourse is somewhat reasonable it might be best to just let the job fail for a bit.

I'm not optimistic on the timeline being reasonably soon so need to do something now.

> Alternatively, can we investigate running lint in the build job?

Nothing wrong with that approach.  It'll just add a tiny bit more work to the build job that could be slightly annoying (e.g. always install the linter package and run the linter... either before or after compilation).  We could also move the lint job off of Concourse and use Github Actions similar to this [commit](https://github.com/greenplum-db/gpbackup/pull/547/commits/ed342ebc0064e9152c8d7ef2b9d90505a135c92f)... but it'll be easier now since they have their own [action](https://github.com/golangci/golangci-lint-action).  Although doing the pinning (with the explanation comment) is the better and fastest solution right now IMO since we're just picking the image that works with our current CI infrastructure... and I don't think it's worth the effort to spend more time on this small lint job.  But if you think it's worth the effort, I can close this PR and create a story in the backlog (for the brainstorming and fixing) while leaving the band-aid in place.

jimmyyih:
Apparently we already have the golangci-lint running as a Github Action.  Not sure how I missed this...  so it should be fine to have that going an let the CI lint job run with the pin (disable it if it fails again) until the prod concourse instance is updated.  Once the update happens, we can unpin and have things go back to normal.  Closing this PR then.

PR #842, opened 2023-06-28
ci: use future minor version RPMs

kalensk:
Releng cannot produce a release candidate RPM that once tagged produces a corresponding RPM with the correct version. Rather another commit is needed to correctly increment the version. This causes lots of headaches for gpupgrade since we compile in a minimum GPDB version causing gpupgrade release to wait for the correct version of GPDB to flow through the pipelines. For now, use a release candidate that has the future minor version rather than the current minor version. This is problematic since the commit sha and this future version do not correctly match reality.

Remove unused majorVersion function

Pipeline: 
https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:rpmResource

Functional Pipeline:
`make PIPELINE_NAME=gpupgrade:rpmResource-functional DUMP_PATH=dump/5X/dump.sql.xz functional-pipeline`
https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:rpmResource-functional

### DIFF
--- a/ci/main/pipeline/1_resources_anchors_groups.yml
+++ b/ci/main/pipeline/1_resources_anchors_groups.yml
@@ -54,14 +54,27 @@ resources:
   type: gcs
   source:
     {{- if .TestRCIdentifier }}
-    # Test release candidate rpms built with --build-test-rc are published to the -dev bucket.
-    bucket: pivotal-gpdb-concourse-resources-dev
-    {{- else }}
-    # Test release candidate rpms built from production pipelines are published to the -prod bucket.
-    bucket: pivotal-gpdb-concourse-resources-prod
+      # Test release candidate rpms built with --build-test-rc are published to the -dev bucket.
+      bucket: pivotal-gpdb-concourse-resources-dev
+      json_key: ((concourse-gcs-resources-service-account-key))
+      regexp: server/published/gpdb{{.GPDBVersion}}/greenplum-db-{{.TestRCIdentifier}}({{escapeVersion .GPDBVersion}}.*)-rhel{{.OSVersionNumber}}-x86_64.debug.rpm
+    {{ continue }}
     {{- end }}
-    json_key: ((concourse-gcs-resources-service-account-key))
-    regexp: server/published/gpdb{{ majorVersion .GPDBVersion }}/greenplum-db-{{.TestRCIdentifier}}({{escapeVersion .GPDBVersion}}.*)-rhel{{.OSVersionNumber}}-x86_64.debug.rpm
+    {{- if eq .GPDBVersion "5" }}
+      bucket: pivotal-gpdb-concourse-resources-prod
+      json_key: ((concourse-gcs-resources-service-account-key))
+      regexp: server/published/gpdb{{.GPDBVersion}}/greenplum-db-{{.TestRCIdentifier}}({{escapeVersion .GPDBVersion}}.*)-rhel{{.OSVersionNumber}}-x86_64.debug.rpm
+    {{ continue }}
+    {{- end }}
+      # Releng cannot produce a release candidate RPM that once tagged produces a corresponding RPM with the correct
+      # version. Rather another commit is needed to correctly increment the version. This causes lots of headaches for
+      # gpupgrade since we compile in a minimum GPDB version causing gpupgrade release to wait for the correct version of
+      # GPDB to flow through the pipelines. For now, use a release candidate that has the future minor version rather than
+      # the current minor version. This is problematic since the commit sha and this future version do not correctly
+      # match reality.
+      bucket: pivotal-gpdb-concourse-resources-prod
+      json_key: ((concourse-gcs-resources-service-account-key))
+      regexp: minor/release-candidates/gpdb{{.GPDBVersion}}/greenplum-db-server-({{escapeVersion .GPDBVersion}}.*)-rhel{{.OSVersionNumber}}-x86_64.rpm
 {{end}}
 
 - name: oss_rpm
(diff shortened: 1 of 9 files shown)


kalensk:
Just discovered that the new location of rpm's is two months old, whereas gpupgrade needs rpm's that are continuously created to catch regressions and test any recent changes made to pg_upgrade, etc. This needs to be addressed before merging as the current location is a no go so far.

jimmyyih:
Not sure how I missed this but the rpm we're pulling now seems to be an optimized build instead of a debug build... I think.  The dev pipeline shows failure for the pg_upgrade tests where the failures are:
```
   test gp_fastsequence          ... FAILED (0.39 sec)  (diff:0.13 sec)
   test ao_indexes               ... FAILED (0.33 sec)  (diff:0.16 sec)

For gp_fastsequence:
===================
-ERROR:  Unexpected internal error (appendonlytid.h:66)
-DETAIL:  FailedAssertion("!(h->bytes_4_5 <= 0x8000)", File: "../../../../src/include/access/appendonlytid.h", Line: 66)
+ i
+---
+(0 rows)

For ao_indexes:
==================
-ERROR:  Unexpected internal error
-DETAIL:  FailedAssertion
+ a
+---
+(0 rows)
```
We expect assertion failures but we don't get that now.  It's unfortunate but that's okay for now.  To fix that issue, you'll need to create an alternative expected out file for the two tests (e.g. `expected/gp_fastsequence_optimized.out` and `expected/ao_indexes_optimized.out`).

test/acceptance/pg_upgrade/5-to-6/upgradeable_tests/target_cluster_regress/expected/ao_indexes.out
```diff
@@ -89,8 +82,9 @@ EXPLAIN (COSTS off) SELECT * FROM aoco_with_gist WHERE a @@ to_tsquery('footext1
  Optimizer: Postgres query optimizer                      
 (6 rows)
 SELECT * FROM ao_with_btree WHERE a > 8;
-ERROR:  Unexpected internal error
-DETAIL:  FailedAssertion
+ a
+---
+(0 rows)
```
jimmyyih, line 87: The idea is to add an alternative expected output file, not modify the current one.  With this change, our local dev test runs will fail.  If you want the optimized builds to be the main expected output then we can have the previous expected output file as `ao_indexes_debug.out` (do same for gp_fastsequence test) instead of the other way around.

kalensk: Does the framework allow do of the same expected files with the same name? I guess maybe we should pair since Im not familiar with that.

jimmyyih: Just to note down what I mentioned while pairing for historical context:
Yes, you can have multiple expected files for the same test (`<test_name>_<prefix>.out`) so in our case we can have `ao_indexes.out` be the CI one for release builds and `ao_indexes_debug.out` be for debug dev/prod builds.  It's practiced fairly often in the regress/isolation/isolation2 frameworks (especially for ORCA vs planner diffs).

kalensk: I wasn't able to get running locally to use the `ao_indexes_debug.out` for example. CI is passing though!

jimmyyih: Sorry, I remembered somewhat incorrectly.  It's actually `<test_name>_[0-9].out` as the generic alternate expected files.  See https://github.com/greenplum-db/gpdb/blob/509887dc0d4be1777a9f87c6011a9bd60bedf49a/src/test/regress/pg_regress.c#L1836-L1868.  ORCA has its own special hack to switch to its expected files.  So we can call the debug expected file `ao_indexes_1.out`.

kalensk: Ah, that seems to work. I _really_ wish the regress/isolation framework had some more user-friendly documentation.

ci/main/pipeline/1_resources_anchors_groups.yml
```diff
@@ -54,14 +54,27 @@ resources:
...
+      json_key: ((concourse-gcs-resources-service-account-key))
+      regexp: server/published/gpdb{{.GPDBVersion}}/greenplum-db-{{.TestRCIdentifier}}({{escapeVersion .GPDBVersion}}.*)-rhel{{.OSVersionNumber}}-x86_64.debug.rpm
+    {{ continue }}
+    {{- end }}
+      # Releng cannot produce a release candidate RPM that once tagged produces a corresponding RPM with the correct
+      # version. Rather another commit is needed to correctly increment the version. This causes lots of headaches for
+      # gpupgrade since we compile in a minimum GPDB version causing gpupgrade release to wait for the correct version of
+      # GPDB to flow through the pipelines. For now, use a release candidate that has the future minor version rather than
+      # the current minor version. This is problematic since the commit sha and this future version do not correctly
+      # match reality.
```
jimmyyih, line 74: Is the commit sha still an issue here?  It seemed fine since the version string had a commit sha that we could look up in the 6X tags/branch.

jimmyyih (approved):

ci/main/pipeline/1_resources_anchors_groups.yml
kalensk: That sentence is just trying to convey that since we are using a different version string from what is produced from 6X_STABLE (ie: "reality") the commit sha and version are mismatched from that perspective. Happy to reword it, as long as that notion is explained in the comment.

jimmyyih: I see; that makes sense.  Sure, we can leave the sentence as-is.

PR #852, opened 2023-07-13
Update grpc and styling

kyeap-vmware:
- lint: only use references to idl structs to avoid copying mutexes
 - lint: use updated grpc insecure credential function calls
 - update error message check on forced shutdown of grpc servers
 - compile and regenerate protobuf and mock files
 - remove old generated protobuf and mock files
 - protobuf style: use pluralized names for repeated fields
 - protobuf style: header ordering
 - protobuf style: use an indent of 2 spaces
 - update grpc and mock compile commands and the proto files
 - go.mod: bump google.golang.org/grpc v1.56.2
 - go.mod: bump protoc-gen-go > protoc-gen-go-grpc v1.3.0
 
 pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:update-grpc


kalensk:
I think a new `go generate ./idl` needs to be run and the newly generated files checked in.

cli/commanders/steps.go
```diff
@@ -43,80 +43,80 @@ func Initialize(client idl.CliToHubClient, request *idl.InitializeRequest, verbo
 	return nil
 }
 
-func InitializeCreateCluster(client idl.CliToHubClient, request *idl.InitializeCreateClusterRequest, verbose bool) (idl.InitializeResponse, error) {
+func InitializeCreateCluster(client idl.CliToHubClient, request *idl.InitializeCreateClusterRequest, verbose bool) (*idl.InitializeResponse, error) {
 	stream, err := client.InitializeCreateCluster(context.Background(), request)
 	if err != nil {
-		return idl.InitializeResponse{}, err
+		return nil, err
```
kalensk, line 49: The reason why all these return an empty struct (ie: `idl.InitializeResponse{}`) instead of `nil` is it protects the caller and following code from de-referencing a nil pointer which causes a panic and everything to blow up. Whereas using an empty struct has default values for the fields. Thus, when the calling code references those fields such as in the `st.Complete` text it won't panic, but instead just have empty values. This is a good thing.

I think it's safe and best practice to return an empty struct than nil.

cli/commanders/steps.go
```diff
@@ -43,80 +43,80 @@ func Initialize(client idl.CliToHubClient, request *idl.InitializeRequest, verbo
...
 	response, err := UILoop(stream, verbose)
 	if err != nil {
-		return idl.InitializeResponse{}, err
+		return nil, err
 	}
 
 	initializeResponse := response.GetInitializeResponse()
 	if initializeResponse == nil {
-		return idl.InitializeResponse{}, xerrors.Errorf("Initialize response is nil")
+		return nil, xerrors.Errorf("Initialize response is nil")
```
kalensk, line 59: I know this is existing code, but since we are here we can easily clean it up.

I don't think we need the following pattern since the underlying GetInitializeResponse() call should do that for us. 
```go
	initializeResponse := response.GetInitializeResponse()
	if initializeResponse == nil {
		return nil, xerrors.Errorf("Initialize response is nil")
	}

	return initializeResponse, nil
```

We should update to this pattern. Thoughts?

```go
return response.GetInitializeResponse(), nil
```

cli/commands/commands.go
```diff
@@ -243,11 +244,11 @@ func stopHubAndAgents() error {
...
-		errCode := grpcStatus.Code(err)
-		errMsg := grpcStatus.Convert(err).Message()
-		// XXX: "transport is closing" is not documented but is needed to uniquely interpret codes.Unavailable
-		// https://github.com/grpc/grpc/blob/v1.24.0/doc/statuscodes.md
-		if errCode != codes.Unavailable || errMsg != "transport is closing" {
+		// XXX: "error reading from server: EOF" is not documented but is
+		// needed to uniquely interpret codes.Unavailable
+		// https://github.com/grpc/grpc/blob/v1.52.0/doc/statuscodes.md
+		errStatus := grpcStatus.Convert(err)
+		if errStatus.Code() != codes.Unavailable || errStatus.Message() != "error reading from server: EOF" {
```
kalensk, line 251: Should we update the link to the latest? https://github.com/grpc/grpc/blob/v1.56.1/doc/statuscodes.md

kalensk, line 251: Now that I look at this if statement closer. Im wondering if it should be an "and" instead of an "or" since won't we want to always ignore when Unavailable with that particular error message?

greenplum/tablespace.go
```diff
@@ -47,7 +47,13 @@ type Tablespace struct {
...
 	Oid  int32
 	Name string
-	Info idl.TablespaceInfo
+	Info *idl.TablespaceInfo
+}
+
+func NewTablespace() Tablespace {
+	var ts Tablespace
+	ts.Info = &idl.TablespaceInfo{}
+	return ts
```
kalensk, line 56: This should be removed.

cli/commanders/steps.go
kalensk: After taking a closer look at this and thinking more about it we unfortunately do need to keep the existing nil check. Ugh. We don't want our callers to have to deal with a nil response. It should be a full valid response object or error out.

cli/commands/commands.go
kyeap-vmware: we refactored this.

cli/commanders/steps.go
kyeap-vmware: we determined that this was done because grpc can return nil.

cli/commanders/steps.go
kyeap-vmware: We discussed this. For our use case it's probably correct to return nil. However, I will change it back to return address to empty struct for codebase consistency. We can choose to change in another commit.

idl/generate.go
```diff
@@ -3,9 +3,14 @@
...
 
-// Creates the .pb.go protobuf definitions.
-//go:generate protoc --plugin=../dev-bin/protoc-gen-go --go_out=plugins=grpc:. common.proto cli_to_hub.proto hub_to_agent.proto
+// Creates the .pb.go protobuf definitions. The use of
+// require_unimplemented_servers=false is to use legacy grpc implementation
+// that does not care about forward compatibility. We are the sole users and
+// maintainers of this grpc definition so this works for us and lets us not
+// need to embed a non-functional interface in our hub and agent servers just
+// to meet a grpc requirement that does not apply to us.
+//go:generate protoc --go_out=. --go_opt=paths=source_relative --go-grpc_out=require_unimplemented_servers=false:. --go-grpc_opt=paths=source_relative common.proto cli_to_hub.proto hub_to_agent.proto
```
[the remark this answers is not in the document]
kyeap-vmware: Specifies the output directory for the generated Go code for gRPC service interfaces and implementations.

kalensk (approved):
Looks good! A very much needed change.

Thank you for updating gRPC version!

PR #856, opened 2023-07-19
remove data migration test directory from rpm

kalensk:
remove following from both the tarball and rpm:
- 6-to-7-seed-scripts/test directory
- data-migration-scripts/README.md

This will also remove any future X-to-Y-seed-scripts/test directories.


kyeap-vmware (approved):
LGTM

PR #858, opened 2023-07-27
port rest of execute.bats to golang

kalensk:
1) port rest of execute.bats to golang

2) bump go test timeout since we are running more and more bats test and hitting the default 10 minute timeout.

3) shift defer to earlier on after initialize in case an error happens earlier on.

4) Fix weird issue in revert.bats by  reindexing the pg_attribute table. See commit for details.

Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:portExecuteBats


jimmyyih:
> Fix weird issue in revert.bats by reindexing the pg_attribute table. See commit for details.

This could be an actual bug that was exposed and may need to be treated as such.  As the workaround should be temporary, we should put a TODO message above it to RCA the issue as soon as we can.

kalensk:
> This could be an actual bug that was exposed and may need to be treated as such. As the workaround should be temporary, we should put a TODO message above it to RCA the issue as soon as we can.

I created a story ("Fix gpupgrade acceptance test issue") with all the details rather than a TODO since soon all the tests will be ported. Its possible to checkout to a specific commit and see the issue in a pipeline.

kyeap-vmware (approved):

PR #860, opened 2023-08-03
Acceptance Testing Refactors

kalensk:
See individual commits for various refactors mostly around acceptance testing.

Pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:prefactors


kyeap-vmware (approved):
That is some tricky bufio and stdin stuff. Luckily it doesn't affect production code and only affects tests that need multiple user inputs.

PR #861, opened 2023-08-04
port revert.bats tests to golang

kalensk:
Port revert.bats tests to golang which so far are the only acceptance tests to run in a multi-host environment. Thus, introduced remote_test_utils.go

The revert_test.go tests should be running in the `5-to-6-centos7-multihost-acceptance-tests` job.

https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:portRevertBats

PR #862, opened 2023-08-05
Consolidate acceptance test directories

kyeap-vmware:
There are two types of acceptance tests, gpupgrade and pg_upgrade. As
    both of these tests are ported from bats to golang, we need to put move
    the respecitve tests out of their subdirectory to put them in the same
    package. These two types of tests use the same set of support functions
    (initialize, execute, finalize, revert, restart, kill-services). As more
    bats tests are ported to golang we needed to make a choice to...
    
    1. Keep the gpupgrade and pg_upgrade tests in separate test packages.
       This would result in many end-to-end specific support functions
       landing in testutils and would result to a LOT of testutils.xxxxxx
       calls in both tests. Dot import could have been used to resolve this,
       but dot import is only recommended for use to break circular imports.
       Keeping the support functions in gpupgrade_tests test package and
       importing into pg_upgrade_tests package is not an option since golang
       does not allows functions to be exported from *_test.go files.
    
    2. Consolidate the gpupgrade and pg_upgrade tests into one testing
       package. This would be achieved by moving the contents of both
       test/acceptance/gpupgrade and test/acceptance/pg_upgrade
       up one directory and remove the nested folders.
    
    We have chosen to consolidate the gpupgrade and pg_upgrade tests.
    
pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:consolidate-test-directories


kalensk:
Looks good! Good change.

1) In the comment message at the end of second line remove the word "put" and fix spelling for "respective".

2) Looks like multihost testing mgiht need golang in PATH. See same change which was needed in the other open [PR](https://github.com/greenplum-db/gpupgrade/pull/861/files#diff-408e290fd4a44b8661e46955e129b8344e760541f6a2c76b0e5a88f9d386638eR11-R24).
In `ci/main/scripts/acceptance-multihost-tests.bash` something like:

```BASH
function run_migration_scripts_and_tests() {
    time ssh cdw '
        set -eux -o pipefail

        export TERM=linux
        export PATH=$PATH:/usr/local/go/bin
        export GOFLAGS="-mod=readonly" # do not update dependencies during build

        source gpupgrade_src/ci/main/scripts/environment.bash
        source "${GPHOME_SOURCE}"/greenplum_path.sh

        echo "Running data migration scripts to ensure a clean cluster..."
        gpupgrade generate --non-interactive --gphome "$GPHOME_SOURCE" --port "$PGPORT"
        gpupgrade apply    --non-interactive --gphome "$GPHOME_SOURCE" --port "$PGPORT" --phase initialize

        cd gpupgrade_src
        go test --cover -count=1 -timeout 30m -v -run "^TestRevert$" ./test/acceptance/gpupgrade
        ./test/acceptance/gpupgrade/revert.bats
  '
}
```

kalensk (approved):
LGTM! With a few minor tweaks for the multi host testing job.

PR #872, opened 2023-09-06
port migratable finalize test heterogeneous_partition_tables

kyeap-vmware:
This test is ported but not added to the schedule because it looked like upgrade is causing data loss in these tables. Partitions are missing data after upgrade happens.

pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:port-heterogenous-partitions

### DIFF
--- a/test/acceptance/5-to-6/migratable_tests/source_cluster_regress/sql/heterogeneous_partition_tables.sql
+++ b/test/acceptance/5-to-6/migratable_tests/source_cluster_regress/sql/heterogeneous_partition_tables.sql
@@ -1,0 +1,136 @@
+-- Copyright (c) 2017-2023 VMware, Inc. or its affiliates
+-- SPDX-License-Identifier: Apache-2.0
+
+--------------------------------------------------------------------------------
+-- Create and setup migratable objects
+--------------------------------------------------------------------------------
+
+-- Heterogeneous partition table with dropped column, constraint, and default
+-- The root and only a subset of children have the dropped column reference.
+CREATE TABLE dropped_column (
+    a int CONSTRAINT positive_int CHECK (b > 0),
+    b int DEFAULT 1,
+    c char,
+    d varchar(50)
+) DISTRIBUTED BY (c)
+PARTITION BY RANGE (a)
+(
+    PARTITION part_1 START(1) END(5),
+    PARTITION part_2 START(5)
+);
+
+ALTER TABLE dropped_column DROP COLUMN d;
+ALTER TABLE dropped_column OWNER TO test_role1;
+
+-- Splitting the subpartition leads to its rewrite, eliminating its dropped column
+-- reference. So, after this, only part_2 and the root partition will have a
+-- dropped column reference.
+ALTER TABLE dropped_column SPLIT PARTITION FOR (1) AT (2)
+INTO (
+    PARTITION split_part_1,
+    PARTITION split_part_2
+);
+
+INSERT INTO dropped_column VALUES (1, 1, 'a');
+INSERT INTO dropped_column VALUES (5, 1, 'a');
+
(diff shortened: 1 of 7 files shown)


kalensk:
Is there a way to add it to the schedule file, but comment it out? Im not sure if the schedule file supports comments... Just an idea.

kalensk:
Actually, thinking a bit more I think we should really add this test to the schedule and just comment out whatever parts don't yet work. That way some of the test is running and its not fully dead code. It also makes it obvious to future developers to fix it since its commented out, etc.

kyeap-vmware:
> Actually, thinking a bit more I think we should really add this test to the schedule and just comment out whatever parts don't yet work. That way some of the test is running and its not fully dead code. It also makes it obvious to future developers to fix it since its commented out, etc.

Done. I made a separate commit that comments out the data checks.

test/acceptance/5-to-6/migratable_tests/source_cluster_regress/sql/heterogeneous_partition_tables.sql
```diff
@@ -0,0 +1,138 @@
...
+);
+
+ALTER TABLE dropped_column DROP COLUMN d;
+ALTER TABLE dropped_column OWNER TO test_role1;
+
+-- Splitting the subpartition leads to its rewrite, eliminating its dropped column
+-- reference. So, after this, only part_2 and the root partition will have a
+-- dropped column reference.
+ALTER TABLE dropped_column
+SPLIT PARTITION FOR (1) AT (2)
```
kalensk, line 29: super minor nit: I find it easier to read if these were on the same line.

test/acceptance/5-to-6/migratable_tests/source_cluster_regress/sql/heterogeneous_partition_tables.sql
```diff
@@ -0,0 +1,138 @@
...
+    a int,
+    b int,
+    c char,
+    d varchar(50),
+    to_drop int
+);
+ALTER TABLE intermediate_table DROP COLUMN to_drop;
+
+ALTER TABLE child_has_dropped_column
+EXCHANGE PARTITION part_1 WITH TABLE intermediate_table;
```
kalensk, line 60: super minor nit: I find it easier to read if these were on the same line.

test/acceptance/5-to-6/migratable_tests/target_cluster_regress/expected/heterogeneous_partition_tables.out
```diff
@@ -0,0 +1,100 @@
...
+
+INSERT INTO dropped_column VALUES (2, 2, 'b');
+INSERT 1
+INSERT INTO dropped_column VALUES (3, 2, 'b');
+INSERT 1
+-- insert should fail due to constraint
+INSERT INTO dropped_column VALUES (4, -1, 'b');
+ERROR:  new row for relation "dropped_column_1_prt_split_part_2" violates check constraint "positive_int"
+DETAIL:  Failing row contains (4, -1, b).
+
```
kalensk, line 70: Nice! Great test and thing to check post-upgrade!

kalensk (approved):
LGTM with some minor tweaks and maybe see if we can add some additional post-upgrade tests around exercising any constraints for the other tables if that makes sense to.

test/acceptance/5-to-6/migratable_tests/target_cluster_regress/expected/heterogeneous_partition_tables.out
kyeap-vmware: This is the only table in this file that has a constraint! Other tests also have constraints that are also being tested in a similar fashion.

PR #874, opened 2023-09-07
port migratable finalize test drop_partition_seg_entries

kyeap-vmware:
commit message from where this test was introduced
```
data-migration-script to drop parent partitions with seg entries

The ICW regression database has more AO and AOCO parent
partitions with seg entries which is currently causing the 5 to 6
upgrade jobs to fail. Thus, the easiest way to address all these tables
seems to add a pre-initialize data migration script. There is no need
for a post-finalize or post-revert script since the seg entries are not
needed.
```

gpupgrade reference commit:
ec53ecd39ba0ec15bb91bfe8f064f02cc3a33873

pipeline:
https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:partition-seg-entries


kalensk (approved):
LGTM!

PR #877, opened 2023-09-16
Port migratable revert tests.

kyeap-vmware:
pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:port-migratable-initialize-revert


test/acceptance/migratable_test.go
```diff
@@ -33,28 +36,98 @@ func TestMigrationScripts(t *testing.T) {
...
-		sourceTestDir := filepath.Join(migratableTestDir, "source_cluster_regress")
-		isolation2_regress(t, source.Version, GPHOME_SOURCE, PGPORT, sourceTestDir, "migratable_source_schedule")
+		isolation2_regress(t, source.Version, GPHOME_SOURCE, PGPORT, sourceTestDir, sourceTestDir, "migratable_source_schedule", false)
 
 		generate(t, migrationDir)
 		apply(t, GPHOME_SOURCE, PGPORT, idl.Step_initialize, migrationDir)
 
 		initialize(t, idl.Mode_link)
 		execute(t)
+		time.Sleep(5 * time.Second)
```
kalensk, line 58: Need to remove these sleep's

test/acceptance/migratable_test.go
```diff
@@ -33,28 +36,98 @@ func TestMigrationScripts(t *testing.T) {
...
 	testDir := filepath.Join(MustGetRepoRoot(t), "test", "acceptance", dir)
-	migratableTestDir := filepath.Join(testDir, "migratable_tests")
+	sourceTestDir := filepath.Join(testDir, "migratable_tests", "source_cluster_regress")
+	targetTestDir := filepath.Join(testDir, "migratable_tests", "target_cluster_regress")
 
 	testutils.MustApplySQLFile(t, GPHOME_SOURCE, PGPORT, filepath.Join(testDir, "setup_globals.sql"))
 	defer testutils.MustApplySQLFile(t, GPHOME_SOURCE, PGPORT, filepath.Join(testDir, "teardown_globals.sql"))
 
 	t.Run("migration scripts generate sql to modify non-upgradeable objects and fix pg_upgrade check errors", func(t *testing.T) {
+		defer killServices(t)
```
kalensk, line 46: We can remove this.

test/acceptance/acceptance_suite_test.go
```diff
@@ -308,9 +312,9 @@ func restoreDemoCluster(t *testing.T, backupDir string, source greenplum.Cluster
 	}
 }
 
-func isolation2_regress(t *testing.T, sourceVersion semver.Version, gphome string, port string, testDir string, schedule string) string {
+func isolation2_regress(t *testing.T, sourceVersion semver.Version, gphome string, port string, testDir string, outputTestDir string, schedule string, useExisting bool) string {
 	var cmdArgs []string
-	if gphome == GPHOME_TARGET {
+	if useExisting {
```
kalensk, line 317: Can't we just always use `"--use-existing" and add that to the default `cmdArgs`? I think it should be fine and keeps things simple.

test/acceptance/acceptance_suite_test.go
```diff
@@ -163,11 +166,12 @@ func finalize(t *testing.T) string {
 	return strings.TrimSpace(string(output))
 }
 
-func revert(t *testing.T) string {
+func revert(t *testing.T, cmdEnv ...string) string {
 	t.Helper()
 
 	cmd := exec.Command("gpupgrade", "revert",
 		"--non-interactive", "--verbose")
+	cmd.Env = cmdEnv
```
kalensk, line 174: This needs to be removed. If there is an issue it needs to be fixed at the root cause.

test/acceptance/acceptance_suite_test.go
kyeap-vmware: `use-existing` prevents isolation2 from dropping the isolation2test database before running. If we want change this to be a default. We'll need each pg_upgrade test to drop database `isolation2test`.

test/acceptance/acceptance_suite_test.go
kyeap-vmware: This isn't a fix. The `migration scripts ignore .psqlrc files` test is specifically testing that gpupgrade doesn't use psqlrc files specified in the PSQLRC environment variable. 

Removing the use of cmd.Env from the helper function means that this test won't be able to use the generate, apply, initialize, revert helper functions. The alternative is code duplication of these 4 functions this test.

test/acceptance/migratable_test.go
kyeap-vmware: This is to cleanup the hub in case initialize, or execute fails. Instead of this we can put the revert back in.

kalensk: As already mentioned offline its not needed since finalize and revert already do that. If those fails that means there is a bug in the underlying revert/finalize code which we want to see and fix.

test/acceptance/acceptance_suite_test.go
kalensk: The cmd.Env is polluting all the core helper functions for a single test. That is a hard no go by me for code maintaince/readablity/etc. If anything that test could be done as a unit test perhaps and not need to change all those core helpers.

test/acceptance/acceptance_suite_test.go
kalensk: Can you elaborate on which specific tests need useExisting since I am confident there is a cleaner way of doing this. As you can see isolation2_regress has a TON of parameters. This is totally a code smell and not the direction we want to keep going. See "long parameter list" in Martin Fowlers awesome book Refactoring.

```
Long Parameter List
A very long list of parameters indicates the method or class has taken on too many responsibilities. It is advisable to break it down into several components, each with a clear cut role.

Why it’s bad: Hurts code readability and reusability
```

kyeap-vmware: Sure! Every test that runs isolation2 twice needs `--use-existing`. This flag tells the isolation2 not to reset the isolation2test database before running tests. Before, gphome was used to determine if isolation2 was running a before or after an upgrade attempt. This worked because running isolation2 the second time happened on either the intermediate or target cluster. The migratable revert tests now introduces tests where the second run of isolation2 runs on the original source cluster (to verify migratable objects are brought back upon revert). The original method of implicitly determining if isolation2 is the second run by checking if gphome is not source is now no longer viable. Since this is a wrapper function I would say it's not surprising that if we need more control over the isolation2 flags, then the number of arguments of its wrapper increases as needed.

test/acceptance/migratable_test.go
kyeap-vmware: As discussed I changed back to original revert method. We will have another commit to change all `defer revert` to be called before initialize so revert will be queued up to clean up test correctly if initialize fails.

test/acceptance/acceptance_suite_test.go
kyeap-vmware: It looks like we should be able to remove this test. There is a unit test that ensures `--no-psql` flag is used.
```
	t.Run("executes sql scripts with correct arguments for the correct database", func(t *testing.T) {
		commanders.SetPsqlFileCommand(exectest.NewCommandWithVerifier(SuccessScript, func(utility string, args ...string) {
			expectedUtility := "/usr/local/gpdb5/bin/psql"
			if utility != expectedUtility {
				t.Errorf("got %q want %q", utility, expectedUtility)
			}

			additionalArgs := args[:6]
			expected := []string{"-v", "ON_ERROR_STOP=1", "--no-align", "--tuples-only", "--no-psqlrc", "--quiet"}
			if !reflect.DeepEqual(additionalArgs, expected) {
				t.Errorf("got args %q want %q", additionalArgs, expected)
			}

```

test/acceptance/acceptance_suite_test.go
kyeap-vmware: arguments were put into an options struct.

test/acceptance/acceptance_suite_test.go
kalensk: The test name should probably be updated to emphasize --no-psqlrc as that is clear documentation to future developers to not remove it or refactor it.

Maybe something like: `"executes sql scripts with correct arguments including --no-psqlrc for the correct database"`

kalensk (approved):

PR #879, opened 2023-09-28
Fix failing tablespace test during finalize

bmdoil:
When running the test
"in link mode gpupgrade finalize should also delete mirror directories and honors --use-hba-hostnames" an error would occur when trying to query tables in a user created tablespace.

The body of the testFinalize function has several deferred functions to clean up the test environment. These will be executed in last in first out (LIFO) order.

Currently this means that the tablespace cleanup functions are run in this order:
1. MustDeleteTablespaces
2. MustRemoveAll(t, tablespaceDir)
3. MustExecuteSQL -- Drops table 'should_be_reverted' created in tablespace

Since we remove the tablespaces and tablespace directory first, the SQL to drop 'should_be_reverted' will fail.

Fix: Remove the calls to MustDeleteTablespaces and 'should_be_reverted' tests.

MustRemoveAll(t, tablespaceDir) takes care of cleaning up the temp tablespace directory and the defer call to restoreDemoCluster will return the state of the source cluster to before the test was run after testFinalize returns or there is a Fatalf called. We don't need to explicity drop the tablespace.

Because restoreDemoCluster is always called after testFinalize runs, creating and dropping the 'should_be_reverted' table within testFinalize isn't necessary.

Authored-by: Brent Doil <bdoil@vmware.com>

### DIFF
--- a/test/acceptance/finalize_test.go
+++ b/test/acceptance/finalize_test.go
@@ -53,15 +53,10 @@ func testFinalize(t *testing.T, mode idl.Mode, useHbaHostnames bool) {
 	createMarkerFilesOnAllSegments(t, source)
 	defer removeMarkerFilesOnAllSegments(t, source)
 
-	table := "public.should_be_reverted"
-	testutils.MustExecuteSQL(t, source.Connection(greenplum.Database("postgres")), fmt.Sprintf(`CREATE TABLE %s (a int); INSERT INTO %s VALUES (1), (2), (3);`, table, table))
-	defer testutils.MustExecuteSQL(t, source.Connection(greenplum.Database("postgres")), fmt.Sprintf(`DROP TABLE %s;`, table))
-
 	tablespaceDir := testutils.GetTempDir(t, "")
 	defer testutils.MustRemoveAll(t, tablespaceDir)
 
 	testutils.MustAddTablespace(t, source, tablespaceDir)
-	defer testutils.MustDeleteTablespaces(t, source, GetTempTargetCluster(t))
 
 	hbaHostnames := ""
 	if useHbaHostnames {
@@ -96,33 +91,14 @@ func testFinalize(t *testing.T, mode idl.Mode, useHbaHostnames bool) {
 
 	verifyMarkerFilesOnAllSegments(t, conf.Intermediate, conf.Target)
 
-	rows := testutils.MustQueryRow(t, conf.Target.Connection(greenplum.Database("postgres")), fmt.Sprintf(`SELECT COUNT(*) FROM %s;`, table))
-	expectedRows := 3
-	if rows != expectedRows {
-		t.Fatalf("got %d want %d rows", rows, expectedRows)
-	}
-
 	testutils.VerifyTablespaceData(t, *conf.Target)
 
-	// FIXME: When this script is run in link mode then the above call to remove tablespaces fails with:
-	//   ERROR: failed to acquire resources on one or more segments (SQLSTATE 58M01)
-	// Specifically, it errors when accessing objects in a database created in a tablespace. That is, accessing a table
-	// from a database created in a tablespace when this test is run in link mode. For example, connecting to foodb and
-	// doing \d+ errors with:
-	//   ERROR:  failed to acquire resources on one or more segments
-	//   DETAIL:  FATAL:  could not read relation mapping file "pg_tblspc/16389/GPDB_6_301908232/16487/pg_filenode.map": Success (relmapper.c:660)
-	// Connecting to one of the segments errors with:
-	//   PGOPTIONS="-c gp_session_role=utility" psql -p 25432 foodb
-	//   psql: FATAL:  could not read relation mapping file "pg_tblspc/16389/GPDB_6_301908232/16487/pg_filenode.map": Success (relmapper.c:660)
-	// Inspecting the tablespace location is extremely chaotic especially after upgrading!
-	if mode != idl.Mode_link {
-		path := filepath.Join(MustGetRepoRoot(t), "test", "acceptance", "helpers", "finalize_checks.bash")
-		script := fmt.Sprintf("source %s; validate_mirrors_and_standby %s %s %s", path, GPHOME_TARGET, conf.Target.CoordinatorHostname(), PGPORT)
-		cmd = exec.Command("bash", "-c", script)
-		output, err = cmd.CombinedOutput()
-		if err != nil {
-			t.Fatalf("unexpected err: %#v stderr %s", err, output)
-		}
+	path := filepath.Join(MustGetRepoRoot(t), "test", "acceptance", "helpers", "finalize_checks.bash")
+	script := fmt.Sprintf("source %s; validate_mirrors_and_standby %s %s %s", path, GPHOME_TARGET, conf.Target.CoordinatorHostname(), PGPORT)
+	cmd = exec.Command("bash", "-c", script)
+	output, err = cmd.CombinedOutput()
+	if err != nil {
+		t.Fatalf("unexpected err: %#v stderr %s", err, output)
 	}
 }
 


test/acceptance/finalize_test.go
```diff
@@ -96,33 +91,14 @@ func testFinalize(t *testing.T, mode idl.Mode, useHbaHostnames bool) {
 
 	verifyMarkerFilesOnAllSegments(t, conf.Intermediate, conf.Target)
 
-	rows := testutils.MustQueryRow(t, conf.Target.Connection(greenplum.Database("postgres")), fmt.Sprintf(`SELECT COUNT(*) FROM %s;`, table))
-	expectedRows := 3
-	if rows != expectedRows {
-		t.Fatalf("got %d want %d rows", rows, expectedRows)
-	}
-
```
kalensk, line 104: Im am confused. We should not be removing some of the core pieces of the test such as the
`CREATE TABLE %s (a int); INSERT INTO %s VALUES (1), (2), (3);`` and the verification check here.

The purpose of this finalize test is to do a simple yet exhaustive test of upgrade end-to-end.

test/acceptance/finalize_test.go
```diff
@@ -53,15 +53,10 @@ func testFinalize(t *testing.T, mode idl.Mode, useHbaHostnames bool) {
...
 
-	table := "public.should_be_reverted"
-	testutils.MustExecuteSQL(t, source.Connection(greenplum.Database("postgres")), fmt.Sprintf(`CREATE TABLE %s (a int); INSERT INTO %s VALUES (1), (2), (3);`, table, table))
-	defer testutils.MustExecuteSQL(t, source.Connection(greenplum.Database("postgres")), fmt.Sprintf(`DROP TABLE %s;`, table))
-
 	tablespaceDir := testutils.GetTempDir(t, "")
 	defer testutils.MustRemoveAll(t, tablespaceDir)
 
 	testutils.MustAddTablespace(t, source, tablespaceDir)
-	defer testutils.MustDeleteTablespaces(t, source, GetTempTargetCluster(t))
```
kalensk, line 64: If we end up removing this `testutils.MustDeleteTablespaces` in the final implementation than there needs to be a comment such as:

```go
testutils.MustAddTablespace(t, source, tablespaceDir) // cleanup is handled by restoreDemoCluster()
```

kalensk:
Part of me thinks there might be another approach. Now that we have identified the issue, can we think of several different types of solutions/implementations to the issue. This PR is just one particular solution, but I am sure their are others each with their own pros and cons. I am hoping/confident there is a solution that might be cleaner.

kalensk:
Given the PR description of the issue it is still not clear to me why this is failing in link mode, but not copy mode. Maybe I just need to think more, but it would be nice to make that crystal clear in the description/commit.

kyeap-vmware (commented):
The regular table `should_be_revert` shouldn't be affected by the tablesapce cleanup, but I think it's an indicator showing that something upstream of the defer statement is not behaving as we expect. We should check the data after upgrade happens, it looks like that portion in VerifyTablespaceData is disabled.

kalensk:
Just to be clear everything in the code was done with intention and on purpose. If things aren't clear I encourage curiosity and discussion to first understand.

> Is the existence of the table should_be_revert to show that regular tables outside of tablespaces are unaffected?

As mentioned in a previous comment in this thread. The purpose is to show that a simple table and a simple tablespace can be upgraded and run through finalize.

> On second look. This entire finalize test seems like it should eventually be broken up into about a dozen smaller tests.
Just scrolling through I can see that this 1 test is...

Correct. This was again done on purpose and with intention. For example, splitting this up will exponentially increase test time. Our acceptance tests already take close to an hour to run. Splitting it up would make that unbearable. We do not want to loose test coverage while still minimizing test time.

kyeap-vmware:
>Can you elaborate on this, because I am not seeing that?

One of the first lines in VerifyTablespaceData is to return immediately if the cluster it is running against is >= 6. Doesn't seem like that is what we want based on the test.
```
func VerifyTablespaceData(t *testing.T, cluster greenplum.Cluster) {
	t.Helper()

	if cluster.Version.Major >= 6 {
		return
	}
...
}
```

bmdoil:
>Part of me thinks there might be another approach. Now that we have identified the issue, can we think of several different types of solutions/implementations to the issue?

I spent more time on debugging the issue and believe I've hit the actual issue. It's not in fact a test issue, I misunderstood the order of operations with the test cleanup. 

The issue is with failover and pg_rewind exercised in the finalize_checks.bash script. 

We fail at step 3 below
```
# |     step                        | cdw         | cmdw        | sdw-p   | sdw-m   |
# |---------------------------------|-------------|-------------|---------|---------|
# | 1:  initial                     | coordinator |   standby   | primary | mirror  |
# | 2a: failover stop               |     -       |   standby   |   -     | mirror  |
# | 2b: failover promote            |     -       | coordinator |   -     | primary |
# | 3:  restore mirrors and standby |   standby   | coordinator | mirror  | primary |<===== fail!
```
Repro:

1. Create filespace/tablespace and table with 100 rows in source cluster
2. upgrade using link mode
3. pg_ctl -m immediate -D primarydatadir stop to trigger failover
4. pg_rewind

The failure manifests during test cleanup when we try to drop the table created in database `foodb` which resides in the tablespace. This results in 0 size pg_filenode.map/*_fsm/*_vm files in databases created in a tablespace. The local pg_filenode.map is checked during backend startup (i.e. when connecting to a database from client) and we fall over in `RelationMapInitializePhase3`.

Before pg_rewind:
```
± bd |tag:5.29.12 ?:5 ✗| → find -L . -name "pg_filenode.map" -exec ls -l {} \; | grep "pg_tbls" | grep demoDataDir0
-rw------- 1 bdoil bdoil 512 Sep 29 02:36 ./datadirs/dbfast1/demoDataDir0/pg_tblspc/16388/GPDB_6_301908232/16486/pg_filenode.map
-rw------- 1 bdoil bdoil 512 Sep 29 02:36 ./datadirs/dbfast1/demoDataDir0/pg_tblspc/16388/GPDB_6_301908232/16485/pg_filenode.map
-rw------- 1 bdoil bdoil 512 Sep 29 02:36 ./datadirs/dbfast_mirror1/demoDataDir0/pg_tblspc/16388/GPDB_6_301908232/16486/pg_filenode.map
-rw------- 1 bdoil bdoil 512 Sep 29 02:36 ./datadirs/dbfast_mirror1/demoDataDir0/pg_tblspc/16388/GPDB_6_301908232/16485/pg_filenode.map
```
After pg_rewind:
```
± bd |tag:5.29.12 ?:5 ✗| → find -L . -name "pg_filenode.map" -exec ls -l {} \; | grep "pg_tbls" | grep demoDataDir0
-rw------- 1 bdoil bdoil 0 Sep 29 02:41 ./datadirs/dbfast1/demoDataDir0/pg_tblspc/16388/GPDB_6_301908232/16486/pg_filenode.map
-rw------- 1 bdoil bdoil 0 Sep 29 02:41 ./datadirs/dbfast1/demoDataDir0/pg_tblspc/16388/GPDB_6_301908232/16485/pg_filenode.map
-rw------- 1 bdoil bdoil 0 Sep 29 02:41 ./datadirs/dbfast_mirror1/demoDataDir0/pg_tblspc/16388/GPDB_6_301908232/16486/pg_filenode.map
-rw------- 1 bdoil bdoil 0 Sep 29 02:41 ./datadirs/dbfast_mirror1/demoDataDir0/pg_tblspc/16388/GPDB_6_301908232/16485/pg_filenode.map
```
fsm and vm files also affected
```
± bd |tag:5.29.12 ?:5 ✗| → find -L . -name "12736_fsm" -exec ls -l {} \; | grep dbfast1
-rw------- 1 bdoil bdoil 98304 Sep 29 02:41 ./datadirs/dbfast1/demoDataDir0/base/1/12736_fsm
-rw------- 1 bdoil bdoil 98304 Sep 29 02:41 ./datadirs/dbfast1/demoDataDir0/base/12812/12736_fsm
-rw------- 1 bdoil bdoil 98304 Sep 29 02:41 ./datadirs/dbfast1/demoDataDir0/base/12809/12736_fsm
-rw------- 1 bdoil bdoil 0 Sep 29 02:41 ./datadirs/dbfast1/demoDataDir0/pg_tblspc/16388/GPDB_6_301908232/16486/12736_fsm
-rw------- 1 bdoil bdoil 0 Sep 29 02:41 ./datadirs/dbfast1/demoDataDir0/pg_tblspc/16388/GPDB_6_301908232/16485/12736_fsm
    
± bd |tag:5.29.12 ?:5 ✗| → find -L . -name "12736_vm" -exec ls -l {} \; | grep dbfast1
-rw------- 1 bdoil bdoil 32768 Sep 29 02:41 ./datadirs/dbfast1/demoDataDir0/base/1/12736_vm
-rw------- 1 bdoil bdoil 32768 Sep 29 02:41 ./datadirs/dbfast1/demoDataDir0/base/12812/12736_vm
-rw------- 1 bdoil bdoil 32768 Sep 29 02:41 ./datadirs/dbfast1/demoDataDir0/base/12809/12736_vm
-rw------- 1 bdoil bdoil 0 Sep 29 02:41 ./datadirs/dbfast1/demoDataDir0/pg_tblspc/16388/GPDB_6_301908232/16486/12736_vm
-rw------- 1 bdoil bdoil 0 Sep 29 02:41 ./datadirs/dbfast1/demoDataDir0/pg_tblspc/16388/GPDB_6_301908232/16485/12736_vm
```

pg_rewind debug logs show the transfer of 0 byte files
```
received chunk for file "pg_tblspc/16388/GPDB_6_301908232/16485/12796_fsm", offset 0, size 0
received chunk for file "pg_tblspc/16388/GPDB_6_301908232/16485/12796_vm", offset 0, size 0
received chunk for file "pg_tblspc/16388/GPDB_6_301908232/16485/pg_filenode.map", offset 0, size 0
```
with link mode primary and mirror point to the same inode for pg_filenode.map
```
11626900 -rw------- 1 bdoil bdoil 512 Sep 29 04:20 ./dbfast1/demoDataDir0/pg_tblspc/16388/GPDB_6_301908232/16485/pg_filenode.map
11626900 -rw------- 1 bdoil bdoil 512 Sep 29 04:20 ./dbfast_mirror1/demoDataDir0/pg_tblspc/16388/GPDB_6_301908232/16485/pg_filenode.map
```
copy mode they don't
```
11633625 -rw------- 1 bdoil bdoil 512 Sep 29 04:33 ./dbfast1/demoDataDir0/pg_tblspc/16388/GPDB_6_301908232/16485/pg_filenode.map
11634369 -rw------- 1 bdoil bdoil 512 Sep 29 04:34 ./dbfast_mirror1/demoDataDir0/pg_tblspc/16388/GPDB_6_301908232/16485/pg_filenode.map
```
pg_rewind works as expected in copy mode
`received chunk for file "pg_tblspc/16388/GPDB_6_301908232/16485/pg_filenode.map", offset 0, size 512`

Based on this evidence, RCA is pg_rewind truncates the target file before copying from the source, since they're the same hard link we end up with a 0 length file.

`pg_tblspc/16388/GPDB_6_301908232/16485/pg_filenode.map (COPY)   <=== `

```
case FILE_ACTION_COPY:
    rewind_copy_file_range(entry->path, 0, entry->source_size, true);
    break;
```

```
/*
 * Copy a file from source to target, between 'begin' and 'end' offsets.
 *
 * If 'trunc' is true, any existing file with the same name is truncated.
 */
static void
rewind_copy_file_range(const char *path, off_t begin, off_t end, bool trunc)
```

I'm going to close this PR since it does not resolve anything. Generally speaking I agree with most of your points. I do think there can be improvements to this test in particular w.r.t clarity and comments, but it'll be more efficient to hash those out offline.

A new PR that resolves the issue will be forthcoming.

PR #885, opened 2023-10-03
Remove migration_scripts.bats

kyeap-vmware:
All migration script tests have been ported to golang isolation2. We don't need the bats test anymore!

pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:remove-bats-migration


kalensk:
Can this PR also remove all references to `bats-core` and 
```
    echo "Installing BATS..."
    ./bats/install.sh /usr/local
```

And the README.md

kalensk (approved):

kalensk:
Awesome. A followup PR or whatever can likely remove the `validate_mirrors_and_standby` directory now that there is just the one file. That is, from
`testutils/validate_mirrors_and_standby/validate_mirrors_and_standby.bash`
to something like:
`testutils/validate_mirrors_and_standby.bash`

PR #886, opened 2023-10-04
Expand gp_upgrade_upgradable_test to exercise finalize

kyeap-vmware:
The typical customer workflow is to run initialize, execute, finalize
and run finalize migration scripts. This change add the two steps to run
finalize and finalize migration scripts as part of the test.
    
In reference to the test output change.
A bitmap of pages can be created dynamically by the planner and is
then discarded at the end of the bitmap index scan (not to be confused
with a BRIN index). Before, the planner was generating this bitmap for
the btree index table_with_name_columns_idx. After rebuilding statistics
the planner chooses a standard Index Scan. This is expected. Since the
finalize rebuilds statistics, the query plan result changed is expected.

pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:upgradeable-finalize


test/acceptance/5-to-6/upgradeable_tests/target_cluster_regress/expected/name_data_type.out
```diff
@@ -28,13 +28,11 @@ SET
...
-         Recheck Cond: (b = 'bbb8'::name)                     
-         ->  Bitmap Index Scan on table_with_name_columns_idx 
-               Index Cond: (b = 'bbb8'::name)                 
- Optimizer: Postgres query optimizer                          
-(6 rows)
+ Gather Motion 3:1  (slice1; segments: 3)                                      
+   ->  Index Scan using table_with_name_columns_idx on table_with_name_columns 
+         Index Cond: (b = 'bbb8'::name)                                        
+ Optimizer: Postgres query optimizer                                           
+(4 rows)
```
bmdoil, line 35: To expand on this, a bitmap of pages can be created dynamically by the planner and is then discarded at the end of the bitmap index scan (not to be confused with a BRIN index). Before, the planner was generating this bitmap for the btree index `table_with_name_columns_idx`. After rebuilding statistics the planner chooses a standard `Index Scan`. This is expected.

test/acceptance/pg_upgrade_upradeable_test.go
```diff
@@ -32,14 +38,23 @@ func Test_PgUpgrade_Upgradeable_Tests(t *testing.T) {
...
+
 		sourceTestDir := filepath.Join(testDir, "upgradeable_tests", "source_cluster_regress")
 		acceptance.Isolation2_regress(t, source.Version, acceptance.GPHOME_SOURCE, acceptance.PGPORT, sourceTestDir, sourceTestDir, idl.Schedule_upgradeable_source_schedule)
 
+		acceptance.Generate(t, migrationDir)
+		acceptance.Apply(t, acceptance.GPHOME_SOURCE, acceptance.PGPORT, idl.Step_initialize, migrationDir)
+
 		acceptance.Initialize(t, idl.Mode_link)
-		defer acceptance.Revert(t)
+		defer revertIgnoreFailures(t)
```
bmdoil, line 51: `revertIgnoreFailures` is being moved to `acceptance.RevertIgnoreFailures(t)` by https://github.com/greenplum-db/gpupgrade/pull/887 so we'll need to rebase.

bmdoil (commented):
>Make the test more inline with what a customer workflow would look like.

Can you add a brief description of the workflow being exercised by `Test_PgUpgrade_Upgradeable_Tests` as a comment to the function so future us can refer to that instead of having to scan through the function?
1. Does the thing
2. Does the other thing
3. Profit

kalensk:
> as a comment to the function 

Just as a overall note in general.

I think its best to actually avoid comments as it forces one to write clear, clean code that is self-documenting. Doing so makes the code super easy to follow and avoids the need for simple comments which can quickly become stale, false, misleading, or not effective. However, there is a time and place for comments, but its fairly rare and usually reserved for when things are not obvious and more explaining "why" or a breadcrumb for future developers. Again needing those are somewhat rare. Maybe take a look at the current code base to get a feel or idea of this concept. I am probably biased, but its served me really well.

test/acceptance/pg_upgrade_upgradeable_test.go
```diff
@@ -32,14 +38,23 @@ func Test_PgUpgrade_Upgradeable_Tests(t *testing.T) {
...
+
 		sourceTestDir := filepath.Join(testDir, "upgradeable_tests", "source_cluster_regress")
 		acceptance.Isolation2_regress(t, source.Version, acceptance.GPHOME_SOURCE, acceptance.PGPORT, sourceTestDir, sourceTestDir, idl.Schedule_upgradeable_source_schedule)
 
+		acceptance.Generate(t, migrationDir)
+		acceptance.Apply(t, acceptance.GPHOME_SOURCE, acceptance.PGPORT, idl.Step_initialize, migrationDir)
+
 		acceptance.Initialize(t, idl.Mode_link)
-		defer acceptance.Revert(t)
+		defer revertIgnoreFailures(t)
```
kalensk, line 51: This `revertIgnoreFailrues` needs a comment. Or  better yet add it to the top of its function definition. Something roughly like:

```
revertIgnoreFailures should primarily be used when BackupDemoCluster and RestoreDemoCluster are used to revert the cluster after finalize. We still want to clean up if the test fails part way through hence we need to run revert while ignoring failures.
```

bmdoil: It's not clear to me why this line needs a comment but we want to avoid a comment explaining the intention of the test driver `Test_PgUpgrade_Upgradeable_Tests`

kalensk: Happy to explain! 

Its incredibly odd to ignore failures. That is super unusual event and unexpected. None of our code really does that. Hence the desire for a comment in the function definition/header to explain "why" that is necessary and "when" to use that function over `revert(t)`.

It can actually be quite harmful to "simple" comments that are mostly just repeating what the clearly laid out code does. Now if the comment is explaining "why", or explaining something unexpected, or something something not fairly obvious that is likely beneficial. But those are fairly rare cases. The reason why having "plain" comments are harmful is because code tends to become much less clean/clear/effective as people use the comment as a crutch. It also quickly leads to misleading code as the comment becomes stale. Rather its useful to let the code lead and drive clarity by not using comments except in rare cases.

I can see how not having comments "goes against the grain" and might not make sense. But its truly been working wonderfully for this code base for many many years through many individuals rotating in and out. I would encourage one to be curious and open to trying it for several months or more.

bmdoil: > It can actually be quite harmful to have "simple" comments that are mostly just repeating what the clearly laid out code does.

Obviously the holy grail is to have self-documenting code. Comments that repeat the code is an anti-pattern. We'll agree to disagree that the test driver is self documenting.

> I can see how not having comments "goes against the grain" and might not make sense. But its truly been working wonderfully for this code base for many many years through many individuals rotating in and out. I would encourage one to be curious and open to trying it for several months or more.

Frankly this is an incredibly condescending statement. 

Kevin, feel free to reword the existing comment we have for `revertIgnoreFailures(t)` to make it more clear if you think it's necessary. Otherwise, LGTM.

bmdoil (approved):

test/acceptance/pg_upgrade_upgradeable_test.go
kalensk: I am sorry that came across that way. That was not my intention. Hopefully, we can sync offline to clarify as I am sure there is a middle ground. 

Maybe a another approach is to see what part of the code is not clear where a comment is useful for you, so that we can strive to make make that part of the code better.

test/acceptance/5-to-6/upgradeable_tests/target_cluster_regress/expected/name_data_type.out
kyeap-vmware: Awesome description of what is going on. I'll add this to the commit's message.

test/acceptance/pg_upgrade_upgradeable_test.go
bmdoil: We synced offline and clarified some misunderstandings. We updated the commit message to cover that we're adding finalize and finalize migration scripts steps to the acceptance test.

PR #887, opened 2023-10-04
split up gpupgrade vs. pg_upgrade acceptance tests

bmdoil:
pipeline: https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:createTestSubDirs


kalensk:
Looks like needs a tweak to the path....

`ci/main/scripts/acceptance-multihost-tests.bash`
```
 cd gpupgrade_src
go test --cover -count=1 -timeout 30m -v -run "^TestRevert$" ./test/acceptance/gupgrade
```

As a side note this is kinda wonky and we should maybe think of a better way of running the multihost tests. Ideally they would all just be run multihost rather than just revert...

bmdoil:
> As a side note this is kinda wonky and we should maybe think of a better way of running the multihost tests. Ideally they would all just be run multihost rather than just revert.

I'll create a story to follow up on this, we definitely want to run all the tests multihost.

kyeap-vmware (approved):
LGTM

PR #892, opened 2023-10-18
enable 6 > 7 and 7 > 7 ci jobs

kyeap-vmware:
The purpose of this is to enable the 6 > 7 and the 7 > 7 jobs without
breaking the current 5 > 6 jobs. Following commits to start making
tweaks so the 6 > 7 pg_upgrade tests are able to run. We expect them to
fail, but we need them to be in a state where ci at least runs them.

new pipeline can be stood up using 
`make pipeline7`

pipelines:
https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:7job-5-to-6-still-works
https://cm.ci.gpdb.pivotal.io/teams/main/pipelines/gpupgrade:7job-6-to-7-pipeline

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -194,19 +194,19 @@ FLY_TARGET ?= cm
 # environment variable JOB_TYPE is used to determine whether a dev or prod
 # pipeline is generated. It is used when go generate runs our yaml parser.
 ifeq ($(FLY_TARGET),prod)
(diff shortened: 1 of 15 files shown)


bmdoil (approved):
LGTM, the added conditionals in the pipeline templates seem in line with our existing patterns.

PR #896, opened 2023-12-06
nonupgradable test to look for tables that use system defined types

kyeap-vmware:
System defined types are not guaranteed to have consistent oids when moving from on version of postgres to the next. The query that looks for these types had to be rewritten for upgrade because the recursive query looking for these types of relations contained a self reference in a subquery. This specific type of query is disabled in 6x so it was rewritten in pgpsql.

gpdb reference commit:
https://github.com/greenplum-db/gpdb/pull/16820

### DIFF
--- a/test/acceptance/pg_upgrade/6-to-7/non_upgradeable_tests/sql/system_defined_types.sql
+++ b/test/acceptance/pg_upgrade/6-to-7/non_upgradeable_tests/sql/system_defined_types.sql
@@ -1,0 +1,76 @@
+-- Copyright (c) 2017-2023 VMware, Inc. or its affiliates
+-- SPDX-License-Identifier: Apache-2.0
+
+-- System defined types are not guaranteed to have consistent oids when moving
+-- from on version of postgres to the next. The query that looks for these
+-- types had to be rewritten for upgrade because the recursive query looking
+-- for these types of relations contained a self reference in a subquery. This
+-- specific type of query is disabled in 6x so it was rewritten in plpgsql.
+
+--------------------------------------------------------------------------------
+-- Create and setup non-upgradeable objects
+--------------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


test/acceptance/pg_upgrade/6-to-7/non_upgradeable_tests/sql/system_defined_types.sql
```diff
@@ -0,0 +1,60 @@
...
+-- specific type of query is disabled in 6x so it was rewritten in pgpsql.
+
+--------------------------------------------------------------------------------
+-- Create and setup non-upgradeable objects
+--------------------------------------------------------------------------------
+-- table that uses system-defined type pg_type
+CREATE TABLE foo (
+    my_id SERIAL PRIMARY KEY,
+    type_info pg_type
+);
```
kalensk, line 17: Curious if it would it make sense to rename the table such that the comment can be removed such as something kind of like: `table_with_system_defined_type`?

Similar for the other comments and objects where appropriate such as `table_with_custom_type` and `table_with_custom_types_dependent_on_system_types`. Again depending exact wording and how complicated things may be.

kyeap-vmware: oh you're completely right. Let me update the table names to be more descriptive and helpful than just foo or foo2.

bmdoil (approved):
LGTM!

kyeap-vmware:
Blocked from merging until plpgsql PR https://github.com/greenplum-db/gpdb/pull/16820 is resolved. There is talks of backporting recursive CTE with self reference subqueries fix from 7X to 6X that is blocking plpgsql PR.

kyeap-vmware:
Recursive CTE fixes will not be backported to 6X. plpgsql fix has been merged. unblocked.

PR #899, opened 2024-01-17
Data type checks for reg*, sql_identifier, and unknown

kyeap-vmware:
The query that looks for these types had to be rewritten for 6 > 7 upgrade
because the recursive query looking for these types of relations contained a
self reference in a subquery. This specific type of query is disabled in 6x
so it was rewritten in plpgsql.


bmdoil (approved):
LGTM.

nit in commit message for `Add pg_upgrade nonupgradable test to detect column type 'unknown'` 

`column types of unknown no upgradable` => `column types of unknown non `upgradable

PR #900, opened 2024-02-13
pg_upgrade nonupgradable test detecting views with removed operators

kyeap-vmware:
Views that use deprecated operators will cause upgrade to fail. This happens during metadata restore on the target cluster because pg_restore will error trying to create a view using types that do not exist anymore. This is not ideal as we could be many hours into upgrade before a view using a deprecated operator causes pg_upgrade to fail. Test that pg_upgrade is looking for views using deprecated operators.


bmdoil (approved):
LGTM

PR #902, opened 2024-03-05
pg_upgrade nonupgradable test detecting views with removed functions

kyeap-vmware:
Views that use removed functions will cause upgrade to fail. This happens during metadata restore on the target cluster because pg_restore will error trying to create a view using functions that do not exist anymore. This is not ideal as we could be many hours into upgrade before a view using a removed function causes pg_upgrade to fail. This check calls a support function on the source cluster to check if views using removed functions exist.

https://github.com/greenplum-db/gpupgrade/pull/902
6x: https://github.com/greenplum-db/gpdb/pull/17152
7x: https://github.com/greenplum-db/gpdb/pull/17153

### DIFF
--- a/test/acceptance/pg_upgrade/6-to-7/non_upgradeable_tests/expected/views_with_removed_functions.out
+++ b/test/acceptance/pg_upgrade/6-to-7/non_upgradeable_tests/expected/views_with_removed_functions.out
@@ -1,0 +1,101 @@
+-- Copyright (c) 2017-2024 VMware, Inc. or its affiliates
+-- SPDX-License-Identifier: Apache-2.0
+
+--------------------------------------------------------------------------------
+-- Create and setup non-upgradeable objects
+--------------------------------------------------------------------------------
+
+DROP SCHEMA IF EXISTS removed_functions CASCADE;
+DROP SCHEMA
+CREATE SCHEMA removed_functions;
+CREATE SCHEMA
+SET search_path to removed_functions;
+SET
+
+CREATE TABLE ao_table(i int) with (appendonly=true);
+CREATE TABLE
+CREATE TABLE rank_table (id int, rank int, year int, gender char(1), count int ) DISTRIBUTED BY (id) PARTITION BY LIST (gender) ( PARTITION girls VALUES ('F'), PARTITION boys VALUES ('M'), DEFAULT PARTITION other );
+CREATE TABLE
+
+CREATE VIEW v01 AS SELECT pg_catalog.pg_current_xlog_insert_location();
+CREATE VIEW
+CREATE VIEW v02 AS SELECT pg_catalog.pg_current_xlog_location();
+CREATE VIEW
+CREATE VIEW v03 AS SELECT pg_catalog.gp_update_ao_master_stats('ao_table'::regclass);
+CREATE VIEW
+CREATE VIEW v04 AS SELECT pg_get_partition_def('rank_table'::regclass);
+CREATE VIEW
+CREATE VIEW v05 AS SELECT pg_get_partition_def('rank_table'::regclass, true);
+CREATE VIEW
+CREATE VIEW v06 AS SELECT pg_get_partition_def('rank_table'::regclass, true, true);
+CREATE VIEW
+CREATE VIEW v07 AS SELECT pg_get_partition_rule_def('rank_table'::regclass);
+CREATE VIEW
+CREATE VIEW v08 AS SELECT pg_get_partition_rule_def('rank_table'::regclass, true);
+CREATE VIEW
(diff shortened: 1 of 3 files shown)


test/acceptance/pg_upgrade/6-to-7/non_upgradeable_tests/sql/views_with_removed_functions.sql
```diff
@@ -0,0 +1,54 @@
...
+CREATE VIEW v04 AS SELECT pg_get_partition_def('rank_table'::regclass);
+CREATE VIEW v05 AS SELECT pg_get_partition_def('rank_table'::regclass, true);
+CREATE VIEW v06 AS SELECT pg_get_partition_def('rank_table'::regclass, true, true);
+CREATE VIEW v07 AS SELECT pg_get_partition_rule_def('rank_table'::regclass);
+CREATE VIEW v08 AS SELECT pg_get_partition_rule_def('rank_table'::regclass, true);
+CREATE VIEW v09 AS SELECT pg_get_partition_template_def('rank_table'::regclass, true, true);
+CREATE VIEW v10 AS SELECT pg_catalog.pg_is_xlog_replay_paused();
+CREATE VIEW v11 AS SELECT pg_catalog.pg_last_xlog_receive_location();
+CREATE VIEW v12 AS SELECT pg_catalog.pg_last_xlog_replay_location();
+CREATE VIEW v13 AS SELECT pg_catalog.pg_switch_xlog();
```
kalensk, line 29: Maybe its just me but it seems like it would be more self documenting if the view was called based on the table. Might also help with debugging and things. Something like `view_pg_current_xlog_insert_location`. Not sure if that is too wordy, but it is more clear than v01

test/acceptance/pg_upgrade/6-to-7/non_upgradeable_tests/sql/views_with_removed_functions.sql
```diff
@@ -0,0 +1,54 @@
+-- Copyright (c) 2017-2024 VMware, Inc. or its affiliates
+-- SPDX-License-Identifier: Apache-2.0
+
+--------------------------------------------------------------------------------
+-- Create and setup non-upgradeable objects
+--------------------------------------------------------------------------------
+
+CREATE TABLE foo(i int) with (appendonly=true);
```
kalensk, line 8: Can we rename `foo` to something like` ao_table`?

test/acceptance/pg_upgrade/6-to-7/non_upgradeable_tests/sql/views_with_removed_functions.sql
kyeap-vmware: The list of views here containing removed functions are arbitrary and non exhaustive so I chose a naming scheme that was easy to verify at a glance rather than explicit in what the views contained. There are over 100 removed functions. I just chose the subset that was the easiest to test.

test/acceptance/pg_upgrade/6-to-7/non_upgradeable_tests/expected/views_with_removed_functions.out
```diff
@@ -0,0 +1,94 @@
...
+DROP VIEW v03;
+DROP VIEW
+DROP VIEW v02;
+DROP VIEW
+DROP VIEW v01;
+DROP VIEW
+DROP TABLE rank_table;
+DROP TABLE
+DROP TABLE foo;
+DROP TABLE
```
soumyadeep2007, line 94: No need to drop these. The "workaround" section should only contain objects that the customer would need to operate on and nothing more.

soumyadeep2007 (approved):
Just a minor thing. Create all the objects in a schema to avoid conflicts with other tests (foo/ao_table/v01 etc aren't very unique).

Otherwise LGTM!

test/acceptance/pg_upgrade/6-to-7/non_upgradeable_tests/expected/views_with_removed_functions.out
kyeap-vmware: > The "workaround" section should only contain objects that the customer would need to operate on and nothing more.

Is there a reason we would need to keep certain objects around after testing? I've been using this section more like a cleanup section. My intention was to keep the tests isolated.

kyeap-vmware: > Create all the objects in a schema to avoid conflicts with other tests

This is also part of the reason why the tests cleanup after themselves. Maybe a "cleanup" section would be more accurate.

PR #903, opened 2024-03-12
add non upgradable test for views with removed types

kyeap-vmware:
Views that use removed types will cause upgrade to fail. This happens during metadata restore on the target cluster because pg_restore will error trying to create a view using types that do not exist anymore. This is not ideal as we could be many hours into upgrade before a view using a removed type causes pg_upgrade to fail. This check calls a support function on the source cluster to check if views using removed types exist.

https://github.com/greenplum-db/gpupgrade/pull/903
https://github.com/greenplum-db/gpdb/pull/17187
https://github.com/greenplum-db/gpdb/pull/17188


bmdoil (approved):
LGTM

PR #905, opened 2024-03-15
Add non-upgradable tests for disallowed operator =>

Chibin:
Postgres 9.5 and above no longer allows operator => to be created by the users. Upgrading from GP6 (9.4) to GP7 will fail due to this.

For example, when running the command `CREATE OPERATOR =>`, the user will see the following error: `ERROR:  syntax error at or near "=>"`.

So during metadata restore on the target cluster, pg_restore will error trying to create an operator that is no longer allowed. This check ensures that the user is alerted that they have an operator that is no longer allowed.

corresponding pg_upgrade change:
https://github.com/greenplum-db/gpdb/pull/17204


kyeap-vmware (approved):
LGTM

PR #906, opened 2024-03-15
Upgrade pg upgrade output dir

Chibin:
With the new `pg_upgrade` `--output-dir` format, the logs by default go
to the data directory. We now need to explicitly use option `--output-dir` to
put log and `*.txt` to `gpAdminLogs/gpupgrade/pg_upgrade/<role><contentid>`.

The new log directory format will now be:
`gpAdminLogs/gpupgrade/pg_upgrade/<role><contentid>/pg_ugprade.d/<timestamp>/[files and log directory]`

The new behavior is introduced with https://github.com/greenplum-db/gpdb/commit/89808b612d593eb508e538eac74e4ccc1e3a93d6

This change is built on-top of https://github.com/greenplum-db/gpupgrade/pull/905

Edit:
Corresponding PR: https://github.com/greenplum-db/gpdb/pull/17234


bmdoil (commented):
Thanks for working on this!

Just confirming, is your test 
`go test -count=1 -v ./test/acceptance/pg_upgrade -run Test_PgUpgrade_NonUpgradeable_Tests` 
passing locally? I'm having trouble with it but it might be unrelated to this PR.
 
I think we should remove some directories. 

Currently we call
```
func GetPgUpgradeDir(role string, contentID int32) (string, error) {
	logDir, err := GetLogDir()
	if err != nil {
		return "", err
	}

	return filepath.Join(logDir, "pg_upgrade", fmt.Sprintf(role+"%d", contentID)), nil
}
```
What about instead using.
```
gpAdminLogs/gpupgrade/pg_upgrade_<timestamp>/<role><contentid>/[files and logs directory]
```
Everything under the `gpAdminLogs/gpupgrade` directory would be handled by pg_upgrade itself. I'm open to suggestions, the main goals are ease of navigation for support/users. As a side note the data migration scripts will end up at the same level as metadata files and logs.

On a related note, we should simplify the timestamps. I think `20240318T233609.231` is excessive. Why not daily, like the agent/hub/cli logs? The pg_upgrade logs themselves should provide the timing precision and if they don't (they don't), we should make them better.

Chibin:
> Thanks for working on this!
> 
> Just confirming, is your test `go test -count=1 -v ./test/acceptance/pg_upgrade -run Test_PgUpgrade_NonUpgradeable_Tests` passing locally? I'm having trouble with it but it might be unrelated to this PR.

Yes, that's correct.
However, the disallowed operator test will fail until my changes are committed into gpdb as well.
I had to rebuild gpdb6 and gpdb7. I also needed to restart the gpupgrade process and clean-up the .gpupgrade for the test to initially work.
If those didn't work, I can double check.

Have we verified that pg_upgrade is creating the correct directories when using the `--output-dir`?
I've verified mine with the following:
```
PGHOST=localhost ${GPHOME_TARGET}/bin/pg_upgrade  -b ${GPHOME_SOURCE}/bin -d ... -D ... --old-port 6000 --new-port 7000 --check --output-dir $HOME/something
```

> 
> I think we should remove some directories.
> 
> Currently we call
> 
> ```
> func GetPgUpgradeDir(role string, contentID int32) (string, error) {
> 	logDir, err := GetLogDir()
> 	if err != nil {
> 		return "", err
> 	}
> 
> 	return filepath.Join(logDir, "pg_upgrade", fmt.Sprintf(role+"%d", contentID)), nil
> }
> ```
> 
> What about instead using.
> 
> ```
> gpAdminLogs/gpupgrade/pg_upgrade_<timestamp>/<role><contentid>/[files and logs directory]
> ```
> 
> Everything under the `gpAdminLogs/gpupgrade` directory would be handled by pg_upgrade itself. I'm open to suggestions, the main goals are ease of navigation for support/users. As a side note the data migration scripts will end up at the same level as metadata files and logs.
> 
Ideally, this is what I would like as well... unfortunately `pg_upgrade` automatically creates the `pg_ugprade.d` for you when specifying a directory.

When using the `--output-dir` option, it creates `pg_ugprade.d/<timestamp>/[files and logs directory]` for us.
If we're allowed to deviate from upstream, I can make the change so that it doesn't create the `pg_ugprade.d` directory when using the `--output-dir` option.

> On a related note, we should simplify the timestamps. I think `20240318T233609.231` is excessive. Why not daily, like the agent/hub/cli logs? The pg_upgrade logs themselves should provide the timing precision and if they don't (they don't), we should make them better.

I also like the daily timestamp idea, if we're okay as a team with that, I can update `pg_upgrade` to be a daily timestamp instead of the current `20240318T233609.231`.

Chibin:
From our offline discussion, we've decided the following:

- From a user's perspective using `gpupgrade`, `pg_upgrade` creating a new timestamp per millisecond can be a bit confusing. This is because if `gpupgrade`, runs an multiple pg_upgrade per segment, the timestamps will all be slightly different.
  - So, we're letting `gpupgrade` dictate the timestamp instead of pg_ugprade doing it automatically.
  - `gpupgrade` will pass the following:
  `--output-dir gpadminlog/gpupgrade/pg_upgrade<timestamp>/<role><contentid>`
   - This will make `pg_ugprade` generate: `gpadminlog/gpupgrade/pg_upgrade<timestamp>/<role+contentid>/[files and logs]`

The timestamp will be in the format of `%Y%m%dT%H%M%S`?

bmdoil (approved):
Tested both pg_upgrade tests and gpupgrade workflow, LGTM. 
nit: The two commits have the same title. Were you meaning to combine them?

Chibin:
> Tested both pg_upgrade tests and gpupgrade workflow, LGTM.

Great, thank you!

> nit: The two commits have the same title. Were you meaning to combine them?

Yes, I plan to squash them together. I kept them separate before so I could quickly revert if the newer approach had issues.

PR #907, opened 2024-03-22
Add `gpupgrade check` to run subset of pg_upgrade checks

kyeap-vmware:
Runs a subset of pg_upgrade checks against a source gpdb6 cluster in preparation for upgrade using gpbackup. Only checks relevant for upgrading using gpbackup.

gpdb commit reference:
https://github.com/greenplum-db/gpdb/pull/xxxxx

gpupgrade check PR https://github.com/greenplum-db/gpupgrade/pull/907
pg_upgrade --check-not-in-place PR https://github.com/greenplum-db/gpdb/pull/17236

```
    kyeap [ main ] ~/workspace/gpbackup
 --- $ gpupgrade check --source-master-port 6000
Performing Consistency Checks on Old Live Server
------------------------------------------------
Checking cluster versions                                   ok
Checking for external tables used in partitioning           ok
Checking for non-covering indexes on partitioned AO tables  ok
Checking for indexes on partitioned tables                  ok
Checking array types derived from partitions                ok
Checking for multi-column LIST partition keys               ok
Checking for functions dependent on plpython2               ok
Checking for views with removed operators                   ok
Checking for views with removed functions                   ok
Checking for SHA-256 hashed passwords                       ok
Checking for removed "abstime" data type in user tables     ok
Checking for removed "reltime" data type in user tables     ok
Checking for removed "tinterval" data type in user tables   ok
Checking for invalid "unknown" user columns                 ok
Checking for roles starting with "pg_"                      ok

*Clusters are compatible*
```

Design doc:
https://docs.google.com/document/d/1r17shvokIRZWWtvWS2OqsqUIS1BwTkitnssWllQx7hw/edit?usp=sharing

### DIFF
--- a/cli/commands/check.go
+++ b/cli/commands/check.go
@@ -1,0 +1,104 @@
+// Copyright (c) 2017-2023 VMware, Inc. or its affiliates
+// SPDX-License-Identifier: Apache-2.0
+
+package commands
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
+	"path"
+	"strconv"
+
+	"github.com/greenplum-db/gpupgrade/greenplum"
+	"github.com/greenplum-db/gpupgrade/greenplum/connection"
+	"github.com/greenplum-db/gpupgrade/idl"
+	"github.com/greenplum-db/gpupgrade/utils"
+	"github.com/greenplum-db/gpupgrade/utils/errorlist"
+	"github.com/spf13/cobra"
+)
+
+func check() *cobra.Command {
+	var sourceGPHome string
+	var targetGPHome string
+	var sourcePort int
+
+	cmd := &cobra.Command{
+		Use:   "check",
+		Short: "Executes a subset of pg_upgrade checks for upgrade from GPDB6 to GPDB7",
+		Long:  CheckHelp,
+		PreRunE: func(cmd *cobra.Command, args []string) error {
+			_, err := os.Stat(sourceGPHome)
+			if os.IsNotExist(err) {
+				return fmt.Errorf(`source GPHOME "%s" does not exist`, sourceGPHome)
+			}
+			_, err = os.Stat(targetGPHome)
+			if os.IsNotExist(err) {
+				return fmt.Errorf(`target GPHOME "%s" does not exist`, targetGPHome)
+			}
+
+			return nil
+		},
+		RunE: func(cmd *cobra.Command, args []string) (err error) {
+			// Get connection to db to get cluster info
+			db, err := connection.Bootstrap(idl.ClusterDestination_source, sourceGPHome, sourcePort)
+			if err != nil {
+				return err
+			}
+			defer func() {
+				if cErr := db.Close(); cErr != nil {
+					err = errorlist.Append(err, cErr)
+				}
+			}()
+
+			source, err := greenplum.ClusterFromDB(db, sourceGPHome, idl.ClusterDestination_source)
+			if err != nil {
+				return err
+			}
+
+			if source.Version.Major != 6 {
+				return fmt.Errorf(`Your cluster version was detected as %s. This command only runs only runs on gpdb6 to look for objects incompatible with GPDB7. `, source.Version.String())
+			}
+
+			const TimeStringFormat = "20060102T150405"
+			pgUpgradeTimestamp := utils.System.Now().Format(TimeStringFormat)
+			pgUpgradeDir, err := utils.GetPgUpgradeDir(greenplum.PrimaryRole, -1, pgUpgradeTimestamp, "7.0.0")
+			if err != nil {
+				return err
+			}
+
+			err = utils.System.MkdirAll(pgUpgradeDir, 0700)
+			if err != nil {
+				return err
+			}
+
+			pgUpgradeArgs := []string{
+				"-c",
+				"--continue-check-on-fatal",
+				"--retain",
+				"--output-dir", pgUpgradeDir,
+				"-d", source.CoordinatorDataDir(),
+				"-b", path.Join(sourceGPHome, "bin"),
+				"-p", strconv.Itoa(sourcePort),
+				"--check-not-in-place",
+			}
+
+			pgUpgradeBinary := path.Join(targetGPHome, "bin", "pg_upgrade")
+			command := exec.Command(pgUpgradeBinary, pgUpgradeArgs...)
+			command.Stdout = os.Stdout
+			command.Stderr = os.Stderr
+
+			pgUpgradeErr := command.Run()
+			if err != nil {
+				return pgUpgradeErr
+			}
+			return nil
+		},
+	}
+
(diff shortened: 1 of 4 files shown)


cli/commands/check.go
```diff
@@ -0,0 +1,84 @@
...
+			logDir, err := utils.GetLogDir()
+			if err != nil {
+				return err
+			}
+
+			pgUpgradeArgs := []string{
+				"-c",
+				"--continue-check-on-fatal",
+				"--retain",
+				"--output-dir", logDir,
```
Chibin, line 55: This will need to change once this branch is rebased to main.

Chibin:
I just want to note that if the target dir doesn't exist, gpupgrade will show nothing.

```
$ gpupgrade check --source-master-port 6000 --target-gphome /usr/local/notreal
$
```

This can get confusing if the default target directory doesn't exist.
```
$ gpupgrade check --source-master-port 6000
$
```

Everything else LGTM.

cli/commands/check.go
kyeap-vmware: why does this need to change? The default logging directory was changed but I'm pulling the logDir using utils function which is where all other logs are landing?

kyeap-vmware: Nice. I see it now. Thanks for the explanation!

kalensk:
Overall, it sounds like product has some new major ideas for upgrade (and gpupgrade) both inplace and not-inplace yet as far as I can tell there is no clear detailed design doc that both product and engineering and others can follow to be on the same page.

This has been discussed within the previous team several times in various forms such as:

- [gpupgrade workflow for pg_upgrade checks for gpupgrade 1.4.0](https://docs.google.com/document/d/1g9OV-JLQ2SUZyWL97eo4JqWnZjwTORKmC3LuVeM2SvI/edit)
- [gpupgrade 1.6.0 workflow proposal](https://docs.google.com/document/d/1dq6ofVF6mzwy5EA-3DnieT81xoAvAJCYKGNZfwIpykQ/edit#heading=h.bb2q6spmcyd2)
- [gpupgrade 1.7.0 workflow proposal](https://docs.google.com/document/d/1OFMZtI_v3Im8ypbRm-aw3LVYpHCh5IgL9qDwtrrOFcI/edit)

Some major considerations:
- Adding a check command misses quite a few other gpupgrade related checks like disk space, ensuring gpupgrade is properly installed on all hosts, a critical check that the customers environment is not mixing GPDB versions which results in crazy errors, and others.
- There is a very strong desire to keep gpupgrade clean and simple as possible, which means sticking to the ~3 phase/command approach of initialize, execute, and finalize. Adding another phase/command _could_ be a slippery slope away from these goals. Being very conscious around this is key.
- Being able to clearly differentiate when downtime occurs for the customer is useful. Or another potential way of saying that is destructive (ie: dropping indexes) vs. non-destructive operations during the upgrade. As well as being able to clearly differentiate when one can revert an upgrade or not.
- When designing gpupgrade we had some fundamental concepts for initialize (or in this case check) which should be carried forward as the implementation changes. That is, initialize/check should be able to run well before the upgrade window such as 1-2 months before, so customers can see what issues they may need to slowly address. However, this surfaces the question that the hub and agents will be left running with the current implementation - is this an issue? Is initializing the target cluster to run the checks and leaving it there for 1-2 months before the upgrade an issue? Just some food for thought.


Rather than introducing the check command as proposed here I would strongly consider the [following proposal](https://docs.google.com/document/d/1OFMZtI_v3Im8ypbRm-aw3LVYpHCh5IgL9qDwtrrOFcI/edit) which is basically renaming initialize to check, splitting execute into upgrade-primaries add upgrade-mirrors as described which also calls gpinitsystem, and finalize as described here. This allows a check command to be used for both in-place and not-inplace upgrades (or if really needed perhaps add a flag such as “gpupgrade check –inplace | –not-inplace”), as well as adding clarity around when downtime occurs, and when one can revert. This also keeps the additional gpupgrade checks available to better catch errors early on such as the environment check, GPDB version, disk space, etc.

cli/commands/check.go
```diff
@@ -0,0 +1,104 @@
...
+			command := exec.Command(pgUpgradeBinary, pgUpgradeArgs...)
+			command.Stdout = os.Stdout
+			command.Stderr = os.Stderr
+
+			pgUpgradeErr := command.Run()
+			if err != nil {
+				return pgUpgradeErr
+			}
+			return nil
+		},
```
kalensk, line 96: We should use the existing pattern of calling `upgrade.Run(streams.Stdout(), streams.Stderr(), opts)`

cli/commands/check.go
```diff
@@ -0,0 +1,104 @@
...
+			}()
+
+			source, err := greenplum.ClusterFromDB(db, sourceGPHome, idl.ClusterDestination_source)
+			if err != nil {
+				return err
+			}
+
+			if source.Version.Major != 6 {
+				return fmt.Errorf(`Your cluster version was detected as %s. This command only runs only runs on gpdb6 to look for objects incompatible with GPDB7. `, source.Version.String())
+			}
```
kalensk, line 61: Is there a way we can plug into the existing VerifyCompatibleGPDBVersions() architecture? This whole check command implementation greatly deviates from all existing patterns which is quite concerning to me as we lose all that great infrastructure, uniformity, included feature set, etc.

cli/commands/check.go
```diff
@@ -0,0 +1,104 @@
...
+
+			const TimeStringFormat = "20060102T150405"
```
kalensk, line 63: Searching the code base for `.Format("2` I see multiple format strings with different precision (on purpose!). Now might be a good time to extract them as constants to like the utils package.

cli/commands/check.go
```diff
@@ -0,0 +1,104 @@
...
+			_, err := os.Stat(sourceGPHome)
+			if os.IsNotExist(err) {
+				return fmt.Errorf(`source GPHOME "%s" does not exist`, sourceGPHome)
+			}
+			_, err = os.Stat(targetGPHome)
+			if os.IsNotExist(err) {
+				return fmt.Errorf(`target GPHOME "%s" does not exist`, targetGPHome)
+			}
+
+			return nil
```
kalensk, line 40: We should be using `upgrade.PathExist()`

cli/commands/check.go
kalensk, line 61: Its a bit weird to me to see this check command hard coded for only 6? Any chance we can make this a bit more generic in a sense or future proof in a way? I don't quite seem the harm overall.

kalensk:
The current implementation of this check command greatly deviates from all existing gpugprade patterns which is quite concerning as we lose all that great infrastructure, uniformity, included feature set, etc.

kyeap-vmware:
Closing this PR. We will be moving in a different direction that does not involve attaching a new command to gpupgrade.

PR #910, opened 2024-04-03
`--output-dir` should only be used for target GP7

Chibin:
`--output-dir` should only be used for target GP7

    Make `--output-dir` option for `pg_upgrade` conditional.
    `--output-dir` option is only introduced to GP 7.2.0, so we should use
    the prior logging behavior for anything below GP 7.


bmdoil (approved):
LGTM

PR #911, opened 2024-04-16
Add gpupgrade_internal_workflow_README.md

kyeap-vmware:
1st draft of the gpupgrade internal workflow for using pg_upgrade.


gpupgrade_internal_workflow_README.md
```diff
@@ -0,0 +1,99 @@
...
+    1. Use pg_upgrade to upgrade the coordinator. **pg_upgrade 1st call**
+    2. Copy the upgraded coordinator to each segment. We use the upgraded coordinator to bootstrap segment upgrade.
+    3. Use pg_upgrade to upgrade the segments **pg_upgrade 2nd call**
+
+1. gpupgrade finalize
+    1. regenerate indexes and stats?
+
+## Why are segments upgraded by copying MDD to segments?
+We would like to address this specific question because it is the biggest reason why pg_upgrade can be complicated for GPDB5 > GPDB6. Theoretically, a better way to upgrade a GPDB cluster would be to run pg_upgrade on coordinator and segments. This is not possible for GPDB4 and GPDB5 segments. Segments do not contain enough schema data to dump and restore GPDB partition tables. This is why segments are upgraded by copying MDD to segments. Unfortunately, this method of upgrading segments comes with a few ramifications.
+
```
kalensk, line 62: Shouldn't we mention the original intention of copying the MDD which was for performance reasons. (Granted that may need to be reevaluated). But mentioning as the main reason why seems important.

jimmyyih: > original intention of copying the MDD which was for performance reasons.

This is also what I remember as the main reason.

bmdoil: That was the main reason, only in retrospect did we learn that
1. It wasn't actually faster in most cases
2. It created several tricky bugs
3. Upgrading segments directly isn't possible for 5->6

If there are relevant PRs or gpdb-dev threads we should pull in the decision making context here.

kyeap-vmware:
I will add another section that will go into more details of the inner working of pg_upgrade.

kyeap-vmware:
Updated 4/25/2024. Should be easier to read in markdown:
https://github.com/kyeap-vmware/gpupgrade/blob/gpupgrade-internal/gpupgrade_internal_workflow_README.md

gpupgrade_internal_workflow_README.md
```diff
@@ -0,0 +1,558 @@
...
+    └── VACUUM FREEZE pg_catalog.pg_database
+```
+In order to make schema on the segments visible we need to freeze the master before copying it over to segments. 
+
+This is done in commit [5583ecdeb](https://github.com/greenplum-db/gpdb/commit/5583ecdeb).
+
+---
+
+#### 2. Do not preserve relfrozenxid of user tables for `pg_dump --binary-upgrade`
+Normally `pg_restore --binary-upgrade` would set a table's relfrozenxid, but
```
bmdoil, line 244: The --binary-upgrade flag for pg_restore was a GPDB addition. See https://github.com/pivotal/gp-gpdb-staging/commit/5844158d125bfc54714d65d6bb2bbf0bebf5fc4d

I've never been convinced this is needed, but it's a lower priority thing to investigate.

gpupgrade_internal_workflow_README.md
```diff
@@ -0,0 +1,558 @@
...
+fails. This means that as soon as 1 check failed, the checks after it would not
+be run. We have made changes so all checks will run even if one fails. This
+used to be much more relevant when checks could run for hours.
+
+**--skip-checks**<br>
+This is used to skip compatibility checks during gpupgrade execute. This is originally put in for performance reasons. Before checks were optimized they were very costly. There is a hidden flag, `gpupgrade executed --skip-pg-upgrade-checks` that will skip all pg_upgrade checks.
+
+**--skip-target-check**<br>
+This doesn't seem to be used anywhere... I also fail to see how running
+only checks for source cluster fits into the current gpupgrade workflow.
```
bmdoil, line 265: Unrelated but I'm considering a scenario where we do use this flag in gpupgrade. If that happens I'll open a separate PR that also updates this README.

gpupgrade_internal_workflow_README.md
```diff
@@ -0,0 +1,558 @@
...
+	 * upgrade segments, coordinator's relfilenodes, _fsm, _vm files will end up
+	 * on segments. Normally this is ok since the files will end up being
+	 * overwritten. However, there is an edge case where there can be data in a
+	 * table on coordinator, but no data on the segment. Examples of such tables
+	 * where this can occur are pg_ao(cs)seg tables. If this edge case happens,
+	 * The new segment's table will end up with old segment's relfilenode and new
+	 * coordinator's _fsm and _vm file. The _fsm and _vm files don't get
+	 * overwritten because they aren't supposed to exist on the segment.
+	 * Attempting to run VACUUM on this table after upgrade completes will
+	 * result in a similar error below.
```
bmdoil, line 384: Food for thought, wouldn't it be more performant and easier to filter out relfilenodes we don't want during copy MDD step instead of copying everything to segments and doing this workaround? Also could  possibly result in removing/simplifying the `parent_partitions_with_seg_entries` data migration script in that case.

gpupgrade_internal_workflow_README.md
```diff
@@ -0,0 +1,558 @@
...
+
+##### 4. pg_aoseg values from coordinator ends up on segments.
+Is this still true? This may have been true we had to rebuild the AO aux
+entries manually using restore_ao_segments(). I'm not so sure this is still
+true. After commit
+[466df4e26a](https://github.com/greenplum-db/gpdb/commit/466df4e26a), these
+values are now transfered using relfilnodes. The values in these tables my be
+overwritten when segment's relfilenodes are moved. If vlues from coordinator
+are still present the effect if unknown. Formal investigation was not done. It
+didn't seem to effect segments before.
```
bmdoil, line 417: Let's try to answer this question before merging README. Should be quick spike.

gpupgrade_internal_workflow_README.md
```diff
@@ -0,0 +1,558 @@
...
+we also freeze coordinator after relfilenode transfer, which should cause the
+`vacuumdb --all --freeze` in `prepare_new_cluster` to be a no-op on segments. 
+
+We may be able to disable it to further simplify the workflow on segments.
+
+### 3. We should be able to skip get_rel_infos if user_opts.check is set for more initialize performance.
+```
+│   ├── get_db_and_rel_infos(&old_cluster)
+│   │   ├── get_db_infos
+│   │   └── get_rel_infos <-------
```
bmdoil, line 524: We should prioritize this, seems an easy performance win.

gpupgrade_internal_workflow_README.md
```diff
@@ -0,0 +1,558 @@
...
+```
+When using pg_upgrade with the `--check` flag, pg_upgrade will run checks and then bail early which means it will not reach `gen_db_file_maps`. We should be able to make checks faster if we disable `get_rel_infos` when `--check` flag is used.
+
+### 4. I see a mixture of styles for skipping code. Do we want to consolidate styles?
+**Style 1:** if statement wraps the functions. This style is likely easier to read at a glance, but produces more diff that need to be handled when merging from upstream.
+```
+if (!skip_checks())
+    check_loadable_libraries();
+```
+**Style 2:** if statement immediately entering the function.
```
bmdoil, line 543: I prefer style 2.

gpupgrade_internal_workflow_README.md
```diff
@@ -0,0 +1,558 @@
...
+check_new_cluster_is_empty(void)
+{
+    if (!is_greenplum_dispatcher_mode())
+        return;
+```
+
+### 5. Investigate this comment.
+```
+TODO: We are currently missing the support to migrate over bitmap indexes.
+Hence, mark all bitmap indexes as invalid.
```
bmdoil, line 555: Let's try to dig up some context here and add it to the README.

gpupgrade_internal_workflow_README.md
```diff
@@ -0,0 +1,558 @@
...
+```
+### 6. get_db_and_rel_infos is called 4 times in coordinator.
+Does this impact our performance in a noticeable way?
```
bmdoil, line 558: It definitely does on large clusters.

bmdoil (commented):
Did an initial pass. I think we should add a section for Deep's gist that really gets in the weeds. Helpful for those with good server context to understand the ins and outs.

gpupgrade_internal_workflow_README.md
kyeap-vmware: I thought about this a little bit. Came up with 2 changes that may make our lives easier.

1. **Remove unwanted files during the COPY MDD step in gpupgrade.**
This would be done in gpupgrade after the pg_upgrade on coordinator is called. Implementation detail question. How would we be able to keep track of wanted/unwanted files/relfilenodes? We would want to filter out the following files coordinator `_fsm`, `_vm`, and relfilenodes.
1.  **Freeze and save current state of new MDD right after coordinator `create_new_objects`.**
I believe there are quite a few wins here with little downside I can see.
    1.  coordinator `_fsm`, `_vm`, and relfilenodes have not been transferred to new segment yet. This resolves all issues we see from COPY MDD step. We would be able to remove most COPY MDD fixes. The only one we can't remove is orphaned ao_blkdir.
    1. Remove `parent_partitions_with_seg_entries` data migration script.

gpupgrade_internal_workflow_README.md
kyeap-vmware: +1. I also prefer style 2.

kalensk (commented):
This is excellent! Very valuable doc that is already paying off with the items you brought up at the end.

I am still digging into it all, and to be honest quite a bit is above my head at the moment. (I'm also busy on other components).

Anyway, thank you!

gpupgrade_internal_workflow_README.md
kyeap-vmware: Before commit
[466df4e26a](https://github.com/greenplum-db/gpdb/commit/466df4e26a) every
pg_aoseg table was TRUNCATED every pg_aoseg table before rebuilding it.

After commit
[466df4e26a](https://github.com/greenplum-db/gpdb/commit/466df4e26a), these
values are now transfered using relfilnodes. The values in pg_aoseg tables that
came from coordinator are overwritten when segment's relfilenodes are moved.

kyeap-vmware:
Thanks for taking a look Kalen!

> I am still digging into it all, and to be honest quite a bit is above my head at the moment. (I'm also busy on other components).

I think this is a sign to me that this README is lacking in some areas. I do still have plans to add sections on relfrozenxids and pg_aoseg.

gpupgrade_internal_workflow_README.md
```diff
@@ -0,0 +1,574 @@
...
+
+1. gpupgrade finalize
+    1. regenerate indexes and stats?
+
+## Why are segments upgraded by copying MDD to segments?
+We would like to address this specific question because it is a big reason why
+pg_upgrade can be complicated. Originally this was done because it was thought
+that this would be more beneficial for performance. It was only later
+discovered that this is likely the only way upgrade could work for 5 > 6
+upgrade because it segments do not contain enough schema data to dump and
```
Chibin, line 77: Did you mean each/the? or something else?
```suggestion
upgrade because each segments do not contain enough schema data to dump and
```

gpupgrade_internal_workflow_README.md
```diff
@@ -0,0 +1,574 @@
...
+restore GPDB partition tables. Unfortunately, this method of upgrading segments
+comes with a few ramifications.
+
+## What are the ramifications of copying MDD to segments to perform segment schema upgrade?
+1. Existing MDD files are now on segments. Issues with this as the root cause has manifests in several ways.
+    1. xids from coordinator ends up on segments, they must be edited to to ensure row visibility on the segment is correct.
+    1. pg_aoseg entries from the coordinator end up on the segments. It is currently unknown if this has an effect on the segments, but it doesn't seem to be causing any issues.
+    1. AO aux VSM and FSM files from the coordinator end up on the segments. We need to ensure that the FSM and VSM of AO aux tables from coordinator do not get used by segments. User tables are not so much of an issue because data is not stored on coordinator so these files sould be empty or not exist.
```
Chibin, line 85: Might be a silly question, what's VSM and FSM?
My guess is VSM is visimap?

Chibin, line 85:
```suggestion
    1. AO aux VSM and FSM files from the coordinator end up on the segments. We need to ensure that the FSM and VSM of AO aux tables from coordinator do not get used by segments. User tables are not so much of an issue because data is not stored on coordinator so these files should be empty or not exist.
```

gpupgrade_internal_workflow_README.md
```diff
@@ -0,0 +1,574 @@
...
+1. pg_upgrade is fragmented because there are different code paths for coordinator and segment upgrades. pg_upgrade can be running in dispatch or segment mode.
+1. Developers must keep in mind that pg_upgrade is run twice during execute. 
+    1. This has performance impacts because master and segments cannot upgrade at the same time.
+    1. Any pg_upgrade bottlenecks can be twice as inefficient. One such example of this is that checks previously ran twice. Once on coordinator and against on segment. Before optimization, view checks were taking up to 2 hours for a single check, which meant that each check added 4 hours to run time.
```
Chibin, line 89: Did you mean again or I'm misreading it.
```suggestion
    1. Any pg_upgrade bottlenecks can be twice as inefficient. One such example of this is that checks previously ran twice. Once on coordinator and again on segment. Before optimization, view checks were taking up to 2 hours for a single check, which meant that each check added 4 hours to run time.
```

Chibin, line 89: > One such example of this is that checks previously ran twice.

Does that mean that it doesn't do this anymore and we only run it once now?

gpupgrade_internal_workflow_README.md
```diff
@@ -0,0 +1,574 @@
...
+target cluster, but can be fixed with a bit a tweaking or just by remaking
+them. To make migration as easy as possible for customers we may generate
+scripts that fix user object to make them upgradable. We also generate the a
+revert script to revert objects back to their original structure if applicable.
+As of today data migration scripts are generated in gpupgrade.
+
+Learn more about these here:
+https://github.com/greenplum-db/gpupgrade/tree/main/data-migration-scripts
+
+## pg_upgrade execution visualized
```
Chibin, line 122: This is great! Thank you for doing this.

gpupgrade_internal_workflow_README.md
```diff
@@ -0,0 +1,574 @@
...
+│   ├── UPDATE pg_class SET relfrozenxid
+│   └── UPDATE pg_class SET relminmxid
+└── freeze_master_data*
+    ├── VACUUM FREEZE
+    └── VACUUM FREEZE pg_catalog.pg_database
+
+* specific to GPDB
+```
+### \*\*\*COPY MDD to segments to bootstrap segments\*\*\*
+At this point the primaries are "upgraded" because they got copied from master. This manner of upgrading a cluster is quite unconventional and the copies of coordinators must now be converted to segments by copying the control values from the old segment cluster.
```
Chibin, line 181:
```suggestion
At this point the primaries are "upgraded" because the data directory were copied from master. This manner of upgrading a cluster is quite unconventional and the copies of coordinators must now be converted to segments by copying the control values from the old segment cluster.
```

gpupgrade_internal_workflow_README.md
kyeap-vmware: I will expand on this in the README. FSM is free space map. and yes! Thank you for catching this mistake, VSM is supposed to be VM for visibility map.

gpupgrade_internal_workflow_README.md
```diff
@@ -0,0 +1,695 @@
...
+#### 1a. Freezing the master after relfilenode transfer
+Since segments are upgraded by copying mdd to bootstrap segment upgrade, we
+must ensure that the schema is visible on segments. GPDB does this by freezing
+the master. Freezing must happen sometime after schema is restored in
+`create_new_objects`.
+```
+pg_upgrade on coordinator
+...
+├── transfer_all_new_tablespaces
+├── update_db_xids* <----- Make relfilenode tuples visible on coordinator, also fixes VACUUM FREEZE error.
```
huansong, line 289: Do you mean the pg_class tuples (which contain the relfilenode)?

gpupgrade_internal_workflow_README.md
```diff
@@ -0,0 +1,695 @@
...
+1. Check for cluster compatibility
+1. Freeze all new cluster rows to remove references to clog entries
+1. Copy control file settings to new cluster (clog, xlog, xid)
+1. Set coordinator's frozenxid and minmxid
+1. Invalidate Indexes
+1. Set coordinator's frozenxid and minmxid again
+1. Copy or link files (relfilenodes) to new cluster
+1. Reset system identifier
+
+## How gpupgrade uses pg_upgrade in it's workflow
```
huansong, line 56: This section seems just a briefed version of `pg_upgrade's modified workflow for GPDB`? Maybe we can just merge the two.

huansong (commented):
Thanks for doing it! Helps a lot. Will continue reviewing as it's being iterated.

gpupgrade_internal_workflow_README.md
```diff
@@ -0,0 +1,695 @@
+# gpupgrade internal workflow with pg_upgrade
+
+**Goal**: The purpose of this README is to capture the process of how in-place
```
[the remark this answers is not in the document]
bmdoil: I think we can add a small section on the not-in-place flag further down, but it's only tangentially related to gpupgrade

gpupgrade_internal_workflow_README.md
```diff
@@ -0,0 +1,695 @@
...
+1. gpugprade execute
+    1. Run pg_upgrade to upgrade the coordinator
+    2. Copy the upgraded coordinator to each segment. We use the upgraded
+       coordinator to bootstrap segment upgrade.
+    3. Run pg_upgrade to upgrade the segments
+
+1. gpupgrade finalize
+    1. regenerate indexes and stats?
+
+## Why are segments upgraded by copying MDD to segments?
```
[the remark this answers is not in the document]
kyeap-vmware: I will clarify this. By copy MDD I mean all of Master Data Directory is rysnc'd to the segments.

gpupgrade_internal_workflow_README.md
```diff
@@ -0,0 +1,695 @@
...
+│   ├── UPDATE pg_database SET datfrozenxid, datminmxid
+│   ├── UPDATE pg_class SET relfrozenxid
+│   └── UPDATE pg_class SET relminmxid
+└── freeze_master_data*
+    ├── VACUUM FREEZE
+    └── VACUUM FREEZE pg_catalog.pg_database
+
+* specific to GPDB
+```
+### \*\*\*COPY MDD to segments to bootstrap segments\*\*\*
```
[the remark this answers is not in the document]
kyeap-vmware: This is part of the gpupgrade orchestration in golang, called in agent/upgrade_primaries.go.
```
func restoreBackup(backupDir string, newDataDir string) error {
	options := []rsync.Option{
		rsync.WithSources(utils.GetCoordinatorPostUpgradeBackupDir(backupDir) + string(os.PathSeparator)),
		rsync.WithDestination(newDataDir),
		rsync.WithOptions("--archive", "--delete"),
		rsync.WithExcludedFiles(
			"internal.auto.conf",
			"postgresql.conf",
			"pg_hba.conf",
			"postmaster.opts",
			"gp_dbid",
			"gpssh.conf",
			"gpperfmon"),
	}

	return rsync.Rsync(options...)
}

```

Segments run rsync to copy the coordinator's Master Data Directory.

gpupgrade_internal_workflow_README.md
```diff
@@ -0,0 +1,695 @@
...
+1. Copy or link files (relfilenodes) to new cluster
+
+
+More about pg_upgrade implementation can be found here:<br>
+https://github.com/greenplum-db/gpdb/blob/6X_STABLE/contrib/pg_upgrade/IMPLEMENTATION
+
+
+## pg_upgrade's modified workflow for GPDB
+#### pg_upgrade check
+1. Check for cluster compatibility
```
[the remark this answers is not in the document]
kyeap-vmware: The check is conducted both on coordinator and segment. There is no difference with the check conducted on coordinator and segment. gpupgrade contains an option to skip both checks.

gpupgrade_internal_workflow_README.md
kyeap-vmware: Story made.

gpupgrade_internal_workflow_README.md
kyeap-vmware: story made

gpupgrade_internal_workflow_README.md
kyeap-vmware: I double checked this. This is not quite true. We do run checks multiple times.
initialize - checks run on coordinator
execute - checks run on coordinator (again) and segments.

During execute we give users the option to skip checks on coordinator and segments.

gpupgrade_internal_workflow_README.md
```diff
@@ -0,0 +1,695 @@
...
+       at the same time.
+    1. Any pg_upgrade bottlenecks can be twice as inefficient. One such example
+       of this is that checks previously ran twice. Once on coordinator and
+       against on segment. Before optimization, view checks were taking up to 2
+       hours for a single check, which meant that each check added 4 hours to
+       run time.
+1. Indexes are marked invalid on the new coordinator which are then copied to
+   new segments. We must reset them temporarily on the new segment before we
+   query for all relations to make sure they are recognized when getting a list
+   of relations for relation matching that happens in gen_db_file_maps.
```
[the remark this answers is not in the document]
kyeap-vmware: bpchar_pattern_ops and bitmap indexes

gpupgrade_internal_workflow_README.md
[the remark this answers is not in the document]
kyeap-vmware: added --run-migration-checks as a GPDB flag in the `Additional flags for pg_upgrade` section.

gpupgrade_internal_workflow_README.md
kyeap-vmware: Make tuples from relfilenode transfer (pg_aoseg, aoblkdir, fastsequence) visible on coordinator, also fixes VACUUM FREEZE error. Tuples are made visible by modifying relfrozenxid of tuples in pg_class.

gpupgrade_internal_workflow_README.md
kalensk: To give some context on that gpupgrade does this on explicitly on purpose and was designed that way in case there are changes to the customers data set between running initialize and execute. That is, customers can and should be running initialize very early on to surface issues and work on them. And when execute runs it runs the checks again in case new data was added that is problematic and to ensure the upgrade succeeds. Meaning is better to error out early rather than fail during the actual upgrade which is quite catastrophic.

gpupgrade_internal_workflow_README.md
kyeap-vmware: My intent for this was to show how the pg_upgrade workflow fit inside of gpupgrade.

gpupgrade_internal_workflow_README.md
kyeap-vmware: Of the 4 times this is called. Only 2 times is it called on cluster's that are filled.
1. called on old_cluster to record relations
2. called to make get databases on new cluster before pg_restore
3. called to make sure new cluster is empty
4. called to get relations in new cluster to compare against old cluster.

We may be fine here unless running this function on empty databases affects performance.

gpupgrade_internal_workflow_README.md
kyeap-vmware: Found. 
This commit broke bitmap indexes for GPDB6.
https://github.com/greenplum-db/gpdb/commit/c249ac7a36d9da3d25b6c419fbd07e2c9cfe954f

They were then fixed in this PR.
AO TID changes breaks bitmap indexes for GPDB6 AOCO tables. PR says they must
be remade after binary upgrade.
https://github.com/greenplum-db/gpdb/pull/8462
https://groups.google.com/a/greenplum.org/g/gpdb-dev/c/p8oTYMbHFaI/m/ZEDxFihlDgAJ

I don't see and bitmap index generation in finalize folder of data migration scripts or a script being generated by pg_upgrade. We may need to add that.

bmdoil (approved):
LGTM! Just a few nits here and there.

gpupgrade_internal_workflow_README.md
```diff
@@ -0,0 +1,721 @@
...
+#### pg_upgrade on coordinator
+1. Check for coordinator compatibility
+1. Use pg_dumpall to dump old cluster schema (no data)
+1. Freeze all new cluster rows to remove references to clog entries
+1. Copy control file settings to new coordinator (clog, xlog, xid)
+1. Recreate database and schemas on new coordinator using pg_restore
+1. Copy or link files (relfilenodes) to new cluster
+1. Set coordinator's frozenxid and minmxid
+1. Freeze coordinator to make everything visible for segments
+
```
Chibin, line 43: Thoughts on making this something like:
#### pg_upgrade on each segment (after pg_upgrade on coordinator)
1. COPY coordinator to segment to bootstrap segment upgrade
1. Check segment for compatibility (same checks that are run on coordinator)

or it's too confusing?

gpupgrade_internal_workflow_README.md
```diff
@@ -0,0 +1,721 @@
...
+## How gpupgrade uses pg_upgrade in it's workflow
+gpupgrade is GPDB's utility for performing in-place upgrades. At a high level
+gpupgrade uses pg_upgrade in the following manner:
+1. gpugprade initialize
+    1. Generate and execute data migration SQL scripts
+    1. Configure target cluster
+    1. Run pg_upgrade checks
+1. gpugprade execute
+    1. Run pg_upgrade to upgrade the coordinator
+    2. Copy the upgraded coordinator to each segment. We use the upgraded
```
Chibin, line 65: Let's use the markdown `1.` for numbering.

PR #912, opened 2024-04-29
pg_upgrade nonupgradable test for removed GUCs

divyeshddv:
Gucs/settings can be applied to certain roles or certain databases and are stored in pg_db_role_settings
Upgrade would also try to restore these roles/settings on target cluster Some of these settings are removed in newer version and will cause upgrade to fail.
This happens during metadata restore on the target cluster because pg_restore will error trying to apply these settings, and would fail since these settings are not present anymore. This is not ideal as we could be many hours into upgrade before a removed guc causes pg_upgrade to fail
This check calls a support function on the source cluster to check if any such removed gucs are applied by the user on any of the dbs/roles.


test/acceptance/pg_upgrade/6-to-7/non_upgradeable_tests/expected/removed_gucs.out
```diff
@@ -0,0 +1,81 @@
...
+ALTER ROLE
+
+CREATE DATABASE all_upgradable_gucs_db;
+CREATE DATABASE
+CREATE DATABASE some_nonupgradable_gucs_db;
+CREATE DATABASE
+CREATE DATABASE all_nonupgradable_gucs_db;
+CREATE DATABASE
+
+ALTER DATABASE all_upgradable_gucs_db SET client_min_messages = debug;
```
bmdoil, line 35: Why do we need to set client_min_messages or log_min_messages? 

Setting client_min_messages to debug will cause the test to fail with all the extra output.

divyeshddv: These are GUCs / settings present on both 6X and 7X. 
Since pg_upgrade should be able to upgrade these, our test also shouldn't fail for these settings.

kyeap-vmware: I changed to use GUCs that still exist to ones that won't affect output so broadly.

